PR #335, opened 2019-04-02
286 virtual collection slicing

hobywan:
##### In a nutshell
I'm finally creating a pull-request for this feature.
It aims to provide a way to slice a given multi-dimensional distributed collection.
The slice is actually a view on the already built collection, and can be accessed through a proxy.
The user supply a **new range** and an **index filtering** function which allows to create **sparse** slices.
It provides **relative indexing** so elements can be accessed as if they were part of a new collection.

##### Design choices:

- the filtering function should be stateless and simple to use.
  It is implemented as a static method instead of a lambda which would allow to capture local variables. It is actually registered as a view handler in the system, and can be invoked in any node of the group created for the view.

- the relative indexing is actually managed by the system not by the user as in the initial design.
Indeed it would be less error-prone (such as multiple elements mapped by a single index), especially on nested slices.  It would be less complicated for the user too.

- views may have different index spaces but should have the same dimension (collapsing will be managed differently). Their proxies keep track of their parents to be able to resolve indices on elements filtering or query. The latter is required for nested views.

- module updates were made as incremental as possible. Hence some parts can definitely be refactored, but I intended to limit changes in the existing code to avoid regressions.

##### Missing features
- reductions: I have to update the manager to support sparse/incomplete collections (_very soon_).
- nested slices: I have to update the indexing scheme to support an arbitrary tree of views (_soon_).
- dimension collapse: I have to manage index mapping of arbitrary dimensions. 
- tests: it should not take too much time to implement.

I made multiple updates on the code, so I would like to have a review before going on.
I intended to make integration tests since some core features were updated.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,39 @@
 *o
 *~
+build
+cmake/turn_on_warnings.cmake
 transport
 lib/*-build
 .emacs.desktop
 project/*
 balance.in
 CMakeSettings.json
+
+# Prerequisites
+*.d
+
+# Compiled Object files
+*.slo
+*.lo
+*.o
+*.obj
+
+# Precompiled Headers
+*.gch
+*.pch
+
+# Compiled Dynamic libraries
+*.so
+*.dylib
+*.dll
+
+# Compiled Static libraries
+*.lai
+*.la
+*.a
+*.lib
+
+# Executables
+*.exe
+*.out
+*.app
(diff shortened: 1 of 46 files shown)


cz4rs:
Closing as there is no urgence to get this updated and merged.
Note: this might contain valid ideas for fixing the issue.

PR #423, opened 2019-08-16
410 Dependent Epochs

lifflander:
This is the full implementation of dependent epochs. TD dump prints on hang could be improved.

- Added a bit to envelope for marking system messages
- Implemented dependent epochs for:
  - General active message handlers;
  - Virtual Collection Elements (with whole collection release)
    - Release, check release, add action on release occurs on indexed proxy
  - ObjGroups
- Integration/Unit Tests for Dependent Epochs
  - General active message handler test
  - *added* ObjGroup test
  - *added* Virtual Collection test

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -2,6 +2,7 @@
 set(
   PROJECT_EXAMPLES_LIST
   hello_world
+  dep_epoch
   ring
   comm_func
   rdma_simple_get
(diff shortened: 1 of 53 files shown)


lifflander:
@PhilMiller @pnstickne @nlslatt @uhetmaniuk I just rebased this and fixed a bunch of problems as the code has changed a lot. I would like to merge this soon into develop so we can move on from it.

PR #424, opened 2019-08-19
410 Integral Set 

lifflander:
- Implemented new integral set vt::IntegralSet, that makes epoch alloc/dealloc far more efficient
    - Uses a sorted, non-overlapping interval tree to implement automatic compaction
    - Holes in the release set of epochs no longer inhibit wrap around (a major problem for tree-structured epochs)

### DIFF
--- a/src/vt/termination/interval/interval.h
+++ b/src/vt/termination/interval/interval.h
@@ -75,7 +75,7 @@ public:
   DomainT lower() const { return lb_; }
   DomainT upper() const { return ub_; }
   DomainT width() const { return (ub_-lb_)+1; }
-  bool    valid() const { return lb_ != sentinel and ub != sentinel and lb_ <= ub_; }
+  bool    valid() const { return lb_ != sentinel and ub_ != sentinel and lb_ <= ub_; }
 
   DomainT get(std::size_t i) const {
     vtAssert(lb_ not_eq sentinel, "Lower bound must be valid");
--- a/src/vt/termination/interval/interval.h
+++ b/src/vt/termination/interval/interval.h
@@ -103,7 +103,7 @@ public:
 
   /*
    * Determine if a input interval is tangent to this interval within a
-   * specified grap.
+   * specified gap.
    *
    *        TangentLeft                   TangentRight
    *       |------i-----| lb_        ub_ |------i-----|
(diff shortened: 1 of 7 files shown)


src/vt/termination/interval/interval.h
```diff
@@ -0,0 +1,226 @@
...
+   *
+   *        TangentLeft                   TangentRight
+   *       |------i-----| lb_        ub_ |------i-----|
+   *                      |----this----|
+   *                     ^              ^
+   *                    gap            gap
+   */
+
+  PositionEnum tangent(IntervalType const& i, DomainT gap = 1) const {
+    if (lb_ == i.ub_ + gap) {
```
PhilMiller: The logic of the tangent being exactly `gap` away doesn't match the comment of "within a specified gap". Without a use case, I can't tell which is right.

src/vt/termination/term_window.cc
```diff
@@ -79,148 +75,68 @@ inline bool EpochWindow::isArchetypal(EpochType const& epoch) {
...
-    /*
-     * Set the first and last epoch, since this is the first time this window
-     * is being initialized for the given epoch configuration
-     */
-    first_unresolved_epoch_ = last_unresolved_epoch_ = epoch;
-  } else {
-    last_unresolved_epoch_ = std::max(last_unresolved_epoch_, epoch);
+  // We should possibly perform some error checking once we wrap around in case
+  // of pending actions
+  if (terminated_epochs_.contains(epoch)) {
```
PhilMiller: I think a possible assertion would be

`epoch > getLast() || terminated_epochs_.contains(epoch)`

I.e. either we haven't wrapped around, or we have, and we're clear to re-use this value. We run into trouble with that when `terminated_epochs_` itself covers a wrapped-around range, though. I.e. [0,100] and [MAX-100, MAX].

PhilMiller:
Looks good enough overall. I'm still uneasy about the unused `gap` parameter and `set` methods, but they're only a small distraction, not breaking anything present.

lifflander:
@PhilMiller Ok, will you approve this? I want to get this merged. I'm tried of rebasing it!

I changed the settings at least one other person has to approve before github will allow merging

PR #454, opened 2019-09-10
450 Fix command line ordering error

lifflander:


src/vt/configs/arguments/args.cc
```diff
@@ -142,8 +142,9 @@ namespace vt { namespace arguments {
     return 0;
   }
 
+  // CLI11 app parser expects to get the arguments in *reverse* order!
   std::vector<std::string> args;
-  for (auto i = 0; i < argc; i++) {
+  for (auto i = argc-1; i > 0; i--) {
```
nlslatt: Was leaving off `argv[0]` deliberate?  I'm not sure what CLI11 expects.

src/vt/configs/arguments/args.cc
```diff
@@ -450,16 +451,18 @@ namespace vt { namespace arguments {
...
-  // Reverse iterate (CLI11 reverses the order when they modify the args)
-  for (auto iter = args.rbegin(); iter != args.rend(); ++iter) {
-    for (auto ii = 0; ii < argc; ii++) {
-      if (std::string(argv[ii]) == *iter) {
+  // Iterate forward (CLI11 reverses the order when it modifies the args)
+  for (auto&& skipped : args) {
+    for (auto ii = item; ii < argc; ii++) {
+      if (std::string(argv[ii]) == skipped) {
         ret_idx.push_back(ii);
+        item++;
```
nlslatt: I'm not sure why `item` is being incremented here, or why it exists at all.

Issue #472, opened 2019-09-24
MPI_Recv -> MPI_Irecv


lifflander:
**What Needs to be Done?**

- [ ] First do this
- [ ] Then do that.
- [ ] Also this other thing.

**Is your feature request related to a problem? Please describe.**

**Describe potential solution outcome**

**Describe alternatives you've considered**

**Additional context**

lifflander:
Merged, closing

PR #473, opened 2019-09-25
472 Convert MPI_Recv to async MPI_Irecv

nlslatt:
#472 changes MPI_Recv to MPI_Irecv to mitigate excessively slow MPI_Recv under EMPIRE that appears to be waiting for the sender to finish unrelated work.

@lifflander @PhilMiller

### DIFF
--- a/src/vt/messaging/active.cc
+++ b/src/vt/messaging/active.cc
@@ -613,76 +613,102 @@ bool ActiveMessenger::tryProcessIncomingMessage() {
 
     NodeType const sender = stat.MPI_SOURCE;
 
-    MPI_Recv(
+    MPI_Request req;
+    MPI_Irecv(
       buf, num_probe_bytes, MPI_BYTE, sender, stat.MPI_TAG,
-      theContext()->getComm(), MPI_STATUS_IGNORE
+      theContext()->getComm(), &req
     );
 
-    auto msg = reinterpret_cast<MessageType>(buf);
-    messageConvertToShared(msg);
-    auto base = promoteMsgOwner(msg);
-
-    auto const is_term = envelopeIsTerm(msg->env);
-    auto const is_put = envelopeIsPut(msg->env);
-    bool put_finished = false;
-
-    if (!is_term || backend_check_enabled(print_term_msgs)) {
-      debug_print(
-        active, node,
-        "tryProcessIncoming: msg_size={}, sender={}, is_put={}, is_bcast={}, "
-        "handler={}\n",
-        num_probe_bytes, sender, print_bool(is_put),
-        print_bool(envelopeIsBcast(msg->env)), envelopeGetHandler(msg->env)
-      );
-    }
-
-    CountType msg_bytes = num_probe_bytes;
-
-    if (is_put) {
-      auto const put_tag = envelopeGetPutTag(msg->env);
-      if (put_tag == PutPackedTag) {
-        auto const put_size = envelopeGetPutSize(msg->env);
-        auto const msg_size = num_probe_bytes - put_size;
-        char* put_ptr = buf + msg_size;
-        msg_bytes = msg_size;
-
-        if (!is_term || backend_check_enabled(print_term_msgs)) {
-          debug_print(
-            active, node,
-            "tryProcessIncoming: packed put: ptr={}, msg_size={}, put_size={}\n",
-            put_ptr, msg_size, put_size
-          );
-        }
-
-        envelopeSetPutPtrOnly(msg->env, put_ptr);
-        put_finished = true;
-      } else {
-        /*bool const put_delivered = */recvDataMsg(
-          put_tag, sender,
-          [=](RDMA_GetType ptr, ActionType deleter){
-            envelopeSetPutPtr(base->env, std::get<0>(ptr), std::get<1>(ptr));
-            handleActiveMsg(base, sender, num_probe_bytes, true);
-          }
-        );
-      }
-    }
-
-    if (!is_put || put_finished) {
-      handleActiveMsg(base, sender, msg_bytes, true);
-    }
-
+    InProgressIRecv recv_holder{buf, num_probe_bytes, sender, req};
+    in_progress_irecv.emplace_back(std::move(recv_holder));
     return true;
   } else {
     return false;
   }
 }
 
+void ActiveMessenger::finishPendingAsyncRecv(InProgressIRecv irecv) {
+  auto buf = irecv.buf;
+  auto num_probe_bytes = irecv.probe_bytes;
+  auto sender = irecv.sender;
+
+  auto msg = reinterpret_cast<MessageType>(buf);
+  messageConvertToShared(msg);
+  auto base = promoteMsgOwner(msg);
+
+  auto const is_term = envelopeIsTerm(msg->env);
+  auto const is_put = envelopeIsPut(msg->env);
+  bool put_finished = false;
+
+  if (!is_term || backend_check_enabled(print_term_msgs)) {
+    debug_print(
+      active, node,
+      "tryProcessIncoming: msg_size={}, sender={}, is_put={}, is_bcast={}, "
+      "handler={}\n",
+      num_probe_bytes, sender, print_bool(is_put),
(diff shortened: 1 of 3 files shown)


nlslatt:
In it's current state, the code fails some tests of the form `DepSendChainInputExplodeDS/TestTermDepSendChain.test_term_dep_send_chain/DStrueK<N>MIGRATEtrue` with processor counts 4 and up on Kahuna.

src/vt/messaging/active.cc
```diff
@@ -613,76 +629,120 @@ bool ActiveMessenger::tryProcessIncomingMessage() {
...
+  if (!is_put || put_finished) {
+    handleActiveMsg(base, sender, msg_bytes, true);
+  }
+}
 
-    if (!is_put || put_finished) {
-      handleActiveMsg(base, sender, msg_bytes, true);
+bool ActiveMessenger::testPendingAsyncRecv() {
+  bool processed = false;
+  for (auto iter = in_progress_irecv.begin(); iter != in_progress_irecv.end(); ++iter) {
```
PhilMiller: Ideally, I think we'd use `MPI_Testsome` here, and loop over just the requests it says have completed.

nlslatt:
> In it's current state, the code fails some tests of the form `DepSendChainInputExplodeDS/TestTermDepSendChain.test_term_dep_send_chain/DStrueK<N>MIGRATEtrue` with processor counts 4 and up on Kahuna.

I'm starting to understand what this test is trying to accomplish.  The error indicates that `MyObjGroup::op3` sends its message to `MyCol::op3`, but it is not processed before the chainer triggers `MyObjGroup::op4`.  This appears to be a real bug.

The message to `MyCol::op3` includes a vector, which I'm guessing means that it is split into two parts, the initial active message and the data message containing the vector.  I suspect that, after the active message is received, we're not enforcing that the chainer wait for the data message to be received and trigger the `MyCol::op3` handler before permitting the chainer to proceed with the next send.

@lifflander and @PhilMiller, does this sound like a deficiency in termination detection or an issue with my implementation of Irecv handling?  Do I need to be producing work when I post the Irecv and consuming when MPI_Test indicates completion in order to hold back the chainer?

PhilMiller:
This sounds like a bug elsewhere. We should find and fix that bug. I believe adding a produce/consume pair here would reliably patch over it, though.

There shouldn't be something consuming on a partially-formed message, but I wouldn't be surprised if there is.

lifflander:
I believe that this is revealing an existing bug in termination detection. I am trying to find the bug.

src/vt/messaging/active.cc
nlslatt: @lifflander The tests pass for me now, but do we want to refactor following @PhilMiller's suggestion?

lifflander:
Yes, I will clean this up

src/vt/messaging/active.cc
lifflander: @PhilMiller If we were to use `MPI_Testsome` we would need to convert the `MPI_Request`s to a C-like array of them, due to MPI's interface. Thus, we would need to break apart the structure we have for holding the `MPI_Request` with meta-data (currently in a `std::list`> and manage the indexing across a new structure. Once priorities are implemented, we won't want to treat all requests as equal, and in fact will want to traverse them in a certain order.

src/vt/serialization/messaging/serialized_messenger.impl.h
```diff
@@ -136,8 +144,19 @@ template <typename UserMsgT>
...
+        theMsg()->pushEpoch(epoch);
+      }
       runnable::Runnable<UserMsgT>::run(handler, nullptr, msg.get(), node);
       action();
+
+      if (is_valid_epoch) {
+        theMsg()->popEpoch(epoch);
+      }
+
+      if (is_valid_epoch) {
```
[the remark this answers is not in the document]
lifflander: This is now moot, now that it's no longer part of this PR.

lifflander:
@PhilMiller Will you look over the changes I made and approve if you feel so inclined?

nlslatt:
The EMPIRE load-balanced tests pass using RotateLB with #508 and this PR together.

nlslatt:
They pass using HierarchicalLB as well.

PhilMiller:
More comments when I'm back at a computer tonight

src/vt/messaging/active.cc
```diff
@@ -441,6 +420,43 @@ bool ActiveMessenger::recvDataMsgBuffer(
...
+      #else
+        std::free(buf);
+      #endif
+    } else if (dealloc_user_buf != nullptr and user_buf != nullptr) {
+      dealloc_user_buf();
+    }
+  };
+
+  if (next != nullptr) {
+    next(RDMA_GetType{buf,num_probe_bytes}, [=]{
```
PhilMiller: Can't this just pass `dealloc_buf` directly as an argument, rather than wrapping the parameter-less lambda in another parameter-less lambda?

lifflander: Good, point

src/vt/messaging/active.cc
```diff
@@ -613,76 +629,120 @@ bool ActiveMessenger::tryProcessIncomingMessage() {
...
+  if (!is_put || put_finished) {
+    handleActiveMsg(base, sender, msg_bytes, true);
+  }
+}
 
-    if (!is_put || put_finished) {
-      handleActiveMsg(base, sender, msg_bytes, true);
+bool ActiveMessenger::testPendingActiveMsgAsyncRecv() {
+  bool processed = false;
+  for (auto iter = in_progress_active_msg_irecv.begin(); iter != in_progress_active_msg_irecv.end(); ++iter) {
```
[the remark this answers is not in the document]
lifflander: Ok, I've implemented this scheme.

src/vt/messaging/active.cc
```diff
@@ -613,76 +629,120 @@ bool ActiveMessenger::tryProcessIncomingMessage() {
...
+    int flag = 0;
+    MPI_Status stat;
+    MPI_Test(&(iter->req), &flag, &stat);
+    if (flag == 1) {
+      auto elm = std::move(*iter);
+      finishPendingActiveMsgAsyncRecv(std::move(elm));
```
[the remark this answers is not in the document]
lifflander: Addressed

src/vt/messaging/active.cc
```diff
@@ -441,6 +420,41 @@ bool ActiveMessenger::recvDataMsgBuffer(
...
+    }
+  };
+
+  if (next != nullptr) {
+    next(RDMA_GetType{buf,num_probe_bytes}, dealloc_buf);
+  } else {
+    dealloc_buf();
+  }
+
+  theTerm()->consume(term::any_epoch_sentinel,1,sender);
```
PhilMiller: I'm probably missing something, but where's the matching produce call?

lifflander: The matching produce is either in `ActiveMessenger::sendMsgBytes` or `ActiveMessenger::sendData` matching respectively with the consume in `ActiveMessenger::deliverActiveMsg` or `ActiveMessenger::finishPendingDataMsgAsyncRecv`

PhilMiller:
The only remaining performance-ish concern I'd have is with passing around `dealloc_user_buf` and `next` by value, rather than moved rvalue reference. If they're consistently just bare functions or a closure with some single stored variable, then it's no big deal.

lifflander:
> The only remaining performance-ish concern I'd have is with passing around `dealloc_user_buf` and `next` by value, rather than moved rvalue reference. If they're consistently just bare functions or a closure with some single stored variable, then it's no big deal.

The predominant use case (by far) is the serialized messenger, which uses a bare data msg to move the serialized contents when not in eager mode. It captures `[handler,recv_tag,node,epoch,is_valid_epoch]`, totaling ~`21` bytes or so: 64, 16, 16, 64, 1 (?).

Do you think that is large enough to be worth moving?

PhilMiller:
Not worth moving, though you may want to separately consider ordering those to avoid padding

lifflander:
Can either of you, @PhilMiller or @nlslatt, approve this if you feel some inclined so I can merge it?

PR #479, opened 2019-10-02
478 Fix tracing bugs in objgroup and colliding method names

lifflander:


src/vt/trace/trace_event.cc
```diff
@@ -49,10 +49,11 @@
...
 namespace vt { namespace trace {
 
-EventClass::EventClass(std::string const& in_event)
-  : event(in_event)
+EventClass::EventClass(
+  std::string const& in_event, std::string const& in_hash_event
+) : event(in_event), hash_event(in_hash_event)
 {
-  auto const& event_hash =  std::hash<std::string>{}(in_event);
+  auto const& event_hash =  std::hash<std::string>{}(in_hash_event);
```
nlslatt: I find it confusing that there are `hash_event` and `event_hash` here.  Is there a way to improve the naming?

lifflander: Fixed.

lifflander:
@nlslatt Please re-review. I've addressed the issue and rebased

PR #485, opened 2019-10-04
484 insertable collection referenced after destruction

nmm0:


lifflander:
Please conform your code to the VT style guidelines. I will make the change for this commit.

PR #468, opened 2019-09-19
457 gcc 9.x warnings and compiler error fixes

nmm0:
Added some commits for preventing warnings for GCC 9.x. Added a workaround for a GCC 9.X bug, and finally fixed an issue where epochs weren't being pushed when handling serialized messages

PR #496, opened 2019-10-08
457 autohandler gcc workaround 2

pnstickne:
Tagged commits, same change as https://github.com/DARMA-tasking/vt/pull/468 otherwise

### DIFF
--- a/cmake/link_vt.cmake
+++ b/cmake/link_vt.cmake
@@ -119,7 +119,7 @@ function(link_target_with_vt)
     if (${ARG_DEBUG_LINK})
       message(STATUS "link_target_with_vt: fmt=${ARG_LINK_FMT}")
     endif()
-    target_compile_definitions(${ARG_TARGET} PUBLIC FMT_HEADER_ONLY=1)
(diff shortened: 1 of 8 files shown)


nmm0:
> All these changes seem reasonable to fix the errors with gcc. But, this seems to break the build in somehow with `test_reduce_op`

I think this might be the no_epoch being pushed bug, I'll push a fix to this branch if so

nmm0:
@lifflander Some of these changes duplicate https://github.com/DARMA-tasking/vt/pull/497 i.e. we both fixed the same bug. Should I go ahead and remove them from this PR? Your fix avoids the no_epoch bug and also is in a contained PR for that issue

lifflander:
@nmm0 Can you rebase the termination fixes out of this PR? I made some other fixes to the test in the other PR so I'd rather keep that separate.

lifflander:
@nmm0 Also, please rebase onto develop so the status checks run.

nmm0:
The most recent version of this pull request is https://github.com/DARMA-tasking/vt/pull/502

PR #508, opened 2019-10-17
507 LB add barriers to synchronize the hierlb tree setup

lifflander:
Fixes a race condition in HierarchicalLB, occurring in EMPIRE runs on kahuna. For now, add barriers to synchronize the setup of the tree with the following use of it. In the future, this could become an asynchronous reduction.

### DIFF
--- a/src/vt/vrt/collection/balance/baselb/baselb.cc
+++ b/src/vt/vrt/collection/balance/baselb/baselb.cc
@@ -299,7 +299,6 @@ NodeType BaseLB::objGetNode(ObjIDType const id) const {
 }
 
 void BaseLB::finishedStats() {
-  theCollective()->barrier();
   this->runLB();
 }
 
--- a/src/vt/vrt/collection/balance/baselb/baselb.cc
+++ b/src/vt/vrt/collection/balance/baselb/baselb.cc
@@ -299,6 +299,7 @@ NodeType BaseLB::objGetNode(ObjIDType const id) const {
 }
 
 void BaseLB::finishedStats() {
+  theCollective()->barrier();
   this->runLB();
 }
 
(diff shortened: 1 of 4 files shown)


PhilMiller:
Do barrier calls inside VT execution mean that we'll get nested handlers, as other things arrive and make progress? Is this something that's done anywhere else in VT currently? If not, I'd lean toward going straight to the asynchronous reduction version, since this seems risky.

nlslatt:
With this diff dropped on top of EMPIRE's vt, I get genuine termination hangs with `HierarchicalLB`, `RotateLB`, and `GreedyLB`.

nlslatt:
> With this diff dropped on top of EMPIRE's vt, I get genuine termination hangs with `HierarchicalLB`, `RotateLB`, and `GreedyLB`.

To expand upon that, this was with `EMWaveInPlasma3DLoadBalanced-vt-parallel` and happened on the first call to LB.  For `HierarchicalLB`, I see both reductions complete on all ranks and all ranks call `setupTree`.  However, it appears that only rank 0 makes it out of the second barrier (when I comment out the node-conditional on the `loadStats` print, still only node 0 prints).  Then, node 2 prints some `lbTreeUp` stuff but no one else does (so, stuck in the barrier but still acting on incoming messages?).  I don't fully understand what's going on, but I'm concerned that @PhilMiller might be right about this approach being risky.

lifflander:
@PhilMiller Barriers call into the scheduler until all nodes reach. They can cause nesting. I'm willing to make this an asynchronous reduction instead. I think that would be better.

lifflander:
Async reduction is implemented, @nlslatt is testing on EMPIRE

Issue #517, opened 2019-10-25
extraneous commas


nlslatt:
**Describe the bug**

`INSTANTIATE_TYPED_TEST_CASE_P` macro calls have an extraneous comma at the end.  For example:
```
INSTANTIATE_TYPED_TEST_CASE_P(test_mem_serial, TestMemoryActive, MsgSerial, );
```

This macro is defined as below in the versions of gtest used by EMPIRE and vt-auto-build:
```
# define INSTANTIATE_TYPED_TEST_CASE_P(Prefix, CaseName, Types)
```

PhilMiller:
See #512 and its fix #513

nlslatt:
Actually, @pbmille, #513 is what caused my compile errors.

lifflander:
Oh, this is surprising. This was causing warnings in the version of gtest we were using. Maybe we should synchronize the versions of gtest so we don't have conflicts like this?

pnstickne:
Link - https://github.com/DARMA-tasking/vt/issues/465

nlslatt:
@lifflander, @pnstickne We decided a long time ago to have vt-auto-build fetch a specific version of gtest due to its ever-changing interface and the need to be compatible with EMPIRE.  We aren't the only ones using EMPIRE's gtest so we might be stuck with the ancient version.  We should document what version is needed so that it can be obtained without vt-auto-build.

PhilMiller:
Would the EMPIRE folks be opposed to that getting updated? Or is it just a matter of the effort involved?

nlslatt:
@PhilMiller I haven't asked.  Even if it was updated, it would still be a fixed version that would eventually become stale.  Is it worth it?

@bathmatt Is there a requirement that EMPIRE's gtest stay at it's current very old version, or is there the possibility of making it more current?  vt doesn't require a newer version, but is forced to require the same version as EMPIRE.

PR #513, opened 2019-10-23
512 Fix clang warning in tests due to gtest macro

lifflander:
This addresses all these warnings that have hung around for a long time:

```bash
/home/travis/build/DARMA-tasking/vt/tests/unit/memory/test_memory_active.cc:163:74: 
warning: ISO C++11 requires at least one argument for the "..." in a variadic macro INSTANTIATE_TYPED_TEST_CASE_P(test_mem_short,  TestMemoryActive, MsgShort);
```

PR #518, opened 2019-10-25
517 remove extraneous commas

nlslatt:
Issue #517. Removed the trailing comma in the argument list for `INSTANTIATE_TYPED_TEST_CASE_P`, which only takes 3 arguments.

### DIFF
--- a/tests/unit/collection/test_broadcast.cc
+++ b/tests/unit/collection/test_broadcast.cc
@@ -148,7 +148,7 @@ using CollectionTestTypes = testing::Types<
 >;
 
 INSTANTIATE_TYPED_TEST_CASE_P(
-  test_bcast, TestBroadcast, CollectionTestTypes,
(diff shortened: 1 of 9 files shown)


PhilMiller:
Jonathan just added those in #513

nlslatt:
Thanks, @pbmille.  My search attempts missed that one.

nlslatt:
Actually, @pbmille, #513 is what **caused** my compile errors.  The macro in question is defined as
```
# define INSTANTIATE_TYPED_TEST_CASE_P(Prefix, CaseName, Types)
```
in the versions of gtest used by EMPIRE and vt-auto-build.

lifflander:
@nlslatt I think we should merge this and start using the same version of gtest that EMPIRE uses for VT (if we care about the warnings). It's just a warning on VT's version of gtest, so I think we are fine. In the latest releases of gtest, this macro completely changes form (and has a updated name) so we eventually won't have this problem.

PR #528, opened 2019-10-29
527 Fix access to null pointer. Add minor clean-ups.

uhetmaniuk:
Access the static variable directly as opposed to using the pointer.

### DIFF
--- a/src/vt/topos/location/location.impl.h
+++ b/src/vt/topos/location/location.impl.h
@@ -88,7 +88,8 @@ EntityLocationCoord<EntityID>::EntityLocationCoord(LocInstType const identifier)
 }
 
 template <typename EntityID>
-/*virtual*/ EntityLocationCoord<EntityID>::~EntityLocationCoord() = default;
+/*virtual*/ EntityLocationCoord<EntityID>::~EntityLocationCoord() {
+}
 
 template <typename EntityID>
 void EntityLocationCoord<EntityID>::registerEntity(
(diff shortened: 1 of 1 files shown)


nlslatt:
Just FYI, I checked to see if this fixes the cuda bug, and it does not.

uhetmaniuk:
Thank you for checking @nlslatt

src/vt/topos/location/location.impl.h
```diff
@@ -88,8 +88,7 @@ EntityLocationCoord<EntityID>::EntityLocationCoord(LocInstType const identifier)
 }
 
 template <typename EntityID>
-/*virtual*/ EntityLocationCoord<EntityID>::~EntityLocationCoord() {
```
lifflander: @uhetmaniuk Just so I understand, is there a reason why the virtual destructor needs to be defaulted vs. empty body? These do have slightly different semantics.

src/vt/topos/location/location.impl.h
```diff
@@ -63,7 +63,7 @@ namespace vt { namespace location {
 
 template <typename EntityID>
 EntityLocationCoord<EntityID>::EntityLocationCoord()
-  : EntityLocationCoord<EntityID>(theLocMan()->cur_loc_inst++)
+  : EntityLocationCoord<EntityID>( LocationManager::cur_loc_inst++ )
```
lifflander: This looks right to me, `cur_loc_inst` is a static variable and should be referenced that way (especially if `theLocMan()` might not be initialized, due to component ordered initialization).

src/vt/topos/location/location.impl.h
uhetmaniuk: @lifflander I did not realize the slight difference. "clang-tidy" suggested the change and I agreed to it. I can revert to the original definition.

uhetmaniuk: Done (reverting to original destructor).

lifflander:
Great, thanks.

PR #531, opened 2019-11-04
521 eliminate T-lambda in Auto Registry

pnstickne:
- Replace the T-lambda with a strong type-erased proxy.
  This avoids signifcant T-cost.

  Uses unique_ptr because shared_ptr is also T-costly.

  (Working on find a middle-ground with C++ and type-erasure
   as well.. Base + Proxy + Impl seems okay, if not verbose.)


lifflander:
For tracking purposes, can we create issues to each of these PRs so we can track them individually? It's going to screw with the release tracker so we can determine which issues/PRs are going into each release.

lifflander:
Closed in place of #542, with a distinct issue to track this

Issue #521, opened 2019-10-28
Reduce lambda usage in high template-variance locations


pnstickne:
In clang (and possibly other compilers), a there will be a DIFFERENT instantiation of a lambda type per template parameters, even if the lambda itself if not dependent upon all the template parameters. In some cases the lambda can be trivially pulled into a less templated function (eg. return a std::function of the lambda) or the construct can be replaced with a different concept.

PR #533, opened 2019-11-04
[#521] remove tuple from debug::useVars

pnstickne:
- What's faster than instantiating a tuple?
  Not instantiating a tuple.

  Tuple instantiation is SLLLLLOW in comparison,
  especially over many more types.

  Compilers may (or may not) optimize out entire construct.


lifflander:
Closed in favor of #543; separate to track this change

PR #534, opened 2019-11-04
[#521] perf. colorization code and unify color selection

pnstickne:
- Rules for configuration are defined (AND followed..);
  logic happens immediately after parsing which ensures
  only one isatty check (ioctl) and centralized logic handling.

  Updated the setting/diagnostic messages to also align
  with the rules.

  Rules:
  - no-color DISABLES color
  - color ENABLES color
  - auto-color (assumed default) enables color in terminal

  (This makes VT consistent with tools like 'ls', or more so..)

- Removes unistd.h bleed - should be removed from other headers

- Consistently use defined colors and other minor formatting;
  swith to 'standard' functions.

### DIFF
--- a/src/vt/configs/debug/debug_colorize.h
+++ b/src/vt/configs/debug/debug_colorize.h
@@ -49,44 +49,36 @@
 #include "vt/configs/types/types_type.h"
 
 #include <string>
-#include <unistd.h>
-// #include <sys/stat.h>
 
 namespace vt { namespace debug {
 
-inline auto istty() -> bool {
-  return isatty(fileno(stdout)) ? true : false;
+inline bool colorizeOutput() {
+  return arguments::ArgConfig::colorize_output;
 }
 
-inline auto ttyc() -> bool {
-  auto nocolor = arguments::ArgConfig::vt_no_color ? false : true;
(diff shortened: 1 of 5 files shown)


src/vt/configs/debug/debug_colorize.h
```diff
@@ -49,44 +49,36 @@
...
 
-inline auto istty() -> bool {
-  return isatty(fileno(stdout)) ? true : false;
+inline bool colorizeOutput() {
+  return arguments::ArgConfig::colorize_output;
 }
 
-inline auto ttyc() -> bool {
-  auto nocolor = arguments::ArgConfig::vt_no_color ? false : true;
-  auto tty = arguments::ArgConfig::vt_auto_color ? istty() : nocolor;
```
pnstickne: Previously vt_auto_color would have precedence to vt_no_color. Now it's strictly; vt_no_color, vt_color, (default = auto)

pnstickne: Also does the isatty (ioctl involved) check ONCE, instead of once-per-usage :|

lifflander:
Closed in favor of #544; new issue for tracking this seperately

PR #535, opened 2019-11-04
[#521] perf. cleanup for demangler.cc

pnstickne:
- Remove <regex> dependency; since this is gone from
  the header as well, saves many hundreds of ms per
  affected TU.. like, literally, 200ms+ per TU :|

- Moved non-effective inline functions to cc.

- Removed assorted aliases and converted to more idiomatic C++.


lifflander:
Closed in favor of #545; new issue for tracking this separately

PR #536, opened 2019-11-04
[#521] move 'pretty pring message' to cc

pnstickne:
- Code does not have benefit being in header so..
  moved to compile-once-CC-file.

  And then FIXED a number of headers/code that broke
  due to having the transitive dependency spider web
  removed.

- Removed last gobal usage of unistd.h and cstdio
  from headers - should add wiki note about the 'allowed headers'
  that may be included in .h files.. so much global pollution.

- As added bonus, uses cached std::string hostname.

- Fix broken header includes exposed when headers stopped
  being included from 'pretty_print_message.h'

  Now 'types_headers.h' should be resolveable at the TOP of the header graph;
  updated bits_counter to accomodate.


lifflander:
Closed in favor of #546; creating separate issue for tracking this

PR #542, opened 2019-11-04
537 eliminate T-lambda in Auto Registry

lifflander:
- Replace the T-lambda with a strong type-erased proxy.
  This avoids signifcant T-cost.

  Uses unique_ptr because shared_ptr is also T-costly.

  (Working on find a middle-ground with C++ and type-erasure
   as well.. Base + Proxy + Impl seems okay, if not verbose.)

### DIFF
--- a/src/vt/registry/auto/auto_registry_common.h
+++ b/src/vt/registry/auto/auto_registry_common.h
@@ -104,12 +104,14 @@ struct RegistrarGenInfoBase {
 };
 
 struct RegistrarGenInfo : RegistrarGenInfoBase {
-  RegistrarGenInfo() {
+
+  /// Create a new object.
+  /// Takes complete ownership of the supplied object (pointer).
+  static RegistrarGenInfo takeOwnership(RegistrarGenInfoBase *owned_proxy) {
+    return RegistrarGenInfo(owned_proxy);
   }
 
-  explicit RegistrarGenInfo(RegistrarGenInfoBase* owned_proxy)
-    : proxy_(owned_proxy)
-  {
+  RegistrarGenInfo() {
   }
 
   // Using unique_ptr; can expand later.
@@ -122,6 +124,12 @@ struct RegistrarGenInfo : RegistrarGenInfoBase {
   virtual HandlerType getRegisteredIndex() override {
     return proxy_->getRegisteredIndex();
   }
+
+private:
+  explicit RegistrarGenInfo(RegistrarGenInfoBase* owned_proxy)
+    : proxy_(owned_proxy)  {
+  }
+
 private:
   std::unique_ptr<RegistrarGenInfoBase> proxy_ = nullptr;
 };
--- a/src/vt/registry/auto/auto_registry_common.h
+++ b/src/vt/registry/auto/auto_registry_common.h
@@ -76,7 +76,7 @@ using AutoActiveIndexType         = std::size_t;
 using AutoActiveObjGroupType      = objgroup::ActiveObjAnyType;
 
 using HandlerManagerType = vt::HandlerManager;
-using AutoHandlerType = int32_t;
+using AutoHandlerType = HandlerType;
 using NumArgsType = int16_t;
 
 enum struct RegistryTypeEnum {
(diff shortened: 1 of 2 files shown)


src/vt/registry/auto/auto_registry_general_impl.h
```diff
@@ -51,17 +51,24 @@
...
 RegistrarGen<ActFnT, RegT, InfoT, FnT>::RegistrarGen() {
   RegT& reg = getAutoRegistryGen<RegT>();
   index = reg.size();
 
-  auto fn = ActFnT::getFunction();
-  auto gen_fn = []() -> AutoHandlerType {
-    return objgroup::registry::makeObjIdx<typename ActFnT::ObjType>();
-  };
+  FnT fn = reinterpret_cast<FnT>(ActFnT::getFunction());
+  RegistrarGenInfo indexAccessor{new RegistrarGenInfoImpl<typename ActFnT::ObjType>()};
```
lifflander: This looks like it leaks memory (unless I'm missing something). Why are we using a pointer here at all? Use a r-val ref here?

Can we do this?
```C++
   explicit RegistrarGenInfo(RegistrarGenInfoBase&& owned_proxy)
     : proxy_(owned_proxy)
   {
   }

//.....

RegistrarGenInfo indexAccessor{RegistrarGenInfoImpl<typename ActFnT::ObjType>()};
```

PhilMiller: That'll give you a sliced object. Need to at least make the LHS a reference assigned from a temporary of the RHS type, to get lifetime extension and dynamic dispatch

src/vt/registry/auto/auto_registry_common.h
```diff
@@ -97,45 +97,77 @@ enum struct RegistryTypeEnum {
...
+  explicit RegistrarGenInfo(RegistrarGenInfoBase* owned_proxy)
+    : proxy_(owned_proxy)
+  {
+  }
+
+  // Using unique_ptr; can expand later.
+  RegistrarGenInfo(RegistrarGenInfo const& in) = delete;
+
+  RegistrarGenInfo(RegistrarGenInfo&& in) {
+    proxy_.swap(in.proxy_);
```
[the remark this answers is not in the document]
PhilMiller: (on my phone, so hard to look in detail)

If the objects in question own anything, the newer style requires that the ownership be expressed in the type of the underlying members, while the older style just requires a generally working destructor

pnstickne:
Added override and "factory" ctor to make it very clear that ownership of the object is assumed - a pointer is required moving across the boundary AFAIK.

It COULD also supply a unique_ptr, except that felt like pushing a detail down (eg. what if such changes?; or maybe a unique_ptr would just be more clear?)

Was not able to use the unique_ptr(..) in the initializer list successfully. I must be doing something incorrectly..

src/vt/registry/auto/auto_registry_general_impl.h
lifflander: Yes, my mistake. It should probably be a template.

```C++
  template <typename T>
   explicit RegistrarGenInfo(T&& owned_proxy)
     : proxy_(std::forward<T>(owned_proxy))
   {
   }

//.....

RegistrarGenInfo indexAccessor{RegistrarGenInfoImpl<typename ActFnT::ObjType>()};
```

src/vt/registry/auto/auto_registry_common.h
```diff
@@ -97,45 +97,77 @@ enum struct RegistryTypeEnum {
...
+  }
+
+  virtual HandlerType getRegisteredIndex() {
```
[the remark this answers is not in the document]
pnstickne: Done.

src/vt/registry/auto/auto_registry_general_impl.h
```diff
@@ -51,17 +51,24 @@
 #include "vt/registry/auto/auto_registry_general.h"
 #include "vt/objgroup/type_registry/registry.h"
 
+#include <functional>
+
 namespace vt { namespace auto_registry {
 
+template <typename RegObjTypeT>
+struct RegistrarGenInfoImpl : RegistrarGenInfoBase {
+  virtual HandlerType getRegisteredIndex() {
```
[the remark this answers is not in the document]
pnstickne: Done.

src/vt/registry/auto/auto_registry_general_impl.h
```diff
@@ -51,17 +51,25 @@
 #include "vt/registry/auto/auto_registry_general.h"
 #include "vt/objgroup/type_registry/registry.h"
 
+#include <functional>
```
pnstickne: Should probably be able to remove this this..

pnstickne:
@lifflander Ready to go!

PR #543, opened 2019-11-04
538 remove tuple from debug::useVars

lifflander:
- What's faster than instantiating a tuple?
  Not instantiating a tuple.

  Tuple instantiation is SLLLLLOW in comparison,
  especially over many more types.

  Compilers may (or may not) optimize out entire construct.

Fixes #583

### DIFF
--- a/src/vt/configs/debug/debug_var_unused.h
+++ b/src/vt/configs/debug/debug_var_unused.h
@@ -52,9 +52,21 @@
 
 namespace vt { namespace debug {
 
+template <typename T>
+int useVarsDummy(T t) {
+  return (void)(t), 0;
+}
+
+/// Attempt to ensure the argument (pack) is used.
+/// This may be optimized away by compilers. C++17 has cleaner methods.
+/// ref. https://stackoverflow.com/questions/25680461/variadic-template-pack-expansion/25683817
 template <typename... Args>
 void useVars(Args&&... args) {
-  (void)(std::make_tuple(std::forward<Args>(args)...));
+  using expander = int[];
+  (void)(expander {
+    0,
+    useVarsDummy(std::forward<Args>(args)) ...
+  });
 }
 
 }} /* end namespace vt::debug */


lifflander:
I think this an improvement, but we need to test this more extensively on ICC and CUDA to see if passes. Thus, I'm not marking this for `1.0.0-beta.4`

src/vt/configs/debug/debug_var_unused.h
```diff
@@ -52,9 +52,21 @@
...
+template <typename T>
+int useVarsDummy(T t) {
+  return (void)(t), 0;
+}
+
+/// Attempt to ensure the argument (pack) is used.
+/// This may be optimized away by compilers. C++17 has cleaner methods.
+/// ref. https://stackoverflow.com/questions/25680461/variadic-template-pack-expansion/25683817
 template <typename... Args>
 void useVars(Args&&... args) {
```
PhilMiller: We may have been over this at some point, but what happens to the warnings if you simply don't give the pack a name `args` at all? That's another common unambiguous way of expressing an unused parameter.

Regardless, if this improves build times, I'm in favor, but I would like to see testing on ICC and Cuda's NVCC, and hence suitability for the next stable tag.

lifflander: Regarding just not referencing `args`, the problem is a bit deeper than that. In these contexts, like:
```
int a = 10;
debug_print(gen, node, "a={}\n", a);
```

If the variable `a` is not explicitly used (read), as the code does, the compiler will complain. So, I believe we need the full expansion.

PhilMiller: It still complains if we define

```
template <typename... Args>
void useVars(Args&&...) { }
```

If so, ugh, but this will have my +1

lifflander: @pnstickne Can you follow up on this?

lifflander:
@PhilMiller @nlslatt Can I get approval to merge this? I tagged it for 1.0.0-beta.4 since it's working with icc.

nlslatt:
This appears to build with cuda as well.

PR #548, opened 2019-11-05
500 Non-heuristic termination hang detector, DOT epoch graphs, performance bugs

lifflander:
This PR provides:
- A non-heuristic hang detector, only detects a hang iff it's impossible for the system to make progress
  - Relies on work priorities, which provides a central queue
  - Term messages are marked as such, and tracked by the schduler
  - Two forms of "idleness" (term and non-term)
- Fixes a major performance bug in collective epochs
  - Optimizes the structure of epoch graphs
- If a hang is *detected*, outputs a DOT file of the hang


lifflander:
Not picking up right commits..closing

PR #549, opened 2019-11-05
500 Non-heuristic termination hang detector, DOT epoch graphs, performance bugs

### LINKED ISSUE #500: Implement non-heuristic-based termination hang detection
**What Needs to be Done?**

- [ ] Add check for queues empty (might want to build off of #474)
- [ ] When many traverses reveal the same counter, do a reduction to check for empty queues also.
- [ ] Reveal the live epochs related to the hang

We recently had a problem where the current hang detection, which is a heuristic, caused us to believe a bug existed that didn't on a very slow over-subscribed machine.

**Additional context**

lifflander:
This PR provides:
- A non-heuristic hang detector, only detects a hang iff it's impossible for the system to make progress
  - Relies on work priorities, which provides a central queue
  - Term messages are marked as such, and tracked by the schduler
  - Two forms of "idleness" (term and non-term)
- Halt all progress on global epoch unless schedule is idle (minus term messages)
- Fixes a major performance bug in collective epochs
  - Create a new API call for creating epoch dependencies (instead of simply a produce/consume), which carry the semantic of the dependency for optimization purposes
  - Do not forward progress epochs that cannot finish due to a dependency
  - Optimizes the structure of epoch graphs, eliding redundant dependencies
- If a hang is *detected*, abort and output a DOT file of the hang
  - Local view of epochs for each node
  - Global view combining the all the local epoch graphs

Fixes #500

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -46,7 +46,7 @@ set(
     termination
       termination/dijkstra-scholten
       termination/interval
-      termination/tree
+      termination/graph
     messaging/envelope messaging/message
     pool/static_sized pool/header
     rdma/channel rdma/collection rdma/group rdma/state
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -43,7 +43,10 @@ set(
       pipe/callback/objgroup_bcast
       pipe/callback/anon
       pipe/callback/cb_union
-    termination/dijkstra-scholten
+    termination
+      termination/dijkstra-scholten
+      termination/interval
+      termination/tree
     messaging/envelope messaging/message
     pool/static_sized pool/header
     rdma/channel rdma/collection rdma/group rdma/state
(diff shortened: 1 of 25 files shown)


lifflander:
Here's an example dot file:
Global view: [outfile.pdf](https://github.com/DARMA-tasking/vt/files/3807568/outfile.pdf)
Local view (e.g, node 1): 
[outfile-1.pdf](https://github.com/DARMA-tasking/vt/files/3807571/outfile-1.pdf)

PhilMiller:
Given the potential to output `dot` graphs of epochs on hangs, might it be worthwhile to associate labels with them as they're created? This would probably be useful in other contexts as well, where we might do timing or other automated analysis, and want to produce more meaningful, easily-interpreted output.

lifflander:
@PhilMiller I completely agree and was considering exactly the same idea. We could add a label to the `nextStep`/`nextStepCollective`.

PR #550, opened 2019-11-05
521 shared msg lambda elim

pnstickne:


lifflander:
@pnstickne Random question, how do you manage to create PRs in "draft" mode. I can't find the way to do that!

pnstickne:
@lifflander Click the little "down arrow" in the right of the PR button - it's a context drop down. Not the most clear.

src/vt/messaging/message/smart_ptr.h
```diff
@@ -48,205 +48,262 @@
...
+  explicit operator T*() const { return get(); }
+
+  /// Access as another (usually base) message type.
   template <typename U>
   MsgSharedPtr<U> to() const {
-    static_assert(
-      std::is_trivially_destructible<T>(),
-      "Message shall not be downcast unless trivially destructible"
-    );
+    // Establish type-erased deref if needed.
```
pnstickne: Could just 'always' pass/down use the deref proxy, especially since such are created with global lifetimes.

src/vt/messaging/message/smart_ptr.h
```diff
@@ -48,205 +48,262 @@
...
+    if (std::is_trivially_destructible<T>() or not shared_) {
+      assert("should not have deref" && not deref_);
+      return MsgSharedPtr<U>(
+        reinterpret_cast<U*>(ptr_), true);
+    }
 
-    return MsgSharedPtr<U>{*this};
+    return MsgSharedPtr<U>(
+      reinterpret_cast<U*>(ptr_), true,
+      deref_ ? deref_ : &statics::StaticMsgDerefs<T>);
```
pnstickne: Basically zero cost here to "create" the proxy.. and if counting CPU cycles, more branches might add up to more.

src/vt/messaging/pending_send.cc
```diff
@@ -49,7 +49,7 @@ namespace vt { namespace messaging {
 
 void PendingSend::sendMsg() {
   if (send_action_ == nullptr) {
-    theMsg()->sendMsgSized(msg_.getShared(), msg_size_);
+    theMsg()->sendMsgSized(msg_, msg_size_);
```
pnstickne: Uh oh. "Feels" like there should be a `to` or the deref proxy should be universally consistent.

pnstickne: Anyway, `to` will have the same effect as `getShared` for a non-trivially-destructible type.

pnstickne:
@lifflander https://github.blog/2019-02-14-introducing-draft-pull-requests/

Also, ready to go in!

src/vt/messaging/message/smart_ptr.h
```diff
@@ -48,205 +48,262 @@
...
   template <typename U>
   MsgSharedPtr<U> to() const {
-    static_assert(
-      std::is_trivially_destructible<T>(),
-      "Message shall not be downcast unless trivially destructible"
-    );
+    // Establish type-erased deref if needed.
+    // Non-shared message are not deref'ed and trivially destructible types
+    // do not really care on which type the delete-expr is invoked.
+    if (std::is_trivially_destructible<T>() or not shared_) {
```
lifflander: Is it ok that this is no longer in a `static_assert`". Does this imply virtualization in byte-copied messages?

src/vt/messaging/message/smart_ptr.h
```diff
@@ -48,205 +48,262 @@
...
+namespace vt { namespace messaging {
+
+struct MsgDerefBase {
+  // Invoke messageDeref on the appropriate type.
+  virtual void messageDeref(void* msg_ptr) = 0;
+  virtual ~MsgDerefBase() {}
+};
+
+template <typename MsgT>
+struct MsgDerefTyped : MsgDerefBase {
```
lifflander: Ah, I see what you're doing here. What about destructors in `MsgSharedPtr<T>?

src/vt/messaging/message/smart_ptr.h
pnstickne: I believe it's OK. The static assert was to prevent certain types from using `to` if it was not the case that such could be guaranteed with MsgSharedPtr due to requiring deref on the appropriate type. With the changes `to` *subsumes* the roll of `virtual`; it will ensure that it will propagate an appropriate deref IFF needed (it would actual be benign to propagate it all cases).

pnstickne: The original message itself is not affected.

src/vt/messaging/message/smart_ptr.h
pnstickne: The destructor will invoke the `derefMessage` of the specific T-implementation (per virtual method), as established on the *original* `MsgSharedPtr<T>` (and propagated through `to`/`virtual`).

pnstickne:
> If you're open to rebasing before integration, there's a typo in the commit message: `toVritual`

I've no problem with such - although I've been scolded before. Amended the commit message for the typo (and improved the explanation in general).

pnstickne:
Now the message-deref will always be done on the original type, regardless of if a type is trivially destructible or not (so there is no special case at all).

Simplifies some checks (replaces one or more conditionals with a single virtual dispatch) and promotes generalized type-erasure support.

Only results in more t-instantiations for message types that do not use `to<T>` or `toVirtual<T>`.

src/vt/messaging/message/smart_ptr.h
lifflander: Ok, I think you are right. I was thinking of another instance of `std::is_trivially_destructible<T>()`, I believe.

lifflander:
@pnstickne Did you try running valgrind after these changes on some examples or tests to make sure that we are not leaking memory as a check?

pnstickne:
> @pnstickne Did you try running valgrind after these changes on some examples or tests to make sure that we are not leaking memory as a check?

Nope. I probably should get in the habit :|
That said, all new allocations are to static space.

Issue #527, opened 2019-10-29
Null Pointer Access of type 'location::LocationManager'


uhetmaniuk:
**Describe the bug**
/Users/ulrich/Dropbox/NGA/vt-examples/vt/src/vt/topos/location/location.impl.h:67:18: runtime error: member access within null pointer of type 'location::LocationManager'

**To Reproduce**
1. Build with trace_enabled

2. Compilation:
platform = macOS Catalina (but happened on macOS Sierra)
compiler = clang 3.9 
flag = "-g -fsanitize=undefined"

2. Run:
./lb_iter.exe --vt_trace

ulrich@Ulrichs-MacBook-Pro vt-examples % ./lb_iter.exe --vt_trace
/Users/ulrich/Dropbox/NGA/vt-examples/vt/src/vt/topos/location/location.impl.h:67:18: runtime error: member access within null pointer of type 'location::LocationManager'
vt: Caught SIGSEGV signal: 11 
0    18   0x10e54154c     vt::debug::stack::dumpStack(int) + 124
1    18   0x10e56977a     vt::runtime::Runtime::sigHandler(int) + 330
2    18   0x7fff6a4a6b1d  _sigtramp + 29
         3 18 0x123094268 3   ???                                 0x0000000123094268 0x0 + 4882776680
4    18   0x10e8200f7     vt::location::LocationManager::LocationManager() + 183
5    18   0x10e820032     vt::location::LocationManager::LocationManager() + 98
6    18   0x10e6070be     vt::runtime::Runtime::initializeComponents() + 140046
7    18   0x10e58b0dc     vt::runtime::Runtime::initialize(bool) + 1484
8    18   0x10e58a671     vt::runtime::Runtime::tryInitialize() + 2337
9    18   0x10e58bcb0     vt::runtime::Runtime::initialize(bool) + 4512
10   18   0x1074b0e99     vt::CollectiveAnyOps<(vt::runtime::eRuntimeInstance)0>::initialize(int&, char**&, short, bool, ompi_communicator_t**) + 13545
11   18   0x104a3f706     main + 246
12   18   0x7fff6a2a5405  start + 1
        13 18 0x2 13  ???                                 0x0000000000000002 0x0 + 2

lifflander:
Merged, closing per #528

PR #551, opened 2019-11-05
241 flush trace

### LINKED ISSUE #241: Flush traces every X steps

uhetmaniuk:
Moved changes for flushing the traces ( Issue [#241](https://github.com/DARMA-tasking/vt/issues/241) ) to avoid formatting issues.

Fixes #241

### DIFF
--- a/src/vt/collective/collective_ops.cc
+++ b/src/vt/collective/collective_ops.cc
@@ -51,7 +51,6 @@
 #include <memory>
 #include <cstdlib>
 #include <mpi.h>
-#include <zlib.h>
 
 namespace vt {
 
@@ -140,7 +139,7 @@ void CollectiveAnyOps<instance>::abort(
   if (myrt) {
 #if backend_check_enabled(trace_enabled)
     //--- Try to flush most of the traces before aborting
-    myrt->theTrace->writeTracesFile(Z_FINISH);
+    myrt->theTrace->cleanupTracesFile();
 #endif
     myrt->abort(str, code);
   } else {
@@ -156,10 +155,6 @@ void CollectiveAnyOps<instance>::output(
   auto tls_rt = curRT;
   auto myrt = tls_rt ? tls_rt : ::vt::rt;
   if (myrt) {
-#if backend_check_enabled(trace_enabled)
-    //--- Try to flush most of the traces
-    myrt->theTrace->writeTracesFile(Z_FULL_FLUSH);
-#endif
     myrt->output(str,code,error,decorate,formatted);
   } else {
     ::fmt::print(str.c_str());
--- a/src/vt/collective/collective_ops.cc
+++ b/src/vt/collective/collective_ops.cc
@@ -137,6 +137,8 @@ void CollectiveAnyOps<instance>::abort(
   auto tls_rt = curRT;
   auto myrt = tls_rt ? tls_rt : ::vt::rt;
   if (myrt) {
+    //--- Try to flush most of the traces before aborting
+    myrt->theTrace->writeTracesFile(Z_FINISH);
     myrt->abort(str, code);
   } else {
     std::_Exit(code);
@@ -151,6 +153,8 @@ void CollectiveAnyOps<instance>::output(
   auto tls_rt = curRT;
   auto myrt = tls_rt ? tls_rt : ::vt::rt;
   if (myrt) {
+    //--- Try to flush most of the traces
+    myrt->theTrace->writeTracesFile(Z_FULL_FLUSH);
     myrt->output(str,code,error,decorate,formatted);
   } else {
     ::fmt::print(str.c_str());
--- a/src/vt/collective/collective_ops.cc
+++ b/src/vt/collective/collective_ops.cc
@@ -137,8 +137,10 @@ void CollectiveAnyOps<instance>::abort(
   auto tls_rt = curRT;
   auto myrt = tls_rt ? tls_rt : ::vt::rt;
   if (myrt) {
+#if backend_check_enabled(trace_enabled)
     //--- Try to flush most of the traces before aborting
     myrt->theTrace->writeTracesFile(Z_FINISH);
+#endif
     myrt->abort(str, code);
   } else {
     std::_Exit(code);
@@ -153,8 +155,10 @@ void CollectiveAnyOps<instance>::output(
   auto tls_rt = curRT;
   auto myrt = tls_rt ? tls_rt : ::vt::rt;
   if (myrt) {
+#if backend_check_enabled(trace_enabled)
     //--- Try to flush most of the traces
     myrt->theTrace->writeTracesFile(Z_FULL_FLUSH);
+#endif
     myrt->output(str,code,error,decorate,formatted);
   } else {
     ::fmt::print(str.c_str());
--- a/src/vt/collective/collective_ops.cc
+++ b/src/vt/collective/collective_ops.cc
@@ -51,6 +51,7 @@
 #include <memory>
 #include <cstdlib>
 #include <mpi.h>
+#include <zlib.h>
 
 namespace vt {
 
--- a/src/vt/topos/location/location.impl.h
+++ b/src/vt/topos/location/location.impl.h
@@ -63,7 +63,7 @@ namespace vt { namespace location {
 
 template <typename EntityID>
 EntityLocationCoord<EntityID>::EntityLocationCoord()
-  : EntityLocationCoord<EntityID>(theLocMan()->cur_loc_inst++)
+  : EntityLocationCoord<EntityID>( LocationManager::cur_loc_inst++ )
 { }
 
 template <typename EntityID>
@@ -74,7 +74,7 @@ EntityLocationCoord<EntityID>::EntityLocationCoord(
 
 template <typename EntityID>
 EntityLocationCoord<EntityID>::EntityLocationCoord(LocInstType const identifier)
-  : this_inst(identifier), recs_(default_max_cache_size)
+  : LocationCoord(), this_inst(identifier), recs_(default_max_cache_size)
 {
   debug_print(
     location, node,
--- a/src/vt/trace/trace.cc
+++ b/src/vt/trace/trace.cc
@@ -134,7 +134,7 @@ void Trace::setupNames(
 }
 
 /*virtual*/ Trace::~Trace() {
-  writeTracesFile();
+  cleanupTracesFile();
 }
 
 void Trace::addUserNote(std::string const& note) {
@@ -359,6 +359,10 @@ void Trace::endProcessing(
   log->idx4 = idx4;
 
   logEvent(log);
+
+  if (traces_.size() > 100) {
+    writeTracesFile();
+  }
 }
 
 void Trace::beginIdle(double const time) {
@@ -596,6 +600,15 @@ void Trace::disableTracing() {
   enabled_ = false;
 }
 
+void Trace::cleanupTracesFile() {
+  if (checkEnabled()) {
+    auto const& node = theContext()->getNode();
+    writeTracesFile();
+    outputFooter(node, start_time_, log_file);
+    gzclose(log_file);
+  }
+}
+
 void Trace::writeTracesFile() {
   auto const node = theContext()->getNode();
 
@@ -610,20 +623,23 @@ void Trace::writeTracesFile() {
 
   if (checkEnabled()) {
     auto path = full_trace_name;
-    gzFile file = gzopen(path.c_str(), "wb");
-    outputHeader(node, start_time_, file);
-    writeLogFile(file, traces_);
-    outputFooter(node, start_time_, file);
-    gzclose(file);
+    if (not file_is_open) {
+      log_file = gzopen(path.c_str(), "wb");
+      outputHeader(node, start_time_, log_file);
+      file_is_open = true;
+    }
+    writeLogFile(log_file, traces_);
+    gzflush(log_file, Z_SYNC_FLUSH);
   }
 
-  if (node == designated_root_node) {
+  if (node == designated_root_node and not wrote_sts_file) {
     std::ofstream file;
     auto name = full_sts_name;
     file.open(name);
     outputControlFile(file);
     file.flush();
     file.close();
+    wrote_sts_file = true;
   }
 }
 
--- a/src/vt/trace/trace.cc
+++ b/src/vt/trace/trace.cc
@@ -126,12 +126,12 @@ void Trace::setupNames(
   auto const prog_name = pc[pc.size()-1];
 
   auto const node_str = "." + std::to_string(node) + ".log.gz";
-  if (ArgType::vt_trace_file != "") {
-    full_trace_name_ = full_dir_name_ + ArgType::vt_trace_file + node_str;
-    full_sts_name_   = full_dir_name_ + ArgType::vt_trace_file + ".sts";
-  } else {
+  if (ArgType::vt_trace_file.empty()) {
     full_trace_name_ = full_dir_name_ + trace_name;
     full_sts_name_   = full_dir_name_ + prog_name + ".sts";
+  } else {
+    full_trace_name_ = full_dir_name_ + ArgType::vt_trace_file + node_str;
+    full_sts_name_   = full_dir_name_ + ArgType::vt_trace_file + ".sts";
   }
 }
 
@@ -437,11 +437,12 @@ void Trace::editLastEntry(std::function<void(LogPtrType)> fn) {
   if (not enabled_ || not traceWritingEnabled(theContext()->getNode())) {
     return;
   }
-
-  auto const trace_cont_size = traces_.size();
-  if (trace_cont_size > 0) {
-    fn(traces_.at(trace_cont_size - 1));
+  if (traces_.empty()) {
+    return;
   }
+  //---
+  auto const trace_cont_size = traces_.size();
+  fn(traces_.at(trace_cont_size - 1));
 }
 
 TraceEventIDType Trace::logEvent(LogPtrType log) {
@@ -578,7 +579,7 @@ TraceEventIDType Trace::logEvent(LogPtrType log) {
   }
 }
 
-bool Trace::traceWritingEnabled(const NodeType node) {
+/*static*/ bool Trace::traceWritingEnabled(NodeType node) {
   if (ArgType::vt_trace) {
     return ((ArgType::vt_trace_mod == 0)
             or (node % ArgType::vt_trace_mod == 1));
@@ -587,7 +588,6 @@ bool Trace::traceWritingEnabled(const NodeType node) {
     return false;
   }
 }
-
 
 void Trace::enableTracing() {
   enabled_ = true;
@@ -599,15 +599,16 @@ void Trace::disableTracing() {
 
 void Trace::cleanupTracesFile() {
   auto const& node = theContext()->getNode();
-  if (traceWritingEnabled(node)) {
-    //--- Sanity check
-    vtAssert(open_events_.empty(), "Trying to dump traces with open events?");
-    cur_stop_ = traces_.size();
-    //--- Dump everything into an output file
-    writeTracesFile();
-    outputFooter(node, start_time_, log_file_);
-    gzclose(log_file_);
+  if (!traceWritingEnabled(node)) {
+    return;
   }
+  //--- Sanity check
+  vtAssert(open_events_.empty(), "Trying to dump traces with open events?");
+  cur_stop_ = traces_.size();
+  //--- Dump everything into an output file
+  writeTracesFile();
+  outputFooter(node, start_time_, log_file_);
+  gzclose(log_file_);
 }
 
 void Trace::flushTracesFile(bool useGlobalSync) {
--- a/src/vt/trace/trace.cc
+++ b/src/vt/trace/trace.cc
@@ -49,8 +49,10 @@
 #include "vt/trace/trace.h"
 #include "vt/utils/demangle/demangle.h"
 
-#include <fstream>
 #include <cinttypes>
+#include <fstream>
+#include <iostream>
+#include <map>
 #include <sys/stat.h>
 #include <unistd.h>
 
@@ -67,6 +69,13 @@ struct vt_gzFile {
 };
 
 using ArgType = vt::arguments::ArgConfig;
+
+template <typename EventT>
+struct TraceEventSeqCompare {
+  bool operator()(EventT* const a, EventT* const b) const {
+    return a->theEventSeq() < b->theEventSeq();
+  }
+};
 
 Trace::Trace(std::string const& in_prog_name, std::string const& in_trace_name)
   : prog_name_(in_prog_name), trace_name_(in_trace_name),
@@ -463,6 +472,12 @@ TraceEventIDType Trace::logEvent(LogPtrType log) {
     return 0;
   }
 
+  vtAssert(
+   log->ep == no_trace_entry_id
+   or TraceRegistry::getEvent(log->ep).theEventId() not_eq no_trace_entry_id,
+    "Event must exist that was logged"
+  );
+
   // close any idle event as soon as we encounter any other type of event
   if (idle_begun_ and
       log->type != TraceConstantsType::BeginIdle and
@@ -472,18 +487,19 @@ TraceEventIDType Trace::logEvent(LogPtrType log) {
 
   auto grouped_begin = [&]() -> TraceEventIDType {
     if (not open_events_.empty()) {
+      LogType* top_event = open_events_.top();
       traces_.push_back(
         new LogType(
           log->time,
-          open_events_.top()->ep,
+          top_event->ep,
           TraceConstantsType::EndProcessing,
-          open_events_.top()->event,
-          open_events_.top()->msg_len,
-          open_events_.top()->node,
-          open_events_.top()->idx1,
-          open_events_.top()->idx2,
-          open_events_.top()->idx3,
-          open_events_.top()->idx4
+          top_event->event,
+          top_event->msg_len,
+          top_event->node,
+          top_event->idx1,
+          top_event->idx2,
+          top_event->idx3,
+          top_event->idx4
         )
       );
     }
@@ -515,18 +531,19 @@ TraceEventIDType Trace::logEvent(LogPtrType log) {
     traces_.push_back(log);
 
     if (not open_events_.empty()) {
+      LogType* top_event = open_events_.top();
       traces_.push_back(
         new LogType(
           log->time,
-          open_events_.top()->ep,
+          top_event->ep,
           TraceConstantsType::BeginProcessing,
-          open_events_.top()->event,
-          open_events_.top()->msg_len,
-          open_events_.top()->node,
-          open_events_.top()->idx1,
-          open_events_.top()->idx2,
-          open_events_.top()->idx3,
-          open_events_.top()->idx4
+          top_event->event,
+          top_event->msg_len,
+          top_event->node,
+          top_event->idx1,
+          top_event->idx2,
+          top_event->idx3,
+          top_event->idx4
         )
       );
     }
@@ -642,8 +659,8 @@ void Trace::writeTracesFile(int flush) {
     "write_traces_file: traces.size={}, "
     "event_type_container.size={}, event_container.size={}\n",
     traces_.size(),
-    TraceContainersType::event_type_container.size(),
-    TraceContainersType::event_container.size()
+    TraceContainersType::getEventTypeContainer()->size(),
+    TraceContainersType::getEventContainer()->size()
   );
 
   if (traceWritingEnabled(theContext()->getNode())) {
@@ -676,7 +693,6 @@ void Trace::writeLogFile(vt_gzFile *file_, TraceContainerType const& traces) {
   for (size_t i = cur_; i < stop_point; i++) {
     auto& log = traces[i];
     auto const& converted_time = timeToInt(log->time - start_time_);
-
     auto const type = static_cast<
       std::underlying_type<decltype(log->type)>::type
     >(log->type);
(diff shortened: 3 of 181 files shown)


src/vt/topos/location/location.impl.h
```diff
@@ -63,7 +63,7 @@ namespace vt { namespace location {
 
 template <typename EntityID>
 EntityLocationCoord<EntityID>::EntityLocationCoord()
-  : EntityLocationCoord<EntityID>(theLocMan()->cur_loc_inst++)
+  : EntityLocationCoord<EntityID>( LocationManager::cur_loc_inst++ )
```
lifflander: This just got merged to develop, can you rebase? Sorry for the moving target!

src/vt/trace/trace.cc
```diff
@@ -42,29 +42,32 @@
...
 #include "vt/trace/trace.h"
-#include "vt/timing/timing.h"
+#include "vt/config.h"
+#include "vt/configs/arguments/args.h"
+#include "vt/runtime/runtime_inst.h"
 #include "vt/scheduler/scheduler.h"
+#include "vt/timing/timing.h"
 #include "vt/utils/demangle/demangle.h"
 
-#include <zlib.h>
```
lifflander: I noticed this include got moved to the header file? What is the reason for that? Can we keep this in the cc file to reduce compilation times?

src/vt/vrt/collection/manager.h
```diff
@@ -870,12 +870,6 @@ namespace details
 
 }}} /* end namespace vt::vrt::collection */
 
-namespace vt {
```
lifflander: Is this forward declaration not needed? If so, I'm cool with the elision.

src/vt/trace/trace.cc
lifflander: Ah, now I see why it got moved...`gzFile` is now in the header. Makes sense.

src/vt/vrt/collection/manager.h
uhetmaniuk: It is already defined in "manager.fwd.h" which get included also here. I did not check what "manager.fwd.h" brings to "manager.h" (besides that forward declaration).

lifflander: Ah cool!

src/vt/topos/location/location.impl.h
uhetmaniuk: @lifflander Done.

lifflander:
@PhilMiller @nlslatt @pnstickne Can I get someone to review this PR?

lifflander:
The commits on this branch were a mess. Commits applied multiple times, merges, etc. I squashed and rebased onto develop.

lifflander:
@uhetmaniuk After re-basing I tried to see if this works if the program crashes. I have the `--vt_trace_mod=1` and LB enabled. However, it's not flushing out the file when I kill the program early. Is there something else I need to do?

uhetmaniuk:
@lifflander Did you enable trace_enabled ? If so, did you get the error message "Trying to dump traces with open events?" ?

lifflander:
Yes, I had `trace_enabled` in cmake. Didn't get any error message.

uhetmaniuk:
@lifflander Could you post the command you ran? 
- I just ran directly from the branch, "mpirun -np 3 ./lb_iter.exe 96 1 16 --vt_trace --vt_lb --vt_trace_flush_mod=1" and the files outputted in "./lb_iter.exe_trace". When I "ctr-c"-ed and exited, I had to use zless to read the "gzip" files (gunzip did not want to uncompress them).

uhetmaniuk:
@lifflander @PhilMiller 
Regarding the flushing of traces when aborting, I am thinking of flushing whatever was  recorded in "vt::CollectiveAnyOps<instance>::abort" and in "vt::CollectiveAnyOps<instance>::output". Are these functions "safe" place for flushing the records?

src/vt/trace/trace.cc
```diff
@@ -42,27 +42,31 @@
...
+
 Trace::Trace(std::string const& in_prog_name, std::string const& in_trace_name)
   : prog_name_(in_prog_name), trace_name_(in_trace_name),
-    start_time_(getCurrentTime())
+    start_time_(getCurrentTime()),
+    log_file_()
 { }
 
-Trace::Trace() { }
+Trace::Trace() : log_file_() { }
```
pnstickne: If setting a default member value there is no need to assign a default in the list initialization.

src/vt/trace/trace.cc
```diff
@@ -102,20 +106,20 @@ void Trace::setupNames(
...
+    full_dir_name_ = ArgType::vt_trace_dir;
   }
   else {
-    full_dir_name = std::string(cur_dir) + "/" + dir_name;
+    full_dir_name_ = std::string(cur_dir) + "/" + dir_name;
   }
 
-  if (full_dir_name[full_dir_name.size()-1] != '/')
-    full_dir_name = full_dir_name + "/";
+  if (full_dir_name_[full_dir_name_.size() - 1] != '/')
```
pnstickne: What if the empty string is specified?

src/vt/trace/trace.cc
```diff
@@ -596,7 +600,38 @@ void Trace::disableTracing() {
...
 
-void Trace::writeTracesFile() {
+void Trace::cleanupTracesFile() {
+  if (checkEnabled()) {
+    auto const& node = theContext()->getNode();
+    //--- Sanity check
+    if (open_events_.empty()) {
+      cur_stop_ = traces_.size();
+    }
+    else {
```
pnstickne: vtAssert should work directly on the assert condition.

pnstickne: Pulling this out also shows a slight conflict within the code:

     vtAssert(!open_events.empty(), "...");

     if (open_events_.empty()) {
       cur_stop_ = traces_.size();
     }

(A local variable could also be used for the condition.)

src/vt/trace/trace.cc
```diff
@@ -643,8 +683,13 @@ void Trace::writeLogFile(gzFile file, TraceContainerType const& traces) {
       "Event must exist that was logged"
     );
 
-    auto const event_seq_id = log->ep == no_trace_entry_id ?
-      no_trace_entry_id : event_iter->second.theEventSeq();
+    TraceEntryIDType event_seq_id;
```
pnstickne: The `vtAssert` already in the code (somewhere in this file) is correct - it is an error if there is no event found here. That is fixed in #575 and this will have a merge conflict.

src/vt/trace/trace.h
```diff
@@ -156,7 +151,9 @@ struct Trace {
   void disableTracing();
   bool checkEnabled();
 
-  void writeTracesFile();
+  void flushTracesFile(bool useGlobalSync = false);
```
pnstickne: Prefer making the caller specify for explicitness.

src/vt/trace/trace.cc
```diff
@@ -577,11 +585,7 @@ bool Trace::checkEnabled() {
     auto const node = theContext()->getNode();
     if (ArgType::vt_trace_mod == 0) {
       return true;
-    } else if (node % ArgType::vt_trace_mod == 1) {
-      return true;
-    } else {
-      return false;
-    }
+    } else return (node % ArgType::vt_trace_mod == 1);
```
pnstickne: `else` layout feels inconsistent.

Can also be simplified:

     return (ArgType::vt_trace_mod
         or node % ArgType::vt_trace_mod == 1);

src/vt/trace/trace.cc
```diff
@@ -609,27 +644,32 @@ void Trace::writeTracesFile() {
...
+    }
+    writeLogFile(log_file_, traces_);
+    gzflush(log_file_, flush);
   }
 
-  if (node == designated_root_node) {
+  if (node == designated_root_node and not wrote_sts_file_) {
     std::ofstream file;
-    auto name = full_sts_name;
+    auto name = full_sts_name_;
```
pnstickne: Assigning local does not enable additional re-use or lifetime.

src/vt/trace/trace.cc
```diff
@@ -609,27 +644,32 @@ void Trace::writeTracesFile() {
...
     file.close();
+    wrote_sts_file_ = true;
   }
 }
 
 void Trace::writeLogFile(gzFile file, TraceContainerType const& traces) {
-  for (auto&& log : traces) {
-    auto const converted_time = timeToInt(log->time - start_time_);
+  auto stop_point = cur_stop_;
+  for (auto i = cur_; i < stop_point; i++) {
```
pnstickne: Consider using explicit types for simple types, as `auto` reduces clarity and gains little in these cases:

     int32_t stop_point = ..;
     for (int32_t i = ...; ; )

src/vt/trace/trace.h
```diff
@@ -185,10 +182,15 @@ struct Trace {
...
-  std::string full_dir_name     = "";
-  UserEventRegistry user_event  = {};
+  std::string full_trace_name_  = "";
+  std::string full_sts_name_    = "";
+  std::string full_dir_name_    = "";
+  UserEventRegistry user_event_ = {};
+  gzFile log_file_;
+  bool file_is_open_            = false;
+  bool wrote_sts_file_          = false;
+  int64_t cur_                  = 0;
```
pnstickne: Use `size_t` for values representing "maximum indexable range".

src/vt/trace/trace.h
```diff
@@ -88,9 +83,9 @@ struct Trace {
 
   friend struct Log;
 
-  std::string getTraceName() const { return full_trace_name; }
-  std::string getSTSName()   const { return full_sts_name;   }
-  std::string getDirectory() const { return full_dir_name;   }
+  std::string getTraceName() const { return full_trace_name_; }
```
pnstickne: Are these needed/used?

src/vt/trace/trace.cc
pnstickne: If an assert fails, the state of the code is ill-defined and no attempt should be made to continue in such a state.

src/vt/trace/trace.cc
```diff
@@ -42,27 +42,31 @@
...
-#include "vt/config.h"
 #include "vt/trace/trace.h"
-#include "vt/timing/timing.h"
+#include "vt/config.h"
+#include "vt/configs/arguments/args.h"
+#include "vt/runtime/runtime_inst.h"
 #include "vt/scheduler/scheduler.h"
+#include "vt/timing/timing.h"
 #include "vt/utils/demangle/demangle.h"
+#include "vt/collective/collective_alg.h"
```
pnstickne: Are all of these additions used directly?

src/vt/trace/trace.cc
pnstickne: If code (regardless of build configuration) is expected to act differently depending on a certain condition that should be captured as ONLY the condition and not an assert.

src/vt/trace/trace.cc
```diff
@@ -609,27 +644,32 @@ void Trace::writeTracesFile() {
...
   }
 
-  if (node == designated_root_node) {
+  if (node == designated_root_node and not wrote_sts_file_) {
     std::ofstream file;
-    auto name = full_sts_name;
+    auto name = full_sts_name_;
     file.open(name);
     outputControlFile(file);
     file.flush();
```
pnstickne: `flush` probably not useful before `close`.

pnstickne: `flush` probably not useful immediately before `close`.

src/vt/trace/trace.cc
```diff
@@ -609,27 +644,32 @@ void Trace::writeTracesFile() {
...
+      log_file_ = gzopen(path.c_str(), "wb");
+      outputHeader(node, start_time_, log_file_);
+      file_is_open_ = true;
+    }
+    writeLogFile(log_file_, traces_);
+    gzflush(log_file_, flush);
   }
 
-  if (node == designated_root_node) {
+  if (node == designated_root_node and not wrote_sts_file_) {
```
pnstickne: This will be incorrect if there are any late-registered events or late name changes.

pnstickne: If wanting to making this a stronger restriction then it needs to be formalized in code and contracts.

src/vt/trace/trace.cc
```diff
@@ -596,7 +600,38 @@ void Trace::disableTracing() {
...
+  if (useGlobalSync) {
+    // Synchronize all the nodes before flushing the traces
+    theCollective()->barrier();
+  }
+  if (traces_.size() > cur_ + ArgType::vt_trace_flush_mod) {
+    writeTracesFile(Z_FULL_FLUSH);
+  }
+}
+
+void Trace::writeTracesFile(int flush) {
```
pnstickne: Flush is `int`? It seems like it should be "yes or no"; consider a boolean or a relective name change.

src/vt/trace/trace.cc
```diff
@@ -596,7 +600,38 @@ void Trace::disableTracing() {
...
+    }
+    //--- Dump everything into an output file
+    writeTracesFile();
+    outputFooter(node, start_time_, log_file_);
+    gzclose(log_file_);
+  }
+}
+
+void Trace::flushTracesFile(bool useGlobalSync) {
+  if (ArgType::vt_trace_flush_mod == 0) {
```
pnstickne: Could use temp here to reduce common ArgType expression.

src/vt/trace/trace.cc
```diff
@@ -596,7 +600,38 @@ void Trace::disableTracing() {
...
+    // Flush the traces at the end only
+    return;
+  }
+  if (useGlobalSync) {
+    // Synchronize all the nodes before flushing the traces
+    theCollective()->barrier();
+  }
+  if (traces_.size() > cur_ + ArgType::vt_trace_flush_mod) {
```
pnstickne: Using mod in an addition feels suspicious.

src/vt/trace/trace.cc
```diff
@@ -787,7 +832,7 @@ void Trace::writeLogFile(gzFile file, TraceContainerType const& traces) {
     delete log;
   }
 
-  traces.empty();
+  cur_ = stop_point;
```
pnstickne: Maybe this - adding in flushing - would be a good case *to* empty (or otherwise occasionally) clean up the trace collection to reduce memory usage.

src/vt/trace/trace.cc
```diff
@@ -596,7 +600,38 @@ void Trace::disableTracing() {
   enabled_ = false;
 }
 
-void Trace::writeTracesFile() {
+void Trace::cleanupTracesFile() {
+  if (checkEnabled()) {
```
pnstickne: `checkEnabled` to use an improved name, like `traceOnCurrentNode` or whatnot.

src/vt/trace/trace.cc
pnstickne: Consider that each node has a separate `traces_` collection. Maybe there should be a DIFFERENT configuration?

src/vt/trace/trace.cc
pnstickne: Also, this check will _probably skip some nodes_; again, pull up the context explicitly (or in name) to make it more obvious as to why.

src/vt/trace/trace.cc
uhetmaniuk: Yes, `gzflush` takes `int` as second parameter.

src/vt/trace/trace.cc
uhetmaniuk: Replace test on vt_trace_dir with vt_trace_dir.empty().

src/vt/trace/trace.cc
uhetmaniuk: OK.

src/vt/trace/trace.cc
uhetmaniuk: Simpified the code. 
---------
Regarding the failure of `vtAssert` (assuming that I understand correctly your comment), it sounds like a bigger issue that is outside the scope of this PR. `vtAssert` is used extensively throughout the code. It may be worth its own issue.

src/vt/trace/trace.cc
uhetmaniuk: Yes, each writing node has its own `traces_` and its own output file.

I do not understand the latest comment.

src/vt/trace/trace.cc
uhetmaniuk: I did not explore the STS information. The test was updated in separate commits.

If this is a potential problem, then a separate issue should be submitted.

src/vt/trace/trace.cc
uhetmaniuk: (Same remark as above).

src/vt/trace/trace.cc
uhetmaniuk: OK.

src/vt/trace/trace.cc
uhetmaniuk: Done.

src/vt/trace/trace.cc
uhetmaniuk: Done.

src/vt/trace/trace.cc
uhetmaniuk: Done.

src/vt/trace/trace.cc
uhetmaniuk: (See above)
It sounds like a bigger issue that is outside the scope of this PR. vtAssert is used extensively throughout the code. It may be worth its own issue.

src/vt/trace/trace.cc
uhetmaniuk: Memory reduction will be part of a spearate PR (if I understood correctly @lifflander ).

src/vt/trace/trace.h
uhetmaniuk: OK.

src/vt/trace/trace.h
uhetmaniuk: They are used in `runtime/runtime.cc`. 
@lifflander may comment whether they can be removed.
Other functions in `trace.h` are also not used.

src/vt/trace/trace.h
uhetmaniuk: OK

src/vt/trace/trace.cc
pnstickne: Okay, so say if the mod is % 2 (for every other node).
Then this will flush when there are 2 extra traces on those nodes..

Yet if this is % 10 (for 1/10 nodes)
Then a node will collect 10 traces before the flush..

pnstickne: Thus, the overall confusion I'm having is the similarity between `vt_trace_mod` and `vt_trace_flush_mod`.

uhetmaniuk: OK. Updated in commit 860212b

src/vt/trace/trace.cc
```diff
@@ -596,7 +597,35 @@ void Trace::disableTracing() {
   enabled_ = false;
 }
 
-void Trace::writeTracesFile() {
+void Trace::cleanupTracesFile() {
+  auto const& node = theContext()->getNode();
+  if (!traceWritingEnabled(node)) {
```
pnstickne: I wonder if it might be worthwhile to not even collect traces on nodes where it's not enabled?

src/vt/trace/trace.cc
pnstickne: Okay, good, because it currently leaks like a sieve..

src/vt/trace/trace.cc
pnstickne: vtAssert isn't an issue - rather, like any assert, it is used to *leave contract guarantees in code*.
This is why it should 1) be primary depending *only* on it's own condition; 2) the it's condition must not cause a side-effect.

src/vt/trace/trace.cc
pnstickne: Can use forward declares..? I really want to say **no** to external libs in headers :|

pnstickne: Even if this means having to add a small proxy, it avoids a global leak that is hard to later remove. fmtlib is another example.

src/vt/trace/trace.h
```diff
@@ -45,26 +45,21 @@
...
 #include <stack>
 #include <string>
+#include <unordered_map>
 #include <vector>
-#include <memory>
-#include <functional>
-#include <iosfwd>
 
 #include <mpi.h>
 #include <zlib.h>
```
pnstickne: This should be removed to avoid leaking zlib as a dependency.

pnstickne: It is *OK* to have internal-linkage only methods which are not part of the interface.

src/vt/trace/trace.cc
pnstickne: More specifically, 

   After a `vtAssert(x, ..)` or `assert(x && ..)`, the contract is that `x` *must* be true-per-contract, regardless of how the build has been generated.

   Therefore, any conditions to check `x` afterward are not honoring the correctness guarantee.

src/vt/trace/trace.h
```diff
@@ -156,7 +151,9 @@ struct Trace {
   void disableTracing();
   bool checkEnabled();
 
-  void writeTracesFile();
+  void flushTracesFile(bool useGlobalSync = false);
+  void writeTracesFile(int flush = Z_FINISH);
+  void cleanupTracesFile();
   void writeLogFile(gzFile file, TraceContainerType const& traces);
```
pnstickne: Can probably avoid most glib leakage in header by removing this from interface.

src/vt/trace/trace.cc
pnstickne: Also, it should be/have been `traces.clear()` :|

src/vt/trace/trace.cc
pnstickne: Also, this check is invalid as it means that writeTracesFile will not be called as it may preclude the root node.

src/vt/trace/trace.cc
pnstickne: This should be resolved prior to merge. While the STS information can be written early on (ie. in case of a program crash), it should ALSO be written on normal (or caught abort) termination to capture the latest state.

pnstickne: It appears as though
- `writeTracesFile` might not always be called for node-0 on cleanup and;
- the guard in place will prevent a new .sts file from being re-generated; thus a file generated early will contain incomplete data in some cases (trivially, the total count of events will be incorrect).

Minimally, something like:
- generate STS file on program termination, REGARDLESS of if the .sts information has already been written (ie. overwrite)

src/vt/trace/trace.h
uhetmaniuk: Updated branch.

src/vt/trace/trace.h
uhetmaniuk: Updated branch.

src/vt/trace/trace.cc
uhetmaniuk: Changed the test as it is inconsistent with the description of `vt_trace_mod` (to `n % mod == 0`).

src/vt/trace/trace.cc
uhetmaniuk: I believe the latest commits answer this comment.

src/vt/trace/trace.cc
uhetmaniuk: Latest commit should address the comment.

src/vt/trace/trace.cc
```diff
@@ -42,24 +42,35 @@
...
+#include "vt/collective/collective_alg.h"
 #include "vt/config.h"
-#include "vt/trace/trace.h"
-#include "vt/timing/timing.h"
 #include "vt/scheduler/scheduler.h"
+#include "vt/timing/timing.h"
+#include "vt/trace/trace.h"
 #include "vt/utils/demangle/demangle.h"
 
-#include <zlib.h>
```
pnstickne: Yay!

src/vt/trace/trace.cc
```diff
@@ -596,7 +604,35 @@ void Trace::disableTracing() {
...
+  if (ArgType::vt_trace_flush_size == 0) {
+    // Flush the traces at the end only
+    return;
+  }
+  if (useGlobalSync) {
+    // Synchronize all the nodes before flushing the traces
+    theCollective()->barrier();
+  }
+  if (traces_.size() > cur_ + ArgType::vt_trace_flush_size) {
+    writeTracesFile(Z_FULL_FLUSH);
```
pnstickne: Use Z_SYNC_FLUSH (or weaker). It is sufficient to ensure the steam-up-until has been written and has significantly
less impact on compression rates.

Z_FULL_FLUSH is ideal for random-access and recovering corrupt streams, neither of which are available here. A downside of Z_FULL_FLUSH is that it resets the dictionary which severely decreases compression.

src/vt/trace/trace.cc
```diff
@@ -596,7 +604,35 @@ void Trace::disableTracing() {
...
+  outputFooter(node, start_time_, log_file_.get());
+  gzclose(log_file_->file_type);
+}
+
+void Trace::flushTracesFile(bool useGlobalSync) {
+  if (ArgType::vt_trace_flush_size == 0) {
+    // Flush the traces at the end only
+    return;
+  }
+  if (useGlobalSync) {
```
pnstickne: Is there a reason to do a global sync when the trace won't be written? (eg. the condition is not met)

src/vt/trace/trace.cc
```diff
@@ -608,43 +644,56 @@ void Trace::writeTracesFile() {
...
+    writeLogFile(log_file_.get(), traces_);
+    gzflush(log_file_->file_type, flush);
+  }
+
+  if (node != designated_root_node) {
+    return;
   }
 
-  if (node == designated_root_node) {
+  if ((flush == Z_FINISH) or (not wrote_sts_file_)) {
```
pnstickne: Can just close the file on finish.

src/vt/trace/trace.h
```diff
@@ -45,32 +45,27 @@
...
 #include <stack>
 #include <string>
+#include <unordered_map>
 #include <vector>
-#include <memory>
-#include <functional>
-#include <iosfwd>
 
-#include <mpi.h>
-#include <zlib.h>
```
pnstickne: Yay!

src/vt/vrt/collection/manager.cc
```diff
@@ -120,7 +124,7 @@ getDispatcher(auto_registry::AutoHandlerType const& han) {
 }
 
 void releaseLBPhase(CollectionPhaseMsg* msg) {
-  theCollection()->releaseLBPhase<>(msg);
+  CollectionManager::releaseLBPhase<>(msg);
```
pnstickne: Hmm, why not using `theCollection()` anymore? Is this any outside change?

src/vt/collective/collective_ops.cc
```diff
@@ -137,6 +138,10 @@ void CollectiveAnyOps<instance>::abort(
   auto tls_rt = curRT;
   auto myrt = tls_rt ? tls_rt : ::vt::rt;
   if (myrt) {
+#if backend_check_enabled(trace_enabled)
+    //--- Try to flush most of the traces before aborting
+    myrt->theTrace->writeTracesFile(Z_FINISH);
```
pnstickne: Might be useful to `cleanupTracesFile` (ensure footer is written, etc.)

src/vt/collective/collective_ops.cc
```diff
@@ -151,6 +156,10 @@ void CollectiveAnyOps<instance>::output(
   auto tls_rt = curRT;
   auto myrt = tls_rt ? tls_rt : ::vt::rt;
   if (myrt) {
+#if backend_check_enabled(trace_enabled)
+    //--- Try to flush most of the traces
+    myrt->theTrace->writeTracesFile(Z_FULL_FLUSH);
```
pnstickne: This will issue an abort in some cases - maybe also issue a cleanup in that case?

From a few lines below:

    if (error and abort_out) {
        vt::abort("Assertion Failed", 129);
     }

src/vt/trace/trace.cc
uhetmaniuk: Updating.

src/vt/collective/collective_ops.cc
uhetmaniuk: OK.

src/vt/collective/collective_ops.cc
uhetmaniuk: OK.

src/vt/trace/trace.cc
uhetmaniuk: The global sync was originally suggested by @lifflander 
Note that a subset of writing nodes may satisfy the test on "traces_.size()" and the other writing nodes may not.
The code does not know the situation without a communication... 
I think the barrier may be replaced with a global communication... 
I am not sure which option is better

src/vt/trace/trace.cc
uhetmaniuk: I do not understand the comment 
The section `if ((flush == Z_FINISH) or (not wrote_sts_file_)) { ... }` is for the STS file. The test on `(flush == Z_FINISH)` is to force the writing the STS at the end (per your previous request).

src/vt/vrt/collection/manager.cc
uhetmaniuk: This change may not matter for this particular line...
The function is static so we may as well access it directly rather than through a pointer (which may not be initialized yet).
There was an [issue](https://github.com/DARMA-tasking/vt/issues/527) in the past for a similar access.
Again, this change may not matter for this particular case...

uhetmaniuk: This change may not matter for this particular line...
The function is static so we may as well access it directly rather than through a pointer (which may not be initialized yet).
There was an issue in the past for a similar access.
Again, this change may not matter for this particular case...

lifflander:
@uhetmaniuk I want to merge this, but I can't rebase it on develop (or even squash it) due to so many conflicts. Please don't merge develop into feature branches---rebase as you go and re-write the history.

Can you rebase this on the latest develop and fix the conflicts?

`git checkout 241-flush-trace`
`git rebase -i develop`, close your editor and save or squash as needed

uhetmaniuk:
Close the pull request as all commits have been migrated to [#606 ] for a cleaner history.

Issue #552, opened 2019-11-05
Add tracing of MPI calls


nlslatt:
**What Needs to be Done?**
- [x] Add runtime option for tracing of MPI calls.
- [x] Add tracing of MPI events conditional on runtime option.

**Is your feature request related to a problem? Please describe.**

We can't currently distinguish between delays due to MPI events and other unattributed time in traces.

nlslatt:
@lifflander: This is done and based off of 1.0.0.  Because it's not meant to cleanly merge into develop, the PR will be a little awkward.

lifflander:
Closing as this is now merged on `1.0.0-beta.4-proposed-update`  (#553) and `develop` (#555)

PR #553, opened 2019-11-05
552 add tracing of mpi calls

nlslatt:
Issue #552.  I have created this PR for merging into 1.0.0 because that's where I branched off.  I don't think it's appropriate to merge directly into 1.0.0, so I have marked this as WIP.  However, this is complete and **ready for review** as far as I am concerned.

This adds tracing of MPI_Isend, MPI_Irecv, and the completion of MPI_Irecv when the new --vt_trace_mpi option is used.  I implemented it differently than @lifflander did in the past due to the conditional (so couldn't use scoping) and the desire to include the message size as a note associated with the MPI_Irecv event itself rather than as a precursor note.


lifflander:
I've changed the base to `1.0.0-beta.4-proposed-update`, turning off WIP status.

PR #555, opened 2019-11-05
552 add tracing of mpi calls (for develop)

lifflander:
Porting #553 for develop


lifflander:
Tested on develop, merging.

Issue #539, opened 2019-11-04
Perf. colorization code and unify color selection


lifflander:
**What Needs to be Done?**

- Rules for configuration are defined (AND followed..);
  logic happens immediately after parsing which ensures
  only one isatty check (ioctl) and centralized logic handling.

  Updated the setting/diagnostic messages to also align
  with the rules.

  Rules:
  - no-color DISABLES color
  - color ENABLES color
  - auto-color (assumed default) enables color in terminal

  (This makes VT consistent with tools like 'ls', or more so..)

- Removes unistd.h bleed - should be removed from other headers

- Consistently use defined colors and other minor formatting;
  swith to 'standard' functions.

lifflander:
This is merged via #544, but caused a bug (see #556). Closing.

PR #557, opened 2019-11-05
#556 restore the original coloring logic

### LINKED ISSUE #556: Color is not output by default on termination anymore (after merging #544)
**Describe the bug**
@pnstickne After merging issue #539, PR #544, the color stopped outputting by default in a normal terminal.

**To Reproduce**
1. Run any test/example.

lifflander:
#539, #544 caused color to *not* be output by default anymore on a my terminal (Mac). Restore the original logic (default on, not relying on `isatty`, unless `--vt_auto_color` is explicit to detect the termination type)

Fixes #556

### DIFF
--- a/src/vt/configs/arguments/args.cc
+++ b/src/vt/configs/arguments/args.cc
@@ -481,7 +481,11 @@ namespace vt { namespace arguments {
   } else if (vt_color) {
     colorize_output = true;
   } else { // assume auto-color
-    colorize_output = isatty(fileno(stdout));
+    if (vt_auto_color) {
+      colorize_output = isatty(fileno(stdout));
+    } else {
+      colorize_output = not vt_no_color;
+    }
   }
 
   /*
--- a/src/vt/configs/arguments/args.cc
+++ b/src/vt/configs/arguments/args.cc
@@ -52,9 +52,8 @@
 
 namespace vt { namespace arguments {
 
-/*static*/ bool        ArgConfig::vt_color              = false;
+/*static*/ bool        ArgConfig::vt_color              = true;
 /*static*/ bool        ArgConfig::vt_no_color           = false;
-/*static*/ bool        ArgConfig::vt_auto_color         = true;
 /*static*/ bool        ArgConfig::vt_quiet              = false;
 
 /*static*/ bool        ArgConfig::colorize_output       = false;
@@ -161,20 +160,16 @@ namespace vt { namespace arguments {
    * Flags for controlling the colorization of output from vt
    */
   auto quiet  = "Quiet the output from vt (only errors, warnings)";
-  auto always = "Colorize output (overrides --vt_auto_color)";
-  auto never  = "Never colorize output (overrides --vt_color)";
-  auto maybe  = "Automatic colorization of output";
+  auto always = "Colorize output (default)";
+  auto never  = "Do not colorize output (overrides --vt_color)";
   auto a  = app.add_flag("-c,--vt_color",      vt_color,      always);
   auto b  = app.add_flag("-n,--vt_no_color",   vt_no_color,   never);
-  auto c  = app.add_flag("-a,--vt_auto_color", vt_auto_color, maybe);
   auto a1 = app.add_flag("-q,--vt_quiet",      vt_quiet,      quiet);
   auto outputGroup = "Output Control";
   a->group(outputGroup);
   b->group(outputGroup);
-  c->group(outputGroup);
   a1->group(outputGroup);
   b->excludes(a);
-  b->excludes(c);
 
   /*
    * Flags for controlling the signals that VT tries to catch
@@ -476,11 +471,9 @@ namespace vt { namespace arguments {
   // Determine the final colorization setting.
   if (vt_no_color) {
     colorize_output = false;
-  } else if (vt_color) {
-    colorize_output = true;
   } else {
-    // Otherwise, currently assume to colorize as VT runs within MPI.
-    // isatty in MPI is always false; mpc_isatty is a vendor extension.
+    // Otherwise, colorize.
+    // (Within MPI there is no good method to auto-detect.)
     colorize_output = true;
   }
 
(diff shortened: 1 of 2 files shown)


src/vt/configs/arguments/args.cc
```diff
@@ -461,7 +461,11 @@ namespace vt { namespace arguments {
   } else if (vt_color) {
     colorize_output = true;
   } else { // assume auto-color
-    colorize_output = isatty(fileno(stdout));
+    if (vt_auto_color) {
+      colorize_output = isatty(fileno(stdout));
+    } else {
+      colorize_output = not vt_no_color;
```
uhetmaniuk: @lifflander My understanding is that CLI::excludes (b->excludes(a); b->excludes(c)) is bi-directional. So only one of the three booleans (vt_color, vt_auto_color, vt_no_color) will be true. How will the line 467 get active?

lifflander: @uhetmaniuk That's actually what runs by default! If the user doesn't specify, `vt_auto_color` is false and VT *does not* check if the terminal is a tty. I had to do this because in many cases `isatty` returns false when it's not true.

uhetmaniuk: @lifflander  I must be missing something... Line 59 of `args.cc` set `vt_auto_color` to `true` (by default). No?

lifflander: No, it's false by default. Meaning that it doesn't check if the terminal is a tty by default.

```C++
/*static*/ bool        ArgConfig::vt_auto_color         = false;
```

Maybe it's different on another branch?

lifflander: Actually, I'm sorry I've been looking at the release branch! It's indeed true by default on develop...

uhetmaniuk: No problem! So many branches to track :)

lifflander: he current develop fails because `isatty(fileno(stdout))` evaluates to false. So along with this maybe we should change the default to false I think.

lifflander: Ok, so the confusing part of this is that the default value for `--vt_auto_color`, is not controlled by the static:
```C++
/*static*/ bool        ArgConfig::vt_auto_color         = true;
```
i.e., This does not actually set `vt_auto_color` to true.

uhetmaniuk: OK

pnstickne:
If a specific environment has issues with terminal detection, that's either 1) the environment issue or 2) the detection of coloring.

This should not dictate expected behavior or options of. Take `ls`, for example. The default to color.

src/vt/configs/arguments/args.cc
```diff
@@ -461,7 +461,11 @@ namespace vt { namespace arguments {
   } else if (vt_color) {
     colorize_output = true;
   } else { // assume auto-color
-    colorize_output = isatty(fileno(stdout));
+    if (vt_auto_color) {
+      colorize_output = isatty(fileno(stdout));
+    } else {
+      colorize_output = true;
```
pnstickne: With no flags, this will result in incorrect colors being emitted to pipes.

pnstickne:
In the case of `ls`, for example, one might add an `alias ls=ls --color=true` for their specific choice without affecting the default behavior of the program itself.

lifflander:
> @lifflander @uhetmaniuk  In the case of `ls`, for example, one might add an `alias ls=ls --color=true` for their specific local preference without affecting the default behavior of the program itself; a similar could apply to vt (eg): `alias vt=vt --vt_color` to force colors locally, regardless of the isatty check. Another option, while allowing standard-auto-coloring-expectations would be to introduce environment variables:
> 
>     `VT_COLOR=1 vt ..`

The problem is that vt is launched with `mpirun`, which may make this difficult.

pnstickne:
For example, the `isatty` call is wrong in (at least some) MPI: https://lost-contact.mit.edu/afs/enea.it/software/doc/poe/ppe.html/d3d70mst35.html (this is an issue in the code, not with default expectations if said code was detecting correctly).

src/vt/configs/arguments/args.cc
pnstickne: In the case of `ls`, for example, one might add an `alias ls=ls --color=true`

> > @lifflander @uhetmaniuk  In the case of `ls`, for example, one might add an `alias ls=ls --color=true` for their specific local preference without affecting the default behavior of the program itself; a similar could apply to vt (eg): `alias vt=vt --vt_color` to force colors locally, regardless of the isatty check. Another option, while allowing standard-auto-coloring-expectations would be to introduce environment variables:
> > ```
> > `VT_COLOR=1 vt ..`
> > ```
> 
> The problem is that vt is launched with `mpirun`, which may make this difficult.

At least for arguments, can be pushed up at point of invocation:

     alias mpirun_vtc=mpirun vt --x

pnstickne:
> The `isatty` call appears to be the 'issue' in (at least some) MPI: https://lost-contact.mit.edu/afs/enea.it/software/doc/poe/ppe.html/d3d70mst35.html (this is an issue in the code, not with default expectations if said code was detecting correctly). Also https://www.mail-archive.com/devel@lists.open-mpi.org/msg17764.html - so, oops :|

Really should probably just address this ..

pnstickne:
@lifflander Let me take this and fix the terminal detection.

pnstickne:
Color is "automatic" by default, which implies "enabled" as there is no appropriate mechanism in MPI (standard) that will answer the question correctly. Notes relating to 'terminal' have been removed and logic remains consistent.

Screen Shot 2019-11-10 at 3 18 03 PM

pnstickne:
Or eliminate the auto-color option.
Or set to false always for auto-color 'check' and then change the default static value at the top of the file back for color-enabled.
Just minimize divergence..

lifflander:
@pnstickne I agree--let's do away with "auto color".

pnstickne:
Screen Shot 2019-11-12 at 8 13 50 PM

lifflander:
@pnstickne Is this ready to merge? If so, I will test it and get it onto mainline.

lifflander:
Rebased onto current develop

Issue #558, opened 2019-11-08
allow removing indices from collection chain set


nlslatt:
**What Needs to be Done?**

- [ ] First do this
- [ ] Then do that.
- [ ] Also this other thing.

**Is your feature request related to a problem? Please describe.**

**Describe potential solution outcome**

**Describe alternatives you've considered**

**Additional context**

PR #560, opened 2019-11-08
#558: Support migrating responsibility for object control flow

PhilMiller:


lifflander:
I think it would be preferable for this to be named after the VT issue #558 so we can track it on the release

PhilMiller:
Does your tracking script look only at non-merge commits? If it considers issue numbers in all commits, then having it in the merge should suffice.

lifflander:
@PhilMiller It looks at both merge and non-merge commits. The problem is that it can't correlate a PR # to the Issue # unless the PR title is prefixed with the issue #. Github does not automatically connect the two in the json.

PhilMiller:
Is there a reason not to go ahead with merging this?

lifflander:
Closing in favor of #562

Issue #563, opened 2019-11-12
Cleanup reduce state before VT shuts down


lifflander:
**What Needs to be Done?**

- [ ] Static data persists in static `ReduceStateHolder<T>`, which causes problems if VT is shutdown because it holds messages which access `thePool`, which will cease to exist.

lifflander:
Closing this as it's not a real problem---just the symptom of another problem.

PR #564, opened 2019-11-12
#563 cleanup ReduceStateHolder<T> at VT finalize

lifflander:


src/vt/collective/reduce/reduce_state_holder.impl.h
```diff
@@ -95,20 +95,31 @@ template <typename T>
...
+/*static*/ void ReduceStateHolder<T>::cleanup() {
+  state_lookup_.clear();
 }
 
 template <typename T>
 /*static*/ typename ReduceStateHolder<T>::GroupLookupType
 ReduceStateHolder<T>::state_lookup_ = {};
 
+template <typename T>
+/*static*/ bool ReduceStateHolder<T>::created_ = false;
```
PhilMiller: I would reverse the sense of this, so it doesn't need to be negated later. Something like `first_insertion_ = true`

src/vt/collective/reduce/reduce.impl.h
```diff
@@ -163,7 +163,15 @@ void Reduce::reduceAddMsg(
     ReduceState<MessageT> state(
       msg->reduce_tag_,msg->reduce_seq_,num_contrib_state
     );
-    ReduceStateHolder<MessageT>::insert(group_,lookup,std::move(state));
+    bool has_elm = ReduceStateHolder<MessageT>::insert(group_,lookup,std::move(state));
+    // For the first time of type MessageT, add a cleanup function to clear out
+    // messages. They must be cleaned up early so message destructors do not run
+    // after VT is shutdown.
+    if (not has_elm) {
+      cleanup_.emplace_back([]{
```
pnstickne: Functions in templates come at cumulatively expensive T-cost.

lifflander:
I think this problem of messages being held in the `ReduceStateHolder<T>` past finalize is actually a symptom of another bug. Thus, I am holding off on merging this.

lifflander:
Closing this as the problem was a consequence of other bugs and is not a real problem

PR #565, opened 2019-11-12
#561 implement new file spec with LB-specific params

### LINKED ISSUE #561: Generalize the parameters to the LB spec file
**What Needs to be Done?**

- [x] Define a new spec. Allow LB-specific parameters to be passed in and parsed.

lifflander:
The new file format allows for different parameters for each LB. Can be passed as a file or using `--vt_lb_args="c=1 k=5 f=2 i=10"`.

Example file:
```
%10 GossipLB c=1 k=5 f=2 i=10
0 HierarchicalLB min=0.9 max=1.1 auto=false
%5 GreedyLB min=1.0
120 GreedyLB c=0 k=2 f=3 i=3
```

Format:
```
[%] [iter] <LB-name> [key=value]...
```

Specific iteration lines override mods. Mods precedence is based on position in file, thus on iteration 10, `GreedyLB` will run---not `GossipLB`.

Fixes #561

### DIFF
--- a/src/vt/vrt/collection/balance/gossiplb/gossiplb.cc
+++ b/src/vt/vrt/collection/balance/gossiplb/gossiplb.cc
@@ -56,6 +56,9 @@ void GossipLB::init(objgroup::proxy::Proxy<GossipLB> in_proxy) {
   proxy = in_proxy;
 }
 
+void GossipLB::inputParams(balance::SpecEntry* spec) {
+}
+
 void GossipLB::runLB() {
   this->inform();
 }
--- a/src/vt/vrt/collection/balance/read_lb.cc
+++ b/src/vt/vrt/collection/balance/read_lb.cc
@@ -46,6 +46,7 @@
 #include "vt/context/context.h"
 #include "vt/vrt/collection/balance/read_lb.h"
 #include "vt/vrt/collection/balance/lb_type.h"
+#include "vt/configs/arguments/args.h"
 
 #include <string>
 #include <fstream>
@@ -57,18 +58,37 @@ namespace vt { namespace vrt { namespace collection { namespace balance {
 
 /*static*/ std::string ReadLBSpec::filename = {};
 /*static*/ SpecIndex ReadLBSpec::num_entries_ = 0;
-/*static*/ std::unordered_map<SpecIndex,SpecEntry> ReadLBSpec::spec_ = {};
-/*static*/ bool ReadLBSpec::has_spec_ = false;
+/*static*/ typename ReadLBSpec::SpecMapType ReadLBSpec::spec_mod_ = {};
+/*static*/ typename ReadLBSpec::SpecMapType ReadLBSpec::spec_exact_ = {};
+/*static*/ std::vector<SpecIndex> ReadLBSpec::spec_prec_ = {};
 /*static*/ bool ReadLBSpec::read_complete_ = false;
+
+/*static*/ bool ReadLBSpec::hasSpec() {
+  if (not ArgType::vt_lb_file) {
+    return false;
+  }
+  if (read_complete_) {
+    return true;
+  } else {
+    auto const file_name = ArgType::vt_lb_file_name;
+    if (file_name == "") {
+      return false;
+    } else {
+      return openFile(file_name);
+    }
+  }
+}
 
 /*static*/ bool ReadLBSpec::openFile(std::string const name) {
   std::ifstream file(name);
   filename = name;
-  has_spec_ = file.good();
   return file.good();
 }
 
 /*static*/ LBType ReadLBSpec::getLB(SpecIndex const& idx) {
+  if (not read_complete_) {
+    readFile();
+  }
   auto const lb = entry(idx);
   if (lb) {
     return lb->getLB();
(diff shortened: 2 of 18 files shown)


lifflander:
@nlslatt @PhilMiller Please review this PR, thanks!

nlslatt:
I haven't been able to play with this to make sure it work because there are unrelated but important bug fixes that haven't been merged into this branch.

lifflander:
@nlslatt Are you trying to test this in the context of priorities?

nlslatt:
When trying to run the `lb_iter` test on this branch using LB flags, I was getting the error `Must have valid root`.  Is that the appropriate vt example to test on?

It would be nice to test within EMPIRE.

lifflander:
> When trying to run the `lb_iter` test on this branch using LB flags, I was getting the error `Must have valid root`. Is that the appropriate vt example to test on?
> 
> It would be nice to test within EMPIRE.

It sounds like this might be missing another bug fix. Which machine did you test on?

lifflander:
@nlslatt I just rebased this PR on develop. Do you need it on a `beta.4` branch?

nlslatt:
> > When trying to run the `lb_iter` test on this branch using LB flags, I was getting the error `Must have valid root`. Is that the appropriate vt example to test on?
> > It would be nice to test within EMPIRE.
> 
> It sounds like this might be missing another bug fix. Which machine did you test on?

I was testing on Mutrino.

nlslatt:
> > > When trying to run the `lb_iter` test on this branch using LB flags, I was getting the error `Must have valid root`. Is that the appropriate vt example to test on?
> > > It would be nice to test within EMPIRE.
> > 
> > 
> > It sounds like this might be missing another bug fix. Which machine did you test on?
> 
> I was testing on Mutrino.

Actually, I think it was Kahuna.  Sorry.

nlslatt:
Okay, I ran on Kahuna again with the rebased version.  I still got the same error, but I realized that I had used an incorrect argument.  The incorrect argument itself seems to have triggered an inappropriate error message, but I'm not sure how.  To reproduce, try something like:
```
srun ./lb_iter --vt_dummy=on
```
We should add a check to vt that warns or aborts if any argument beginning with "--vt_" is not recognized.  We also need to figure out how an incorrect argument is leading to the assertion below, but that's probably should be a separate issue (but I haven't checked to make sure the problem also exists outside of this branch).
```
vt: [0] ------------------------------------------------------------------------------------------------------------------------
vt: [0] ------------------------------------------- Runtime Error: System Aborting! --------------------------------------------
vt: [0] ------------------------------------------------ Fatal Error on Node 0 -------------------------------------------------
vt: [0] ------------------------------------------------------------------------------------------------------------------------
vt: [0] 
vt: [0]              Reason: Must have valid root
vt: [0]    Assertion failed: (root != uninitialized_destination)
vt: [0]           Num Nodes: 1
vt: [0]                File: /home/nlslatt/vt-auto-build/vt-script-build5/vt/vt/src/vt/group/group_manager.cc
vt: [0]                Line: 146
vt: [0]            Function: groupRoot
vt: [0]           Build SHA: 80556d4abc9dd01bea8c2555af15b3add4172327
vt: [0]           Build Ref: refs/heads/561-new-lb-file-spec
vt: [0]         Description: heads/561-new-lb-file-spec-0-g80556d4abc
vt: [0]            GIT Repo: clean
vt: [0]            Hostname: kn2.eth
```

nlslatt:
Yes, it would be great to see a version of this rebased on beta.4.

lifflander:
@nlslatt Should we open an issue for the incorrect argument error message bug?

src/vt/vrt/collection/balance/gossiplb/gossiplb.cc
```diff
@@ -56,6 +56,9 @@ void GossipLB::init(objgroup::proxy::Proxy<GossipLB> in_proxy) {
   proxy = in_proxy;
 }
 
+void GossipLB::inputParams(balance::SpecEntry* spec) {
+}
```
nlslatt: Is this really supposed to be a no-op?

lifflander: This is fixed on the branch with GossipLB. Since this is off of develop and it doesn't have the implementation it doesn't pull the parameters.

lifflander:
Fixed a bug in how the parser was eating up newlines and merging lines that should not have been merged thus silently parsing incorrectly.

nlslatt:
@lifflander A critical bug fix is still missing.  `read_lb.cc:77` silently returns false if the requested spec file doesn't exist.  This should be an error condition.

lifflander:
Fixed the bug, added a bunch more checks and 2 new tests for the file reader.

lifflander:
Updated 1.0.0-beta.4 with upstream changes on this PR

src/vt/vrt/collection/balance/read_lb.cc
```diff
@@ -77,118 +106,193 @@ namespace vt { namespace vrt { namespace collection { namespace balance {
...
+ReadLBSpec::parseParams(std::vector<std::string> params) {
+  ParamMapType param_map;
+
+  /*
+   * Split params into 'key=value'
+   */
+  for (auto&& p : params) {
+    std::string key, value;
+    bool found = false;
+    for (int i = 0; i < p.size(); i++) {
```
nlslatt: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]

src/vt/vrt/collection/balance/read_lb.cc
```diff
@@ -77,118 +106,193 @@ namespace vt { namespace vrt { namespace collection { namespace balance {
...
+/*static*/ typename ReadLBSpec::ParamMapType
+ReadLBSpec::parseParams(std::vector<std::string> params) {
+  ParamMapType param_map;
+
+  /*
+   * Split params into 'key=value'
+   */
+  for (auto&& p : params) {
+    std::string key, value;
+    bool found = false;
```
nlslatt: error: variable ‘found’ set but not used [-Werror=unused-but-set-variable]

We should error out if `!found`.

src/vt/vrt/collection/balance/read_lb.cc
lifflander: Fixe

src/vt/vrt/collection/balance/read_lb.cc
lifflander: Add check

PR #569, opened 2019-11-13
#568 group fix incorrect subtree size

### LINKED ISSUE #568: Group subtree reporting incorrect size
**Describe the bug**
The subtree size used to determine if the `default_group` is going to be used is wrong thus leading the less regular communication patterns (not using standard binary tree). While it's technically not incorrect (it runs fine), it is using the wrong spanning tree.

lifflander:
Fixes #568

### DIFF
--- a/src/vt/group/collective/group_info_collective.cc
+++ b/src/vt/group/collective/group_info_collective.cc
@@ -370,7 +370,7 @@ void InfoColl::upTree() {
 
     debug_print(
       group, node,
-      "InfoColl::upTree: case 2: sub={}\n",
+      "InfoColl::upTree: case 2: sub={} (total is same)\n",
       subtree_
     );
 
@@ -400,8 +400,8 @@ void InfoColl::upTree() {
 
     debug_print(
       group, node,
-      "InfoColl::upTree: case 3: sub={}\n",
-      subtree_
+      "InfoColl::upTree: case 3: sub={}, total={}\n",
+      subtree_, total_subtree
     );
 
     auto msg = makeSharedMessage<GroupCollectiveMsg>(
--- a/src/vt/group/collective/group_info_collective.cc
+++ b/src/vt/group/collective/group_info_collective.cc
@@ -222,14 +222,13 @@ void InfoColl::upTree() {
     "Must be equal"
   );
   decltype(msgs_) msg_in_group = {};
-  std::size_t subtree = 0;
+  subtree_ = 0;
   for (auto&& msg : msgs_) {
     if (msg->isInGroup()) {
       msg_in_group.push_back(msg);
-      subtree += msg->getSubtreeSize();
+      subtree_ += msg->getSubtreeSize();
     }
   }
-  subtree_ = subtree;
 
   auto& span_children_ = collective_->span_children_;
   auto const& is_root = collective_->isInitialRoot();
@@ -242,7 +241,7 @@ void InfoColl::upTree() {
     "InfoColl::upTree: is_in_group={}, msgs.size()={}, msg_in_group.size()={}, "
     "extra={}, wait={}, group={:x}, op={:x}, is_root={}, subtree={}\n",
     is_in_group, msgs_.size(), msg_in_group.size(), extra_count_,
-    coll_wait_count_, group, op, is_root, subtree
+    coll_wait_count_, group, op, is_root, subtree_
   );
 
   if (is_root) {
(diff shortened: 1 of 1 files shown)


src/vt/group/collective/group_info_collective.cc
```diff
@@ -375,10 +391,21 @@ void InfoColl::upTree() {
...
+    auto const child = theContext()->getNode();
+    auto const extra = 1;
+
+    subtree_ -= msg_in_group[0]->getSubtreeSize();
+
+    auto const total_subtree = static_cast<NodeType>(subtree_ + subtree_this);
+
+    debug_print(
+      group, node,
+      "InfoColl::upTree: case 3: sub={}\n",
```
pnstickne: Non `total` in this debug_print (parity break).

lifflander: Ok, I've addressed this in my new commit.

src/vt/group/collective/group_info_collective.cc
```diff
@@ -375,10 +391,21 @@ void InfoColl::upTree() {
...
+      "InfoColl::upTree: case 3: sub={}, total={}\n",
+      subtree_, total_subtree
+    );
+
     auto msg = makeSharedMessage<GroupCollectiveMsg>(
-      group,op,is_in_group,sub,child,0,extra
+      group,op,is_in_group,total_subtree,child,0,extra
     );
     theMsg()->sendMsg<GroupCollectiveMsg,upHan>(p, msg);
     theMsg()->sendMsg<GroupCollectiveMsg,upHan>(p, msg_in_group[0].get());
```
nlslatt: Is the `.get()` here and in other `sendMsg` invocations safe?

lifflander: Yes, the `.get()` should be fine here. The node that receives `msg_in_group[0]` owns it now and so is allowed to forward it off.

lifflander:
> I could better review if you could explain why you're subtracting to compute the subtree size.

When a node receives some subtrees, it decides to forward some on `upHan`. Before, the size was incorrect because they were being counted twice---once by the node that originally received the subtree and again if that node decided to forward them on.

Forwarding occurs when a node receives more than a some max (>3) of children. In that case, the spanning tree is flattening at that node so it forwards these on. In the second phase, if nodes higher in the original spanning tree are also over-subscribed (> than the max), they will be pushed down the tree in the second phase.

The new subtree counting code handles all these cases.

src/vt/group/collective/group_info_collective.cc
```diff
@@ -354,12 +359,23 @@ void InfoColl::upTree() {
...
      */
-    auto const& child = theContext()->getNode();
-    auto const& extra = static_cast<GroupCollectiveMsg::CountType>(
+    auto const child = theContext()->getNode();
+    auto const extra = static_cast<GroupCollectiveMsg::CountType>(
       msg_in_group.size()
     );
+
+    for (std::size_t i = 0; i < msg_in_group.size(); i++) {
+      subtree_ -= msg_in_group[i]->getSubtreeSize();
```
lifflander: These are subtracted off because they are forwarded up the tree on line 369/385.

src/vt/group/collective/group_info_collective.cc
```diff
@@ -375,10 +391,21 @@ void InfoColl::upTree() {
      * child in the spanning tree because that will create a stick-like graph,
      * loosing efficiency!
      */
-    auto const& child = theContext()->getNode();
-    auto const& extra = 1;
+    auto const child = theContext()->getNode();
+    auto const extra = 1;
+
+    subtree_ -= msg_in_group[0]->getSubtreeSize();
```
lifflander: `msg_in_group[0]->getSubtreeSize()` is subtracted because it forwarded up the tree on line 384/411.

src/vt/group/collective/group_info_collective.cc
```diff
@@ -395,10 +422,26 @@ void InfoColl::upTree() {
...
+
+    auto const extra =
       static_cast<GroupCollectiveMsg::CountType>(msg_in_group.size() / 2);
-    auto const& child = theContext()->getNode();
-    auto const& total_subtree = static_cast<NodeType>(sub + subtree);
+    auto const child = theContext()->getNode();
+
+    auto tree_iter = msg_list.rbegin();
+    for (int i = 0; i < extra; i++) {
+      subtree_ -= (*tree_iter)->getSubtreeSize();
```
lifflander: These are subtracted because they are forwarded on on line 421/463.

PR #573, opened 2019-11-13
#572 fix race condition in broadcast delivery

### LINKED ISSUE #572: Fix race condition in collection broadcast
**Describe the bug**
This has long been a race condition that I meant to fix in the past that hasn't manifested in the virtual collections until priorities are added.

lifflander:
Fix a race condition in collections manifested with priorities.

Fixes #572

### DIFF
--- a/src/vt/vrt/collection/manager.impl.h
+++ b/src/vt/vrt/collection/manager.impl.h
@@ -445,68 +445,65 @@ template <typename ColT, typename IndexT, typename MsgT>
         "broadcast: apply to element: epoch={}, bcast_epoch={}\n",
         msg->bcast_epoch_, base->cur_bcast_epoch_
       );
-      if (base->cur_bcast_epoch_ == msg->bcast_epoch_ - 1) {
-        vtAssert(base != nullptr, "Must be valid pointer");
-        base->cur_bcast_epoch_++;
+      vtAssert(base != nullptr, "Must be valid pointer");
+      base->cur_bcast_epoch_++;
 
-        #if backend_check_enabled(lblite)
-          debug_print(
-            vrt_coll, node,
-            "broadcast: apply to element: instrument={}\n",
-            msg->lbLiteInstrument()
-          );
-          if (msg->lbLiteInstrument()) {
-            recordStats(base, msg);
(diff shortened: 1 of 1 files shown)


src/vt/vrt/collection/manager.impl.h
```diff
@@ -445,68 +445,65 @@ template <typename ColT, typename IndexT, typename MsgT>
         "broadcast: apply to element: epoch={}, bcast_epoch={}\n",
         msg->bcast_epoch_, base->cur_bcast_epoch_
       );
-      if (base->cur_bcast_epoch_ == msg->bcast_epoch_ - 1) {
```
nlslatt: @lifflander Can you explain what this conditional was originally intended to do?

lifflander: Originally, there was a constraint that broadcasts must be received in order by each collection element. The reason being that with arbitrary migrations or unstructured collection insertions (for dynamically insertable collection), it's impossible to know (without keeping full accounting) if a broadcast has been delivered. Imagine a bunch of broadcasts in flight while an element is migrating. How does the element know if it received them? By enforcing an order on reception the problem is solved but over-synchronized.

This problem was solved in a couple ways:
  - Insertable collections are now a special case and require structured insertions based on insertion epochs. This means that they get the proper broadcast epoch stamp to ensure they receive all causally-following broadcasts---no more, no less. For insertable, there is special code that buffers broadcasts to handle this case. For normal collections, this is not needed.
  - For normal collections, there is a new constraint that migration must happen causally after any preceding broadcasts. By using the load balancing framework to migrate, this is upheld. The special case here is the user manually migrating collection elements, which is now an "expert-only" mode. The user of this expert mode must ensure that they follow the rules---only migrate when no broadcasts are outstanding or expect to miss broadcasts.

lifflander: Eventually this restriction on manual migrations may be lifted for special types of collection or other state that indicate that broadcasts need to be buffered and tracked.

nlslatt: Thank you.

Issue #554, opened 2019-11-05
Trace builds of VT crash early in startup when certain callback kinds are registered (eg. reduce.cc, ~1/8 tests),


lifflander:
**Describe the bug**
This is related to #342, with an expanded explanation of the problem. When compiling in trace mode, the tracing component in conjunction with the registry, generates a `*.sts` file that holds all the object types and method types for reference in the projections trace.

Here's an example sts file (I've changed the extension to txt since github does not support unknown file types): [EMPIRE_PIC-NLS-release-beta3.exe.sts.txt](https://github.com/DARMA-tasking/vt/files/3811293/EMPIRE_PIC-NLS-release-beta3.exe.sts.txt)

The log files for each node refer to the sts file when tracing/logging events that occur. To see a full trace in projections, build with `-Dvt_trace_enabled=true`, run `examples/hello_world` with "--vt_trace --vt_trace_dir=testdir`. Download projections [Projections](http://charm.cs.illinois.edu/distrib/binaries/projections/projections_6.8.0.tar.gz). Open the sts file in the directory and you can look at the "Timeline" tool.

The demangler tries to parse out the object and function/functor type using compiler-specific functions for extracting. Snippet from `src/vt/utils/demangle/get_type_name.h`:

```
#if defined(__clang__)
  StaticString p = __PRETTY_FUNCTION__;
  return StaticString(p.data() + 31, p.size() - 31 - 1);

# elif defined(__GNUC__)
  StaticString p = __PRETTY_FUNCTION__;

#   if __cplusplus < 201402
    return StaticString(p.data() + 36, p.size() - 36 - 1);
#   else
    return StaticString(p.data() + 46, p.size() - 46 - 1);
#   endif

# elif defined(_MSC_VER)
  StaticString p = __FUNCSIG__;
  return StaticString(p.data() + 38, p.size() - 38 - 7);
#endif
```

This is therefore inconsistent and compiler-specific. Not only that, but the demangler does not actually parse the type but try to pull it apart with heuristics.

**To Reproduce**
1. Build VT in trace mode (`-Dvt_trace_enabled=true`)
2. Run `make test`

**Expected behavior**
Consistent types for all compilers on all tests.

**Platform (please complete the following information):**
 - All platforms

pnstickne:
"All compilers" is not reasonable as is is a compiler-specific detail (*maybe* improved in C++2a). Can support primary compilers and perform better extraction.

Support will be limited to approximately, as it is currently:

[...]

PR #576, opened 2019-11-19
575 event registry bugs

### LINKED ISSUE #575: Events not found (and causing crash on trace write)
During the trace write, all events must exist. However the events (that were previously registered) are not being found.

This appears to be an issue with initialization. The original code looks like it attempts to address the issue with template initialization ordering.

pnstickne:
"Works better". Pre-work to make #554 easier, as in 'not going to run into these issues'.

This will be part of #554 and as it contains trace-related bugs still.

- fixed (apparent?) initialization ordering issue where events add during initialization could not be found later in certain cases (differed by example project).
- fixed cross-access of iterators
- a bunch of updates for consistency, responsibility hiding, etc.

Fixes #575

### DIFF
--- a/src/vt/messaging/active.cc
+++ b/src/vt/messaging/active.cc
@@ -265,7 +265,7 @@ EventType ActiveMessenger::sendMsgSized(
     auto const handler = envelopeGetHandler(msg->env);
     bool const is_auto = HandlerManagerType::isHandlerAuto(handler);
     if (is_auto) {
-      trace::TraceEntryIDType ep = auto_registry::theTraceID(
+      trace::TraceEntryIDType ep = auto_registry::handlerTraceID(
         handler, auto_registry::RegistryTypeEnum::RegGeneral
       );
       if (not is_bcast) {
--- a/src/vt/messaging/active.cc
+++ b/src/vt/messaging/active.cc
@@ -298,19 +298,12 @@ EventType ActiveMessenger::sendMsgSized(
     base, uninitialized_destination, msg_size, true, &deliver
   );
 
-  EventType ret = no_event;
-
   if (deliver) {
-    EventType event = no_event;
-
     sendMsgBytesWithPut(dest, base, msg_size, send_tag);
-
-    ret = event;
-  } else {
-    ret = ret_event;
+    return no_event;
   }
 
-  return ret;
+  return ret_event;
 }
 
 ActiveMessenger::SendDataRetType ActiveMessenger::sendData(
--- a/src/vt/registry/auto/auto_registry_impl.h
+++ b/src/vt/registry/auto/auto_registry_impl.h
@@ -155,28 +155,7 @@ void setHandlerTraceName(std::string const& name, std::string const& parent) {
 inline void setTraceName(
   trace::TraceEntryIDType id, std::string const& name, std::string const& parent
 ) {
-#if backend_check_enabled(trace_enabled)
-  using TraceContainersType = trace::TraceRegistry::TraceContainersType;
-  auto event_iter = TraceContainersType::getEventContainer().find(id);
-  vtAssertExpr(event_iter != TraceContainersType::getEventContainer().end());
-  if (event_iter != TraceContainersType::getEventContainer().end()) {
-    if (name != "") {
-      event_iter->second.setEventName(name);
-    }
-    if (parent != "") {
-      auto type_id = event_iter->second.theEventTypeId();
-      auto iter = TraceContainersType::getEventTypeContainer().find(type_id);
-      vtAssertInfo(
-        iter != TraceContainersType::getEventTypeContainer().end(),
-        "Event type must exist",
-        name, parent, id, type_id
-      );
-      if (iter != TraceContainersType::getEventTypeContainer().end()) {
-        iter->second.setEventName(parent);
-      }
-    }
-  }
-#endif
+  trace::TraceRegistry::setTraceName(id, name, parent);
 }
 
 }} // end namespace vt::auto_registry
--- a/src/vt/registry/auto/auto_registry_impl.h
+++ b/src/vt/registry/auto/auto_registry_impl.h
@@ -130,7 +130,7 @@ void setHandlerTraceNameObjGroup(
 #if backend_check_enabled(trace_enabled)
   auto const handler = makeAutoHandlerObjGroup<ObjT,MsgT,f>(ctrl);
   auto const trace_id = theTraceID(handler, RegistryTypeEnum::RegObjGroup);
-  setTraceName(trace_id, name, parent);
+  trace::TraceRegistry::setTraceName(trace_id, name, parent);
 #endif
 }
 
@@ -139,7 +139,7 @@ void setHandlerTraceName(std::string const& name, std::string const& parent) {
 #if backend_check_enabled(trace_enabled)
   auto const handler = makeAutoHandler<MsgT,f>(nullptr);
   auto const trace_id = theTraceID(handler, RegistryTypeEnum::RegGeneral);
-  setTraceName(trace_id, name, parent);
+  trace::TraceRegistry::setTraceName(trace_id, name, parent);
 #endif
 }
 
@@ -148,14 +148,8 @@ void setHandlerTraceName(std::string const& name, std::string const& parent) {
 #if backend_check_enabled(trace_enabled)
   auto const handler = makeAutoHandlerParam<T,value>();
   auto const trace_id = theTraceID(handler, RegistryTypeEnum::RegGeneral);
-  setTraceName(trace_id, name, parent);
+  trace::TraceRegistry::setTraceName(trace_id, name, parent);
 #endif
-}
-
-inline void setTraceName(
-  trace::TraceEntryIDType id, std::string const& name, std::string const& parent
-) {
-  trace::TraceRegistry::setTraceName(id, name, parent);
 }
 
 }} // end namespace vt::auto_registry
--- a/src/vt/registry/auto/auto_registry_impl.h
+++ b/src/vt/registry/auto/auto_registry_impl.h
@@ -129,7 +129,7 @@ void setHandlerTraceNameObjGroup(
 ) {
 #if backend_check_enabled(trace_enabled)
   auto const handler = makeAutoHandlerObjGroup<ObjT,MsgT,f>(ctrl);
-  auto const trace_id = theTraceID(handler, RegistryTypeEnum::RegObjGroup);
+  auto const trace_id = handlerTraceID(handler, RegistryTypeEnum::RegObjGroup);
   trace::TraceRegistry::setTraceName(trace_id, name, parent);
 #endif
 }
@@ -138,7 +138,7 @@ template <typename MsgT, ActiveTypedFnType<MsgT>* f>
 void setHandlerTraceName(std::string const& name, std::string const& parent) {
 #if backend_check_enabled(trace_enabled)
   auto const handler = makeAutoHandler<MsgT,f>(nullptr);
-  auto const trace_id = theTraceID(handler, RegistryTypeEnum::RegGeneral);
+  auto const trace_id = handlerTraceID(handler, RegistryTypeEnum::RegGeneral);
   trace::TraceRegistry::setTraceName(trace_id, name, parent);
 #endif
 }
@@ -147,7 +147,7 @@ template <typename T, T value>
 void setHandlerTraceName(std::string const& name, std::string const& parent) {
 #if backend_check_enabled(trace_enabled)
   auto const handler = makeAutoHandlerParam<T,value>();
-  auto const trace_id = theTraceID(handler, RegistryTypeEnum::RegGeneral);
+  auto const trace_id = handlerTraceID(handler, RegistryTypeEnum::RegGeneral);
   trace::TraceRegistry::setTraceName(trace_id, name, parent);
 #endif
 }
--- a/src/vt/trace/trace.cc
+++ b/src/vt/trace/trace.cc
@@ -643,11 +643,8 @@ void Trace::writeLogFile(gzFile file, TraceContainerType const& traces) {
       std::underlying_type<decltype(log->type)>::type
     >(log->type);
 
-    auto* events = TraceContainersType::getEventContainer();
-    auto event_iter = events->find(log->ep);
-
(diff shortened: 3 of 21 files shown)


src/vt/trace/trace.cc
```diff
@@ -817,15 +822,15 @@ void Trace::outputControlFile(std::ofstream& file) {
   ContainerEventSortedType sorted_event;
   ContainerEventTypeSortedType sorted_event_type;
 
-  for (auto&& elem : TraceContainersType::getEventContainer()) {
+  for (auto&& elem : *TraceContainersType::getEventContainer()) {
```
pnstickne: Probably a better way to do this than `*`. Copy-constructor in that case?

src/vt/trace/trace_containers.cc
```diff
@@ -0,0 +1,54 @@
...
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+//
+// Questions? Contact darma@sandia.gov
+//
+// *****************************************************************************
+//@HEADER
+*/
+
+#include "vt/config.h"
```
pnstickne: Should remove non-necessary headers.

src/vt/trace/trace_containers.h
```diff
@@ -70,32 +70,29 @@ using TraceContainerEventClassType = EventLookupType<TraceEntryIDType, EventClas
 
 struct Trace;
 
-// Use static template initialization pattern to deal with ordering issues with
-// auto-registry
-template <typename = void>
+// Container types are created-as-needed, which occurs during
```
pnstickne: "After making these changes, events are found in the programs that don't crash". As such, I suspect it's due to an initialization ordering issue.

src/vt/trace/trace_containers.h
```diff
@@ -70,32 +70,29 @@ using TraceContainerEventClassType = EventLookupType<TraceEntryIDType, EventClas
...
+// intialization to avoid intiailization ordering issues.
 class TraceContainers {
  public:
-  static TraceContainerEventClassType& getEventTypeContainer(){
-    return event_type_container;
+  static TraceContainerEventClassType* getEventTypeContainer(){
+    if (not event_type_container_)
```
pnstickne: Could hide the pointers behind a wrapper, although seems legit as so.

src/vt/trace/trace_event.cc
```diff
@@ -55,7 +55,7 @@ EventClass::EventClass(
 ) : event(in_event), hash_event(in_hash_event)
 {
   auto const& event_hash =  std::hash<std::string>{}(in_hash_event);
-  this_event_ = event_hash;
+  this_event_ = event_hash == 0 ? 1 : event_hash;
```
pnstickne: Can *technically* occur. Also future changes planned to ensure that the ID is guaranteed unique instead of probably unique. Many sides, still dice.

src/vt/trace/trace_registry.cc
```diff
@@ -0,0 +1,120 @@
...
+#include "vt/trace/trace_event.h"
+#include "vt/trace/trace_containers.h"
+#include "vt/trace/trace_registry.h"
+
+namespace vt { namespace trace {
+
+/*static*/ TraceEntryIDType
+TraceRegistry::registerEventHashed(
+    std::string const& event_type_name, std::string const& event_name
+  ) {
```
pnstickne: Extra space.

src/vt/trace/trace_registry.cc
```diff
@@ -0,0 +1,120 @@
...
+  } else {
+    event_seq = event_iter->second.theEventSeq();
+  }
+
+  return new_event.theEventId();
+}
+
+/*static*/ void
+TraceRegistry::setTraceName(
+  TraceEntryIDType id, std::string const& name, std::string const& parent) {
```
pnstickne: `)` should trail for consistency.

src/vt/trace/trace_registry.cc
```diff
@@ -0,0 +1,120 @@
...
+      );
+      if (iter != event_types->end()) {
+        iter->second.setEventName(parent);
+      }
+    }
+  }
+#endif
+}
+
+/*static*/ bool TraceRegistry::getEventSequence(TraceEntryIDType id, TraceEntryIDType &seq) {
```
pnstickne: Formatting..

pnstickne: Maybe expose the event object directly.

src/vt/trace/trace_containers.cc
```diff
@@ -0,0 +1,54 @@
...
+#include "vt/trace/trace_common.h"
+#include "vt/trace/trace_containers.h"
+
+namespace vt { namespace trace {
+
+/*static*/ TraceContainerEventClassType* TraceContainers::event_type_container_ = nullptr;
```
pnstickne: Maybe brace-initializer. Not quite sure how this is guaranteed pre-C++17.

src/vt/trace/trace_containers.cc
```diff
@@ -0,0 +1,54 @@
+/*
+//@HEADER
+// *****************************************************************************
+//
+//                              trace_containers.h
```
pnstickne: Lies..

src/vt/trace/trace.cc
```diff
@@ -629,19 +631,20 @@ void Trace::writeTracesFile() {
...
 
-    auto event_iter = TraceContainersType::getEventContainer().find(log->ep);
-
-    vtAssert(
-      log->ep == no_trace_entry_id or
-      event_iter != TraceContainersType::getEventContainer().end(),
-      "Event must exist that was logged"
-    );
+    auto* events = TraceContainersType::getEventContainer();
+    auto event_iter = events->find(log->ep);
```
pnstickne: Can just use `getEventSequence`..

src/vt/trace/trace.cc
```diff
@@ -452,18 +462,19 @@ TraceEventIDType Trace::logEvent(LogPtrType log) {
 
   auto grouped_begin = [&]() -> TraceEventIDType {
     if (not open_events_.empty()) {
+      LogType* top_event = open_events_.top();
```
pnstickne: Could unify with following..

src/vt/registry/auto/auto_registry_impl.h
```diff
@@ -155,28 +155,7 @@ void setHandlerTraceName(std::string const& name, std::string const& parent) {
...
-        "Event type must exist",
-        name, parent, id, type_id
-      );
-      if (iter != TraceContainersType::getEventTypeContainer().end()) {
-        iter->second.setEventName(parent);
-      }
-    }
-  }
-#endif
+  trace::TraceRegistry::setTraceName(id, name, parent);
```
pnstickne: Could pull up/out further.

src/vt/trace/trace_event.cc
pnstickne: Actually, since the MSVC 10 issue it seems like all hashes are FNV64 or-better, so assuming size_t is 2^64..

src/vt/trace/trace.cc
pnstickne: Done.

src/vt/trace/trace_containers.h
pnstickne: (Where 'these changes' are specifically related to returning a new'ed pointer instead of a reference to a static object created during initialization somewhere..)

src/vt/trace/trace.cc
pnstickne: Actually, "shouldn't" make a copy in this case. Seems legit.

src/vt/trace/trace_registry.cc
```diff
@@ -0,0 +1,128 @@
...
+      if (iter != event_types->end()) {
+        iter->second.setEventName(type_name);
+      }
+    }
+  }
+#endif
+}
+
+/*static*/ bool
+TraceRegistry::getEventSequence(TraceEntryIDType id, TraceEntryIDType &seq) {
```
pnstickne: No "0 sentinel" for TraceEntryIDType.. maybe it should be a TraceEntrySeqType split on *signed* int so < 0 is sentinel.

src/vt/trace/trace_containers.h
lifflander: With your changes, is all this still happening at static init time? It must execute then for correctness, except for the user's override `setTraceName`, if the user desires to override the given name.

src/vt/trace/trace_containers.h
```diff
@@ -50,70 +50,36 @@
...
-// Use static template initialization pattern to deal with ordering issues with
-// auto-registry
-template <typename = void>
+// Container types are created-as-needed, which occurs during
+// intialization to avoid intiailization ordering issues.
 class TraceContainers {
  public:
-  static TraceContainerEventClassType& getEventTypeContainer(){
-    return event_type_container;
+  static TraceContainerEventClassType* getEventTypeContainer(){
```
lifflander: What is the motivation of hoisting this to a pointer?

src/vt/trace/trace_containers.cc
```diff
@@ -0,0 +1,12 @@
+#include "vt/trace/trace_common.h"
+#include "vt/trace/trace_containers.h"
+
+namespace vt { namespace trace {
+
+/*static*/ TraceContainerEventClassType*
```
lifflander: I think you should initialize this variable to `nullptr`

src/vt/messaging/active.cc
```diff
@@ -298,19 +298,12 @@ EventType ActiveMessenger::sendMsgSized(
     base, uninitialized_destination, msg_size, true, &deliver
   );
 
-  EventType ret = no_event;
-
   if (deliver) {
-    EventType event = no_event;
-
     sendMsgBytesWithPut(dest, base, msg_size, send_tag);
```
lifflander: I think there was a bug here before, or maybe the code change. I thought that `sendMsgBytesWithPut` should return an event. I think the code was refactored a long time ago not to, but this code didn't get updated.

src/vt/trace/trace.cc
```diff
@@ -573,21 +588,11 @@ TraceEventIDType Trace::logEvent(LogPtrType log) {
...
-      return true;
-    } else if (node % ArgType::vt_trace_mod == 1) {
-      return true;
-    } else {
-      return false;
-    }
-  } else {
-    return false;
-  }
+  return ArgType::vt_trace
```
lifflander: Nice cleanup!

src/vt/trace/trace_containers.h
```diff
@@ -50,70 +50,36 @@
...
+    if (not event_type_container_)
+      event_type_container_ = new TraceContainerEventClassType();
```
lifflander: I don't see any `delete` almost sure that this leaks memory. Why are you using a manually allocated object? I would steer away from this. Ideally, C++ code should never include a `new`

src/vt/trace/trace_containers.cc
pnstickne: Without the assignment it's guaranteed to be assigned all-zeros in the STATIC initialization. It's not until C++17 that assignment (in certain cases) is done during static initialization.

src/vt/trace/trace_containers.h
pnstickne: Never deleted. Global statics. Used when events are written.

pnstickne: `static_lifetime_events_` (eg.)

src/vt/trace/trace_containers.h
pnstickne: This could probably also be done with initialization of a non-pointer type (and separate variable capturing initialization state). Main change is it allows detection of initialization.

pnstickne: if (not event_container_intiailized_) {
       event_container_ = EventContainerType{};
       event_container_intiailized_ = true;
   }
   return event_container_;

src/vt/trace/trace_containers.cc
pnstickne: (I had it first assigning to nullptr and then removed it - a comment would probably be useful.)

src/vt/trace/trace_containers.h
pnstickne: However, even then my concern was it might be *re-initialized* later on.

src/vt/trace/trace_containers.cc
pnstickne: @lifflander It looks like a `unique_ptr` would still work with the default ctor (as it is `constexpr`); don't quite know what C++ level guarantees constexpr as static non-local initialization. Would still be returned as a raw ptr, accessed from `get` to usages.

pnstickne: - https://en.cppreference.com/w/cpp/memory/unique_ptr/unique_ptr

pnstickne: > Constant initialization is performed after (until C++14)instead of (since C++14) zero initialization of the static and thread-local objects and before all other initialization. Only the following variables are constant initialized:

> 2) .. Static or thread-local object of class type that is initialized by a constructor call, if the constructor is constexpr ..

pnstickne: Also then a shared_ptr: `constexpr shared_ptr() noexcept;`.

pnstickne: Violated by unordered_map: https://en.cppreference.com/w/cpp/container/unordered_map/unordered_map

nlslatt:
@lifflander I don't know if it's related to this PR or not (I'm using beta.4 plus this PR on Mutrino), but I'm seeing an entry method hidden by Projections because there are no tracing events associated with it, even though it should have run many times.  I'm specifically talking about Frontend::requestFields, which is called unconditionally by Backend::updateFieldSolutions (which itself does appear in the traces many times).

lifflander:
> @lifflander I don't know if it's related to this PR or not (I'm using beta.4 plus this PR on Mutrino), but I'm seeing an entry method hidden by Projections because there are no tracing events associated with it, even though it should have run many times. I'm specifically talking about Frontend::requestFields, which is called unconditionally by Backend::updateFieldSolutions (which itself does appear in the traces many times).

Was this also happening before 575 was being used?

PR #578, opened 2019-11-19
554 demangler parsing fixes for tracing

### LINKED ISSUE #554: Trace builds of VT crash early in startup when certain callback kinds are registered (eg. reduce.cc, ~1/8 tests),

pnstickne:
Basic idea of unification and pushing out responsibilities. Still need to wire in the arguments.

Drafted to 554 as a pre-req.

Fixes #554


src/vt/registry/auto/auto_registry_general.h
```diff
@@ -47,19 +47,93 @@
...
+template <typename ObjTypeT, typename... Args>
+struct FunctorAdapterArgs {
+  using FunctionPtrType = void (*)(Args...);
+  using ObjType = ObjTypeT;
+
+  static constexpr FunctionPtrType getFunction() {
+    return &proxyOperatorToNewInstanceReg<ObjType, Args...>;
+  }
+
+  static std::string traceGetEventType() {
```
pnstickne: Extraction of trace information pushed as close to types as seems practical. `__PRETTY_PRINT__` are Clang/GCC/ICC(?) specific, but one might be able to get traces (FSVO) with some implementations of `type_id` (and a CXX type demangler).

The parsing code is very minimal (or rather, it's just a few dress-up helpers) and avoids some SEGV's what were occurring.

src/vt/registry/auto/auto_registry_general.h
```diff
@@ -47,19 +47,93 @@
...
+    using TE = vt::util::demangle::TemplateExtract;
+    using DU = vt::util::demangle::DemanglerUtils;
+    auto ns = TE::getTypeName<ObjTypeT>();
+    return DU::removeSpaces(ns);
+  }
+
+  static std::string traceGetEventName() {
+    using TE = vt::util::demangle::TemplateExtract;
+    using DU = vt::util::demangle::DemanglerUtils;
+    //auto args = DU::join(",", TE::getTypeNames<Args...>());
```
pnstickne: Recursive template fail.. TODO.

src/vt/registry/auto/auto_registry_general.h
```diff
@@ -69,10 +143,24 @@ struct FunctorAdapterMember {
...
+  static std::string traceGetEventName() {
+    using TE = vt::util::demangle::TemplateExtract;
+    using DU = vt::util::demangle::DemanglerUtils;
+    auto barename = TE::getBarename(TE::getValueName<F,f>());
+    auto args = TE::getVoidFuncStrArgs(TE::getTypeName<F>());
+    return barename + "(" + DU::removeSpaces(args) + ")";
+  }
+
+  static size_t getNumArgs() {
+    return 0; // lies - see NumArgsTag, perhaps
```
pnstickne: Is used/relevant for these types? :-/

src/vt/registry/auto/auto_registry_general_impl.h
```diff
@@ -46,53 +46,52 @@
...
 
-template <typename ActFnT, typename RegT, typename InfoT, typename FnT>
+template <typename RunnableT, typename RegT, typename InfoT, typename FnT>
 AutoHandlerType registerActiveGen() {
-  return RegistrarWrapperGen<ActFnT, RegT, InfoT, FnT>().registrar.index;
+  return RegistrarWrapperGen<RunnableT, RegT, InfoT, FnT>().registrar.index;
 }
 
-template <typename ActFnT, typename RegT, typename InfoT, typename FnT>
+template <typename RunnableT, typename RegT, typename InfoT, typename FnT>
```
pnstickne: Runnable-needs-Adapter (RunnableT::AdapterType), both are separate template types.

src/vt/registry/auto/functor/auto_registry_functor_impl.h
```diff
@@ -58,80 +57,38 @@ namespace vt { namespace auto_registry {
...
-template <typename FunctorT, typename... Args>
-static inline auto functorHandlerWrapperRval(Args&&... args) {
-  typename FunctorT::FunctorType instance;
-  return instance.operator()(std::forward<Args>(args)...);
-}
+template <typename RunnableT, typename RegT, typename InfoT, typename FnT>
+RegistrarFunctor<RunnableT, RegT, InfoT, FnT>::RegistrarFunctor() {
+  using AdapterType = typename RunnableT::AdapterType;
 
-template <typename FunctorT, typename... Args>
```
pnstickne: Now contained in the Adapter.

pnstickne: Was able to (had to?) eliminate the msg differentiation on forwarding.. not.. sure if such is required.

src/vt/registry/auto/auto_registry_general.h
```diff
@@ -69,10 +143,24 @@ struct FunctorAdapterMember {
...
+    auto ns = TE::getNamespace(TE::getValueName<F,f>());
+    return DU::removeSpaces(ns);
+  }
+
+  static std::string traceGetEventName() {
+    using TE = vt::util::demangle::TemplateExtract;
+    using DU = vt::util::demangle::DemanglerUtils;
+    auto barename = TE::getBarename(TE::getValueName<F,f>());
+    auto args = TE::getVoidFuncStrArgs(TE::getTypeName<F>());
+    return barename + "(" + DU::removeSpaces(args) + ")";
```
pnstickne: Should wrap all bar ".." in std::string("..") or, does even matter? Or is there a better way to handle string concat (as in, easier).

src/vt/registry/auto/auto_registry_general.h
```diff
@@ -69,10 +143,24 @@ struct FunctorAdapterMember {
 
   static constexpr FunctionPtrType getFunction() { return f; }
 
-  template <typename... A>
-  auto operator()(A&&... a) -> decltype(auto) {
-    return f(std::forward<A>(a)...);
-   }
+  static std::string traceGetEventType() {
```
pnstickne: Should/could be wrapped in `#if .. trace` perhaps.

src/vt/registry/auto/auto_registry_general.h
```diff
@@ -47,19 +47,93 @@
...
+static inline auto proxyOperatorToNewInstanceReg(Args... args) {
+  T instance;
+  return instance.operator()(args...);
+}
+
+/// All Functor Adapters require:
+/// - FunctionPtrType
+/// - ObjType
+/// - getFunction
+/// - traceGetEventType
```
pnstickne: ..for when trace is enabled.

src/vt/registry/auto/auto_registry_general.h
```diff
@@ -47,19 +47,93 @@
...
+/// - traceGetEventName
+/// (The 'Functor' naming is a bit of a misnomer as they are not functors
+///  nor do they strictly wrap functors..)
+
+/// Adapt a type that provides operator()(..)
+/// NOTE:
+/// MULTIPLE INSTANCES of the type will be created and discarded.
+/// This cannot be used for a stateful instance.
+/// This is an implementation detail that could be reconsidered.
+template <typename ObjTypeT, typename... Args>
```
pnstickne: Pulled from the functor header so that adapters are side x side.

pnstickne: This adapter is used 'very infrequently' in the examples.

pnstickne:
And, now collects `operator(Args..)` case .. GCC builds still hosed; warnings, and more importantly: I've got my lib's crossed >_<.

pnstickne:
Tests still don't pass .. all examples run to completion good STS output.

pnstickne:
@lifflander This is ready to go. I would merge this as it also includes the #575 issues and fixes SEGV's that occur in several examples (including 'reduce).

Only one test fails locally, and I believe that to be a test consistency issue (sometimes is it transpose_4, transpose_8, or both).

> 99% tests passed, 1 tests failed out of 681
> 
> Total Test time (real) =  89.61 sec
> 
> The following tests FAILED:
> 	132 - vt:transpose_8 (Failed)
> Errors while running CTest
> make: *** [test] Error 8

src/vt/registry/auto/functor/auto_registry_functor_impl.h
```diff
@@ -58,80 +57,38 @@ namespace vt { namespace auto_registry {
...
-static inline auto functorHandlerWrapperReg(Args... args) {
-  typename FunctorT::FunctorType instance;
-  return instance.operator()(args...);
-}
+  RegT& reg = getAutoRegistryGen<RegT>();
+  index = reg.size(); // capture current index
```
pnstickne: Uh, how is this code compiling? What `index` is being used?

pnstickne:
@lifflander Back-merged in 575.

src/vt/registry/auto/functor/auto_registry_functor_impl.h
pnstickne: Derp.

lifflander:
I just rebased this on latest develop and changed the target.

```
/Users/jliffla/codes/vt/virtual-transport/src/vt/utils/demangle/demangle.cc:159:9: warning: comparison of unsigned expression >= 0 is always true [-Wtautological-compare]
  if (e >= 0 && typestr[e] == ')') {
      ~ ^  ~
1 warning generated.
```

@pnstickne Can you fix this warning? I'm not sure if there is actually a bug here or if this just needs to be removed?

src/vt/registry/auto/auto_registry_general.h
lifflander: Indeed, it's an alternative interface.

src/vt/registry/auto/auto_registry_general.h
lifflander: I don't think it matters. We are not worried about speed to much in this case---otherwise we would use `std::stringstream`

src/vt/registry/auto/functor/auto_registry_functor_impl.h
lifflander: Great

pnstickne:
Oh goodness. That's a useful warning as std::nopos is the correct compare.. still used to -1 for not found. I'll get in correction.

pnstickne:
Warning should be gone.

pnstickne:
Super super recommend. It also means that I can run tests with tracing mode enabled 💃

pnstickne:
Having this in whatever main WIP would be so handy to avoid "random yet predictable crashes" failures during startup of various programs.

PR #590, opened 2019-11-25
#589 Do not call LB on single node (fixes FPE in stats)

### LINKED ISSUE #589: Stop calling into LB with 1 node (FPE in stats)
**Describe the bug**

lifflander:
Fixes #589


pnstickne:
Maybe prevent all VT one one node? :D

PR #593, opened 2019-11-26
#592 vtAbort if LB compile-time disabled but runtime enabled

### LINKED ISSUE #592: Error out if LB is compile-time disabled and runtime-enabled
**Describe the bug**

lifflander:
Fixes #592

### DIFF
--- a/src/vt/runtime/runtime.cc
+++ b/src/vt/runtime/runtime.cc
@@ -711,6 +711,12 @@ void Runtime::printStartupBanner() {
 
   //fmt::print("{}\n", reset);
   fmt::print(reset);
+
+  // Enqueue a check for later in case arguments are modified before work
+  // actually executes
+  theSched->enqueue([this]{
+    this->checkForArgumentErrors();
+  });
 }
 
 void Runtime::printShutdownBanner(term::TermCounterType const& num_units) {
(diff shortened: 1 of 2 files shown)


src/vt/runtime/runtime.cc
```diff
@@ -711,6 +711,12 @@ void Runtime::printStartupBanner() {
 
   //fmt::print("{}\n", reset);
   fmt::print(reset);
+
+  // Enqueue a check for later in case arguments are modified before work
+  // actually executes
+  theSched->enqueue([this]{
+    this->checkForArgumentErrors();
+  });
```
PhilMiller: Isn't this a bit racy? How do you ensure this is after whatever last point that arguments could safely be modified?

The principled solution to that issue would distinguish the name and type of "structure containing arguments as they are being processed" from "structure containing arguments as they should be used to initialize and execute", with the latter of course being immutable.

nlslatt: @pbmille You want beta.4 merged soon, right?

lifflander: Yes, this only works if EMPIRE doesn't call into the scheduler before modifying the args. We are looking for a solution where EMPIRE doesn't have to change.

PR #594, opened 2019-11-26
#591 Fix bad captured stack pointer

### LINKED ISSUE #591: Bug in group (tutorial 32 ranks crash)
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

lifflander:
Fixes #591

### DIFF
--- a/src/vt/group/group_manager.cc
+++ b/src/vt/group/group_manager.cc
@@ -375,9 +375,13 @@ EventType GroupManager::sendGroupCollective(
       return no_event;
     }
   } else if (in_group && !group_ready) {
-    local_collective_group_info_.find(group)->second->readyAction([=]{
-      theGroup()->sendGroupCollective(base,from,size,is_root,deliver);
-    });
+    local_collective_group_info_.find(group)->second->readyAction(
+      [base,from,size,is_root]{
+        // Do not capture deliver, it's a pointer to stack memory
+        bool dummy;
+        theGroup()->sendGroupCollective(base,from,size,is_root,&dummy);
+      }
+    );
     *deliver = true;
     return no_event;
   } else {


src/vt/group/group_manager.cc
```diff
@@ -375,9 +375,13 @@ EventType GroupManager::sendGroupCollective(
       return no_event;
     }
   } else if (in_group && !group_ready) {
-    local_collective_group_info_.find(group)->second->readyAction([=]{
-      theGroup()->sendGroupCollective(base,from,size,is_root,deliver);
-    });
+    local_collective_group_info_.find(group)->second->readyAction(
```
pnstickne: Did I call it or what? :D

pnstickne: (Well, tangentially anyway..)

src/vt/group/group_manager.cc
```diff
@@ -375,9 +375,13 @@ EventType GroupManager::sendGroupCollective(
...
+      [base,from,size,is_root]{
+        // Do not capture deliver, it's a pointer to stack memory
+        bool dummy;
```
pnstickne: Can be any value in? Although with [=] it could still have done a *copy* capture, no?

PhilMiller: I'd stick with strictly referring to that as capture *by value* - in this case, the value of the pointer, which contained the address of a local stack variable. Subsequent dereferencing of such a pointer is, of course, an error, since it occurs after the end of the lifetime of the object referenced.

I don't think a C++ compiler could reliably warn about this, since it can't tell when the lambda will be called. The Rust compiler probably could, since it would require that the lifetime of the lamdba not exceed the lifetime of any object on which it holds a borrowed reference. The C++ Core Guidelines effort may facilitate doing more Rust-like analysis in C++, but that's not here yet.

src/vt/group/group_manager.cc
lifflander: Yes, you did

PR #596, opened 2019-11-29
595 support gtest 110

pnstickne:
"Compatibility shim" to enable VT to work sans-warnings (and sans-errors!) targeting either test 1.8 or gtest 1.10.

I usually merge this branch in locally as some compiler / gtest versions do not compile otherwise.

### DIFF
--- a/tests/unit/active/test_active_bcast_put.cc
+++ b/tests/unit/active/test_active_bcast_put.cc
@@ -42,7 +42,7 @@
 //@HEADER
 */
 
-#include <gtest/gtest.h>
+#include "vt_gtest.h"
 
 #include "test_parallel_harness.h"
 #include "data_message.h"
@@ -142,9 +142,9 @@ TEST_P(TestActiveBroadcastPut, test_type_safe_active_fn_bcast2) {
   }
 }
 
-INSTANTIATE_TEST_CASE_P(
+INSTANTIATE_TEST_SUITE_P_VT(
   InstantiationName, TestActiveBroadcastPut,
-  ::testing::Range(static_cast<NodeType>(0), static_cast<NodeType>(16), 1),
+  ::testing::Range(static_cast<NodeType>(0), static_cast<NodeType>(16), 1)
 );
 
 }}} // end namespace vt::tests::unit
--- a/tests/unit/termination/test_termination_action_nested_rooted.cc
+++ b/tests/unit/termination/test_termination_action_nested_rooted.cc
@@ -72,17 +72,17 @@ struct TestTermNestedRooted : action::BaseFixture {
   }
 };
 
-TEST_P(TestTermNestedRooted, test_term_detect_nested_rooted_epoch) /* NOLINT */{
+TEST_P(TestTermNestedRooted, test_term_detect_nested_rooted_epoch) {
   kernel(depth_, no_epoch);
 }
 
-INSTANTIATE_TEST_CASE_P /*NOLINT*/(
+INSTANTIATE_TEST_SUITE_P_VT(
   InstantiationName, TestTermNestedRooted,
   ::testing::Combine(
     ::testing::Range(0, 3),
     ::testing::Bool(),
     ::testing::Range(2, 10, 2)
-  ),
+  )
 );
 
 }}} // end namespace vt::tests::unit
--- a/tests/unit/vt_gtest.h
+++ b/tests/unit/vt_gtest.h
@@ -1,0 +1,101 @@
+/*
+//@HEADER
+// *****************************************************************************
+//
+//                                vt_gtest.h
+//                           DARMA Toolkit v. 1.0.0
+//                       DARMA/vt => Virtual Transport
+//
+// Copyright 2019 National Technology & Engineering Solutions of Sandia, LLC
+// (NTESS). Under the terms of Contract DE-NA0003525 with NTESS, the U.S.
+// Government retains certain rights in this software.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+// * Redistributions of source code must retain the above copyright notice,
+//   this list of conditions and the following disclaimer.
+//
+// * Redistributions in binary form must reproduce the above copyright notice,
+//   this list of conditions and the following disclaimer in the documentation
+//   and/or other materials provided with the distribution.
+//
+// * Neither the name of the copyright holder nor the names of its
+//   contributors may be used to endorse or promote products derived from this
+//   software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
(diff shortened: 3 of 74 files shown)


tests/unit/vt_gtest.h
```diff
@@ -0,0 +1,101 @@
...
+///
+/// Use
+///     #include "vt_gtest.h"
+/// instead of
+///     #include <gtest/gtest.h>
+///
+/// This is secondary from election to use a standard VT harness.
+///
+
+#ifndef __VIRTUAL_TRANSPORT_VT_GTEST__
```
pnstickne: Should regenerate these guards.

tests/unit/vt_gtest.h
```diff
@@ -0,0 +1,101 @@
...
+
+#define TYPED_TEST_SUITE_P TYPED_TEST_CASE_P
+#define REGISTER_TYPED_TEST_SUITE_P REGISTER_TYPED_TEST_CASE_P
+#define INSTANTIATE_TYPED_TEST_SUITE_P INSTANTIATE_TYPED_TEST_CASE_P
+#define INSTANTIATE_TEST_SUITE_P INSTANTIATE_TEST_CASE_P
+
+// Unified macros to avoid having to specify a trailing comma to avoid
+// pedantic warning about variadic arguments relating to name generation.
+// Use the originals if wishing for name generators.
+#define INSTANTIATE_TYPED_TEST_SUITE_P_VT(Prefix, Suite, Types) \
```
pnstickne: Unfortunately, can't rename macros..

tests/unit/termination/test_termination_action_nested_rooted.cc
```diff
@@ -72,17 +72,17 @@ struct TestTermNestedRooted : action::BaseFixture {
   }
 };
 
-TEST_P(TestTermNestedRooted, test_term_detect_nested_rooted_epoch) /* NOLINT */{
+TEST_P(TestTermNestedRooted, test_term_detect_nested_rooted_epoch) {
   kernel(depth_, no_epoch);
 }
 
-INSTANTIATE_TEST_CASE_P /*NOLINT*/(
+INSTANTIATE_TEST_SUITE_P_VT(
```
pnstickne: "_VT" postfix forms don't take in a ... for name generation and thus don't need (and can't accept) a trailing comma to avoid va-arg/list warning.

tests/unit/active/test_active_bcast_put.cc
```diff
@@ -142,9 +142,9 @@ TEST_P(TestActiveBroadcastPut, test_type_safe_active_fn_bcast2) {
   }
 }
 
-INSTANTIATE_TEST_CASE_P(
+INSTANTIATE_TEST_SUITE_P_VT(
```
pnstickne: Macro also provides compatibility between 1.8 and 1.10 by not exposing the REMOVED support for name-generation in this form.

lifflander:
@nlslatt Can you comment on how this will effect EMPIRE and possibly conflict with Trilinos?

nlslatt:
> @nlslatt Can you comment on how this will effect EMPIRE and possibly conflict with Trilinos?

This will break EMPIRE.  I can provide more details later.

EMPIRE requires that the gtest within the Trilinos install be turned off so as to avoid a conflict.  Do you want to make that same requirement of the Trilinos install used to provide Kokkos to vt?

pnstickne:
@nlslatt Is there something specific in this PR that will break EMPIRE, or is that more specific to a later switch to GoogleTest 1.10+?

These changes are a small compatibility shim to eliminate all warnings (and some errors :-/) for both gtest 1.8 and gtest 1.10 targets. Only VT code should use the "vt_gtest.h" header..

I have no data-points for other gtest version compatibility or what "gtest/gtest.h" resolve to in the EMPIRE build process.

pnstickne:
Uh oh. I didn't mean to leave the test discovery on this branch. Pulling.

tests/unit/vt_gtest.h
```diff
@@ -0,0 +1,101 @@
...
+  INSTANTIATE_TYPED_TEST_CASE_P(Prefix, Suite, Types, /*...*/)
```
nlslatt: This will break EMPIRE because EMPIRE's gtest only expects 3 arguments:
`# define INSTANTIATE_TYPED_TEST_CASE_P(Prefix, CaseName, Types)`

tests/unit/vt_gtest.h
```diff
@@ -0,0 +1,101 @@
...
+// (The resulting tests are ill-defined against an incompatile gtest lib.)
+//#define GTEST_REMOVE_LEGACY_TEST_CASEAPI_ 1
+
+#include <gtest/gtest.h>
+
+// 1.8 -> 1.10 changes "Test Case" to "Test Suite" and deprecates
+// the former; these are forward-compatible shims for gtest 1.8 builds.
+// Larger changes to method names and such can't be done pre-1.10.
+
+#ifndef TYPED_TEST_SUITE_P // gtest < 1.10 (should be 1.8)
```
nlslatt: EMPIRE uses a version at least compatible with gtest sha `43863938377a9ea` (dated Wed Sep 20 14:32:57 2017 -0400), which is what `vt-auto-build` uses.  I'm not sure what this corresponds to in terms of releases.

nlslatt:
> Yes, there are specific things in this PR that will break EMPIRE. Please see review comments for details.

When I say it will break EMPIRE, I mean that turning on checkpoint/vt tests in an EMPIRE build (which happens automatically when gtest is enabled in EMPIRE) will cause the build to fail.

I'm clarifying this because there is technically more than one solution to this problem.  It would be very nice to maintain the testability of vt itself within an EMPIRE build.  This could be important for cases when EMPIRE+vt is behaving badly on some platform and we want to know if that is because vt is behaving badly.

pnstickne:
> I'm clarifying this because there is technically more than one solution to this problem. It would be very nice to maintain the testability of vt itself within an EMPIRE build. This could be important for cases when EMPIRE+vt is behaving badly on some platform and we want to know if that is because vt is behaving badly.

I support that end goal with a more robust build. This set of changes should be more accurately summarized as "reduce friction when updating 1.8/1.8.1 through 1.10", without actually mandating the gtest version dependency is increased. (VT tests would be expected to compile and run against 1.8..1.10 versions.)

It appears 43863938377a9ea is between gtest versions 1.8 and 1.8.1:

    $ git tag --contains 43863938377a9ea
    release-1.10.0
    release-1.8.1
    v1.10.x

I'll verify compatibility against the specific commit.

tests/unit/vt_gtest.h
```diff
@@ -0,0 +1,101 @@
...
+///     #include <gtest/gtest.h>
+///
+/// This is secondary from election to use a standard VT harness.
+///
+
+#ifndef __VIRTUAL_TRANSPORT_VT_GTEST__
+#define __VIRTUAL_TRANSPORT_VT_GTEST__
+
+// This define detects some future-breaking API without updating gtest target.
+// (The resulting tests are ill-defined against an incompatile gtest lib.)
```
PhilMiller: Typo here: incompatible

pnstickne:
Going to move gtest into VT. Then this is not (strictly) necessary.

Issue #599, opened 2019-12-02
Event logging leaks memory


pnstickne:
- Events written to nodes which do not emit traces are leaked

Does *NOT* address the issue of growing-events-until-termination, which *should be handled as follow-up to the event flushing*.

pnstickne:
To be precise, this is to enable support for no-leaks when flushing will incrementally clear the stored event log.

pnstickne:
Closing, as integrated into #241 on #551

PR #600, opened 2019-12-02
#599 ensure events Log objects are properly managed

pnstickne:
- Avoids valgrind warnings - mostly merge-friendly with #241.

- Defines when ownership is accepted and is future-ready
  for a change to unique_ptr, if such is desired.

- Eliminates most post-construction modification of events.

### DIFF
--- a/src/vt/trace/trace.cc
+++ b/src/vt/trace/trace.cc
@@ -58,6 +58,8 @@
 
 namespace vt { namespace trace {
 
+using LogType = Trace::LogType;
+
 template <typename EventT>
 struct TraceEventSeqCompare {
   bool operator()(EventT* const a, EventT* const b) const {
@@ -154,12 +156,7 @@ void Trace::addUserNote(std::string const& note) {
   auto const type = TraceConstantsType::UserSuppliedNote;
   auto const time = getCurrentTime();
 
-  LogPtrType log = new LogType(time, type);
-  logEvent(log);
-
-  editLastEntry([note](LogPtrType entry) {
-    entry->setUserNote(note);
-  });
+  logEvent(new LogType(time, type, note, Log::UserDataType{}));
 }
 
 void Trace::addUserData(int32_t data) {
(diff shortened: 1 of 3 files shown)


pnstickne:
Ended up switching to unique_ptr as it should make memory ownership slightly more clear between traces and the begin/end event stack. The location where `traces_.clear()` is done should probably not be ignored ignore the merge to #241; it's also a trivial location to perform as-flushed memory management.

pnstickne:
This can go in as it is while decreases the "true" memory leaks, but will require integration with #241 (mainly, collection management) to actually see reduction. (#241 as it is now will bring some memory reduction by side-effect to the non-unique_ptr approach, although still leave dangling pointers.)

src/vt/trace/trace.cc
```diff
@@ -154,12 +156,7 @@ void Trace::addUserNote(std::string const& note) {
   auto const type = TraceConstantsType::UserSuppliedNote;
   auto const time = getCurrentTime();
 
-  LogPtrType log = new LogType(time, type);
-  logEvent(log);
-
-  editLastEntry([note](LogPtrType entry) {
-    entry->setUserNote(note);
-  });
+  logEvent(new LogType(time, type, note, Log::UserDataType{}));
```
lifflander: Can we make this `std:make_unique` to preserve ownership semantics instead of constructing the `std::unique_ptr` later?

src/vt/trace/trace.cc
```diff
@@ -348,106 +326,79 @@ void Trace::endProcessing(
...
+TraceEventIDType Trace::logEvent(LogType* log) {
+  assert(log != nullptr && "log cannot be null");
 
-  auto const trace_cont_size = traces_.size();
-  if (trace_cont_size > 0) {
-    fn(traces_.at(trace_cont_size - 1));
-  }
-}
+  // Honor promise of ownership; log remains valid in scope.
+  auto log_uptr = std::unique_ptr<LogType>{log};
```
lifflander: Instead of promotion here, should come in as a `std::unique_ptr`

pnstickne:
@lifflander @uhetmaniuk  Changes made on #606

pnstickne:
Changes on https://github.com/DARMA-tasking/vt/pull/606

pnstickne:
Closing as on #241 work.

PR #602, opened 2019-12-02
241 flush trace shifts

pnstickne:


src/vt/trace/trace.cc
```diff
@@ -637,18 +636,55 @@ void Trace::writeTracesFile(int flush) {
...
     auto path = full_trace_name_;
-    if (not file_is_open_) {
-      log_file_ = gzopen(path.c_str(), "wb");
-      outputHeader(node, start_time_, log_file_);
-      file_is_open_ = true;
+
+    gzFile file = static_cast<gzFile>(log_file_);
+    if (not file) {
+      file = gzopen(path.c_str(), "wb");
+      log_file_ = static_cast<void*>(file);
```
pnstickne: Don't actually need that static_cast..

src/vt/trace/trace.h
```diff
@@ -150,30 +147,52 @@ struct Trace {
...
+  ///
+  /// Normally, when tracing is not finalized:
+  /// - All 'closed' events are written.
+  /// - The file stream is flushed, such that it is in a valid state.
+  ///
+  /// If tracing is finalized:
+  /// - All events are written, even non-closed events.
+  /// - The file is CLOSED and all future trace collection is disabled.
+  /// - Should only be done/requested on termination as tracing
+  ///   as tracign cannot be resumed afterward.
```
pnstickne: Typos..

src/vt/trace/trace.h
```diff
@@ -150,30 +147,52 @@ struct Trace {
...
+  void writeTracesFile(bool finalizeTracing);
+
   bool inIdleEvent() const;
 
   static double getCurrentTime();
   void outputControlFile(std::ofstream& file);
   static TimeIntegerType timeToInt(double const time);
   static void traceBeginIdleTrigger();
   static void outputHeader(
```
pnstickne: Should be private..

src/vt/trace/trace.h
```diff
@@ -187,11 +206,13 @@ struct Trace {
   std::string full_sts_name_    = "";
   std::string full_dir_name_    = "";
   UserEventRegistry user_event_ = {};
-  gzFile log_file_;
-  bool file_is_open_            = false;
+
+  /*gzFile*/ void* log_file_;
```
pnstickne: Valid, as `gzFile` is a pointer type.

src/vt/trace/trace.h
```diff
@@ -187,11 +206,13 @@ struct Trace {
...
+  bool tracing_closed_          = false;
+
   bool wrote_sts_file_          = false;
-  size_t cur_                  = 0;
-  size_t cur_stop_             = 0;
+  size_t cur_                   = 0;
```
pnstickne: TODO: should remove from container once written. Need to keep track of total (as per written + retained) and to ensure that no events on the open event stack are removed in non-finalization cases.

src/vt/trace/trace.cc
```diff
@@ -658,7 +694,10 @@ void Trace::writeTracesFile(int flush) {
   }
 }
 
-void Trace::writeLogFile(gzFile file, TraceContainerType const& traces) {
+void Trace::writeTracesToLogFile(void* fileVP) {
```
pnstickne: Accept fileVP for parity with static methods; otherwise, don't accept arguments because it's NOT a static method and partial-acceptance is confusing, especially when the object passed around owns the objects..

src/vt/trace/trace.cc
```diff
@@ -637,18 +636,55 @@ void Trace::writeTracesFile(int flush) {
...
+
   if (traceWritingEnabled(theContext()->getNode())) {
     auto path = full_trace_name_;
-    if (not file_is_open_) {
-      log_file_ = gzopen(path.c_str(), "wb");
-      outputHeader(node, start_time_, log_file_);
-      file_is_open_ = true;
+
+    gzFile file = static_cast<gzFile>(log_file_);
+    if (not file) {
```
pnstickne: Maybe explicit nullptr compare as it's not immediately obvious from the type.

src/vt/trace/trace.cc
```diff
@@ -637,18 +636,55 @@ void Trace::writeTracesFile(int flush) {
...
+        trace, node,
+        "Trying to dump traces with open events - abnormal termination?"
+      );
+    }
+
+    //--- Dump everything into an output file
+    cur_stop_ = traces_.size();
+  }
+
   if (traceWritingEnabled(theContext()->getNode())) {
```
pnstickne: Could use `node` from above.

src/vt/trace/trace.cc
```diff
@@ -637,18 +636,55 @@ void Trace::writeTracesFile(int flush) {
     TraceContainersType::event_container.size()
   );
 
+  if (finalizeTracing) {
+    disableTracing(); // Accept no more traces.
+
+    if (not open_events_.empty()) {
+      // During an abort there might be open events.
```
pnstickne: This is awkward. However, it's marginally less awkward that failing with an abort from within a possible abort.. or just failing in general during termination,

src/vt/trace/trace.cc
```diff
@@ -637,18 +636,55 @@ void Trace::writeTracesFile(int flush) {
     TraceContainersType::event_container.size()
   );
 
+  if (finalizeTracing) {
+    disableTracing(); // Accept no more traces.
```
pnstickne: Technically code could still re-enable tracing. Tracing should be prohibited from being re-enabled once finalization occurs.

src/vt/trace/trace.cc
```diff
@@ -621,13 +616,17 @@ void Trace::flushTracesFile(bool useGlobalSync) {
...
-    writeTracesFile(Z_FULL_FLUSH);
+    writeTracesFile(false);
   }
 }
 
-void Trace::writeTracesFile(int flush) {
+void Trace::writeTracesFile(bool finalizeTracing) {
   auto const node = theContext()->getNode();
 
+  if (tracing_closed_) {
```
pnstickne: Avoids unlikely edge-case of multiple finalization attempts. Contract says it's a once-only operation.

src/vt/trace/trace.cc
```diff
@@ -621,13 +616,17 @@ void Trace::flushTracesFile(bool useGlobalSync) {
     theCollective()->barrier();
```
pnstickne: The "useGlobalSync" should probably be moved INSIDE the check, such that it's only done occasionally. Also, maybe it can eliminated.. not exactly sure the case here.

src/vt/trace/trace.h
```diff
@@ -150,30 +147,52 @@ struct Trace {
   void enableTracing();
   void disableTracing();
 
+  /// Flush traces, IF needed.
+  /// Flushing only occurs if other constraints are met;
+  /// it might be good to rename this method to reflect.
```
pnstickne: Name feels off. Maybe `flushTracesIncremental` or .. something to signify that it's really a REQUEST to flush traces. Also, maybe it would be relevant to call this as a result of scheduler and/or on a periodic schedule?

src/vt/trace/trace.h
```diff
@@ -87,10 +84,10 @@ struct Trace {
   std::string getSTSName()   const { return full_sts_name_;   }
   std::string getDirectory() const { return full_dir_name_;   }
 
-  void initialize();
```
pnstickne: Externally there is now only initializeTracing; so the public contract decreases.

src/vt/trace/trace.h
```diff
@@ -61,9 +61,6 @@
 #include <unordered_map>
 #include <vector>
 
-#include <mpi.h>
-#include <zlib.h>
```
pnstickne: Huzzah! Gone!

src/vt/collective/collective_ops.cc
```diff
@@ -157,7 +158,8 @@ void CollectiveAnyOps<instance>::output(
   if (myrt) {
 #if backend_check_enabled(trace_enabled)
     //--- Try to flush most of the traces
-    myrt->theTrace->writeTracesFile(Z_FULL_FLUSH);
+    bool finalizeTracing = error and abort_out;
+    myrt->theTrace->writeTracesFile(finalizeTracing);
```
pnstickne: See condition below for when an abort is scheduled. If vt::abort happens to call abort above (does it?!?!), the trace-write guards will prevent any such double finalization.

src/vt/runtime/runtime.cc
```diff
@@ -970,8 +968,10 @@ void Runtime::initializeTrace() {
     if (ArgType::vt_trace) {
       std::string name = user_argc_ == 0 ? "prog" : user_argv_[0];
       auto const& node = theContext->getNode();
-      theTrace->setupNames(
-        name, name + "." + std::to_string(node) + ".log.gz", name + "_trace"
+      theTrace->initializeTracing(
```
pnstickne: It would be more transparent to accept 'the scheduler' as an argument in the function and then supply it here as well.

src/vt/trace/trace.cc
```diff
@@ -637,18 +636,55 @@ void Trace::writeTracesFile(int flush) {
...
+      // however, it comes with severe compression penalities when used often.
+      // Z_SYNC_FLUSH is sufficient to ensure data-until-here can be decompressed.
+      gzflush(file, Z_SYNC_FLUSH);
     }
-    writeLogFile(log_file_, traces_);
-    gzflush(log_file_, flush);
   }
 
-  if (node == designated_root_node and not wrote_sts_file_) {
+  // STS file is written on first write request, and then (possibly again)
```
pnstickne: This could be done smarter. However, really want to ensure that there is (an attempt) to capture the STS from the END of the program termination.

src/vt/trace/trace.cc
```diff
@@ -637,18 +636,55 @@ void Trace::writeTracesFile(int flush) {
...
+  // during finalization to ensure at least partial data and ultimately
+  // full summary data: eg. new event counts, new event types, renames..
+  if (node == designated_root_node
+      and (not wrote_sts_file_ or finalizeTracing)) {
```
pnstickne: Order seems reversed.

src/vt/trace/trace.h
pnstickne: It would also be possible to separate out different external methods and use the same internal method for implementation (it was just too juicy to combine the code with a parameter switch instead of the nested calls).

src/vt/trace/trace.cc
pnstickne: Removed needless cast.

src/vt/trace/trace.h
pnstickne: Made outputHeader/outputFooter/outputControlFile private as they are related.

src/vt/trace/trace.h
pnstickne: Added additional TODO inside the writeTracesToLogFile method.

src/vt/trace/trace.cc
pnstickne: Done; it is more clear, me thinks.

src/vt/trace/trace.cc
pnstickne: Could, and do.

src/vt/trace/trace.cc
pnstickne: It still is awkward. However, after finalization there is also the guarantee that all future events are suppressed so even if there are dangling open events they will be benign.

src/vt/trace/trace.cc
pnstickne: Note anymore.. added guard to `enableTracing`.

src/vt/trace/trace.cc
pnstickne: Reversed (un-reversed?)

src/vt/runtime/runtime.cc
```diff
@@ -935,13 +935,15 @@ void Runtime::initializeComponents() {
   theEvent = std::make_unique<event::AsyncEvent>();
   thePool = std::make_unique<pool::Pool>();
 
-  // Initialize tracing, when it is enabled; used in the AM constructor
+  theSched = std::make_unique<sched::Scheduler>();
+
+  // Initialize tracing, when it is enabled;
+  // used in the AM constructor, depends on the scheduler.
   initializeTrace();
```
pnstickne: Tracing deemed important.. and relies on the scheduler (this could be made more explicit!). The scheduler (actor) does not rely on any of the following objects.

src/vt/trace/trace.cc
```diff
@@ -637,28 +640,82 @@ void Trace::writeTracesFile(int flush) {
...
+    if (finalizeTracing) {
+      // Closing zip stream/file permantently.
+      outputFooter(node, start_time_, file);
+
+      gzclose(file);
+      log_file_ = nullptr;
+      tracing_closed_ = true;
+    } else {
+      // Z_FULL_FLUSH is useful if previous data is corrupted or random access;
+      // Z_SYNC_FLUSH is sufficient to ensure data-until-here is written.
```
pnstickne: Typos/grammer.

src/vt/trace/trace.cc
```diff
@@ -825,6 +882,9 @@ void Trace::writeLogFile(gzFile file, TraceContainerType const& traces) {
     delete log;
```
pnstickne: `delete log` - should also eject from collection. Otherwise, will only save memory by effect of dangling pointers while not realizing true benefit of a pruned collection.

src/vt/trace/trace.h
pnstickne: VT should be really considerate about what it exposes due to "one global smackaroo world" sucked in from every assorted header file.

pnstickne:
Oops, this is mostly a POC. The changes on the other branch already deal with the zlib stuff and do so in a different way. I've transferred my findings back to comments on there.

uhetmaniuk:
@pnstickne @lifflander 
Should we close this pull request?

PR #604, opened 2019-12-03
603 dladdr link

pnstickne:
- Builds on Ubuntu now (not really sure *why* it was building on OS X - LD pulled in otherwise?)

- Split to stack_out.cc to reduce global VT pollution.

### DIFF
--- a/cmake/link_vt.cmake
+++ b/cmake/link_vt.cmake
@@ -30,6 +30,7 @@ function(link_target_with_vt)
     LINK_CHECKPOINT
     LINK_DETECTOR
     LINK_CLI11
+    LINK_DL
   )
   set(
     multiValueArg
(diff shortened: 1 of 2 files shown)


cmake/link_vt.cmake
```diff
@@ -76,6 +77,16 @@ function(link_target_with_vt)
     endif()
   endif()
 
+  if (NOT DEFINED ARG_LINK_DL AND ${ARG_DEFAULT_LINK_SET} OR ARG_LINK_DL)
```
pnstickne: `dumpStack` does not appear to be compile-time guarded. On a normal build, `dladdr` linking is failing on my Ubuntu VM.

src/vt/configs/error/stack_out.h
```diff
@@ -45,16 +45,9 @@
...
 #define INCLUDED_VT_CONFIGS_ERROR_STACK_OUT_H
 
-#include "vt/config.h"
-
 #include <cstdlib>
 #include <vector>
 #include <tuple>
-#include <sstream>
-
-#include <execinfo.h>
```
pnstickne: Able to pull all these from header (bled externally though other error headers).

cmake/link_vt.cmake
lifflander: This should definitely been enabled

PR #606, opened 2019-12-11
241 flush cleaner

uhetmaniuk:
- Incorporate the changes from [#241 ](https://github.com/DARMA-tasking/vt/issues/241) to get a cleaner history.
- Merge "manually" the changes from [#600 ]
- Close the pull request [#551 ]

### DIFF
--- a/src/vt/runtime/runtime.cc
+++ b/src/vt/runtime/runtime.cc
@@ -510,6 +510,17 @@ void Runtime::printStartupBanner() {
       auto f12 = opt_on("--vt_trace_mod", f11);
       fmt::print("{}\t{}{}", vt_pre, f12, reset);
     }
+    if (ArgType::vt_trace_flush_size != 0) {
+      auto f11 = fmt::format("Flush output incrementally with a buffer of,"
+                             " at least, {} record(s)",
+                             ArgType::vt_trace_flush_size);
+      auto f12 = opt_on("--vt_trace_flush_size", f11);
+      fmt::print("{}\t{}{}", vt_pre, f12, reset);
+    } else {
+      auto f11 = fmt::format("Flushing traces at end of run");
+      auto f12 = opt_inverse("--vt_trace_flush_size", f11);
+      fmt::print("{}\t{}{}", vt_pre, f12, reset);
+    }
   }
   #endif
 
@@ -821,7 +832,7 @@ void Runtime::abort(std::string const abort_str, ErrorCodeType const code) {
     auto const comm = theContext->getComm();
     MPI_Abort(comm, 129);
   } else {
-	std::_Exit(code);
+    std::_Exit(code);
     // @todo: why will this not compile with clang!!?
     //quick_exit(code);
   }
--- a/src/vt/runtime/runtime.cc
+++ b/src/vt/runtime/runtime.cc
@@ -128,6 +128,12 @@ void Runtime::pauseForDebugger() {
   if (node == 0 || node == -1) {
     ::fmt::print("{}Caught SIGINT signal: {} \n", prefix, sig);
   }
+  // Try to flush out all logs before dying
+# if backend_check_enabled(trace_enabled)
+  if (vt::theTrace()) {
+    vt::theTrace()->cleanupTracesFile();
+  }
+# endif
   if (Runtime::nodeStackWrite()) {
     auto stack = debug::stack::dumpStack();
     auto stack_pretty = debug::stack::prettyPrintStack(std::get<1>(stack));
@@ -149,6 +155,12 @@ void Runtime::pauseForDebugger() {
   auto vt_pre    = debug::vtPre();
   auto bred      = debug::bred();
   ::fmt::print("{}Caught SIGSEGV signal: {} \n", vt_pre, sig);
+  // Try to flush out all logs before dying
+# if backend_check_enabled(trace_enabled)
+  if (vt::theTrace()) {
+    vt::theTrace()->cleanupTracesFile();
+  }
+# endif
   if (Runtime::nodeStackWrite()) {
     auto stack = debug::stack::dumpStack();
     if (ArgType::vt_stack_file != "") {
--- a/src/vt/trace/trace.cc
+++ b/src/vt/trace/trace.cc
@@ -734,6 +734,7 @@ void Trace::writeTracesFile(int flush) {
     auto name = full_sts_name_;
     file.open(name);
     outputControlFile(file);
+    file.flush();
     file.close();
     wrote_sts_file_ = true;
   }
--- a/src/vt/trace/trace.cc
+++ b/src/vt/trace/trace.cc
@@ -904,6 +904,9 @@ void Trace::writeTracesFile(int flush) {
     // Poof!
     traces.pop();
   }
+
+  // Actually call flush to get it written to disk
+  gzflush(gzfile, flush);
 }
 
 void Trace::outputControlFile(std::ofstream& file) {
--- a/src/vt/trace/trace.cc
+++ b/src/vt/trace/trace.cc
@@ -179,7 +179,9 @@ void Trace::addUserNote(std::string const& note) {
   auto const type = TraceConstantsType::UserSuppliedNote;
   auto const time = getCurrentTime();
 
-  logEvent(new LogType(time, type, note, Log::UserDataType{}));
+  logEvent(
+    std::make_unique<LogType>(time, type, note, Log::UserDataType{})
+  );
 }
 
 void Trace::addUserData(int32_t data) {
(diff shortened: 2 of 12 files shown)


src/vt/trace/trace.cc
```diff
@@ -154,12 +173,7 @@ void Trace::addUserNote(std::string const& note) {
   auto const type = TraceConstantsType::UserSuppliedNote;
   auto const time = getCurrentTime();
 
-  LogPtrType log = new LogType(time, type);
-  logEvent(log);
-
-  editLastEntry([note](LogPtrType entry) {
-    entry->setUserNote(note);
-  });
+  logEvent(new LogType(time, type, note, Log::UserDataType{}));
```
uhetmaniuk: @pnstickne
When inserting #600 , I did not address the comment from @lifflander about std::unique_ptr.
Please advise during the review.

src/vt/trace/trace.cc
```diff
@@ -788,10 +802,11 @@ void Trace::writeLogFile(gzFile file, TraceContainerType const& traces) {
       vtAssertInfo(false, "Unimplemented log type", converted_time, log->node);
     }
 
-    delete log;
+    LogPtrType log_p = traces[i].release();
+    delete log_p;
```
uhetmaniuk: @pnstickne 
The traces are partially cleaned as soon as they are flushed to the 'gzFile'.
We are left with a vector of null pointers.
@lifflander Do we need a more appropriate release of memory after flushing (and before merging with develop) ?

pnstickne: With #600 there should be no need to do anything except remove the unique_ptr from the collection.

pnstickne:
It might be useful to consider a queue in place of the vector for FIFO - http://www.cplusplus.com/reference/queue/queue/ 
All events are processed FIFO with the little in-open-stack caveat.

pnstickne:
The other interesting caveat about vector is that it's continuous, so it has to pay copy-costs 'on occasion'. This is an interesting read:

https://www.codeproject.com/Articles/5425/An-In-Depth-Study-of-the-STL-Deque-Container

src/vt/trace/trace.cc
pnstickne: I updated to `reset` in a newer commit.

src/vt/trace/trace.cc
```diff
@@ -463,134 +462,115 @@ TraceEventIDType Trace::logEvent(LogPtrType log) {
...
     return 0;
   }
+
+  // Normal case of event emitted at end
+  TraceEventIDType event = log->event;
+  traces_.push(std::move(log));
+
+  // If auto-flush, can flush immediately.
+  // TODO: log time of flushing; unify with group-end.
+  if (ArgType::vt_trace_flush_size not_eq 0
```
pnstickne: This will cause trace flushes immediately on flush limit, without requiring waiting on any external event.

That's probably a Good Thing (and is very much easier to demonstrate when running trace-enabled programs).

It would probably be good to include trace timings to ensure that such artifacts do not add unexpected timing impacts; as well as to open up more tracing in general.

src/vt/trace/trace.cc
```diff
@@ -601,61 +581,110 @@ void Trace::disableTracing() {
...
+  while (not traces.empty()) {
+    std::unique_ptr<LogType> const& log = traces.front();
 
+    auto const& converted_time = timeToInt(log->time - start_time);
     auto const type = static_cast<
       std::underlying_type<decltype(log->type)>::type
     >(log->type);
 
-    // force guaranteed type (used in format specifiers)
-    size_t event_seq_id = TraceRegistry::getEvent(log->ep).theEventSeq();
```
pnstickne: This code should probably have static guards in place to ensure that types have a valid size for the format specifiers used and/or explicit conversions to the proper types to mitigate any accidental misbliefs of changes between the format specifiers and externally-defined types.

This case is already "incorrect" as TraceEntrySeqType is defined as:

     using TraceEntrySeqType   = uint32_t;

This was the reason for the previous addition that had been added, along with nice comment; I've re-applied the missing previous changes, although left the type and this comment to explicitly highly the previous code choice.

src/vt/trace/trace.cc
```diff
@@ -601,61 +581,110 @@ void Trace::disableTracing() {
...
+  gzclose(log_file_.get()->file_type);
+  log_file_ = nullptr;
+}
+
+void Trace::flushTracesFile(bool useGlobalSync) {
+  if (ArgType::vt_trace_flush_size == 0) {
+    // Flush the traces at the end only
+    return;
+  }
+  if (useGlobalSync) {
```
pnstickne: This bit feels off. It seems like it could introduce unnecessary waits between nodes while traces are per-node artifacts, especially when it *could* be called from 'any domain'.

Having an external flush request also seems less useful when flush-on-emit is used (see below). The flush-on-emit can be easily reverted.

lifflander:
@uhetmaniuk Can you address @pnstickne comments?

uhetmaniuk:
@pnstickne @lifflander 
A change in Trace::Log and/or TraceContainer has broken the flushing of traces. A member of `traces` is not found in the trace container.
I will keep looking for the fix.

src/vt/trace/trace.cc
pnstickne: I have a separate PR incoming that will address format width specifiers.

src/vt/trace/trace.cc
```diff
@@ -348,106 +365,88 @@ void Trace::endProcessing(
...
-  if (not enabled_ || not checkEnabled()) {
-    return;
-  }
-
-  auto const trace_cont_size = traces_.size();
-  if (trace_cont_size > 0) {
-    fn(traces_.at(trace_cont_size - 1));
-  }
-}
+TraceEventIDType Trace::logEvent(std::unique_ptr<LogType> log) {
```
pnstickne: @lifflander @PhilMiller @uhetmaniuk I have a set of changes to eliminate unique_ptr entirely _through_ the container. (std::queue still does standard heap allocs by default). It may be possible to replace the queue alloc or change to an implementation with a different allocator or a once-allocated circular or .. so many options.

Also have WIP changes that can reduce the size of each log message (barring user string) a fair bit through unionizing as either 'system' or 'union' data.

src/vt/trace/trace.cc
pnstickne: **See note above about possible-flush-after-every-emit**

src/vt/trace/trace.cc
pnstickne: Oops, the bit above about `%zu` was incorrect. It should be promoted to `unsigned long`.

pnstickne: (Mainly I was grumpy the call to `getEvent` disappeared.)

uhetmaniuk:
@lifflander @pnstickne 
I confirm that the fix in #613 will make the flushing of traces work again (I tested it on a local branch).

src/vt/trace/trace.cc
uhetmaniuk: Sounds good.

src/vt/trace/trace.cc
uhetmaniuk: @pnstickne @lifflander @ppebay @PhilMiller 
I am sorry but I have a hard time processing a lot of these comments. 
- Are they actionable items that I, @uhetmaniuk, need to modify within the original scope of the PR (before approval)?
- Are you, @pnstickne, handling these changes?
- Are these comments just documenting the thought process for future developments?

pnstickne: @uhetmaniuk 

- Are they actionable items that I, @uhetmaniuk, need to modify within the original scope of the PR (before approval)?

I do no believe so. This is historic behavior, not new. While it is _technically_ UB, there have been no known/reported issues here to date.

- Are you, @pnstickne, handling these changes?

Yes. I have some code that uses explicit conversions that will ensure type widening (as required) and warn on type narrowing (as currently implicitly happens, which will also be addressed). It's off the current PR as it is non-essential to dealing with issues relating to beta-5 and/or limiting overall memory usages.

- Are these comments just documenting the thought process for future developments?

Yes. I do not believe there is any imminent actionable work. If anyone *wanted* to make an immediate change it would be for:

     // now guaranteed to be width of %lu format specifier
     unsigned long event_seq_id = TraceRegistry::getEvent(log->ep).theEventSeq();

pnstickne: @uhetmaniuk 
Actually, I like think change to `unsigned long` would be *good* and I fully support such! I won't be back home until tonight to make that change though.

src/vt/runtime/runtime.cc
```diff
@@ -149,6 +155,12 @@ void Runtime::pauseForDebugger() {
   auto vt_pre    = debug::vtPre();
   auto bred      = debug::bred();
   ::fmt::print("{}Caught SIGSEGV signal: {} \n", vt_pre, sig);
+  // Try to flush out all logs before dying
+# if backend_check_enabled(trace_enabled)
+  if (vt::theTrace()) {
+    vt::theTrace()->cleanupTracesFile();
+  }
+# endif
```
PhilMiller: I would have done this step *after* printing the stack dump, just in case part of the crash relates to the serialization and output to the files. We can maybe adjust that going forward.

PhilMiller:
@lifflander I believe this can all be applied to 1.0.0-beta.5-proposed-update

Issue #567, opened 2019-11-12
--vt_lb_stats_dir not being parsed correctly in EMPIRE on Mutrino


lifflander:
**Describe the bug**
Try to specify a directory using `--vt_lb_stats_dir`

pnstickne:
Another issue (related?) seems to be that unknown `--vt_` (or args in general) are causing "various program crashes"; if a build doesn't have a particular option available it could thus have similar results. :|

pnstickne:
Also, `--help` (and maybe a `--vt_help`) should probably terminate the program..

@lifflander In addition to resolving above, what about:

     mpirun vt/program --vt_arg_1 --vt_arg_2 -- --arg-to-prog1 --arg-to-prog2

Issue #444, opened 2019-09-03
Split stage runtime initialization and argument resolution


lifflander:
Split runtime init into two phases. Resolve argument conflicts with clear precedence.

lifflander:
@uhetmaniuk I think I agree with Nicole. I think we should parse the YAML in EMPIRE and then pass the `ArgConfig` object to VT. I want to the user to be able to supply multiple configurations at different levels in the code that get resolved. 

A couple minor additions:
- I want to adding an option in VT to dump the final configuration file (CLI11 does bidirectional config file input and output) using a opt-in argument. Maybe `--vt_dump_config`.
- We should add an option for specifying a config file as CLI11 can take that directly (it supports ini files without any extra work!).  https://github.com/CLIUtils/CLI11#configuration-file

lifflander:
I would also like you to write some general code that handles printing configuration at startup (`vt::Runtime::printStartupBanner`). Right now, there's a lot of specialized code for that that is error prone. This will also help with correctly/automatically outputting precedence resolution decisions.

I'm thinking:

```C++
struct ArgumentBase {
  virtual void addToAppParse() = 0;
};

template <typename T>
struct Argument : ArgumentBase {

  void addToAppParse() override {

  }

  static constexpr bool isFlag() { return std::is_same<T,bool>::value; }

  std::string arg_;
  std::string desc_;
  T value = {};
  T default = {};
  std::function<std::string(T const& value)> print_value = nullptr;
  std::function<bool(T const& value) print_condition = nullptr;
  bool always_print_startup_;
};



```

uhetmaniuk:
@lifflander I would like to clarify a few questions:
- why do we update argc and argv after parsing them? (I could not find a place where the updated values are needed);
- do you envision "resolveOptions" to be called after calling the CLI parser?
- the Config object will combine, for each parameter, the default value (possibly updated by EMPIRE), the command-line value, and the input file (YAML) value. CLI will only be aware of the command-line value and the default value. The output from CLI will not see the input file (YAML) commands. Am I missing something?

lifflander:
- We must update argc/argv because they get passed to EMPIRE after VT uses them. If the options are not removed, EMPIRE will see unexpected arguments.

- Yes, `resolveOptions` will be called during `Runtime::initialize`, in the new split design, after the allocation and various overloads are added.

[...]

PR #607, opened 2019-12-11
ISSUE? improved arg handling

pnstickne:
Basically:

- All arguments must be accounted for and correctly parsed.
  This mitigates "oops" cases when arguments are not correctly
  handled by appropriate domain.

  To supply MPI_Init args (still does not address issue of late MPI_Init), use:
    `--vt_mpi_args mpiarg1 mpiarg2..`
  To supply user args (returned up through initialize):
    `-- userarg1 userarg2 ..`
  To switch back to VT args, use `--vt_args`

- Also supports 'auto pass-through' args for historic reasons.
  That is, any arguments before a "--vt_" arg including "--vt_help"
  are automatically classified as a pass-through arguments and any
  "--vt_" or argument will implicitly switch into vt-arg mode.

    `1 2 3 --vt_no_color` is the same as
    `--vt_no_color -- 1 2 3` while
    `--vt_no_color 1 2 3` is invalid (left in vt-arg context where such are invalid).

   Most notable, short vt args like '-c' or '-q' do NOT initiate
   vt-arg mode and will subject to pass-through behavior.
   (Short args have been *removed* to avoid this edge-case.)

- The syntax can also be extended indefinitely (although auto-mode only applies once at the start), for whatever use it may have:
   `1 --vt_no_color -- 2 --vt_args --vt_trace -- 3` is the same as
   `1 2 3 --vt_no_color --vt_trace` and
   `--vt_no_color --vt_trace -- 1 2 3`

   It is recommended that parameters that take values do so with the `--param=value` form.

This addresses some PEBCAK errors locally and provides a base for some future changes with directory arg handling. There is a chance similar issues are accidentally encountered on other systems.

Pass-through arguments (and the program name) are displayed in the banner for clarity.

It is a "breaking" change in that some previously valid parameters invocations are not forbidden, ie. needing to use a `--` to switch out of vt-arg context. However changes required are an artifact of the parameters supplied during invocation (as opposed to requiring code changes).

When in vt-arg context, unknown arguments will *abort the program* with an argument error.

The `--vt_help` (replaces `--help`) parameter is also more useful (and program-aborting).


src/vt/configs/arguments/args.cc
```diff
@@ -486,40 +481,92 @@ namespace vt { namespace arguments {
...
+    int result = app.exit(ex);
+    // ..CLI11 returns 0 on --help, which is still terminal. Great job.
+    return result ? result : 1;
+  }
+
+  // Get the clean prog name; don't allow path bleed in usages.
+  // std::filesystem is C++17.
+  std::string prog_name = argv[0];
+  size_t l = prog_name.find_last_of("/\\");
+  if (l < prog_name.size() - 1) {
```
pnstickne: Overflow on `""` as prog_name .. should never happen, still technically incorrect.

src/vt/configs/arguments/args.h
```diff
@@ -147,8 +152,22 @@ struct ArgConfig {
   static std::string vt_user_str_2;
   static std::string vt_user_str_3;
 
+  /// Name of the program launched (excluding any path!)
+  static std::string prog_name;
+
+  /// Name of the program launched, aka argv[0].
+  /// Original char* objects.
```
pnstickne: "object" (singular).

src/vt/configs/arguments/args.h
```diff
@@ -147,8 +152,22 @@ struct ArgConfig {
...
+  static char* argv_prog_name;
+
+  /// Arguments to pass to MPI Init.
+  /// Does not include argv[0]. Original char* objects.
+  static std::vector<char*> mpi_init_args;
+  /// Arguments are being ref-returend as the result of parse(..).
+  /// Does not include argv[0]. Orignal char* objects.
```
pnstickne: Typo.

src/vt/runtime/runtime.cc
```diff
@@ -267,493 +271,20 @@ bool Runtime::tryFinalize() {
...
     initializeComponents();
     initializeOptionalComponents();
     initializeErrorHandlers();
     sync();
     if (theContext->getNode() == 0) {
       printStartupBanner();
+      // Enqueue a check for later in case arguments are modified before work
+      // actually executes
+      theSched->enqueue([this]{
+        this->checkForArgumentErrors();
```
pnstickne: Probably should be done up in ArgParse.. really *shouldn't* be able to modify.. not sure why comment.

src/vt/runtime/runtime.cc
```diff
@@ -928,8 +459,23 @@ void Runtime::setup() {
   debug_print(runtime, node, "end: setup\n");
 }
 
-void Runtime::initializeContext(int argc, char** argv, MPI_Comm* comm) {
-  theContext = std::make_unique<ctx::Context>(argc, argv, is_interop_, comm);
+void Runtime::initializeContext() {
+  std::vector<char*>& mpi_args = ArgType::mpi_init_args;
```
pnstickne: `const`.

src/vt/runtime/runtime.cc
```diff
@@ -928,8 +459,23 @@ void Runtime::setup() {
...
+  int argc = mpi_args.size() + 1;
+  char* *const argv = new char*[argc + 1];
```
pnstickne: This could be unified with code in arg parsing..

src/vt/configs/arguments/args.cc
```diff
@@ -141,70 +144,56 @@ namespace vt { namespace arguments {
...
   auto dir    = "Name of directory to write stack files";
-  auto mod    = "Write stack dump if (node % vt_stack_mod) == 0";
-  auto g = app.add_flag("--vt_no_warn_stack",   vt_no_warn_stack,   warn);
-  auto h = app.add_flag("--vt_no_assert_stack", vt_no_assert_stack, assert);
-  auto i = app.add_flag("--vt_no_abort_stack",  vt_no_abort_stack,  abort);
-  auto j = app.add_flag("--vt_no_stack",        vt_no_stack,        stack);
-  auto k = app.add_option("--vt_stack_file",    vt_stack_file,      file, "");
-  auto l = app.add_option("--vt_stack_dir",     vt_stack_dir,       dir,  "");
-  auto m = app.add_option("--vt_stack_mod",     vt_stack_mod,       mod,  1);
+  auto mod    = "Write stack dump if (node % ArgConfig::vt_stack_mod) == 0";
```
pnstickne: Oops - remove.

src/vt/configs/arguments/args.cc
```diff
@@ -213,34 +202,32 @@ namespace vt { namespace arguments {
...
+    "trace_enabled)";
   auto tfile     = "Name of trace files";
   auto tdir      = "Name of directory for trace files";
-  auto tmod      = "Output trace file if (node % vt_stack_mod) == 0";
-  auto n  = app.add_flag("--vt_trace",           vt_trace,           trace);
-  auto nm = app.add_flag("--vt_trace_mpi",       vt_trace_mpi,       trace_mpi);
-  auto o  = app.add_option("--vt_trace_file",    vt_trace_file,      tfile, "");
-  auto p  = app.add_option("--vt_trace_dir",     vt_trace_dir,       tdir,  "");
-  auto q  = app.add_option("--vt_trace_mod",     vt_trace_mod,       tmod,  1);
+  auto tmod      = "Output trace file if (node % ArgConfig::vt_stack_mod) == 0";
```
pnstickne: Oops - remove.

lifflander:
@nlslatt Can you advise how this will affect EMPIRE? I'm worried about specializing the handling of arguments since the order in which VT is initialized with respect to other argument parsers might interact badly.

pnstickne:
> @nlslatt Can you advise how this will affect EMPIRE? I'm worried about specializing the handling of arguments since the order in which VT is initialized with respect to other argument parsers might interact badly.

Indeed interested - these changes are to make Life Better overall and to avoid edge-case with mis-beliefs around arguments supplied; although it may require some ordering differences as-passed-to-VT. In particular, adding `--` to switch out of VT-mode and/or explicitly switching to MPI mode if any MPI parameters should be passed downward.

The overall ordering wrt MPI and startup is the same: that is, MPI_Init call remains at intact location (and it would be nice to call MPI_Init first to let it scrub all MPI-specific args, although such alters the larger contract). The caller is still guaranteed a argv/argc "cleaned" from VT arguments before the return, subject to the noted extraction differenences.

pnstickne:
- Rebased up taking in collection and trace changes.

- Only displays help (or argument error on one node).

- Help message reads:

```
vt (Virtual Transport)

Usage:
[APP-ARGS..] [VT-ARGS] [--vt_mpi_args MPI-ARGS..] [-- APP-ARGS..]

Arguments up until the first '--vt_*' are treated as application arguments.

After the first '--vt_*', additional arguments are treated as VT arguments
unless '--vt_mpi_args' or '--' are used to switch the argument mode.
The '--vt_args' flag can be used to switch back into VT argument mode.
Modes can be switched indefinitely.

Application pass-through arguments are supplied to the host program
for further processing and are not used by VT for any configuration.
MPI (and only MPI) arguments are given directly to MPI_Init.

It is an error if an unexpected argument is encountered in VT argument mode.
The currently recognized VT arguments are listed below; availability varies
based on build and compilation settings.

```

pnstickne:
So that's like 100% more of a help message and useful behavior than before, along with sane argument checking and a pass-through mode to hand down arguments to whatever consume changes there are.

If/as such goes in develop, it would be nice to piece up some of the runtime initialization-split changes.

pnstickne:
Checks are "failing for reasons".

pnstickne:
@uhetmaniuk ^_^ Is the concept of an 'anchor' transferable to argument scopes? The goal here is to *limit*, not to *add*.

Also, CLI11 1.8 has new support for sub commands; are those similar to 'anchors'? A little bit of documentation / examples would really help the ability to compare and unify.

pnstickne:
@uhetmaniuk I think it would be beneficial to pull up the MPI_Init "prior to" returning arguments to client implementations (such that MPI_Init implementations can silently add or deduct arguments, as was the original MPI design intent). See comments in PR. Not sure if/how the goal is considered elsewhere. Such changes are no present on this branch and is considered out of scope.

uhetmaniuk:
@pnstickne 
The issue #444 separates the creation of Runtime object and the parsing of command-line arguments (via CLI and, also, an input file from an external software).
You can see the separation in the commit 
https://github.com/DARMA-tasking/vt/commit/39b4bac3e595737098eecb4e0f1fa661acc01d79
Let me know if this is not clear.
(PS: I am in the process of cleaning-up branch #444-split-runtime-initialization into #444-split-runtime-init-cleaner).

lifflander:
> Checks are "failing for reasons".

I'm trying to add more checks and things are a bit unstable right now. Sorry about that. Just look at Travis for now. I'm "allowed" to override checks when I merge.

lifflander:
> @uhetmaniuk ^_^ Is the concept of an 'anchor' transferable to argument scopes? The goal here is to _limit_, such as to prevent (current and future) argument errors, not to add new functionality.
> 
> Also, CLI11 1.8 (??) has new/improved support for sub-commands; are those similar to 'anchors'? A little bit of documentation / examples would really help the ability to compare and unify.

Anchors are there to solve another problem: providing a configuration from multiple sources with a hierarchy. For instance, EMPIRE uses a YAML file and we need to allow args to come from there. Those are lower priority than command line. It also allows application-specific defaults to be specified. I think that adding those features is orthogonal to this, which is more around making the current functionality better.

Also, I agree we should switch to the new version 1.8 of CLI. They add a lot of new features

lifflander:
@pnstickne What is the correct issue number of this?

pnstickne:
@lifflander I've a suspicious it may 'address' https://github.com/DARMA-tasking/vt/issues/567 - in particular it is designed to rule out some issues that might cause 'unexpected' program termination, which was the original impetus.

pnstickne:
I think it would be good to git a split-phase init (requires different loading of VT as current loading is really 'too late' for MPI args and 'too early' for additional inputs); as orthogonal, I wonder if just that can be split off from #444 - @uhetmaniuk 

That is, split-loading that does not at phase #1 deal with any loading or config changes per-se. This would require a different initialization pattern. Phase #2 would build on that with additional loading/config changes.

lifflander:
The wrong issue number is throwing off all my analysis for the release. I'm closing this and re-opening it with the correct issue number on the branch.

PR #609, opened 2019-12-11
608 gossiplb bugs

### LINKED ISSUE #608: Fix bugs in GossipLB
**Describe the bug**
Performance bugs

lifflander:
Fix bugs in GossipLB causing bad load distributions.

Fixes #608

### DIFF
--- a/src/vt/vrt/collection/balance/gossiplb/gossiplb.cc
+++ b/src/vt/vrt/collection/balance/gossiplb/gossiplb.cc
@@ -116,8 +116,8 @@ void GossipLB::doLBStages() {
 
     debug_print(
       gossiplb, node,
-      "GossipLB::doLBStages: running iter_={}, num_iters_={}\n",
-      iter_, num_iters_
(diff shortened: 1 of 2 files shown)


src/vt/vrt/collection/balance/gossiplb/gossiplb.cc
```diff
@@ -392,17 +392,23 @@ void GossipLB::decide() {
         // the rest of the stats framework
         auto obj_load = loadMilli(iter->second);
 
+        bool eval = Criterion(criterion_)(this_new_load_, selected_load, obj_load, avg);
```
pnstickne: Does `Criterion(criterion_)` const/reusable? If so, does it add any value?

PhilMiller: It's only used once within this entire function, so it doesn't much matter. IIRC, it should be cheap to compute anyway

PR #613, opened 2019-12-16
#611 Move statics for trace to function to fix bug w/clang 3.9

### LINKED ISSUE #611: Fix Clang 3.9 initialization bug with event containers
Clang 3.9 does not work with tracing, full stop.
There may be an issue with the unique_ptr not being correctly constant-initialized.

lifflander:
Fixes #611

### DIFF
--- a/src/vt/trace/trace_containers.cc
+++ b/src/vt/trace/trace_containers.cc
@@ -10,11 +10,10 @@ namespace vt { namespace trace {
 // that is when auto-handler events are registered.
 // Using a constexpr constructor / initializer list is also relevant.
 // - https://en.cppreference.com/w/cpp/language/constant_initialization
-static std::unique_ptr<TraceContainerEventClassType> event_type_container_{};
-static std::unique_ptr<TraceContainerEventType> event_container_{};
 
 /*static*/ TraceContainerEventClassType*
 TraceContainers::getEventTypeContainer(){
+  static std::unique_ptr<TraceContainerEventClassType> event_type_container_ = nullptr;
   if (event_type_container_ == nullptr) {
     event_type_container_ = std::make_unique<TraceContainerEventClassType>();
   }
@@ -23,6 +22,7 @@ TraceContainers::getEventTypeContainer(){
 
 /*static*/ TraceContainerEventType*
 TraceContainers::getEventContainer(){
+  static std::unique_ptr<TraceContainerEventType> event_container_ = nullptr;
   if (event_container_ == nullptr) {
     event_container_ = std::make_unique<TraceContainerEventType>();
   }
--- a/src/vt/trace/trace_containers.cc
+++ b/src/vt/trace/trace_containers.cc
@@ -5,15 +5,13 @@
 
 namespace vt { namespace trace {
 
-// These MUST have static-lifetime and be initialized during constant
-// initialization as they are modified during dynamic initialization as
-// that is when auto-handler events are registered.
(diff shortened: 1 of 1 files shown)


src/vt/trace/trace_containers.cc
```diff
@@ -10,11 +10,10 @@ namespace vt { namespace trace {
 // that is when auto-handler events are registered.
 // Using a constexpr constructor / initializer list is also relevant.
 // - https://en.cppreference.com/w/cpp/language/constant_initialization
-static std::unique_ptr<TraceContainerEventClassType> event_type_container_{};
-static std::unique_ptr<TraceContainerEventType> event_container_{};
 
 /*static*/ TraceContainerEventClassType*
 TraceContainers::getEventTypeContainer(){
+  static std::unique_ptr<TraceContainerEventClassType> event_type_container_ = nullptr;
```
pnstickne: Does this would with the *other suggestions* while leaving in place?

While I can see Clang having an issue with one form, if it has an issue with all 3, then it is *horribly horrible wrong*.

lifflander: No, I tried all the other suggestions. None of them work.

pnstickne: Here are the explicit forms again:

      // default const-expression ctor
      static std::unique_ptr<TraceContainerEventClassType> event_type_container_; 
      // initialization list of single nullptr (const-expr) to const-expr ctor
      static std::unique_ptr<TraceContainerEventClassType> event_type_container_{nullptr};
      // should be same as above per braces-or-equals
      static std::unique_ptr<TraceContainerEventClassType> event_type_container_ = nullptr;
      // this is the form that Clang 3.9 does not work with. This should be a value-assignment,
      // although it *differs* from the other 3 above and might be the *explicit case to avoid*
      // regardless of the definition occurring in a function body.
      static std::unique_ptr<TraceContainerEventClassType> event_type_container_{};

lifflander: All three of those other options failed.

pnstickne: :|

pnstickne: Okay, something seems super seriously wrong. Although that seems *too major* ..

pnstickne: Maybe: https://bugs.llvm.org/show_bug.cgi?id=19491 (limited to OS X?); links are broken.

pnstickne: Please add comments at the declarations inside the functions it is a Clang 3.9 work-about and must remain even though constant-initialization is well-defined in C++14 :|

I'll do it when I get home later if left alone..

src/vt/trace/trace_containers.cc
```diff
@@ -23,6 +20,11 @@ TraceContainers::getEventTypeContainer(){
 
 /*static*/ TraceContainerEventType*
 TraceContainers::getEventContainer(){
+  // n.b. Container MUST be constant-initialized as this function is used
```
pnstickne: Technically the comment not quite correct as it's not a requirement that it's constant-initialized in context (a C++ compiler *may* choose do so) as long as it's initialized-before-use..

src/vt/trace/trace_containers.cc
PhilMiller: Could we get a link to the Clang bug that's forcing this in the comment? Is 19491 actually the culprit?

PR #616, opened 2019-12-16
#615 Termination use DS-epoch by default for collection chain nextStep

### LINKED ISSUE #615: Make rooted epoch DS default for collection chain set
**What Needs to be Done?**

- [ ] First do this
- [ ] Then do that.
- [ ] Also this other thing.

**Additional context**

lifflander:
Fixes #615

### DIFF
--- a/src/vt/messaging/collection_chain_set.h
+++ b/src/vt/messaging/collection_chain_set.h
@@ -77,7 +77,9 @@ class CollectionChainSet final {
       auto& idx = entry.first;
       auto& chain = entry.second;
 
-      EpochType new_epoch = theTerm()->makeEpochRooted();
+      // The parameter `true` here tells VT to use an efficient rooted DS-epoch
+      // by default. This can still be overridden by command-line flags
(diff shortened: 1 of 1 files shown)


src/vt/messaging/collection_chain_set.h
```diff
@@ -77,7 +77,9 @@ class CollectionChainSet final {
       auto& idx = entry.first;
       auto& chain = entry.second;
 
-      EpochType new_epoch = theTerm()->makeEpochRooted();
+      // The parameter `true` here tells VT to use an efficient rooted DS-epoch
+      // by default. This can still be overridden by command-line flags
```
PhilMiller: Really? Even if code explicitly says it needs a DS rooted epoch, the command-line flag would still override that? That seems insane, and potentially very broken.

Consider the corresponding place for collection initialization epochs, where you need to add an explicit `false` to force a wave TD on an epoch - will setting the command line argument the other way just break that, too?

PhilMiller:
I agree with the code, but think the comment reflects insanity

PR #622, opened 2019-12-17
#621 Termination clear out state after term detected

### LINKED ISSUE #621: Epochs not being cleared for state in TD
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

lifflander:
Fixes #621

### DIFF
--- a/src/vt/termination/dijkstra-scholten/comm.cc
+++ b/src/vt/termination/dijkstra-scholten/comm.cc
@@ -92,6 +92,26 @@ namespace vt { namespace term { namespace ds {
   theTerm()->epochTerminated(epoch, true);
 }
 
+/*static*/ void StateDS::disengage(EpochType epoch) {
+  debug_print(
+    termds, node,
+    "StateDS::disengage: epoch={:x}\n", epoch
+  );
+
+  // We are likely inside code that references the map structure that holds DS
+  // terminators. We can remove it safely. So we will enqueue an action to do
+  // the cleanup. This action must see if the DS term is still disengaged (it
+  // can easily be re-engaged after it disengages)
+  theSched()->enqueue([epoch]{
+    auto ptr = theTerm()->getDSTerm(epoch);
+    if (ptr != nullptr) {
+      if (not ptr->isEngaged()) {
+        theTerm()->cleanupEpoch(epoch, false);
+      }
+    }
+  });
+}
+
 /*static*/ StateDS::TerminatorType*
 StateDS::getTerminator(EpochType const& epoch) {
   auto term_iter = theTerm()->term_.find(epoch);
--- a/src/vt/termination/dijkstra-scholten/comm.cc
+++ b/src/vt/termination/dijkstra-scholten/comm.cc
@@ -89,7 +89,7 @@ namespace vt { namespace term { namespace ds {
     termds, node,
     "StateDS::rootTerminated: epoch={:x}\n", epoch
   );
-  theTerm()->epochTerminated(epoch);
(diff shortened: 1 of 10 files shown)


src/vt/termination/dijkstra-scholten/comm.cc
```diff
@@ -89,7 +89,27 @@ namespace vt { namespace term { namespace ds {
...
+}
+
+/*static*/ void StateDS::disengage(EpochType epoch) {
+  debug_print(
+    termds, node,
+    "StateDS::disengage: epoch={:x}\n", epoch
+  );
+
+  // We are likely inside code that references the map structure that holds DS
+  // terminators. We can remove it safely. So we will enqueue an action to do
```
PhilMiller: Is "can" a typo for "can't"?

src/vt/termination/dijkstra-scholten/comm.cc
```diff
@@ -89,7 +89,27 @@ namespace vt { namespace term { namespace ds {
...
+  // the cleanup. This action must see if the DS term is still disengaged (it
+  // can easily be re-engaged after it disengages)
```
PhilMiller: This comment on outstanding references doesn't give me a whole lot of confidence that we won't have dangling reference bugs anyway. 

If we want to be safe against that, we can work with objects whose lifetimes actually track such references to specific epochs, rather than the raw epoch numbers.

Alternately, the interface presented by StateDS can be better encapsulated so that references don't leak out, and it handles operations on epochs by number regardless of what state they've ended up in.

src/vt/termination/term_window.h
```diff
@@ -105,6 +105,11 @@ struct EpochWindow {
    */
   void closeEpoch(EpochType const& epoch);
 
+  /*
+   * Terminated an active epoch by adding it to the sett
```
PhilMiller: Typo here: 'sett'

src/vt/termination/termination.h
```diff
@@ -157,7 +157,7 @@ struct TerminationDetector :
 
 private:
   TermStateType& findOrCreateState(EpochType const& epoch, bool is_ready);
-  void cleanupEpoch(EpochType const& epoch);
+  void cleanupEpoch(EpochType const& epoch, bool isRoot);
```
PhilMiller: I'd be much happier if this were an `enum {ROOT, NON_ROOT}`

PhilMiller:
I think your fixes are OK, but I'll have to spend a bunch more time with the surrounding code to be more confident.

src/vt/termination/dijkstra-scholten/comm.cc
lifflander: Yes, should be can't

src/vt/termination/term_window.h
lifflander: Fixed, this was a copy-paste error also

src/vt/termination/termination.h
lifflander: Switched to enum

src/vt/termination/dijkstra-scholten/comm.cc
lifflander: I've improved the comment in the code. I agree with you about lifetimes, but that would be a major refactor. Are you comfortable with the code as is?

PhilMiller: If it holds up in practice for now, yeah, I'm fine going ahead with this.

PR #624, opened 2019-12-18
Refactor an explicit 'new' to clarify that it's a shared_ptr

### LINKED ISSUE #627: SeqList: Initialize shared_ptr as such, not with bare new
Use of bare `new` in `SeqList` makes it hard to see explicit ownership semantics, and costs a second allocation for the `shared_ptr` control block.

PhilMiller:
Fixes #627


PhilMiller:
Sure, though I don't see the functional aspect of this - consolidating the control block allocation with the object - as all that important. This needn't be ported over to the release branch. The clarity that this allocation has an owner is the important aspect, and just needs to be carried forward for new development.

PR #629, opened 2019-12-18
628 pending send bug

### LINKED ISSUE #628: PendingSend/CollectionChainSet fails for simple sends
**Describe the bug**
Try a simple non-collection send with collection chain set or dependent send chain. It will fail to sequence properly.

lifflander:
Fixes #628

### DIFF
--- a/src/vt/messaging/active.impl.h
+++ b/src/vt/messaging/active.impl.h
@@ -116,11 +116,7 @@ ActiveMessenger::sendMsgSz(
   ByteType const& msg_size
 ) {
   envelopeSetup(msg->env, dest, han);
-
-  auto const is_epoch = envelopeIsEpochType(msg->env);
-  if (is_epoch) {
-    setupEpochMsg(msg);
-  }
+  setupEpochMsg(msg);
 
   auto base = promoteMsg(msg).template to<BaseMsgType>();
   return PendingSendType(base, msg_size);
@@ -207,11 +203,7 @@ ActiveMessenger::sendMsgSz(
   if (tag != no_tag) {
     envelopeSetTag(msg->env, tag);
   }
-
-  auto const is_epoch = envelopeIsEpochType(msg->env);
-  if (is_epoch) {
-    setupEpochMsg(msg);
-  }
+  setupEpochMsg(msg);
 
   auto base = promoteMsg(msg).template to<BaseMsgType>();;
   return PendingSendType(base, msg_size);
@@ -260,11 +252,7 @@ ActiveMessenger::PendingSendType ActiveMessenger::sendMsg(
   if (tag != no_tag) {
     envelopeSetTag(msg->env, tag);
   }
-
-  auto const is_epoch = envelopeIsEpochType(msg->env);
-  if (is_epoch) {
-    setupEpochMsg(msg);
-  }
+  setupEpochMsg(msg);
 
   auto base = promoteMsg(msg).template to<BaseMsgType>();
   return PendingSendType(base, sizeof(MessageT));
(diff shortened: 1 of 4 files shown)


src/vt/messaging/active.impl.h
```diff
@@ -159,6 +159,12 @@ ActiveMessenger::PendingSendType ActiveMessenger::broadcastMsgSz(
   if (tag != no_tag) {
     envelopeSetTag(msg->env, tag);
   }
+
```
lifflander: We must put the epoch on early---before `PendingSend` gets returned.

src/vt/messaging/pending_send.cc
```diff
@@ -53,8 +53,28 @@ void PendingSend::sendMsg() {
...
+  if (msg_ != nullptr) {
+    auto const is_epoch = envelopeIsEpochType(msg_->env);
+    auto const is_term = envelopeIsTerm(msg_->env);
+    if (is_epoch and not is_term) {
+      bool const is_produce = op == PendingTermEnum::Produce;
+      auto const ep = is_produce ? envelopeGetEpoch(msg_->env) : epoch_produced_;
+      if (ep != no_epoch and ep != term::any_epoch_sentinel) {
+        if (is_produce) {
+          theTerm()->produce(ep,1);
+          epoch_produced_ = ep;
```
lifflander: This machination is due to some inconsistencies in the serialized messenger. Eventually these will be solved, but now this solves the problem.

lifflander: In particular, saving the produced epoch vs. just reading it again from the envelope.

src/vt/messaging/pending_send.cc
```diff
@@ -53,8 +53,28 @@ void PendingSend::sendMsg() {
...
+        } else {
+          theTerm()->consume(ep,1);
+        }
+      }
+    }
+  }
+}
+
```
PhilMiller: Can we rewrite this as an equivalent of

```c++
bool getProduceEpoch() {
  if (msg_ == nullptr
    || !envelopeIsEpochType(msg_->env)
    || envelopeIsTerm(msg_->env)
  )
    return no_epoch;

  return envelopeGetEpoch(msg_->env);
}

// in PendingSend::PendingSend
epoch_produced_ = getProduceEpoch();
if (epoch_produced_  != no_epoch)
  theTerm()->produce(epoch_produced_, 1);

// in sendMsg()
if (epoch_produced_  != no_epoch)
  theTerm()->consume(epoch_produced_, 1);
```

src/vt/messaging/active.impl.h
```diff
@@ -159,6 +159,12 @@ ActiveMessenger::PendingSendType ActiveMessenger::broadcastMsgSz(
...
+  auto const is_epoch = envelopeIsEpochType(msg->env);
+  if (is_epoch) {
+    setupEpochMsg(msg);
+  }
+
   return sendMsgSz(this_node, han, msg, msg_size);
```
PhilMiller: Why not add that code in `sendMsgSz` - it would be easier to validate if epoch setting on the message were exactly alongside construction of each `PendingSend`

src/vt/messaging/pending_send.cc
lifflander: Done.

src/vt/messaging/active.impl.h
lifflander: Good point, I actually missed some of the callsites.

src/vt/messaging/active.impl.h
```diff
@@ -116,6 +116,12 @@ ActiveMessenger::sendMsgSz(
   ByteType const& msg_size
 ) {
   envelopeSetup(msg->env, dest, han);
+
+  auto const is_epoch = envelopeIsEpochType(msg->env);
+  if (is_epoch) {
```
PhilMiller: I went through the whole file to confirm that every site where we construct a `PendingSend` has the epoch being set. That's all good.

All of those sites have a redundancy, though - `setupEpochMsg` does the check of the envelope type, so they can all just be bare, unconditional calls to that.

lifflander: Good point. I will remove that redundancy.

lifflander: Fixed

PR #632, opened 2019-12-18
630 progress function

### LINKED ISSUE #630: Poll MPI progress functions less often based on wall time and handlers processed
**What Needs to be Done?**

When you do that, note that it can't be strictly per trip through the scheduler, there also needs to be a time aspect - either k handlers run, or t ms since the last poll

lifflander:
Fixes #630

### DIFF
--- a/src/vt/configs/arguments/args.cc
+++ b/src/vt/configs/arguments/args.cc
@@ -470,8 +470,8 @@ namespace vt { namespace arguments {
    */
 
   auto nsched = "Number of times to run the progress function in scheduler";
-  auto ksched = "Run the MPI progress function every k handlers that run";
-  auto ssched = "Run the MPI progress function every s seconds";
+  auto ksched = "Run the MPI progress function at least every k handlers that run";
+  auto ssched = "Run the MPI progress function at least every s seconds";
   auto sca = app.add_option("--vt_sched_num_progress", vt_sched_num_progress, nsched, 2);
   auto hca = app.add_option("--vt_sched_progress_han", vt_sched_progress_han, ksched, 0);
   auto kca = app.add_option("--vt_sched_progress_sec", vt_sched_progress_sec, ssched, 0.0);
--- a/src/vt/configs/arguments/args.cc
+++ b/src/vt/configs/arguments/args.cc
@@ -58,6 +58,8 @@ namespace vt { namespace arguments {
 
 /*static*/ bool        ArgConfig::colorize_output       = false;
 /*static*/ int32_t     ArgConfig::vt_sched_num_progress = 2;
+/*static*/ int32_t     ArgConfig::vt_sched_progress_han = 1;
+/*static*/ double      ArgConfig::vt_sched_progress_sec = 0.0;
 /*static*/ bool        ArgConfig::vt_no_sigint          = false;
 /*static*/ bool        ArgConfig::vt_no_sigsegv         = false;
 /*static*/ bool        ArgConfig::vt_no_terminate       = false;
@@ -468,9 +470,15 @@ namespace vt { namespace arguments {
    */
 
   auto nsched = "Number of times to run the progress function in scheduler";
+  auto ksched = "Run the MPI progress function every k handlers that run";
+  auto ssched = "Run the MPI progress function every s seconds";
   auto sca = app.add_option("--vt_sched_num_progress", vt_sched_num_progress, nsched, 2);
+  auto hca = app.add_option("--vt_sched_progress_han", vt_sched_progress_han, ksched, 1);
+  auto kca = app.add_option("--vt_sched_progress_sec", vt_sched_progress_sec, ssched, 0.0);
   auto schedulerGroup = "Scheduler Configuration";
   sca->group(schedulerGroup);
+  hca->group(schedulerGroup);
+  kca->group(schedulerGroup);
 
   /*
    * Run the parser!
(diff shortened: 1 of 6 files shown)


src/vt/configs/arguments/args.cc
```diff
@@ -468,9 +470,15 @@ namespace vt { namespace arguments {
    */
 
   auto nsched = "Number of times to run the progress function in scheduler";
+  auto ksched = "Run the MPI progress function every k handlers that run";
```
PhilMiller: "Run the MPI progress function *at least* every ..." please

src/vt/scheduler/scheduler.cc
```diff
@@ -142,6 +143,27 @@ void Scheduler::scheduler(bool msg_only) {
...
+  // By default, `vt_sched_progress_han` is 1 and will happen every time we go
+  // through the scheduler
+  bool k_handlers_executed =
+    processed_after_last_progress_ >= ArgType::vt_sched_progress_han;
+  bool enough_time_passed =
+    progress_time_enabled_ and
+    TimeType::getCurrentTime() - last_progress_time_ > ArgType::vt_sched_progress_sec;
+
+  if ((not progress_time_enabled_ and k_handlers_executed) or enough_time_passed) {
+    runProgress(msg_only);
```
PhilMiller: There's a missing `processed_after_last_progress_ = 0` here

src/vt/scheduler/scheduler.cc
```diff
@@ -142,6 +143,27 @@ void Scheduler::scheduler(bool msg_only) {
...
+
+  // By default, `vt_sched_progress_han` is 1 and will happen every time we go
+  // through the scheduler
+  bool k_handlers_executed =
+    processed_after_last_progress_ >= ArgType::vt_sched_progress_han;
+  bool enough_time_passed =
+    progress_time_enabled_ and
+    TimeType::getCurrentTime() - last_progress_time_ > ArgType::vt_sched_progress_sec;
+
+  if ((not progress_time_enabled_ and k_handlers_executed) or enough_time_passed) {
```
PhilMiller: The nested conditions here confuse me.

You probably want `or work_queue_.empty()` in the mix somewhere as well

src/vt/scheduler/scheduler.cc
lifflander: Look above this in `runProgress`

src/vt/configs/arguments/args.cc
```diff
@@ -468,9 +470,15 @@ namespace vt { namespace arguments {
...
+  auto ssched = "Run the MPI progress function every s seconds";
   auto sca = app.add_option("--vt_sched_num_progress", vt_sched_num_progress, nsched, 2);
+  auto hca = app.add_option("--vt_sched_progress_han", vt_sched_progress_han, ksched, 0);
```
PhilMiller: This is defaulted to 0, though the comment below says it expect it to be 1. I don't see it being set somewhere else.

src/vt/runtime/runtime.cc
```diff
@@ -94,6 +94,21 @@ Runtime::Runtime(
...
   ArgType::parse(argc, argv);
+
+  // Must do this here before anything else to avoid unreported hangs
+  if (ArgType::vt_sched_progress_han != 0 and ArgType::vt_sched_progress_sec == 0.0) {
+    // vtAbort does not work here because we are uninitialized
+    auto vt_pre = debug::vtPre();
+    auto bred   = debug::bred();
+    fmt::print(
+      "{}{}If --vt_sched_progress_han != 0, "
+      "you must set --vt_sched_progress_sec to avoid hangs\n",
```
PhilMiller: I don't see the logic that leads to this conclusion.

src/vt/configs/arguments/args.cc
lifflander: I just fixed the comment. It's supposed to say 0.

tests/unit/scheduler/test_scheduler_progress.cc
```diff
@@ -0,0 +1,139 @@
...
+#include "vt/transport.h"
+#include "test_parallel_harness.h"
+
+#include <memory>
+#include <chrono>
+#include <thread>
+
+namespace vt { namespace tests { namespace unit {
+
+struct TestSchedProgress : TestParallelHarness { };
```
PhilMiller: Do these tests potentially get run concurrently with one another? That risks them stepping on each others' toes with settings of the argument variables, since those are global.

tests/unit/scheduler/test_scheduler_progress.cc
```diff
@@ -0,0 +1,139 @@
...
+
+TEST_F(TestSchedProgress, test_scheduler_progress_1) {
+  using namespace std::this_thread;
+  using namespace std::chrono;
+  using namespace std::chrono_literals;
+
+  // Run scheduler every millisecond
```
PhilMiller: I think this comment is wrong?

src/vt/runtime/runtime.cc
lifflander: Removed

src/vt/scheduler/scheduler.cc
lifflander: Added

src/vt/configs/arguments/args.cc
lifflander: Fixed

tests/unit/scheduler/test_scheduler_progress.cc
lifflander: No, great does not allow them to ever run concurrently.

tests/unit/scheduler/test_scheduler_progress.cc
lifflander: FIxed

Issue #635, opened 2019-12-19
Tracing drops off first two chars of method on bare handlers no namespace


lifflander:
**Describe the bug**
Try running `./examples/term_epoch_advanced`, look at produced sts file for the handler `propagateHandler(PropagateMsg)`. It shows as `opagateHandler(PropagateMsg`

Here's a patch that seems to solve the problem:

```diff
commit 90afedbc0c7275e58480fb0f81db9cea1fa876d7 (HEAD -> 631-trace-system-events)
Author: Jonathan Lifflander <jliffla@sandia.gov>
Date:   Thu Dec 19 12:27:56 2019 -0800

    fix small bug in demangling

diff --git a/src/vt/utils/demangle/demangle.cc b/src/vt/utils/demangle/demangle.cc
index cd2b79cf..0e0034b1 100644
--- a/src/vt/utils/demangle/demangle.cc
+++ b/src/vt/utils/demangle/demangle.cc
@@ -134,7 +134,10 @@ TemplateExtract::getBarename(std::string const& typestr) {
   size_t d = getNameDivide(typestr, s, s, 0);

   // skip "::" divider
-  d += 2;
+  if (typestr[d] == ':') {
+    d += 2;
+  }
+
   if (d >= typestr.length())
     return typestr;

```

pnstickne:
This was actually 'by design'. However, it can be .. confusing.

Will get in a change to have namespace-less types go into a "(none)" namespace for clarity.

lifflander:
Merged, closing

PR #639, opened 2019-12-20
#635 fix barename and namespace extraction

pnstickne:
- Fixes bug where barenames outside of any namespace could have
  leading characters removed.

- The "(none)" namespace placeholer is used for types that
  do not belong to any namespace.


pnstickne:
These "circleci" checks..

pnstickne:
Merge is onto existing branch tagged for 1-beta5.

lifflander:
I tried to add those checks...trying to set them up

PR #640, opened 2019-12-20
#631 trace system events

### LINKED ISSUE #631: Traces are hard to understand with system level handlers
**Describe the bug**

For every message application code sends, the resulting traces of the corresponding handler will show several events, logically nested but actually broken up by the unstacking code in the trace writer. This makes application performance analysis difficult, since none of those handlers are things the user explicitly called for, and computed message receive counts are a variable multiple of send counts.

**To Reproduce**

Run an application sending messages to handlers that are methods of process, objgroup, or collections, and look at the pile of beginProcess/endProcessing events wrapped around each application-level handler.

**Expected behavior**

Unless the user asks otherwise, all of those system-level wrapper handlers should be treated as overhead, or fully attributed to the innermost handler.

**Screenshots**

lifflander:
Fixes #631

### DIFF
--- a/src/vt/configs/arguments/args.cc
+++ b/src/vt/configs/arguments/args.cc
@@ -78,6 +78,11 @@ namespace vt { namespace arguments {
 /*static*/ std::string ArgConfig::vt_trace_dir          = "";
 /*static*/ int32_t     ArgConfig::vt_trace_mod          = 0;
 /*static*/ int32_t     ArgConfig::vt_trace_flush_size   = 0;
+/*static*/ bool        ArgConfig::vt_trace_sys_all        = false;
+/*static*/ bool        ArgConfig::vt_trace_sys_term       = false;
+/*static*/ bool        ArgConfig::vt_trace_sys_location   = false;
+/*static*/ bool        ArgConfig::vt_trace_sys_collection = false;
+/*static*/ bool        ArgConfig::vt_trace_sys_serial_msg = false;
 
 /*static*/ bool        ArgConfig::vt_lb                 = false;
 /*static*/ bool        ArgConfig::vt_lb_file            = false;
@@ -230,6 +235,11 @@ namespace vt { namespace arguments {
   auto tdir      = "Name of directory for trace files";
   auto tmod      = "Output trace file if (node % vt_stack_mod) == 0";
   auto tflushmod = "Flush output trace every (vt_trace_flush_size) trace records";
+  auto tsysall   = "Trace all system events";
+  auto tsysTD    = "Trace system termination events";
+  auto tsysloc   = "Trace system location manager events";
+  auto tsyscoll  = "Trace system virtual context collection events";
+  auto tsyssmsg  = "Trace system serialization manager events";
   auto n  = app.add_flag("--vt_trace",              vt_trace,           trace);
   auto nm = app.add_flag("--vt_trace_mpi",          vt_trace_mpi,       trace_mpi);
   auto o  = app.add_option("--vt_trace_file",       vt_trace_file,      tfile, "");
@@ -237,6 +247,11 @@ namespace vt { namespace arguments {
   auto q  = app.add_option("--vt_trace_mod",        vt_trace_mod,       tmod,  1);
   auto qf = app.add_option("--vt_trace_flush_size", vt_trace_flush_size,tflushmod,
     0);
+  auto qt = app.add_flag("--vt_trace_sys_all",        vt_trace_sys_all,        tsysall);
+  auto qw = app.add_flag("--vt_trace_sys_term",       vt_trace_sys_term,       tsysTD);
+  auto qx = app.add_flag("--vt_trace_sys_location",   vt_trace_sys_location,   tsysloc);
+  auto qy = app.add_flag("--vt_trace_sys_collection", vt_trace_sys_collection, tsyscoll);
+  auto qz = app.add_flag("--vt_trace_sys_serial_msg", vt_trace_sys_serial_msg, tsyssmsg);
   auto traceGroup = "Tracing Configuration";
   n->group(traceGroup);
   nm->group(traceGroup);
@@ -244,6 +259,11 @@ namespace vt { namespace arguments {
   p->group(traceGroup);
   q->group(traceGroup);
   qf->group(traceGroup);
+  qt->group(traceGroup);
+  qw->group(traceGroup);
+  qx->group(traceGroup);
+  qy->group(traceGroup);
+  qz->group(traceGroup);
 
 
   /*
--- a/src/vt/configs/arguments/args.h
+++ b/src/vt/configs/arguments/args.h
@@ -80,6 +80,11 @@ public:
 
   static bool vt_trace;
   static bool vt_trace_mpi;
+  static bool vt_trace_sys_all;
+  static bool vt_trace_sys_term;
+  static bool vt_trace_sys_location;
+  static bool vt_trace_sys_collection;
+  static bool vt_trace_sys_serial_msg;
   static std::string vt_trace_file;
   static std::string vt_trace_dir;
   static int32_t vt_trace_mod;
@@ -159,6 +164,19 @@ private:
 inline bool user1() { return ArgConfig::vt_user_1; }
 inline bool user2() { return ArgConfig::vt_user_2; }
 inline bool user3() { return ArgConfig::vt_user_3; }
+
+inline bool traceTerm() {
+  return ArgConfig::vt_trace_sys_term or ArgConfig::vt_trace_sys_all;
+}
+inline bool traceLocation() {
+  return ArgConfig::vt_trace_sys_location or ArgConfig::vt_trace_sys_all;
+}
+inline bool traceCollection() {
+  return ArgConfig::vt_trace_sys_collection or ArgConfig::vt_trace_sys_all;
+}
+inline bool traceSerialMsg() {
+  return ArgConfig::vt_trace_sys_serial_msg or ArgConfig::vt_trace_sys_all;
+}
 
 }} /* end namespace vt::arguments */
 
--- a/src/vt/handler/handler.h
+++ b/src/vt/handler/handler.h
@@ -68,11 +68,7 @@ static constexpr HandlerType const blank_handler = 0;
 static constexpr BitCountType const auto_num_bits = 1;
 static constexpr BitCountType const functor_num_bits = 1;
 static constexpr BitCountType const objgroup_num_bits = 1;
-
-#if backend_check_enabled(trace_enabled)
 static constexpr BitCountType const trace_num_bits = 1;
-#endif
-
 static constexpr BitCountType const control_num_bits = 20;
 static constexpr BitCountType const handler_id_num_bits =
  BitCounterType<HandlerType>::value - (
@@ -80,9 +76,7 @@ static constexpr BitCountType const handler_id_num_bits =
    + functor_num_bits
    + objgroup_num_bits
    + control_num_bits
-#if backend_check_enabled(trace_enabled)
    + trace_num_bits
-#endif
  );
 
 // eHandlerBits::ObjGroup identifies the handler as targeting an objgroup; the
@@ -91,12 +85,8 @@ enum eHandlerBits {
   ObjGroup   = 0,
   Auto       = eHandlerBits::ObjGroup   + objgroup_num_bits,
   Functor    = eHandlerBits::Auto       + auto_num_bits,
-#if backend_check_enabled(trace_enabled)
   Trace      = eHandlerBits::Functor    + functor_num_bits,
   Control    = eHandlerBits::Trace      + trace_num_bits,
-#else
-  Control    = eHandlerBits::Functor    + functor_num_bits,
-#endif
   Identifier = eHandlerBits::Control    + control_num_bits
 };
 
--- a/src/vt/handler/handler.h
+++ b/src/vt/handler/handler.h
@@ -68,10 +68,21 @@ static constexpr HandlerType const blank_handler = 0;
 static constexpr BitCountType const auto_num_bits = 1;
 static constexpr BitCountType const functor_num_bits = 1;
 static constexpr BitCountType const objgroup_num_bits = 1;
+
+#if backend_check_enabled(trace_enabled)
+static constexpr BitCountType const trace_num_bits = 1;
+#endif
+
 static constexpr BitCountType const control_num_bits = 20;
 static constexpr BitCountType const handler_id_num_bits =
  BitCounterType<HandlerType>::value - (
-   auto_num_bits + functor_num_bits + objgroup_num_bits + control_num_bits
+     auto_num_bits
+   + functor_num_bits
+   + objgroup_num_bits
+   + control_num_bits
+#if backend_check_enabled(trace_enabled)
+   + trace_num_bits
+#endif
  );
 
 // eHandlerBits::ObjGroup identifies the handler as targeting an objgroup; the
@@ -80,7 +91,12 @@ enum eHandlerBits {
   ObjGroup   = 0,
   Auto       = eHandlerBits::ObjGroup   + objgroup_num_bits,
   Functor    = eHandlerBits::Auto       + auto_num_bits,
+#if backend_check_enabled(trace_enabled)
+  Trace      = eHandlerBits::Functor    + functor_num_bits,
+  Control    = eHandlerBits::Trace      + trace_num_bits,
+#else
   Control    = eHandlerBits::Functor    + functor_num_bits,
+#endif
   Identifier = eHandlerBits::Control    + control_num_bits
 };
 
@@ -91,7 +107,8 @@ struct HandlerManager {
 
   static HandlerType makeHandler(
     bool is_auto, bool is_functor, HandlerIdentifierType id,
-    bool is_objgroup = false, HandlerControlType control = 0
+    bool is_objgroup = false, HandlerControlType control = 0,
+    bool is_trace = true
   );
   static void setHandlerIdentifier(HandlerType& han, HandlerIdentifierType id);
   static void setHandlerControl(HandlerType& han, HandlerControlType control);
@@ -104,6 +121,10 @@ struct HandlerManager {
   static bool isHandlerAuto(HandlerType han);
   static bool isHandlerFunctor(HandlerType han);
   static bool isHandlerObjGroup(HandlerType han);
+#if backend_check_enabled(trace_enabled)
+  static void setHandlerTrace(HandlerType& han, bool is_trace);
+  static bool isHandlerTrace(HandlerType han);
+#endif
 };
 
 } //end namespace vt
--- a/src/vt/messaging/active.cc
+++ b/src/vt/messaging/active.cc
@@ -262,21 +262,26 @@ EventType ActiveMessenger::sendMsgSized(
   auto const is_epoch = envelopeIsEpochType(msg->env);
 
   #if backend_check_enabled(trace_enabled)
-    auto const handler = envelopeGetHandler(msg->env);
-    bool const is_auto = HandlerManagerType::isHandlerAuto(handler);
-    if (is_auto) {
-      trace::TraceEntryIDType ep = auto_registry::handlerTraceID(
-        handler, auto_registry::RegistryTypeEnum::RegGeneral
+    // We are not allowed to hold a ref to anything in the envelope, get this,
+    // modify it and put it back
+    auto handler = envelopeGetHandler(msg->env);
+
+    // If trace_this is set locally on the envelope (which was set when the
+    // message was created before the handler was setup based on configuration),
+    // we need to set the trace bits on the handler. This enables the
+    // handler-inside-handler to correctly enable tracing on a very specific
+    // part of the execution, configured by the user (e.g., location tracing
+    // enabled, termination tracing disabled)
+    HandlerManagerType::setHandlerTrace(handler, envelopeGetTraceThis(msg->env));
+
+    // Update the handler with trace bits now set appropriately
+    envelopeSetHandler(msg->env, handler);
+
+    if (not is_bcast or (is_bcast and dest == this_node_)) {
+      makeTraceCreationSend(
+        base, handler, auto_registry::RegistryTypeEnum::RegGeneral,
+        msg_size, is_bcast
       );
-      if (not is_bcast) {
-        trace::TraceEventIDType event = theTrace()->messageCreation(ep, msg_size);
-        envelopeSetTraceEvent(msg->env, event);
-      } else if (is_bcast and dest == this_node_) {
-        trace::TraceEventIDType event = theTrace()->messageCreationBcast(
-          ep, msg_size
-        );
-        envelopeSetTraceEvent(msg->env, event);
-      }
     }
   #endif
 
--- a/src/vt/messaging/active.cc
+++ b/src/vt/messaging/active.cc
@@ -266,13 +266,15 @@ EventType ActiveMessenger::sendMsgSized(
     // modify it and put it back
     auto handler = envelopeGetHandler(msg->env);
 
-    // If trace_this is set locally on the envelope (which was set when the
-    // message was created before the handler was setup based on configuration),
-    // we need to set the trace bits on the handler. This enables the
-    // handler-inside-handler to correctly enable tracing on a very specific
-    // part of the execution, configured by the user (e.g., location tracing
-    // enabled, termination tracing disabled)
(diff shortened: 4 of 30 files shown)


src/vt/configs/arguments/args.cc
```diff
@@ -78,6 +78,11 @@ namespace vt { namespace arguments {
 /*static*/ std::string ArgConfig::vt_trace_dir          = "";
 /*static*/ int32_t     ArgConfig::vt_trace_mod          = 0;
 /*static*/ int32_t     ArgConfig::vt_trace_flush_size   = 0;
+/*static*/ bool        ArgConfig::vt_trace_sys_all        = false;
```
pnstickne: It feels like this is overall N-flags is a less-than-ideal approach to handle sets..

src/vt/configs/arguments/args.h
```diff
@@ -158,6 +163,19 @@ inline bool user1() { return ArgConfig::vt_user_1; }
 inline bool user2() { return ArgConfig::vt_user_2; }
 inline bool user3() { return ArgConfig::vt_user_3; }
 
+inline bool traceTerm() {
```
pnstickne: Non-consistency here :|

It seems like the final 'state' if internal state variables should be handled separately from the source of the variable in a post-argument stage. There is a certain branch that I'd love to *continue* such work on.. 🔢

src/vt/messaging/active.impl.h
```diff
@@ -62,6 +63,53 @@ void ActiveMessenger::setTermMessage(MsgPtrT const msg) {
...
+
+template <typename MsgPtrT>
+void ActiveMessenger::makeTraceCreationSend(
+  MsgPtrT msg, HandlerType const handler, auto_registry::RegistryTypeEnum type,
+  MsgSizeType msg_size, bool is_bcast
+) {
+  #if backend_check_enabled(trace_enabled)
+    auto cur_event = envelopeGetTraceEvent(msg->env);
+    if (cur_event == trace::no_trace_event) {
+      trace::TraceEntryIDType ep = auto_registry::handlerTraceID(handler, type);
```
pnstickne: These names.. always double-take on Entry vs Event :|

src/vt/runnable/collection.impl.h
```diff
@@ -69,7 +69,12 @@ template <typename MsgT, typename ElementT>
     trace::TraceEntryIDType trace_id = auto_registry::handlerTraceID(
       handler, reg_enum
     );
-    trace::TraceEventIDType trace_event = theMsg()->getCurrentTraceEvent();
+    // Note: get the event from the envelope to get the semantic connection for
+    // collections---no the current thing running in the queue
```
pnstickne: "not"

src/vt/runnable/general.impl.h
```diff
@@ -121,7 +126,9 @@ template <typename MsgT>
   }
 
   #if backend_check_enabled(trace_enabled)
-    theTrace()->endProcessing(trace_id, msg_size, trace_event, from_node);
+    if (is_traced) {
+      theTrace()->endProcessing(trace_id, msg_size, trace_event, from_node);
```
pnstickne: Would be nice to `theTrace()->endProcessing(id_from_before)` :|

src/vt/utils/demangle/demangle.cc
```diff
@@ -134,7 +134,10 @@ TemplateExtract::getBarename(std::string const& typestr) {
   size_t d = getNameDivide(typestr, s, s, 0);
 
   // skip "::" divider
-  d += 2;
+  if (typestr[d] == ':') {
```
pnstickne: Other PR is over - just note for rebase.

src/vt/vrt/collection/manager.impl.h
```diff
@@ -643,6 +643,7 @@ template <typename>
   );
   if (msg->isRoot()) {
     auto new_msg = makeMessage<CollectionConsMsg>(*msg);
+    theMsg()->setCollectionMessage(new_msg);
```
pnstickne: I really want to flip these .. part of SharedMsgPtr work.

src/vt/trace/trace.cc
```diff
@@ -124,10 +124,21 @@ Trace::Trace() { }
...
+      theTrace()->endIdle();
+    }
+  #endif
+}
+
 void Trace::initialize() {
   theSched()->registerTrigger(
     sched::SchedulerEvent::BeginIdle, traceBeginIdleTrigger
   );
+  theSched()->registerTrigger(
```
pnstickne: There is other code in trace.cc which will 'end a trace event when another even is encountered'.

If the begin/end are accurate it might worthwhile to eliminate that logic. (Any overlaps would represent ACTUAL overlaps then.)

src/vt/messaging/active.impl.h
```diff
@@ -62,6 +63,53 @@ void ActiveMessenger::setTermMessage(MsgPtrT const msg) {
...
+      if (not is_bcast) {
+        trace::TraceEventIDType event = theTrace()->messageCreation(ep, msg_size);
+        envelopeSetTraceEvent(msg->env, event);
+      } else {
+        trace::TraceEventIDType event = theTrace()->messageCreationBcast(
+          ep, msg_size
+        );
+        envelopeSetTraceEvent(msg->env, event);
```
pnstickne: Common to both branches.

pnstickne: trace::TraceEventIDType event;
      if (not is_bcast) {
        event = theTrace()->messageCreation(ep, msg_size);
      } else {
        event = theTrace()->messageCreationBcast(ep, msg_size);
      }
      envelopeSetTraceEvent(msg->env, event);

src/vt/messaging/active.impl.h
```diff
@@ -62,6 +63,53 @@ void ActiveMessenger::setTermMessage(MsgPtrT const msg) {
...
   envelopeSetPriority(msg->env, sys_min_priority);
 #endif
+#if backend_check_enabled(trace_enabled)
+  //fmt::print("arguments::traceTerm()={}\n",arguments::traceTerm());
+  envelopeSetTraceThis(msg->env, arguments::traceTerm());
+#endif
+}
+
+template <typename MsgPtrT>
+void ActiveMessenger::setLocationMessage(MsgPtrT const msg) {
```
pnstickne: These names were/are confusing to me, as as reader. The information that it is trace-related is lost.

Compare:
      
      enableLocationTrace

Or

      enableTrace(msg, additive_flags)

Or, pretty much anything indicating it's a trace :|

src/vt/messaging/envelope/envelope_get.impl.h
```diff
@@ -87,6 +87,11 @@ template <typename Env>
 inline trace::TraceEventIDType envelopeGetTraceEvent(Env& env) {
   return reinterpret_cast<Envelope*>(&env)->trace_event;
 }
+
+template <typename Env>
+inline bool envelopeGetTraceThis(Env& env) {
+  return reinterpret_cast<Envelope*>(&env)->trace_this;
```
pnstickne: Possibly `trace_enabled` instead, for parity and to break association with `this`.

src/vt/messaging/envelope/envelope_setup.impl.h
```diff
@@ -70,6 +70,10 @@ inline void envelopeInit(Env& env) {
   envelopeSetPriority(env, min_priority);
   envelopeSetPriorityLevel(env, 0);
 # endif
+# if backend_check_enabled(trace_enabled)
+  envelopeSetTraceThis(env, true);
+  envelopeSetTraceEvent(env, trace::no_trace_event);
```
pnstickne: From previous comment (which I now partially retract), it would then be something like:

     envelopeSetTraceEvent(env, trace::enable_trace_sentinel);

src/vt/trace/trace.cc
pnstickne: Also, the event could be emitted *immediately* and thus avoid the internal trace idle flags.

src/vt/configs/arguments/args.cc
pnstickne: e.g, consider arg syntax:

     -vt_trace sys,user,..

pnstickne:
Codacy wants to steal my information to login :|

pnstickne:
I still believe the event output stream can be .. 'optimized' in terms of how writes occur.

That would likely require a new contract definition and client handling.

src/vt/messaging/active.impl.h
pnstickne: Uhg.. now I see the naming :|

src/vt/trace/trace.cc
lifflander: There is a reason this is required now. We are not emitting system events, so the trace should show overhead (black) instead of ideal. Thus we need to end the idle explicitly.

src/vt/configs/arguments/args.cc
pnstickne: Later time..

src/vt/runnable/general.impl.h
pnstickne: Incoming 🍡

src/vt/trace/trace.cc
pnstickne: @lifflander Alright, I can see for the second case: what about the first?

That is, if begin + end idle events mark the actual idle time, use them as points to emit events. A new event (even if it were somehow to be trigged 'when idle') would not implicit trigger and end idle event. Any overlaps would represent event emission errors (or even possibly valid overlaps).

pnstickne: Wrt above https://github.com/DARMA-tasking/vt/blob/b53076be583d79a989ccd108ad504119484e863a/src/vt/trace/trace.cc#L579

Current implicit-end (which works on state, vs immediate event emitting):

    // close any idle event as soon as we encounter any other type of event
    if (idle_begun_ and
        log->type != TraceConstantsType::BeginIdle and
        log->type != TraceConstantsType::EndIdle) {
      endIdle();
    }

pnstickne: Assuming that no other traces can be written WHILE in idle, between the idle-start and idle-end callbacks, the above code / state management is stale. (See new vs potentially different forms above.)

pnstickne: And with the state reduction:

```
/*static*/ void Trace::traceBeginIdleTrigger() {
    theTrace()->beginIdle();
  #endif
}

/*static*/ void Trace::traceEndIdleTrigger() {
    theTrace()->endIdle();
  #endif
}
```

src/vt/messaging/active.impl.h
```diff
@@ -62,6 +63,51 @@ void ActiveMessenger::setTermMessage(MsgPtrT const msg) {
...
+#if backend_check_enabled(trace_enabled)
```
pnstickne: Okay, these names are still confusing - slept on it a day, although the previous `traceXYZ` was incomplete in scope.

A name similar to `markAsLocationMessage` would read easier for me overall, and fit in with `setTermMessage` which does more than tracing. Giving it a go:

- markAsTermMessage
- makeAsLocationMessage
- markAsSerialMsgMessage
- markAsCollectionMessage

lifflander: Ok, that's a good idea. Will update

pnstickne:
Any reason not to move to develop? (minus checks)

src/vt/messaging/envelope/envelope_set.impl.h
```diff
@@ -123,6 +123,11 @@ template <typename Env>
 inline void envelopeSetTraceEvent(Env& env, trace::TraceEventIDType const& evt) {
   reinterpret_cast<Envelope*>(&env)->trace_event = evt;
 }
+
+template <typename Env>
+inline void envelopeSetTraceThis(Env& env, bool in) {
+  reinterpret_cast<Envelope*>(&env)->trace_this = in;
```
PhilMiller: Should this be `static_cast` as seen in other recent issues?

lifflander:
> Any reason not to move to develop? (minus checks)

I fix the small issues and get this on develop.

src/vt/messaging/envelope/envelope_get.impl.h
PhilMiller: `static_cast` here - let's not make things worse. Or just add a commit correcting all of these

pnstickne:
@lifflander Ready to move to PR-ready status?

lifflander:
@PhilMiller @pnstickne I plan to address these and get this merged today.

src/vt/configs/arguments/args.h
lifflander: I agree this needs to be addressed more broadly, tabling this for now so i can move forward on this PR.

src/vt/runnable/collection.impl.h
lifflander: Fixed.

src/vt/utils/demangle/demangle.cc
lifflander: Rebased out

src/vt/messaging/active.impl.h
lifflander: Fixed.

src/vt/messaging/envelope/envelope_get.impl.h
lifflander: Addressed this

src/vt/messaging/active.impl.h
lifflander: Fixed

src/vt/messaging/envelope/envelope_set.impl.h
lifflander: Yes, but I'm going to leave this for another PR to cleanup and test.

src/vt/messaging/envelope/envelope_get.impl.h
lifflander: Again, I don't want to mess with now and make sure the compiler doesn't produce a bad offset in some way.

src/vt/trace/trace.cc
lifflander: I agree, but want this done in a separate issue. Created one for this; https://github.com/DARMA-tasking/vt/issues/685

src/vt/messaging/envelope/envelope_get.impl.h
lifflander: Created a separate issue for this: #684

src/vt/handler/handler.h
```diff
@@ -68,10 +68,21 @@ static constexpr HandlerType const blank_handler = 0;
 static constexpr BitCountType const auto_num_bits = 1;
 static constexpr BitCountType const functor_num_bits = 1;
 static constexpr BitCountType const objgroup_num_bits = 1;
+
+#if backend_check_enabled(trace_enabled)
+static constexpr BitCountType const trace_num_bits = 1;
```
pnstickne: Might be reasonable to leave in all builds.

src/vt/messaging/envelope/envelope_set.h
```diff
@@ -186,6 +186,15 @@ inline void envelopeSetPriorityLevel(Env& env, PriorityLevelType priority_level)
...
 inline void envelopeSetTraceEvent(Env& env, trace::TraceEventIDType const& evt);
+
+/**
+ * \brief Set whether this envelope is traced at runtime
+ *
+ * \param[in,out] env the envelope
+ * \param[in] in whether it should be traced
+ */
+template <typename Env>
+inline void envelopeSetTraceRuntimeEnabled(Env& env, bool in);
```
pnstickne: `in` could be a wee bit more descriptive.

src/vt/runnable/collection.h
```diff
@@ -54,7 +54,8 @@ struct RunnableCollection {
   static void run(
```
pnstickne: Might be worth considering overloads so the basic forms (no/reduced optionals) don't deal with extra information.

src/vt/runnable/general.impl.h
```diff
@@ -76,16 +76,17 @@ template <typename MsgT>
...
 #if backend_check_enabled(trace_enabled)
   trace::TraceProcessingTag processing_tag;
-  {
+  bool is_traced = HandlerManagerType::isHandlerTrace(handler);
+  if (is_traced) {
     trace::TraceEntryIDType trace_id = auto_registry::handlerTraceID(
       handler, auto_registry::RegistryTypeEnum::RegGeneral
     );
-    trace::TraceEventIDType trace_event = trace::no_trace_event;
+    trace::TraceEventIDType  trace_event = trace::no_trace_event;
```
pnstickne: Extra space.

src/vt/runnable/general.impl.h
```diff
@@ -123,7 +124,9 @@ template <typename MsgT>
   }
 
 #if backend_check_enabled(trace_enabled)
-  theTrace()->endProcessing(processing_tag);
+  if (is_traced) {
+    theTrace()->endProcessing(processing_tag);
```
pnstickne: If the processing_tag is the default value then it will suppress inside the `endProcessing` function, so the condition here can be removed.

src/vt/runnable/general.impl.h
lifflander: Oops, whitespace here is off

src/vt/runtime/runtime.cc
```diff
@@ -573,6 +573,32 @@ void Runtime::printStartupBanner() {
       auto f12 = opt_inverse("--vt_trace_flush_size", f11);
       fmt::print("{}\t{}{}", vt_pre, f12, reset);
     }
+    if (not ArgType::vt_trace_sys_all) {
```
pnstickne: Could simplify lots of this logic if just specify state (no on/off etc.).

src/vt/serialization/messaging/serialized_messenger.impl.h
```diff
@@ -309,8 +309,10 @@ template <typename MsgT, typename BaseT>
   auto const& total_size = ptr_size + msg_size + han_size + size_size;
   auto const& tag = no_tag;
   if (is_bcast) {
+    theMsg()->markAsSerialMsgMessage(msg);
```
pnstickne: Could be moved outside of condition.

src/vt/serialization/messaging/serialized_messenger.impl.h
```diff
@@ -415,13 +418,24 @@ template <typename MsgT, typename BaseT>
...
     return theMsg()->broadcastMsg<PayloadMsg,payloadMsgHandler>(
       payload_msg.get()
     );
   } else {
     auto const& total_size = ptr_size + sys_size;
 
-    sys_msg->handler = han;
+    auto han_modified = han;
+
+#   if backend_check_enabled(trace_enabled)
```
pnstickne: Several different layout variations of `#` used.

src/vt/serialization/messaging/serialized_messenger.impl.h
```diff
@@ -415,13 +418,24 @@ template <typename MsgT, typename BaseT>
...
 
+    theMsg()->markAsSerialMsgMessage(payload_msg);
     return theMsg()->broadcastMsg<PayloadMsg,payloadMsgHandler>(
       payload_msg.get()
     );
   } else {
     auto const& total_size = ptr_size + sys_size;
 
-    sys_msg->handler = han;
+    auto han_modified = han;
```
pnstickne: Maybe `traceable_han` or some such. Definitely good to have a different variable in either case.

src/vt/runtime/runtime.cc
lifflander: I like the symmetry here. Breaking out earlier is easier to read

src/vt/runnable/general.impl.h
lifflander: I like the symmetry here. Breaking out earlier is easier to read I think

src/vt/topos/location/location.impl.h
```diff
@@ -406,6 +408,7 @@ void EntityLocationCoord<EntityID>::routeMsgNode(
     envelopeGetRef(msg->env), msg->getLocFromNode(), print_ptr(msg.get()),
     epoch
   );
+  theMsg()->markAsLocationMessage(msg);
```
pnstickne: Seems odd snuggled to the debug (instead of the following if).

src/vt/topos/location/location.impl.h
```diff
@@ -716,6 +737,7 @@ template <typename EntityID>
           inst, entity, event_id, ask_node, home_node
         );
         msg2->setResolvedNode(node);
+        theMsg()->markAsLocationMessage(msg2);
```
pnstickne: Going all the way back, these to be exposed as `theMsg()` for consumers? If not, having a clear `\internal` mark on the Doxygen would be helpful encouragement to not use.

src/vt/topos/location/location.impl.h
lifflander: Fixed

src/vt/serialization/messaging/serialized_messenger.impl.h
lifflander: Fixed

src/vt/serialization/messaging/serialized_messenger.impl.h
lifflander: I recently started switching to starting with `#` and then indenting because I think techically that's the correct way according to the standard. I haven't gone through and updated all the code. I will create an issue to unify the style across the codebase.

https://github.com/DARMA-tasking/vt/issues/686

src/vt/serialization/messaging/serialized_messenger.impl.h
lifflander: Done

src/vt/runnable/general.impl.h
lifflander: Fixed

lifflander: Fixed

src/vt/runnable/collection.h
lifflander: Agreed, good task for the future

src/vt/messaging/envelope/envelope_set.h
lifflander: Done

src/vt/runtime/runtime.cc
lifflander: Agreed, let's do this in a separate issue

src/vt/handler/handler.h
lifflander: Done

src/vt/topos/location/location.impl.h
lifflander: I think this will require a larger refactor into a new category (internal to VT access to `ActiveMessenger`). I agree this needs to be done. For now, I am tabling this.

lifflander: https://github.com/DARMA-tasking/vt/issues/687

pnstickne:
Phone (mobile hotspot) ran out of power earlier D:

src/vt/messaging/active.cc
```diff
@@ -262,21 +262,31 @@ EventType ActiveMessenger::sendMsgSized(
...
+    HandlerManagerType::setHandlerTrace(
+      handler, envelopeGetTraceRuntimeEnabled(msg->env)
+    );
+
+    // Update the handler with trace bits now set appropriately
+    envelopeSetHandler(msg->env, handler);
+
+    if (not is_bcast or (is_bcast and dest == this_node_)) {
+      // auto cur_event = envelopeGetTraceEvent(msg->env);
+      // if (cur_event == trace::no_trace_event) {
```
PhilMiller: What's with the commented out code left here?

src/vt/vrt/collection/manager.impl.h
```diff
@@ -1002,6 +1014,19 @@ messaging::PendingSend CollectionManager::broadcastMsgUntypedHandler(
 
   auto msg = promoteMsg(raw_msg);
 
+  #if backend_check_enabled(trace_enabled)
+    // Create the trace creation event for the broadcast here to connect it a
```
PhilMiller: Missing a word "connect it **to** a higher semantic level"

src/vt/vrt/collection/manager.impl.h
```diff
@@ -1424,6 +1449,21 @@ messaging::PendingSend CollectionManager::sendMsgUntypedHandler(
       }
     #endif
 
+    #if backend_check_enabled(trace_enabled)
+      // Create the trace creation event here to connect it a higher semantic
```
PhilMiller: Again, missing 'to'

PhilMiller:
Would it be worth considering this for beta.5 as well? Besides improving the quality of the traces, it would also sharply reduce their size when not tracing system-level events separately.

PR #644, opened 2019-12-21
643 trace log no uptr

pnstickne:


src/vt/trace/trace.cc
```diff
@@ -551,42 +551,38 @@ bool Trace::checkDynamicRuntimeEnabled() {
   return enabled_ and traceWritingEnabled(theContext()->getNode());
 }
 
-TraceEventIDType Trace::logEvent(std::unique_ptr<LogType> log) {
-  assert(log != nullptr && "log cannot be null");
-
+TraceEventIDType Trace::logEvent(LogType&& log) {
```
pnstickne: All call-sites now temp objects-to-rvalue-refs.

src/vt/trace/trace.cc
```diff
@@ -551,42 +551,38 @@ bool Trace::checkDynamicRuntimeEnabled() {
...
-      auto end_log = std::make_unique<LogType>(
-        open_events_.top(),
-        logTime,
-        TraceConstantsType::EndProcessing
-      );
-      traces_.push(std::move(end_log));
+      double logTime = log.time;
+      traces_.push(
+        LogType(open_events_.top(), logTime, TraceConstantsType::EndProcessing
+      ));
```
pnstickne: Parenthesis placement incorrect.

src/vt/trace/trace_log.h
```diff
@@ -52,59 +52,166 @@
...
-  TraceConstantsType type = TraceConstantsType::InvalidTraceType;
-  TraceEventIDType event = no_trace_event;
-  NodeType node = uninitialized_destination;
-  TraceMsgLenType msg_len = 0;
-  uint64_t idx1 = 0, idx2 = 0, idx3 = 0, idx4 = 0;
+  union Data {
+
+    // System event data; cannot be used in 'user events'.
+    // Type is simple with auto-defined ctor/copy/move..
+    // Expected ~40 bytes
```
pnstickne: Overlap of approximately 40 bytes, depending on std::string implementation.

src/vt/trace/trace_log.h
```diff
@@ -52,59 +52,166 @@
 #include <cstdint>
 #include <vector>
 #include <memory>
+#include <new>
```
pnstickne: For placement-new.

src/vt/trace/trace_log.h
```diff
@@ -52,59 +52,166 @@
...
+    Data(UserData const& user_data) {
+      new (&user) UserData{user_data};
+    }
+
+    Data(SysData const& sys_data) : sys(sys_data) {
+    }
+
+    ~Data() {
+      // Doesn't (can't) actually cleanup due to no access to which
+      // union is activated. See destroy() which MUST be called by ~Log().
```
pnstickne: "..union member.."

src/vt/trace/trace_log.h
```diff
@@ -113,42 +220,94 @@ struct Log {
...
-      NodeType in_node)
-    : time(in_time), type(in_type),
-      node(in_node)
-  { }
+  Log(
+    double const in_time, TraceConstantsType const in_type,
+    NodeType in_node
+  ) : time(in_time), type(in_type),
+      node(in_node),
+      data(Data::SysData{0, 0, 0, 0, 0}),
```
pnstickne: Can just be `Data::StsData{}`.

src/vt/trace/trace_log.h
```diff
@@ -113,42 +220,94 @@ struct Log {
...
   Log(
     double const in_time, TraceEntryIDType const in_ep, TraceConstantsType const in_type,
     NodeType in_node,
     TraceMsgLenType const in_msg_len
-  ) : time(in_time), ep(in_ep), type(in_type),
+  ) : time(in_time), type(in_type), ep(in_ep),
       node(in_node),
-      msg_len(in_msg_len)
-  { }
+      data(Data::SysData{in_msg_len, 0, 0, 0, 0}),
```
pnstickne: Can probably just be `Data::SysData{in_msg_len}`.

src/vt/trace/trace_log.h
```diff
@@ -113,42 +220,94 @@ struct Log {
...
+  // NOT VALID FOR USER EVENTS
   Log(
     Log const& in, double in_time, TraceConstantsType in_type
-  ) : time(in_time), ep(in.ep), type(in_type), event(in.event),
-      node(in.node), msg_len(in.msg_len),
-      idx1(in.idx1), idx2(in.idx2), idx3(in.idx3), idx4(in.idx4)
-  { }
+  ) : time(in_time), type(in_type), ep(in.ep), event(in.event),
+      node(in.node),
+      data(in.data.sys),
```
pnstickne: It's "safe" to access data.sys, although such will result in ill-defined output, hence the assert too.
Might also just be `data(in.sys_data())` which has an assert inside it..

src/vt/trace/trace_log.h
```diff
@@ -113,42 +220,94 @@ struct Log {
...
+      data_type(LogDataType::sys)
+  {
+  }
+
+  ~Log() {
+    data.destroy(data_type);
+  }
+
+  inline Data::UserData const& user_data() const {
+    assert(data_type == LogDataType::user && "Expecting user data-type");
```
pnstickne: Alternative to assert, return dummy/no-data references.

src/vt/trace/trace_log.h
```diff
@@ -113,42 +220,94 @@ struct Log {
...
+  TraceEventIDType event = no_trace_event;
+  // If the event relates to a DIFFERENT node.
+  NodeType node = uninitialized_destination;
+
+private:
+
+  // Union of sys/user data
+  /*union*/ Data data;
+
+  // Active union struct..
```
pnstickne: Might miss alignment - possible move above `data` to see if it can snuggle up to one of the previous members.

pnstickne: Meh, don't care.

src/vt/trace/trace_log.h
```diff
@@ -52,59 +52,166 @@
...
   Log(
     double const in_begin_time, double const in_end_time,
     TraceConstantsType const in_type, std::string const& in_note,
     TraceEventIDType const in_event
   ) : time(in_begin_time), end_time(in_end_time),
-      type(in_type), event(in_event), user_supplied_note(in_note)
-  { }
+      type(in_type), event(in_event),
+      data(Data::UserData{in_note, 0, 0, false}),
+      data_type(LogDataType::user)
```
pnstickne: Could potentially pass reference into `data` and avoid `data_type` although such feels "wrong". Anyway, the union itself cannot capture the 'type' of union it is.

src/vt/trace/trace_log.h
```diff
@@ -52,59 +52,166 @@
...
+        sys = data.sys;
+      }
+    }
 
-  std::string user_supplied_note = "";
-  UserDataType user_supplied_data = 0;
-  UserEventIDType user_event = 0;
-  bool user_start = false;
+    // Move based on type
+    Data(LogDataType data_type, Data&& data) {
```
pnstickne: Doesn't seem possible to use initialization list in this case..

src/vt/trace/trace_log.h
```diff
@@ -52,59 +52,166 @@
...
+      UserData& operator=(UserData const&) = delete;
+      UserData& operator=(UserData&&) = delete;
+
+      ~UserData() = default;
+    } user;
+
+    // Copy based on type
+    Data(LogDataType data_type, Data const& data) {
+      if (data_type == Log::LogDataType::user) {
+        new (&user) UserData{data.user};
```
pnstickne: User data requires placement-new as it is uninitialized storage.

src/vt/trace/trace_log.h
```diff
@@ -52,59 +52,166 @@
...
+
+      UserData(
+        std::string const& user_note, UserDataType user_data,
+        UserEventIDType user_event, bool user_start
+      ) : user_note(user_note), user_data(user_data),
+          user_event(user_event), user_start(user_start)
+      {
+      }
+
+      UserData(UserData const&) = default;
```
pnstickne: // Used for placement-new

src/vt/trace/trace_log.h
```diff
@@ -52,59 +52,166 @@
...
+      TraceMsgLenType msg_len;
+      uint64_t idx1;
+      uint64_t idx2;
+      uint64_t idx3;
+      uint64_t idx4;
+    } sys;
+
+    // User event data; cannot be used in 'normal events'.
+    // Extra hoopla for non-trivial std::string type involved.
+    // Expected ~30-40 bytes.
```
pnstickne: `std::string` storage varies a bit based on implementation. In MSVC++ (2019, 64-bit) it is 28 bytes. YMMV.

src/vt/trace/trace.h
```diff
@@ -69,9 +69,8 @@ struct vt_gzFile;
 struct Trace {
   using LogType             = Log;
   using TraceConstantsType  = eTraceConstants;
-  using TraceContainersType = TraceContainers;
   using TimeIntegerType     = int64_t;
-  using TraceContainerType  = std::queue<std::unique_ptr<LogType>>;
```
pnstickne: Now that there is no keeping around the LogType* for the open-close stack.. can drop all unique_ptr and push through copy-move.

The dequeue implementation (or "storage container used") probably itself uses dynamic memory allocations (subject to whatever internal optimizations); those are a separate concern.

src/vt/trace/trace.cc
```diff
@@ -774,65 +766,71 @@ void Trace::writeTracesFile(int flush) {
...
+    TraceEntrySeqType event_seq_id = log.ep == no_trace_entry_id
       ? 0 // no_trace_entry_seq != 0 (perhaps shift offsets..).
-      : TraceRegistry::getEvent(log->ep).theEventSeq();
+      : TraceRegistry::getEvent(log.ep).theEventSeq();
 
-    switch (log->type) {
-    case TraceConstantsType::BeginProcessing:
+    switch (log.type) {
+    case TraceConstantsType::BeginProcessing: {
+      auto const& sdata = log.sys_data();
```
pnstickne: Events can have either system-data (`sys_data()`) or user-data (`user_data()`).

src/vt/trace/trace.cc
```diff
@@ -610,17 +606,14 @@ TraceEventIDType Trace::logEvent(std::unique_ptr<LogType> log) {
...
+    double logTime = log.time;
     traces_.push(std::move(log));
 
     // Emit a '[re]start' event for the reactivated stack item.
-    auto begin_log = std::make_unique<LogType>(
-      open_events_.top(),
-      logTime,
-      TraceConstantsType::BeginProcessing
+    traces_.push(
+      LogType(open_events_.top(), logTime, TraceConstantsType::BeginProcessing)
```
pnstickne: This is rv-ref, and thus move-semantics. How cool are temporary objects?

src/vt/trace/trace.cc
```diff
@@ -551,42 +551,38 @@ bool Trace::checkDynamicRuntimeEnabled() {
...
-      traces_.push(std::move(end_log));
+      double logTime = log.time;
+      traces_.push(
+        LogType(open_events_.top(), logTime, TraceConstantsType::EndProcessing
+      ));
     }
 
-    // Saved copy has independent lifetime
-    open_events_.push(LogType(*log.get()));
+    open_events_.push(log /* copy, not forwarding r-ref */);
```
pnstickne: "rv-ref".

src/vt/trace/trace.cc
```diff
@@ -642,23 +635,23 @@ TraceEventIDType Trace::logEvent(std::unique_ptr<LogType> log) {
...
     break;
   default:
     vtAssert(0, "Not implemented");
     return 0;
   }
 
   // Normal case of event emitted at end
-  TraceEventIDType event = log->event;
+  TraceEventIDType event = log.event;
   traces_.push(std::move(log));
```
pnstickne: And forward rv-ref for move semantics - for sys-data there is little benefit to copy-semantics due to POD. However, for user-data and std::string .. maybe some.

Overall, I'd expect a normal event to have [one] copy/move here .. and MAYBE temp object elision. For small objects (< ~80 bytes with union, should be A-OK).
(A few more copy/moves if a begin..end event.)

src/vt/trace/trace.cc
```diff
@@ -610,17 +606,14 @@ TraceEventIDType Trace::logEvent(std::unique_ptr<LogType> log) {
       break;
```
pnstickne: Missing comment explaining `break;` case.

    // No events to restart - add log to traces normally.

src/vt/trace/trace_log.h
pnstickne: Nope. Things start to get "confusing" with POD types and exactly which rule applies. The simple value-initialization case is very much more clear.

src/vt/trace/trace.cc
pnstickne: Fixed.

src/vt/trace/trace_log.h
pnstickne: Switched to using a "common initial sequence" which is a much simpler way to deal with the problem..

src/vt/trace/trace_log.h
pnstickne: (Using an initialization list and-then-copy would be slightly UB due to access a potentially non-active union member; it would require additional logic pushed in to technically avoid UB and thus likely defeat potential compiler optimizations..)

src/vt/trace/trace_log.h
pnstickne: Not doing. Usage only as needed.

src/vt/trace/trace_log.h
pnstickne: Okay, great. Can probably figure that out..

src/vt/trace/trace.cc
pnstickne: Added comment (back). Yay.

src/vt/trace/trace_log.h
```diff
@@ -52,59 +52,163 @@
...
-  uint64_t idx1 = 0, idx2 = 0, idx3 = 0, idx4 = 0;
+  union Data {
+
+    // System event data; cannot be used in 'user events'.
+    // Type is simple with auto-defined ctor/copy/move..
+    // Expected ~40 bytes
+    struct SysData {
+    public:
+      // "Common initial sequence"
+      LogDataType data_type;
```
pnstickne: Special C+11 guarantee:

"One special guarantee is made in order to simplify the use of unions: if a union contains several structures that share a common initial sequence (see below), and if the union object currently contains one of these structures, it is permitted to inspect the common initial part of any of them anywhere that a declaration of the completed type of the union is visible. Two structures share a common initial sequence if corresponding members have compatible types (and, for bit-fields, the same widths) for a sequence of one or more initial members."

src/vt/trace/trace_log.h
lifflander: `std::string` will vary also based on short string optimization. It maybe larger than expected to accommodate the short string non-heap allocation.

src/vt/trace/trace_log.h
```diff
@@ -52,59 +52,163 @@
...
+  Log(
+    Log const& in
+  ) : time(in.time), end_time(in.end_time),
+      type(in.type), ep(in.ep), event(in.event),
+      // copy any kind of data
+      data(in.data)
+  {
+  }
+
+  Log(
```
lifflander: Can we call the default move constructor here?

src/vt/trace/trace_log.h
lifflander: I rely strongly on this guarantee for envelopes in messages which carefully order the members for casting when the message arrives without the specific type on the other end of the network.

src/vt/trace/trace_log.h
pnstickne: @lifflander The sizes above look large enough to SBO, excluding the old GCC method which was a pointer-to-object. (For some reason that appears no longer strictly permitted for std::string these days..)

Also, pushed a few amends just a minute ago ^_^

pnstickne:
@lifflander Alright, been sitting on this too long. Ready for review.

Soft-blocks https://github.com/DARMA-tasking/vt/issues/647 as there will be annoying same-line merge conflicts.

src/vt/trace/trace_log.h
pnstickne: Yay! And it works with Clang 3.9 ;-)
Finding such documentation also helps me reinforce 🖌

src/vt/trace/trace_log.h
pnstickne: That's a good question..

pnstickne: And "yes". Looks like I didn't clean these up after making Log::Data 'self aware'.

src/vt/trace/trace.h
lifflander: This is great. This simplifies this code a lot.

src/vt/trace/trace_log.h
lifflander: This will help with EMPIRE memory usage.

src/vt/trace/trace_log.h
```diff
@@ -52,59 +52,165 @@
...
+      uint64_t idx4;
+
+    private:
+      friend struct Log;
+
+      SysData(
+        TraceMsgLenType msg_len,
+        uint64_t in_idx1 = 0, uint64_t in_idx2 = 0,
+        uint64_t in_idx3 = 0, uint64_t in_idx4 = 0
+      ) : msg_len(msg_len),
```
nlslatt: Constructor argument shadows data member, leading to incorrect behavior.

src/vt/trace/trace_log.h
```diff
@@ -52,59 +52,165 @@
...
+      // user_data() or sys_data(), which return references,
+      // is internal to Log. No external copies for you.
+      friend struct Log;
 
-  // [obsolete] - use appropriate ctor
+      UserData(
+        std::string const& user_note, UserDataType user_data,
+        UserEventIDType user_event, bool user_start
+      ) : user_note(user_note), user_data(user_data),
+          user_event(user_event), user_start(user_start)
```
nlslatt: Constructor arguments shadows data members, leading to incorrect behavior.

PR #475, opened 2019-10-01
444 split runtime initialization

uhetmaniuk:
Split the runtime initialization.
Allow for a third-party specification of arguments.
Set precedence rule for each argument.
Generate the startup banner.
Give the option to create a ".INI" file with the simulation parameters.

PR #646, opened 2019-12-26
Split stage runtime initialization and argument resolution (Cleaned)

uhetmaniuk:
** Replaces closed pull-request #475  (with a cleaner history for a simpler rebase)

Split the runtime initialization.
Allow for a third-party specification of arguments.
Set precedence rule for each argument.
Generate the startup banner.
Give the option to create a ".INI" file with the simulation parameters.


lifflander:
Way too complicated.

PR #648, opened 2020-01-07
645 Doxygen

### LINKED ISSUE #645: Make doxygen work
**What Needs to be Done?**

- [ ] First do this
- [ ] Then do that.
- [ ] Also this other thing.

**Additional context**

lifflander:
Fixes #645


src/vt/context/context.h
```diff
@@ -54,33 +54,105 @@
 
 namespace vt {  namespace ctx {
 
+/** \file */
```
pnstickne: Would be useful to have wiki page on what format(s) of Doxygen are preferred.
It doesn't have to be exclusive... I like simple `/// briefs` myself.

PR #655, opened 2020-01-26
633 unify msgptr 2

pnstickne:


examples/trace_ring.cc
```diff
@@ -89,7 +89,7 @@ static void ring(RingMsg* msg) {
 }
 
 static void sendToNext() {
-  RingMsg* msg = new RingMsg(my_node);
+  auto msg = makeSharedMessage<RingMsg>(my_node);
```
pnstickne: Any 'new Msg' from the outside is now considered a program flaw and WILL FAIL programs with a vtAssert.

This was the only such usages remaining in the examples.

src/vt/messaging/active.cc
```diff
@@ -180,9 +179,14 @@ EventType ActiveMessenger::sendMsgBytes(
 ) {
   auto const& msg = base.get();
 
+  // Ensure valid-ref message.
```
pnstickne: Probably not necessary.

src/vt/messaging/active.h
```diff
@@ -155,7 +155,7 @@ struct BufferedActiveMsg {
 
 struct ActiveMessenger {
   using BufferedMsgType      = BufferedActiveMsg;
-  using MessageType          = ShortMessage*;
+  using RecvMsgType          = ShortMessage;
```
pnstickne: Explicit `*` at usage sites much more clear.

src/vt/messaging/active.h
```diff
@@ -179,14 +179,41 @@ struct ActiveMessenger {
...
+  /*!
+    \internal
+    Invoke message sending.
+    Should be funnel-through method.
+  */
+  template <typename MessageT>
+  ActiveMessenger::PendingSendType sendMsgImpl(
+    NodeType dest,
+    HandlerType han,
+    MsgPtr<MessageT>& msg,
```
pnstickne: Most MsgPtr should be & at call interfaces to avoid ref management work. Take a copy inside if/as relvant.

src/vt/messaging/active.h
```diff
@@ -208,45 +235,40 @@ struct ActiveMessenger {
...
-    TagType const& tag
-  );
-
-  template <typename MsgT>
-  PendingSendType sendMsg(
-    NodeType const& dest, HandlerType const& han, MsgSharedPtr<MsgT> const& msg
+    NodeType dest,
+    HandlerType han,
+    MessageT* msg,
+    TagType tag = no_tag
```
pnstickne: External methods all take tag as optional (no overloads).

src/vt/messaging/active.h
```diff
@@ -208,45 +235,40 @@ struct ActiveMessenger {
...
   );
 
   template <typename MsgT>
   PendingSendType sendMsg(
-    NodeType const& dest, HandlerType const& han, MsgSharedPtr<MsgT> const& msg,
-    TagType const& tag
+    NodeType dest,
+    HandlerType han,
+    MsgPtr<MsgT>& msg,
```
pnstickne: Most external calls currently also accept a MsgT* although some accept a MsgPtr<MsgT>& .. there should be consistent overloads for the MsgPtr form such that makeMessage can be used without an additional 'get' operation on it .. it's awkward to see `sendMsg(..., makeMessage<...>(..).get())`. However, if accepting a & it's still not possible to accept an rv-ref directly..

src/vt/event/event_record.cc
```diff
@@ -87,7 +87,13 @@ bool EventRecord::testMPIEventReady() {
 
   bool const mpiready = flag == 1;
 
-  if (mpiready and msg_ != nullptr and isSharedMessage(msg_.get())) {
+  if (mpiready and msg_ != nullptr) {
+    // Ensure valid-ref message.
+    vtAssertInfo(
```
pnstickne: Probably should be removed. See other comments.

src/vt/messaging/active.cc
```diff
@@ -197,9 +201,7 @@ EventType ActiveMessenger::sendMsgBytes(
     );
   }
 
-  if (is_shared) {
```
pnstickne: All messages should be 'shared' and use same ref-counting measures.
Adding wrapping type and/or construct to manage epoch is separate.

src/vt/messaging/active.cc
```diff
@@ -774,13 +776,17 @@ bool ActiveMessenger::tryProcessIncomingActiveMsg() {
...
 void ActiveMessenger::finishPendingActiveMsgAsyncRecv(InProgressIRecv* irecv) {
-  auto buf = irecv->buf;
+  char* buf = irecv->buf;
   auto num_probe_bytes = irecv->probe_bytes;
   auto sender = irecv->sender;
 
-  auto msg = reinterpret_cast<MessageType>(buf);
-  messageConvertToShared(msg);
-  auto base = promoteMsgOwner(msg);
+  auto msg = static_cast<RecvMsgType*>(static_cast<void*>(buf));
```
pnstickne: These casts are done in several different ways.. anyway, this is the relevant rule of the reinterpret_cast.

src/vt/messaging/active.cc
```diff
@@ -774,13 +776,17 @@ bool ActiveMessenger::tryProcessIncomingActiveMsg() {
...
+  // Reset envelope to avoid ref-count=not_shared_msg if the data
+  // comes from a 'new RecvMsgType(..)'.
+  // The MsgPtr will then acquire a ref-count itself.
+  // TODO: Maybe envelope ref-count should be 0 on initialize?
```
pnstickne: If it was 0 on initialize then one could still be allowed to supply a `new Message` from the outside. However, the goal is to eliminate this option.

src/vt/messaging/active.h
```diff
@@ -273,43 +295,50 @@ struct ActiveMessenger {
 
   template <typename MessageT, ActiveTypedFnType<MessageT>* f>
   PendingSendType broadcastMsgSz(
```
pnstickne: 'Sz' methods are currently be considered internal (tag required).

src/vt/messaging/active.h
```diff
@@ -273,43 +295,50 @@ struct ActiveMessenger {
...
   );
 
   template <typename MessageT, ActiveTypedFnType<MessageT>* f>
   PendingSendType sendMsgSz(
-    NodeType const& dest, MessageT* const msg, ByteType const& msg_size,
-    TagType const& tag = no_tag
+    NodeType dest,
+    MessageT* msg,
+    ByteType msg_size,
+    TagType tag = no_tag
```
pnstickne: Exception not always? :|

src/vt/messaging/active.h
```diff
@@ -273,43 +295,50 @@ struct ActiveMessenger {
...
 
   /*
    *  Auto method for dispatching to the serialization framework if required
    *  based on examining compile-time traits of the message
    */
+
   template <typename MessageT, ActiveTypedFnType<MessageT>* f>
-  PendingSendType sendMsgAuto(
-    NodeType const& dest, MessageT* const msg, TagType const& tag
+  PendingSendType broadcastMsgAuto(
```
pnstickne: Groups in broadcast.. send.. ordering.

src/vt/messaging/active.h
```diff
@@ -431,49 +452,46 @@ struct ActiveMessenger {
...
-    HandlerType const& han, MessageT* const msg, TagType const& tag
+    HandlerType han,
+    MsgPtr<MsgT>& msg,
+    TagType tag = no_tag
   );
 
-  template <typename MsgT>
-  PendingSendType broadcastMsg(
-    HandlerType const& han, MsgSharedPtr<MsgT> const& msg
+  template <typename MessageT>
```
pnstickne: These methods feel very out-of-place, especially as they only support very limited handlers.

src/vt/messaging/active.impl.h
```diff
@@ -56,276 +56,272 @@
...
+}
+
+template <typename MessageT>
+ActiveMessenger::PendingSendType ActiveMessenger::sendMsgAuto(
+  NodeType dest,
+  HandlerType han,
+  MessageT* msg,
+  TagType tag
+) {
+  return ActiveSendHandler<MessageT>::sendMsg(dest,msg,han,tag);
```
pnstickne: Formatting. Also, divergence in direct funnel.

src/vt/messaging/active.impl.h
```diff
@@ -56,276 +56,272 @@
...
-  return PendingSendType(base, msg_size);
+template <typename MessageT, ActiveTypedFnType<MessageT>* f>
+ActiveMessenger::PendingSendType ActiveMessenger::sendMsg(
+  NodeType dest,
+  MessageT* msg,
+  TagType tag
+) {
+  auto const han = auto_registry::makeAutoHandler<MessageT,f>(msg);
+  MsgPtr<MessageT> msgptr = promoteMsg(msg);
+  return sendMsgImpl<MessageT>(dest, han, msgptr, -1, tag);
```
pnstickne: Use constant/constrexpr instead of -1 for clarity. Also avoid implicit conversion.

src/vt/messaging/envelope/envelope_base.h
```diff
@@ -58,6 +58,10 @@ struct ActiveEnvelope {
   EnvelopeDataType type : envelope_num_bits;
   NodeType dest         : node_num_bits;
   HandlerType han       : handler_num_bits;
+  // Reference count for envelope for lifetime management.
```
pnstickne: Semantic change to drop 'shared' nature. It is *probably* valid to consider a deref *from* zero to be invalid, as it would have been be done manually.

src/vt/messaging/message/refs.impl.h
```diff
@@ -72,7 +72,11 @@ void messageDeref(MsgT* msg) {
     print_ptr(msg), refcount
   );
 
-  if (refcount == 0) {
+  // N.B. makeSharedMessage does NOT increase the envelope ref-count anymore.
```
pnstickne: Semantic change to drop 'shared' nature. It is probably valid to consider a deref from zero to be invalid, as it would have been be done manually (that is: not with MsgPtr, makeMessage, or makeSharedMessage+promoteMessage).

src/vt/messaging/message/shared_message.impl.h
```diff
@@ -70,54 +70,48 @@ template <typename MsgT, typename... Args>
 MsgT* makeSharedMessage(Args&&... args) {
   messageTypeChecks<MsgT>();
   MsgT* msg = new MsgT{std::forward<Args>(args)...};
-  envelopeSetRef(msg->env, 1);
-  msg->has_owner_ = false;
+  // n.b. do NOT actually take a ref here.
```
pnstickne: Taking a ref here would be 1->2(promoteMsg to MsgPtr)->1(non-release)

pnstickne: Also, should have less code duplication if/as possible

src/vt/messaging/message/shared_message.impl.h
```diff
@@ -70,54 +70,48 @@ template <typename MsgT, typename... Args>
...
 
-///[obsolete] Use makeMessage instead.
 template <typename MsgT, typename... Args>
-MsgSharedPtr<MsgT> makeMsg(Args&&... args) {
+[[deprecated("Use makeMessage instead")]]
+MsgPtr<MsgT> makeMsg(Args&&... args) {
   return makeMessage<MsgT>(std::forward<Args>(args)...);
 }
 
-template <typename MsgT>
```
pnstickne: All messages are 'shared'.

src/vt/messaging/message/shared_message.impl.h
```diff
@@ -70,54 +70,48 @@ template <typename MsgT, typename... Args>
...
-void messageConvertToShared(MsgT* msg) {
-  envelopeSetRef(msg->env, 1);
-}
-
-template <typename MsgT>
-void messageSetUnmanaged(MsgT* msg) {
-  envelopeSetRef(msg->env, not_shared_message);
-}
-
-template <typename MsgT>
```
pnstickne: Folded into specific use-case for clarity (with comments and code redux).

src/vt/messaging/message/smart_ptr.h
```diff
@@ -52,13 +52,13 @@
 #include <iosfwd>
 #include <cassert>
 
-namespace vt { namespace messaging {
+namespace vt { namespace impl { namespace messaging {
```
pnstickne: "Meh". Not really sure about this specific change - pushing to perhaps separate the 'impl' in VT from types users might decide to use even if they aren't explicit designed as the public API.

src/vt/messaging/message/smart_ptr.h
```diff
@@ -187,116 +190,108 @@ struct MsgSharedPtr final {
...
-        "MsgSmartPtr: (auto) init(), ptr={}, envRef={}, takeRef={} address={}\n",
-        print_ptr(msgPtr), envelopeGetRef(msgPtr->env), takeRef, print_ptr(this)
-      );
-
-      if (takeRef) {
-        // Could be moved to type-erased impl..
-        messageRef(msgPtr);
-      }
+
+    vtAssertInfo(
```
pnstickne: Add explicit check for -1000 (not_shared_message) in a vtAssert for a better diagnostic message as (when) programs fail due to use of `new Message..`.

src/vt/messaging/message/smart_ptr.h
```diff
@@ -187,116 +190,108 @@ struct MsgSharedPtr final {
...
 
   /// Clear all internal state. Effectively destructor in operation,
   /// with guarantee that init() can be used again after.
   void clear() {
-    bool shared = shared_;
-
-    if (shared) {
-      assert("shared -> message ptr" && ptr_);
-      T* msgPtr = get();
+    if (not ptr_) {
```
pnstickne: Idempotent for cases like `MsgPtr x = ..; x = nullptr; /* x leaves scope */`

src/vt/messaging/message/smart_ptr.h
```diff
@@ -167,136 +167,123 @@ struct MsgSharedPtr final {
...
     assert("given message" && msgPtr);
     assert("not initialized" && not ptr_);
     assert("given impl" && impl);
 
     ptr_ = msgPtr;
     impl_ = impl;
-    bool shared = (shared_ = isSharedMessage<T>(msgPtr));
-
-    if (shared) {
-      vtAssertInfo(
```
pnstickne: Asserts part of core envelop ref/deref.

src/vt/serialization/messaging/serialized_messenger.impl.h
```diff
@@ -62,6 +63,17 @@
 
 namespace vt { namespace serialization {
 
+// Deserializes a Msg (owned by a MsgPtr) from the given data.
+template <typename MsgT>
+static MsgPtr<MsgT> deserializeFullMessage(SerialByteType* source, size_t len) {
+  auto msg = makeMessage<MsgT>();
+  deserializeInPlace<MsgT>(source, len, msg.get());
+  // Deserialization overwrites the envelop ref-count;
+  // as it was guaranteed to be 1 previously, reset the count.
```
pnstickne: ..guaranteed by makeMessage..

In this case it would probably be more clear to use makeSharedMessage, then wrap in promoteMsg at end - no need to call envelopeSetRef then.

pnstickne: Done.

src/vt/messaging/active.h
pnstickne: Tag optional for consistency.

src/vt/messaging/active.h
pnstickne: Tag now optional for consistency.

src/vt/messaging/envelope/envelope_base.h
pnstickne: Done. Previous 0->defref case NOT supported.

src/vt/messaging/envelope/envelope_ref.impl.h
```diff
@@ -52,11 +52,24 @@ namespace vt {
 
 template <typename Env>
 inline void envelopeRef(Env& env) {
+  vtAssert(
```
pnstickne: It doesn't make sense to assert ">=0" specifically because at -1 there is UB due to expected release..

src/vt/messaging/envelope/envelope_ref.impl.h
```diff
@@ -52,11 +52,24 @@ namespace vt {
...
+    "instead of a MsgPtr/makeMessage/makeSharedMessage construct."
+  );
+
   (reinterpret_cast<Envelope*>(&env))->ref++;
 }
 
 template <typename Env>
 inline RefType envelopeDeref(Env& env) {
+  vtAssertInfo(
+    (reinterpret_cast<Envelope*>(&env))->ref >= 1,
```
pnstickne: Now strengthen the contract, such that a deref without a pared ref (even initially) is invalid. Previous commits (amended) allowed such..

src/vt/messaging/message/smart_ptr.h
```diff
@@ -167,136 +167,123 @@ struct MsgSharedPtr final {
...
-        envelopeGetRef(msgPtr->env) > 0, "Bad Ref (before deref)",
-        shared, envelopeGetRef(msgPtr->env)
-      );
-      debug_print(
-        pool, node,
-        "MsgSmartPtr: (auto) clear(), ptr={}, envRef={}, address={}\n",
-        print_ptr(msgPtr), envelopeGetRef(msgPtr->env), print_ptr(this)
-      );
+    vtAssertInfo(
```
pnstickne: Should be removed (now part of core deref..)

src/vt/serialization/messaging/serialized_messenger.impl.h
```diff
@@ -130,12 +142,10 @@ template <typename UserMsgT>
...
     [handler,recv_tag,node,epoch,is_valid_epoch]
     (RDMA_GetType ptr, ActionType action){
-      // be careful here not to use "msg", it is no longer valid
-      auto raw_ptr = reinterpret_cast<SerialByteType*>(std::get<0>(ptr));
-      auto ptr_size = std::get<1>(ptr);
-      auto msg = makeMessage<UserMsgT>();
-      deserializeInPlace<UserMsgT>(raw_ptr, ptr_size, msg.get());
-      messageResetDeserdes(msg);
+      // be careful here not to use "sys_msg", it is no longer valid
+      auto msg_data = reinterpret_cast<SerialByteType*>(std::get<0>(ptr));
```
pnstickne: Example of different (or now different) conversion method - also to static_cast paring, or unified call?

src/vt/serialization/messaging/serialized_messenger.impl.h
```diff
@@ -54,6 +54,7 @@
 #include "vt/runnable/general.h"
 #include "vt/serialization/messaging/serialized_data_msg.h"
 #include "vt/serialization/messaging/serialized_messenger.h"
+#include "vt/messaging/envelope/envelope_set.h" // envelopeSetRef
```
pnstickne: Would be ideal to use 'message' (not 'envelop') methods..

src/vt/messaging/active.h
```diff
@@ -153,9 +153,18 @@ struct BufferedActiveMsg {
   { }
 };
 
+/*!
+ * Primary interface to send messages, as through 'theMsg()'.
+ *
+ * Message ownership semantics on sending:
+ *
+ * Any call to a sendMsg* or broadcastMsg* method LOSES TRANSFER OWNERSHIP of the
```
pnstickne: !!!!

src/vt/messaging/active.h
pnstickne: Might be worth considering accepting MsgT&, MsgT&& and then having the caller use that directly:

  auto msg = MyMsg{1,2,3,4};
  sendMsg(msg)  // const&, requires temp
  sendMsg(std::move(msg)) // &&, via std::move
  sendMsg(MyMsg{1,2,3,4}) // &&, via rv(?)-ref
  sendMsg(&msg)  // * -- BAD

Then the external API would change to accept such and MsgPtr would be more of an internal concern. The code in send/broadcast msg could perform immediate wire serialization to avoid internal copies - this could also avoid heap allocations of messages themselves; would have to play with acq-release semantics.

pnstickne: There is awkwardness with 'env' (in particular, it should generally NOT be copied) - however, that could be (and is currently) addressed with being elided from a copy. So the only way to pass around a message internally is with MsgPtr<MsgT> or MsgT*.

src/vt/messaging/active.impl.h
pnstickne: Done.

src/vt/messaging/message/refs.impl.h
pnstickne: Done.

src/vt/messaging/message/smart_ptr.h
pnstickne: Done - part of base envelope.

src/vt/messaging/active.h
pnstickne: Another alternative is to accept MsgPtr&& as secondary and switch usages of 'makeSharedMsg<T>(...)' to `MsgPtr<T>{...}`. As an xref that would be cleanly accepted as well:

    sendMsg<F,..>(MsgPtr<F>{..})

pnstickne:
See #665 (draft PR #667) for cleaner branch.

PR #658, opened 2020-02-03
657 clean active message

pnstickne:
Flatten types, combine methods with optionals, simplify call-trees, remove non-necessary code. There is a plan.. largely a cleaner base for an improved API.

Also, fun change: **sendMsg and broadcastMsg will automatically serialize, as needed, thus making sendMsgAuto/broadcastMsgAuto not-needed.** (If an application wants to ensure a type will be serialized or not, it should add it's own static checks over the types it defines.)

The 'eliminates' the ability to send a message without serialization if it has serialization, if such is even generally valid. Messages and message hierarchies should be non-conflicting in this sense.

---

Each message that has a serialization function must define if it **requires** or **supports** serialization or **prohibits** serialization. Messages are required to call the serialize function of their base message type. Validation of serialization calling through the base-type is handled as a run-time construct while other validation is performed with static compilation checks.

The rules for requires/supports/prohibits are validated at compile time (Clang can also validate if a type defines / doesn't define a serialize function. GCC, not so much..).
- A type that supports serialization must be trivially copyable and implement a serialize function. It can be implemented by a type that supports/requires/prohibits serialization, as it is a super set of such. _A type that supports serialization will still be sent without using serialization._
- A type that prohibits serialization must/should not have a serialize function. It must be byte copyable (defined as being trivially copyable and trivially destructible). It can only derive from a type which supports or prohibits serialization.
- A type that requires serialize must have a serialize function. It does not have to be byte copyable. It can only derive from a type which supports or requires serialization.

There a several different flavors, and this branch currently contains two.

(This is the  proposed macro form, which opts for convention, eg. `MessageParentType` or similar, instead of the current commits that use some "macro magic".)

```
template <typename T>
struct EpochGraphMsg : collective::reduce::operators::ReduceTMsg<T> {
  using MessageParentType = collective::reduce::operators::ReduceTMsg<T>;

  // Specify the serialization mode. In this case serialization is enabled if the base type
  // (which in turn will require serialization if T requires to be serialized).
  // All messages with a serialize function must specific their serialization desires.
  // Messages without serialization default to byte-copyable and are RECOMMENDED
  // to explicit declare themselves as serialization-prohibited.
  msg_serialize_if_needed_by_base();

    MessageParentType::serialize(s); // always serialize parent
    s | ...
  }
};
```

The second is to use CRTP, although the effect is the same. The code is definitely easier to write, maintain, and it will catch type "miss base-d" type errors even without a serialize function.

However, it also is a little bit longer for client types to implement and introduces additional types.. granted such are merely a compilation concerns as they should be optimized out well.

```
struct ReduceMsg : SerializeSupported<
  ::vt::Message,
  ReduceMsg
>, ReduceLink
{
  using Base = SerializeSupported<::vt::Message, ReduceMsg>;

    Base::serialize(s);
    s | ...;
  }
};
```

In addition to directly declaring support/required/prohibited, a message may also define if it must be CONDITIONALLY required, based on either the parent type or a template parameter. This provides the changes with sufficient flexibility to handle the case of `ReduceDataMsg<DataType>` and derivates. This special case has been called out in subsequent PR comments.

The two methods are not mutually exclusive and one could be used for internal vs. external (to reduce types introduce via CRTP and forwarding methods, for example).


src/vt/vrt/collection/manager.impl.h
```diff
@@ -2113,7 +2111,7 @@ CollectionManager::constructMap(
     "construct_map: range={}\n", range.toString().c_str()
   );
 
-  SerdesMsg::broadcastSerialMsg<MsgType,distConstruct<MsgType>>(
+  theMsg()->broadcastMsg<MsgType,distConstruct<MsgType>>(
```
pnstickne: "Auto serialize magic".

tests/unit/scheduler/test_scheduler_priorities.cc
```diff
@@ -71,7 +71,7 @@ TEST_F(TestSchedPriorities, test_scheduler_priorities_1) {
 
   // Make a term message
   auto msg = vt::makeMessage<vt::term::TermMsg>(no_epoch);
-  vt::theMsg()->setTermMessage(msg);
+  vt::theMsg()->setTermMessage(msg.get());
```
pnstickne: Tiny breaking change: only accepts Msg*

tests/unit/serialization/test_serialize_messenger.cc
```diff
@@ -133,9 +133,8 @@ TEST_F(TestSerialMessenger, test_serial_messenger_1) {
       using TupleType = std::tuple<int, int, int>;
 
       auto msg = makeSharedMessage<DataMsg<TupleType>>(TupleType{val1,val2,val3});
-      SerializedMessenger::sendSerialMsg<
-        DataMsg<TupleType>, testHandler<TupleType>
-      >(1, msg);
+      auto han = auto_registry::makeAutoHandler<DataMsg<TupleType>,testHandler<TupleType>>(nullptr);
```
pnstickne: From user-code, should only be from `theMsg()->send..` or similar.

pnstickne:
> The 'eliminates' the ability to send a message without serialization if it has serialization, if such is even generally valid. Messages and message hierarchies should be non-conflicting in this sense.

A type is considered serializable if the MsgT conforms to the serializabion traits.
For type inheritance is means that all base messages (up until one that DOESN'T support serialization) will follow serialization..

So this type hierarchy is valid and results in the expected behavior:
- Child(S) : ActiveMessage(S,C) -> serialized
- GrandChild(S) : Child(S) : ActiveMessage(S,C) -> serialized
- GrandChild(S) : Child(S,C) : ActiveMessage(S,C) -> serialized
- Child(C) : ActiveMessage(S,C) -> byte-copied 
- GrandChild(C) : Child(C) ActiveMessage(S,C) -> byte-copied
- GrandChild(C) : Child(S,C) ActiveMessage(S,C) -> byte-copied

While these type hierarchies are not supported:
- GrandChild(S): Child(C): ActiveMessage(S,C) -> ill-defined: child should be serializable
- GrandChild(C): Child(S): ActiveMessage(S,C) -> ill-defined: child is not trivially copyable

S = serializable
C = non-serializable (AND trivially copyable)

src/vt/messaging/message/message_serialize.h
```diff
@@ -0,0 +1,151 @@
...
+/** \file */
+
+// These macros are always available, regardless of serialization support.
+// (Otherwise they would have to be guarded in non-serialization builds.)
+
+#define msg_serialize_mode_(MODE) \
+  static constexpr ::vt::messaging::SerializationMode vt_serialize_mode \
+    = ::vt::messaging::SerializationMode::MODE
+
+/**
```
pnstickne: Require _explicitly_ marking serializable message with their serialization expectations.
- Resolves the issue where an intermediate type that had a serializer could be 'accidentally serialized' when not needed to be.
- Cross-verifies that a serialization method has actually be implemented on the message
- Allows serialization to be 'required' or 'supported' based on base message type and/or fields from template parameters

This does not yet check / prevent a non-serialized type from inheriting from a type that must be serialized.

pnstickne: Currently the setup allows a message to check the base message for serialization.. however, if every method explicitly defined/indicated a base/parent message then the enforcement along type hierarchies (that is, a non-serializable type cannot inherit from a serialization-required type).. that might a good positive step and eliminate some additional have-to-manually-do complexities.

pnstickne: The strict requirement can be relaxed (eg. with a define only set by default in core-VT builds) that will set the default to 'allow' which results in lack-of-validation.

pnstickne:
Hmm. Some of the test executables are .. "missing"?

examples/comm_func.cc
```diff
@@ -49,13 +49,17 @@
 #include <array>
 
 struct PingMsg : vt::Message {
+  msg_parent_type = ::vt::Message;
```
pnstickne: Basic idea is that each message that defines a `serialize` (or non-intrusive version) is required to define if it's **required** (will be used to transmit) or **supported** (called by subtypes that require serialization). If a type is not to be serialized it can be marked as **prohibited**.

examples/comm_func.cc
```diff
@@ -49,13 +49,17 @@
...
+  msg_serialize_required(); // by payload_
+
   PingMsg() = default;
   explicit PingMsg(int64_t size) {
     payload_.resize(size);
   }
 
   template <typename SerializerT>
   void serialize(SerializerT& s) {
+    msg_serialize_parent(s);
```
pnstickne: All types a responsible to call serialize on their parent. As all types that **support** serialization must provide this method there is a clear call graph back up to ActiveMsg.

examples/comm_func.cc
```diff
@@ -75,7 +79,7 @@ static void handler(PingMsg*) {
   count++;
   if (count == pings) {
     auto msg = vt::makeMessage<PingMsg>(1);
-    vt::theMsg()->sendMsgAuto<PingMsg,done>(0, msg.get());
```
pnstickne: There is never a need for call-sites to make a determination of how to handle messages.

examples/jacobi1d_vt.cc
```diff
@@ -193,9 +193,8 @@ struct LinearPb1DJacobi : vt::Collection<LinearPb1DJacobi,Index1D> {
 
 
   struct VecMsg : vt::CollectionMessage<LinearPb1DJacobi> {
-
-    IdxBase from_index = 0;
-    double val = 0.0;
+    msg_parent_type = vt::CollectionMessage<LinearPb1DJacobi>;
+    msg_serialize_if_needed_by_base_or_type1(IdxBase);
```
pnstickne: Messages serialization can be selected based on parent/and or (template) types.

src/vt/messaging/message/message_serialize.h
```diff
@@ -0,0 +1,305 @@
...
+ * \brief Mark the message as requiring serialization.
+ *
+ * It is an error if the message does not provide serialization support.
+ */
+#define msg_serialize_required()                                          \
+  msg_serialize_mode_(require);                                           \
+  static_assert(                                                          \
```
pnstickne: Static asserts check parent type to ensure a valid call tree.

src/vt/messaging/message/message_serialize.h
```diff
@@ -0,0 +1,305 @@
...
+  )
+
+/**
+ * \brief Mark message as serializable if needed by any of the types.
+ *
+ * If the message does not need to be serialized it will be treated
+ * as supporting serialization.
+ */
+#define msg_serialize_if_needed_by_type1(TYPE1)                           \
+  void vt_serialize_defined_on_type() {}                                  \
```
pnstickne: TODO: verify compatible parent.

src/vt/messaging/message/message_serialize.h
```diff
@@ -0,0 +1,305 @@
...
+     : ::vt::messaging::SerializationMode::support)
+
+namespace vt { namespace messaging {
+
+enum SerializationMode
+{
+  not_specified,     // default value
+  support,           // this has a serialization function that DERIVED types can use.
+                     // this type itself is NOT SERIALIZED for transmission.
+  allow,             // allow serialization, if supported by the type
```
pnstickne: TODO: eliminate allow (no longer useful/used).

src/vt/pipe/interface/callback_direct.h
```diff
@@ -86,9 +89,8 @@ struct CallbackDirect : RemoteContainerMsg<MsgT,std::tuple<CallbackT>> {
 
   template <typename SerializerT>
   void serialize(SerializerT& s) {
-    BaseType::serialize(s);
```
pnstickne: Perhaps require that messages define a `use BaseType` (replace `msg_parent_type`)..

src/vt/pipe/interface/callback_direct.h
```diff
@@ -61,6 +61,9 @@ namespace vt { namespace pipe { namespace interface {
 
 template <typename MsgT, typename CallbackT>
 struct CallbackDirect : RemoteContainerMsg<MsgT,std::tuple<CallbackT>> {
```
pnstickne: TODO: verify if this is _actually_ a message. If not, update the name and remove the msg_* usages.

src/vt/pipe/interface/callback_direct_multi.h
```diff
@@ -64,6 +64,8 @@ static struct CallbackDirectSendMultiTagType {} CallbackDirectSendMultiTag { };
 
 template <typename MsgT, typename TupleT>
 struct CallbackDirectSendMulti : RemoteContainerMsg<MsgT,TupleT> {
```
pnstickne: TODO: verify is used as a message and update accordingly.

src/vt/pipe/interface/remote_container_msg.h
```diff
@@ -59,6 +59,9 @@ namespace vt { namespace pipe { namespace interface {
 
 template <typename MsgT, typename TupleT>
 struct RemoteContainerMsg : BaseContainer<MsgT> {
```
pnstickne: TODO: ..

src/vt/serialization/messaging/serialized_messenger.impl.h
```diff
@@ -186,95 +186,35 @@ template <typename UserMsgT, typename BaseEagerMsgT>
...
-template <typename MsgT, ActiveTypedFnType<MsgT> *f, typename BaseT>
-/*static*/ messaging::PendingSend
- SerializedMessenger::broadcastParserdesMsg(MsgT* msg) {
-  debug_print(
-    serial_msg, node,
-    "broadcastParserdesMsg: msg={}\n", print_ptr(msg)
-  );
-  return parserdesMsg<MsgT,f>(msg,true);
+  return parserdesMsgSendImpl<MsgT>(msg, handler, dest);
 }
```
pnstickne: TODO: remove all parserdes (currently not called in from ActiveMessage anyway).

src/vt/vrt/collection/balance/hierarchicallb/hierlb_msgs.h
```diff
@@ -64,7 +67,6 @@ struct LBTreeUpMsg : HierLBTypes, ::vt::Message {
       child_size_(in_child_size)
   { }
 
-
   #if hierlb_use_parserdes
```
pnstickne: TODO: remove parserdes

src/vt/vrt/collection/balance/stats_msg.h
```diff
@@ -154,8 +154,16 @@ struct LoadData {
   int32_t  P_ = 0;
 };
 
+static_assert(
```
pnstickne: TODO: collect/unify type-checks. See if it is work exposing as a macro.

src/vt/vrt/collection/destroy/destroy_msg.h
```diff
@@ -53,14 +53,18 @@ namespace vt { namespace vrt { namespace collection {
 
 template <typename ColT, typename IndexT>
 struct DestroyMsg final : ::vt::Message {
+  msg_parent_type = ::vt::Message;
+  using CollectionProxyType = CollectionProxy<ColT,IndexT>;
```
pnstickne: Could also use `decltype(proxy_)` if placing macro after member declaration.

src/vt/vrt/collection/messages/system_create.h
```diff
@@ -76,11 +79,15 @@ struct CollectionCreateMsg : ::vt::Message {
 
   template <typename SerializerT>
   void serialize(SerializerT& s) {
+    msg_serialize_parent(s);
     s | info | tup | map;
   }
 };
 
 struct CollectionConsMsg : ::vt::collective::reduce::ReduceMsg {
+  msg_parent_type        = ::vt::collective::reduce::ReduceMsg;
```
pnstickne: Messages that don't define a serialization function can explicitly mark themselves as prohibiting serialization. This mandates that derived types also prohibit serialization.

src/vt/vrt/collection/messages/user_wrap.h
```diff
@@ -80,17 +83,7 @@ struct ColMsgWrap : CollectionMessage<ColT,BaseMsgT> {
...
-    CollectionMessage<ColT,BaseMsgT>::serializeParent(s);
-    CollectionMessage<ColT,BaseMsgT>::serializeThis(s);
-  }
-
-  template <typename SerializerT>
-  void serializeThis(SerializerT& s) {
-    s | msg_;
-  }
-
+  // TODO - does not align with 'serialize if needed' above.
```
pnstickne: Comment outdated.

src/vt/vrt/context/context_vrtmanager.impl.h
```diff
@@ -159,10 +159,12 @@ messaging::PendingSend VirtualContextManager::sendSerialMsg(
     auto promoted_msg = promoteMsg(msg);
     messaging::PendingSend pending(
       promoted_msg, [=](MsgVirtualPtr<BaseMsgType> mymsg){
-        SerializedMessenger::sendSerialMsg<
-          MsgT, virtualTypedMsgHandler<MsgT>, VirtualMessage
-        >(
-          reinterpret_cast<MsgT*>(mymsg.get()),
+        // Uses special implementation overload not exposed in theMsg..
```
pnstickne: Next step might be to flatten out serialization or first step in any message transmission is generation of byte payload + headers.

tests/unit/collection/test_broadcast.cc
```diff
@@ -73,15 +73,21 @@ struct TestCol : Collection<TestCol<Args...>,TestIndex> {
...
+  msg_parent_type = CollectionMessage<TestCol<Args...>>;
   using TupleType = std::tuple<Args...>;
+  msg_serialize_if_needed_by_base_or_type1(TupleType);
+
   ColMsg() = default;
   explicit ColMsg(TupleType in_tup) : tup(in_tup) {}
-  TupleType tup;
+
   template <typename SerializerT>
   void serialize(SerializerT& s) {
```
pnstickne: Add call to serialize parent (oops!). Also add serialization check to ensure message is serialized, to catch such as Runtime Assert Failures.

tests/unit/collection/test_index_types.cc
```diff
@@ -68,13 +68,18 @@ struct TestCol : Collection<TestCol<IndexT>,IndexT> {
...
 struct ColMsg : CollectionMessage<TestCol<IndexT>> {
+  msg_parent_type = CollectionMessage<TestCol<IndexT>>;
+  msg_serialize_if_needed_by_base();
+
   ColMsg() = default;
   explicit ColMsg(int in_data) : data_(in_data) {}
-  int data_;
+
   template <typename SerializerT>
   void serialize(SerializerT& s) {
```
pnstickne: Invoke parent serialize.

tests/unit/collection/test_send.cc
```diff
@@ -76,16 +76,22 @@ struct TestCol : Collection<TestCol<Args...>,TestIndex> {
...
+  msg_parent_type = CollectionMessage<TestCol<Args...>>;
   using TupleType = std::tuple<Args...>;
+  msg_serialize_if_needed_by_base_or_type1(TupleType);
+
   ColMsg() = default;
   explicit ColMsg(TupleType in_tup) : tup(in_tup) {}
-  TupleType tup;
+
   template <typename SerializerT>
   void serialize(SerializerT& s) {
```
pnstickne: Invoke parent serialize.

tests/unit/data_message.h
```diff
@@ -71,13 +81,20 @@ struct TestStaticBytesMsg : MessageT {
...
 template <typename MessageT, NumBytesType num_bytes>
 struct TestStaticSerialBytesMsg : TestStaticBytesMsg<MessageT,num_bytes> {
+  msg_parent_type               = TestStaticBytesMsg<MessageT,num_bytes>;
+  msg_serialize_required();
+
+  // Derived type serializes parent type's members.. (don't do this)
   template <typename SerializerT>
   void serialize(SerializerT& s) {
-    MessageT::serializeThis(s);
+    MessageT::serialize(s);
```
pnstickne: Skip parent's (missing) serialization and call appropriate method..

tests/unit/memory/test_memory_lifetime.cc
```diff
@@ -67,6 +68,12 @@ struct SerialTrackMsg : ::vt::Message {
   ~SerialTrackMsg() { --alloc_count; }
 
   static int32_t alloc_count;
+
+public:
```
pnstickne: Unify location of check.

tests/unit/termination/test_term_dep_send_chain.cc
```diff
@@ -90,15 +90,21 @@ struct MyCol : vt::Collection<MyCol,vt::Index2D> {
...
+
     template <typename SerializerT>
     void serialize(SerializerT& s) {
       s | vec | step;
     }
+
     std::vector<double> vec = {};
     int step = 0;
+
+  public:
```
pnstickne: Update placement.

tutorial/tutorial_1c.h
```diff
@@ -118,7 +127,7 @@ static inline void activeMessageSerialization() {
   (void)num_nodes;  // don't warn about unused variable
 
   /*
-   * The theMsg()->sendMsgAuto(..) will serialize the message sent to the
+   * The theMsg()->sendMsg(..) will serialize the message sent to the
```
pnstickne: Improve comment.

src/vt/messaging/message/message_serialize.h
```diff
@@ -47,6 +47,46 @@
...
+  std::enable_if_t<
+    std::is_same<void (U::*)(::serdes::Serializer&), decltype(&U::template serialize<::serdes::Serializer&>)>::value
+  >>
+  : std::true_type
+{};
+
+// Either declared out-of-line (no inheritence in play)
+// or a serialize member is declared on the precise type.
+template <typename T>
+static constexpr auto const has_own_serialize =
```
PhilMiller: Convention would name this `has_own_serialize_v` I think

src/vt/messaging/message/message_serialize.h
```diff
@@ -0,0 +1,305 @@
...
+};
+
+template <typename T>
+struct byte_copyable {
+  constexpr static bool value = std::is_trivially_copyable<T>::value and not std::is_pointer<T>::value;
+};
+
+// ref. https://en.cppreference.com/w/cpp/types/void_t
+template<typename... Ts>
+struct cxx14_make_void { typedef void type;};
```
lifflander: Look at the `detector` component. It includes this.

src/vt/messaging/message/message_serialize.h
```diff
@@ -0,0 +1,305 @@
...
+ *
+ * The message CANNOT support a serialization function
+ * and the message is always sent without serialization.
+ * Standard copyable transmission rules apply.
+ */
+#define msg_serialize_prohibited()                                        \
+  msg_serialize_mode_(prohibit);                                          \
+  static_assert(                                                          \
+    not ::vt::messaging::msg_serialization_mode<MessageParentType>::required, \
+    "Message cannot prohibt serialization because serialization is required by the parent type" \
```
lifflander: Typo here: "prohibt"

src/vt/messaging/message/message_serialize.h
```diff
@@ -47,6 +47,46 @@
...
+// They are lifted for demonstration purposes.
+#define HAS_DETECTION_COMPONENT 1
+#include "serialization_library_headers.h"
+#include "traits/serializable_traits.h"
+
+namespace vt { namespace messaging {
+
+// has_own_serialize_member<T>::value is true iff the type
+// of T, without consideration of parent types, defines a serialize
+// method. Should work on GCC 4.9+, Clang 3.9+, ICC 19+ (18 might not work..)
```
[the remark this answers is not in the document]
pnstickne: > Lack of ICC 18 support might be problematic on target systems :-(

Hmm, works on ICC 17. It appears it's a linking issue with Compiler Explorer, not a compilation issue.

tests/unit/collection/test_broadcast.cc
pnstickne: Done.

tests/unit/memory/test_memory_lifetime.cc
pnstickne: Done. It was.

tests/unit/termination/test_term_dep_send_chain.cc
pnstickne: Done. Also fixed calling parent serializer.

src/vt/messaging/active.h
```diff
@@ -221,35 +221,33 @@ struct ActiveMessenger {
   virtual ~ActiveMessenger();
 
   /**
-   * \brief Mark a message as a termination message.
```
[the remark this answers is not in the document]
pnstickne: Merge conflict. Should be updated when rebased..

src/vt/messaging/active.impl.h
```diff
@@ -76,7 +74,82 @@ void ActiveMessenger::setTagMessage(MsgT* msg, TagType tag) {
   envelopeSetTag(msg->env, tag);
 }
 
-template <typename MessageT>
+#if HAS_SERIALIZATION_LIBRARY
+
+// Conditionally-enabled methods for message types that support serialization.
+// The current implementation of these methods eventually call BACK into
+// sendMsgImpl with a NON-SERIALIZAD MESSAGE.
```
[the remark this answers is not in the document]
pnstickne: Comment updated.

examples/comm_func.cc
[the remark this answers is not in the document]
pnstickne: It's a fun macro. Used to do a little bit more. The "advantage" of a macro is it will catch trivial typing errors. It isn't a function-like macro due to the notorious issue of dealing with types containing commas.

     // Usage: msg_parent_type = vt::Message.
    #define msg_parent_type using MessageParentType /* leave space */

src/vt/messaging/message/message.h
```diff
@@ -203,11 +205,31 @@ struct ActiveMsg : BaseMsg {
...
+   * \brief Explicitly serialize this message.
+   *
+   * Should be called froms derived type that support/require serialization.
+   *
+   * \param[in] s the serializer
+   */
+  template <typename SerializerT>
+  void serialize(SerializerT& s) {
+    // n.b. not actually used, as extracted during transmission.
+    // In the future can be used to trace serialization that do not serializer
```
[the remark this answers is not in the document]
pnstickne: Updated (eliminated).

src/vt/pipe/interface/callback_direct.h
pnstickne: It is not a real message.

src/vt/pipe/interface/callback_direct_multi.h
pnstickne: It is not..

src/vt/messaging/message/message_serialize.h
pnstickne: Removed.

src/vt/serialization/messaging/serialized_messenger.impl.h
pnstickne: Done!

src/vt/serialization/messaging/serialized_messenger.impl.h
```diff
@@ -186,95 +186,35 @@ template <typename UserMsgT, typename BaseEagerMsgT>
...
-  return parserdesMsgHandler(msg,h,is_bcast,dest);
-}
-
-template <typename MsgT, ActiveTypedFnType<MsgT> *f, typename BaseT>
-/*static*/ messaging::PendingSend SerializedMessenger::parserdesMsg(
-  MsgT* msg, bool is_bcast, NodeType dest
-) {
-  auto const& h = auto_registry::makeAutoHandler<MsgT,f>(nullptr);
-  return parserdesMsgHandler(msg,h,is_bcast,dest);
+  return parserdesMsgSendImpl<MsgT>(msg, han, uninitialized_destination);
```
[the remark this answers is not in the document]
pnstickne: There is in the Active Messaging component (local only) to improve clarity there.

src/vt/messaging/message/message_serialize.h
pnstickne: Fixed.

pnstickne:
@lifflander @PhilMiller Alright - so what about the overall direction? Does this seem like a good pattern/requirement to adopt in VT? How can it be improved?

src/vt/messaging/message/message_serialize.h
```diff
@@ -0,0 +1,265 @@
...
+
+#if !defined INCLUDED_MESSAGING_MESSAGE_MESSAGE_SERIALIZE_H
+#define INCLUDED_MESSAGING_MESSAGE_MESSAGE_SERIALIZE_H
+
+#include <type_traits>
+
+/** \file */
+
+// Usage: msg_parent_type = vt::Message.
+#define msg_parent_type using MessageParentType /* leave space */
```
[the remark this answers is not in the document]
pnstickne: @PhilMiller What about if the usage was `msg_parent_type XYZ;`?

The primary (only?) advantage of a macro is that it hides the implementation of the type name and will fail compilation if typed incorrectly (eg. `msg_base_type XYZ;`).

lifflander:
> @lifflander @PhilMiller Alright - so what about the overall direction? Does this seem like a good pattern/requirement to adopt in VT? How can it be improved?
> 
> (There is some wonkadoo with the CI builds.. a different Checkpoint version?)

Does this depend on changes in checkpoint?

src/vt/messaging/message/message_serialize.h
```diff
@@ -0,0 +1,267 @@
...
+       and vt::messaging::byte_copyable<TYPE1>::value                     \
+       and vt::messaging::byte_copyable<TYPE2>::value                     \
+       and vt::messaging::byte_copyable<TYPE3>::value                     \
+     )                                                                    \
+     ? ::vt::messaging::SerializationMode::require                        \
+     : ::vt::messaging::SerializationMode::support)
+
+namespace vt { namespace messaging {
+
+enum SerializationMode
```
lifflander: Since we are using macros anyway, should be make this a strongly typed enum?

```c++
enum struct SerializationMode : int {
  //...
};
```

src/vt/messaging/message/message_serialize.h
```diff
@@ -206,33 +206,190 @@ static constexpr auto const has_own_serialize =
...
+// Allows testing of is_base_of<DefinesSerializationMode<Msg>,Msg> to determine
+// if one of the CRTP types is used, as the templated type is invariant.
+// All of the CRTP types are required to define the serialize mode which will
+// be picked up via normal inheritance.
+template <typename MsgT>
+struct DefinesSerializationMode {
+};
+
+template <typename MsgT, typename SelfT>
+struct NonSerializedMsg : MsgT, DefinesSerializationMode<SelfT>
```
pnstickne: @lifflander @PhilMiller CRTP implementation as suggested by Jonathan. Pros and cons. Both could use with some recommended pattern and both involve duplication.

src/vt/collective/reduce/operators/default_msg.h
```diff
@@ -59,19 +59,25 @@ template <typename>
 struct ReduceCombine;
 
 template <typename DataType>
-struct ReduceDataMsg : ReduceMsg, ReduceCombine<void> {
-  msg_parent_type    = ReduceMsg;
-  msg_serialize_if_needed_by_base_or_type1(DataType);
+struct ReduceDataMsg : SerializeIfNeeded<
```
pnstickne: And there be usages.. this is one of the gotcha types in VT (it is expected to be NOT serialize in LoadStatsMsg, eg).

src/vt/collective/reduce/operators/default_msg.h
```diff
@@ -95,63 +101,69 @@ struct ReduceDataMsg : ReduceMsg, ReduceCombine<void> {
 };
 
 template <typename T>
-struct ReduceTMsg : ReduceDataMsg<T> {
-  msg_parent_type = ReduceDataMsg<T>;
-  msg_serialize_if_needed_by_base();
+struct ReduceTMsg : SerializeIfNeeded<
```
pnstickne: `SerializeIfNeeded` also (and implicitly) follows the base serialization mode.

src/vt/collective/reduce/operators/default_msg.h
```diff
@@ -95,63 +101,69 @@ struct ReduceDataMsg : ReduceMsg, ReduceCombine<void> {
...
 
 template <typename T, std::size_t N>
-struct ReduceArrMsg : ReduceDataMsg<std::array<T,N>> {
-  msg_parent_type   = ReduceDataMsg<std::array<T,N>>;
-  msg_serialize_if_needed_by_base();
+struct ReduceArrMsg : SerializeIfNeeded<
+  ReduceDataMsg<std::array<T, N>>,
+  ReduceArrMsg<T, N>
+> {
+  using Parent = ReduceDataMsg<std::array<T, N>>;
```
pnstickne: One method, just pull out Self/Parent types.

src/vt/collective/reduce/reduce_msg.h
```diff
@@ -72,9 +72,12 @@ struct ReduceLink {
...
 
-struct ReduceMsg : ::vt::Message, ReduceLink {
-  msg_parent_type = ::vt::Message;
-  msg_serialize_supported();
+struct ReduceMsg : SerializeSupported<
+  ::vt::Message,
+  ReduceMsg
+>, ReduceLink
+{
+  using Base = SerializeSupported<::vt::Message, ReduceMsg>;
```
pnstickne: Another is to just define the Base once.

src/vt/vrt/collection/balance/stats_msg.h
```diff
@@ -175,17 +176,20 @@ struct LoadStatsMsg : CollectionMessage<ColT>, LoadData {
   PhaseType cur_phase_ = fst_lb_phase;
 };
 
-struct ProcStatsMsg : collective::ReduceTMsg<LoadData> {
-  msg_parent_type   = collective::ReduceTMsg<LoadData>;
-  msg_serialize_prohibited();
+struct ProcStatsMsg : NonSerialized<
```
pnstickne: Requires a non-serialized `ReduceTMsg<..>`

src/vt/collective/reduce/operators/default_msg.h
```diff
@@ -59,19 +59,25 @@ template <typename>
 struct ReduceCombine;
 
 template <typename DataType>
-struct ReduceDataMsg : ReduceMsg, ReduceCombine<void> {
-  msg_parent_type    = ReduceMsg;
-  msg_serialize_if_needed_by_base_or_type1(DataType);
+struct ReduceDataMsg : SerializeIfNeeded<
+  ReduceMsg,
```
pnstickne: The base message is always first type parameter as it's "more important".

src/vt/collective/reduce/operators/default_msg.h
```diff
@@ -95,63 +101,69 @@ struct ReduceDataMsg : ReduceMsg, ReduceCombine<void> {
...
+    SerializeIfNeeded<Parent, Self>::serialize(s);
   }
 };
 
 template <typename T>
-struct ReduceVecMsg : ReduceDataMsg<std::vector<T>> {
-  msg_parent_type   = ReduceDataMsg<std::vector<T>>;
-  msg_serialize_required(); // stl collection used in base, always required.
-
+struct ReduceVecMsg : SerializeRequired<
```
pnstickne: While this would also work with `SerializeIfNeeded` (due to std::vector being supplied to the base), using `SerializeRequired` very clearly indicates the coding expectations.

src/vt/vrt/collection/balance/stats_msg.h
```diff
@@ -160,10 +160,11 @@ static_assert(
...
 
 template <typename ColT>
-struct LoadStatsMsg : CollectionMessage<ColT>, LoadData {
-  msg_parent_type   = CollectionMessage<ColT>;
-  msg_serialize_prohibited();
-
+struct LoadStatsMsg : NonSerialized<
+  CollectionMessage<ColT>,
+  LoadStatsMsg<ColT>
+>, LoadData
```
pnstickne: Not a big fan of this particular formatting layout.

src/vt/vrt/collection/balance/gossiplb/gossip_msg.h
```diff
@@ -53,7 +53,7 @@
 namespace vt { namespace vrt { namespace collection { namespace balance {
 
 struct GossipMsg : vt::Message {
-  msg_parent_type = vt::Message;
+  using MessageParentType = vt::Message;
```
pnstickne: @lifflander @PhilMiller Another options is to eliminate some of the macros with a convention. Still requires a macro call to specify the type.

src/vt/vrt/collection/balance/gossiplb/gossip_msg.h
```diff
@@ -75,7 +75,7 @@ struct GossipMsg : vt::Message {
 
   template <typename SerializerT>
   void serialize(SerializerT& s) {
-    msg_serialize_parent(s);
```
pnstickne: The macro call is not needed now. A much earlier form "did some magic" in the macro.

src/vt/messaging/message/message_serialize.h
```diff
@@ -206,33 +206,190 @@ static constexpr auto const has_own_serialize =
...
 };
 
+// Trivial test to ensure that the type is ELIGIBLE to be byte-copied, regardless of
+// if such supports explicit serialization. The assertion is that if serialization is
+// not required then it should be skipped.
+// This does not honor suppression (eg. no support of handling isByteCopyable == false).
+// It also implicitly prohibits types with virtual members, which is arguably
+// an artificial limitation of is_trivially_copyable.
 template <typename T>
 struct byte_copyable {
```
pnstickne: Might also allow isByteCopyable to SUPPRESS if a byte-copy can be used:

     constexpr static bool value = std::is_trivially_copyable<T>::value
         and not std::is_pointer<T>::value
         and not suppress_byte_copy_when_isByteCopyable_is_false<T>::value;

pnstickne: That said, there is currently no `isByteCopyable = false` in the current VT and envelops must meet the definition of is_trivially_copyable anyway.

pnstickne:
@lifflander https://github.com/DARMA-tasking/checkpoint/search?q=contiguousTyped&unscoped_q=contiguousTyped - only "contiguous" currently in Checkpoint.

pnstickne:
Well, compiles here on Clang.. should try GCC. It appears as though GCC might be unnecessarily attempting to instantiate a template on `constexpr const bool vt::messaging::has_own_serialize<vt::collective::barrier::BarrierMsg>`, eg.

pnstickne:
Regardless, appears like it might be an edge-case in Checkpoint that might not have been previously caught when the type is not ACTUALLY serialized:

```
/home/travis/build/DARMA-tasking/vt/checkpoint/install/include/serializers/base_serializer.h: In instantiation of ‘static void serdes::Serializer::contiguousTyped(SerializerT&, T*, serdes::SerialSizeType) [with SerializerT = serdes::Serializer; T = bool; serdes::SerialSizeType = long unsigned int]’:
1185/home/travis/build/DARMA-tasking/vt/checkpoint/install/include/dispatch/dispatch_serializer_byte.h:100:33:   required from ‘void serdes::SerializerDispatchByte<SerializerT, T>::operator()(SerializerT&, T*, serdes::SerialSizeType, serdes::SerializerDispatchByte<SerializerT, T>::isByteCopyType<U>*) [with U = bool; SerializerT = serdes::Serializer; T = bool; serdes::SerialSizeType = long unsigned int; serdes::SerializerDispatchByte<SerializerT, T>::isByteCopyType<U> = bool]’
```

There is indeed no serdes::Serializer::contiguousBytes method defined (see base_serializer.h).

src/vt/messaging/message/message_serialize.h
pnstickne: Hrrmp, Checkpoint has a specific definition of what isByteCopyable means :|

pnstickne:
@PhilMiller @lifflander @nlslatt I've updated the primary header comment and clear most PR comments that don't show the end effect.

src/vt/messaging/message/message_serialize.h
pnstickne: Switched.

lifflander:
> @PhilMiller @lifflander @nlslatt I've updated the primary header comment and clear most PR comments that don't show the end effect.
> 
> There is some cleanup to do (which is mostly just a menial task) based on feedback and then general prep and minor cleanup that shouldn't have any relevant impact on overall design goals on the usages VT is hoping to encourage/require of client implementations.

Here are the responses from Nicole and Phil collected from chat:

Nicole: 
> I don't think I have a strong opinion on which serialization approach.  Probably CRTP.

Phil: 
> Formally, I think the CRTP serialization approach is probably stronger. Stylistically, it feels much heavier than the inline macro approach, though

lifflander:
@pnstickne Based on the feedback, and considering that you have both implementations done, can we provide both with the caveat that we put the macros in a separate header file that is explicitly included as needed by a user?

I can see users liking one over the other for various reasons. Since the implementations work together, I don't see any downside to providing both as potential options.

What do you think?

Let me know when you're ready for a full review by me and I will look over the code carefully.

src/vt/collective/reduce/operators/default_msg.h
lifflander: Agreed

lifflander:
Need to test with ICC before merging into develop.

lifflander:
Develop and 1.0.0 currently fail with Intel 19 compilers. We do work with Intel 18, which is currently being used. So I'm testing this PR with 18.

PhilMiller:
What fails on icc 19? We should probably fix that for 1.0 just so as not to
impede progress for users. I think we want to encourage compiler upgrades
as rapidly as we can. Eventually, that gets us c++17 features.

lifflander:
We have a lot of errors. A least some of them seem to come from FMT.

lifflander:
Here's the first failure for this PR for Intel 18. I'm not sure why this is happening, given that I believe we has trivially copyable in VT before.

```
[  0%] Building CXX object src/CMakeFiles/vt.dir/vt/collective/basic.cc.o
cd /lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt-build/src && /opt/cray/pe/craype/2.5.18/bin/CC  -DFMT_HEADER_ONLY=1 -DFMT_USE_USER_DEFINED_LITERALS=0 -DHAS_DETECTION_COMPONENT=1 -I/lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/lib/fmt -I/lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/lib/CLI -I/lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt-build/debug -I/lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src -isystem /lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/checkpoint/checkpoint-install/include -isystem /lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/detector/detector-install/include  -Wall -pedantic -Wshadow -Wno-unknown-pragmas -g   -fPIC -qopenmp -std=gnu++14 -o CMakeFiles/vt.dir/vt/collective/basic.cc.o -c /lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/collective/basic.cc
In file included from /lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/messaging/message/message.h(52),
                 from /lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/messaging/message.h(48),
                 from /lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/activefn/activefn.h(49),
                 from /lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/sequence/seq_common.h(53),
                 from /lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/sequence/sequencer_fwd.h(49),
                 from /lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/runtime/runtime_component_fwd.h(50),
                 from /lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/runtime/runtime.h(50),
                 from /lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/runtime/runtime_inst.impl.h(50),
                 from /lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/runtime/runtime_inst.h(76),
                 from /lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/runtime/runtime_headers.h(49),
                 from /lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/collective/startup.h(49),
                 from /lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/collective/basic.cc(47):
/lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/messaging/message/message_serialize.h(227): error: namespace "std" has no member "is_trivially_copyable"
    constexpr static bool value = std::is_trivially_copyable<T>::value and not std::is_pointer<T>::value;
                                       ^

In file included from /lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/messaging/message/message.h(52),
                 from /lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/messaging/message.h(48),
                 from /lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/activefn/activefn.h(49),
                 from /lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/sequence/seq_common.h(53),
                 from /lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/sequence/sequencer_fwd.h(49),
                 from /lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/runtime/runtime_component_fwd.h(50),
                 from /lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/runtime/runtime.h(50),
                 from /lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/runtime/runtime_inst.impl.h(50),
                 from /lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/runtime/runtime_inst.h(76),
                 from /lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/runtime/runtime_headers.h(49),
                 from /lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/collective/startup.h(49),
                 from /lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/collective/basic.cc(47):
/lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/messaging/message/message_serialize.h(227): error: type name is not allowed
    constexpr static bool value = std::is_trivially_copyable<T>::value and not std::is_pointer<T>::value;
                                                             ^

In file included from /lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/messaging/message/message.h(52),
                 from /lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/messaging/message.h(48),
                 from /lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/activefn/activefn.h(49),
                 from /lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/sequence/seq_common.h(53),
                 from /lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/sequence/sequencer_fwd.h(49),
                 from /lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/runtime/runtime_component_fwd.h(50),
                 from /lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/runtime/runtime.h(50),
                 from /lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/runtime/runtime_inst.impl.h(50),
                 from /lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/runtime/runtime_inst.h(76),
                 from /lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/runtime/runtime_headers.h(49),
                 from /lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/collective/startup.h(49),
                 from /lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/collective/basic.cc(47):
/lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/messaging/message/message_serialize.h(227): error: the global scope has no "value"
    constexpr static bool value = std::is_trivially_copyable<T>::value and not std::is_pointer<T>::value;
                                                                 ^
```

lifflander:
Created a new issue for the Intel 19 problems: https://github.com/DARMA-tasking/vt/issues/728

PhilMiller:
Are those missing traits an issue of what version of libstdc++ is getting
used? I think you need to do something to get the matching gcc also loaded
to fix that.

lifflander:
Maybe, it's using a version of gcc for the stdlib from 4.9.3

lifflander:
I made progress on Intel 19, but now am stuck at a internal compiler error.

lifflander:
Intel 18 is making progress after the swap: `module swap gcc/4.9.3 gcc/5.3.0`

lifflander:
First failure with newer gcc stdlib:

```
/lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/messaging/active.impl.h(163): error: static assertion failed with "Message sent without serialization must be trivially copyable."
    static_assert(
    ^
          detected during:
            instantiation of "vt::messaging::ActiveMessenger::PendingSendType vt::messaging::ActiveMessenger::sendMsgCopyableImpl(vt::NodeType={vt::PhysicalResourceType={int16_t={short}}}, vt::HandlerType={int64_t={long}}, vt::messaging::MsgSharedPtr<MessageT> &, vt::ByteType={uint64_t={unsigned long}}, vt::TagType={int32_t={int}}) [with MessageT=StartWorkMsg]" at line 351 of "/lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/messaging/active.h"
            instantiation of "vt::messaging::ActiveMessenger::PendingSendType vt::messaging::ActiveMessenger::sendMsgImpl(vt::NodeType={vt::PhysicalResourceType={int16_t={short}}}, vt::HandlerType={int64_t={long}}, vt::messaging::MsgSharedPtr<MessageT> &, vt::ByteType={uint64_t={unsigned long}}, vt::TagType={int32_t={int}}) [with MessageT=StartWorkMsg, <unnamed>=0]" at line 269
            instantiation of "vt::messaging::ActiveMessenger::PendingSendType vt::messaging::ActiveMessenger::broadcastMsg<MessageT,f>(MessageT *, vt::TagType={int32_t={int}}) [with MessageT=StartWorkMsg, f=startJacobi1dHandler]" at line 169 of "/lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/examples/jacobi1d_node.cc"

compilation aborted for /lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/examples/jacobi1d_node.cc (code 2)
examples/CMakeFiles/jacobi1d_node.dir/build.make:62: recipe for target 'examples/CMakeFiles/jacobi1d_node.dir/jacobi1d_node.cc.o' failed
make[2]: *** [examples/CMakeFiles/jacobi1d_node.dir/jacobi1d_node.cc.o] Error 2
CMakeFiles/Makefile2:1485: recipe for target 'examples/CMakeFiles/jacobi1d_node.dir/all' failed
make[1]: *** [examples/CMakeFiles/jacobi1d_node.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```

lifflander:
A lot more errors, here's another one:

```
/lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/vrt/collection/manager.impl.h(650): error: no instance of overloaded function "vt::messaging::ActiveMessenger::broadcastMsg" matches the argument list
            argument types are: (vt::vrt::collection::CollectionConsMsg *)
            object type is: vt::messaging::ActiveMessenger
      theMsg()->broadcastMsg<CollectionConsMsg,collectionFinishedHan>(
                ^
/lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/messaging/active.impl.h(418): note: this candidate was rejected because there is a type mismatch after argument substitution
  ActiveMessenger::PendingSendType ActiveMessenger::broadcastMsg(
                                                    ^
/lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/messaging/active.impl.h(429): note: this candidate was rejected because there is a type mismatch after argument substitution
  ActiveMessenger::PendingSendType ActiveMessenger::broadcastMsg(
                                                    ^
/lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/messaging/active.impl.h(342): note: this candidate was rejected because there is a type mismatch after argument substitution
  ActiveMessenger::PendingSendType ActiveMessenger::broadcastMsg(
                                                    ^
/lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/messaging/active.impl.h(319): note: this candidate was rejected because there is a type mismatch after argument substitution
  ActiveMessenger::PendingSendType ActiveMessenger::broadcastMsg(
                                                    ^
/lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/messaging/active.impl.h(261): note: this candidate was rejected because there is a type mismatch after argument substitution
  ActiveMessenger::PendingSendType ActiveMessenger::broadcastMsg(
                                                    ^
          detected during:
            instantiation of "void vt::vrt::collection::CollectionManager::distConstruct(SysMsgT *) [with SysMsgT=vt::vrt::collection::CollectionCreateMsg<vt::vrt::collection::CollectionInfo<vt::tutorial::MyCol, vt::Index1D>, std::tuple<>, vt::tutorial::MyCol, vt::Index1D>]" at line 2163
            instantiation of "vt::vrt::collection::CollectionManager::CollectionProxyWrapType<ColT, ColT::IndexType> vt::vrt::collection::CollectionManager::constructMap<ColT,Args...>(ColT::IndexType, const vt::HandlerType={int64_t={long}} &, Args &&...) [with ColT=vt::tutorial::MyCol, Args=<>]" at line 2110
            instantiation of "vt::vrt::collection::CollectionManager::CollectionProxyWrapType<ColT, ColT::IndexType> vt::vrt::collection::CollectionManager::construct<ColT,Args...>(ColT::IndexType, Args &&...) [with ColT=vt::tutorial::MyCol, Args=<>]" at line 91 of "/lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/examples/../tutorial/tutorial_2a.h"
```

lifflander:
Looks like the new overloads are failing

lifflander:
For reference, here are my loaded modules:

```
jliffla@mutrino:/lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt-build> module list
Currently Loaded Modulefiles:
  1) modules/3.2.11.1                                  9) ugni/6.0.14.0-6.0.7.1_3.15__gea11d3d.ari         17) rca/2.2.18-6.0.7.1_5.52__g2aa4f39.ari            25) boost/1.66.0/intel/17.0.4
  2) eproxy/2.0.22-6.0.7.1_7.5__g1ebe45c.ari          10) pmi/5.0.14                                       18) atp/2.1.3                                        26) hdf5/1.10.1/cray/7.6.2/intel/17.0.4
  3) cray-mpich/7.7.6                                 11) dmapp/7.1.1-6.0.7.1_6.6__g45d1b37.ari            19) perftools-base/7.0.6                             27) pnetcdf-exo/1.8.1/cray/7.6.2/intel/17.0.4
  4) intel/18.0.5                                     12) gni-headers/5.0.12.0-6.0.7.1_3.13__g3b1768f.ari  20) PrgEnv-intel/6.0.5                               28) netcdf-exo/4.4.1.1/cray/7.6.2/intel/17.0.4
  5) craype-network-aries                             13) xpmem/2.2.15-6.0.7.1_5.13__g7549d06.ari          21) totalview/2019.0.4                               29) yamlcpp/0.5.3/intel/17.0.4
  6) craype/2.5.18                                    14) job/2.2.3-6.0.7.1_5.48__g6c4e934.ari             22) craype-haswell                                   30) devpack/20180124/cray/7.6.2/intel/17.0.4
  7) cray-libsci/19.02.1                              15) dvs/2.7_2.2.120-6.0.7.1_12.5__g74cb2cc4          23) craype-hugepages2M                               31) gcc/5.3.0
  8) udreg/2.3.2-6.0.7.1_5.15__g5196236.ari           16) alps/6.6.43-6.0.7.1_5.51__ga796da32.ari          24) slurm/19.05.1                                    32) cmake/3.14.6
```

pnstickne:
Previously VT did not assert trivially copyable (at least in those places). It asserted only trivially destructible. I added a number of trivially copyable checks here, so that might have over-strengthened beyond some .. compilers.

lifflander:
I found one of the bugs in ICC. ICC does not consider this struct to be trivially copyable, due to the const member.

```C++
struct StartWorkMsg : ShortMessage {
  bool const converged;

  StartWorkMsg(bool const& in_conv) : converged(in_conv) { }
};
```

This is a bug in the compiler: https://stackoverflow.com/questions/19526526/is-const-int2-trivially-copyable

lifflander:
Another bug that I have yet to pin down exactly is lb_iter. It fails to compile while simple_collection does compile. After minimizing it, it seems that without a virtual destructor lb_iter fails to compile, whereas simple_collection works because it has one.

Not sure what is going on with this. None of the serialization code seems incorrectly conditionalized on the collection type during construction.

```
Scanning dependencies of target lb_iter
[100%] Building CXX object examples/CMakeFiles/lb_iter.dir/lb_iter.cc.o
/lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/vrt/collection/manager.impl.h(650): error: no instance of overloaded function "vt::messaging::ActiveMessenger::broadcastMsg" matches the argument list
            argument types are: (vt::vrt::collection::CollectionConsMsg *)
            object type is: vt::messaging::ActiveMessenger
      theMsg()->broadcastMsg<CollectionConsMsg,collectionFinishedHan>(
                ^
/lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/messaging/active.impl.h(418): note: this candidate was rejected because there is a type mismatch after argument substitution
  ActiveMessenger::PendingSendType ActiveMessenger::broadcastMsg(
                                                    ^
/lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/messaging/active.impl.h(429): note: this candidate was rejected because there is a type mismatch after argument substitution
  ActiveMessenger::PendingSendType ActiveMessenger::broadcastMsg(
                                                    ^
/lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/messaging/active.impl.h(342): note: this candidate was rejected because there is a type mismatch after argument substitution
  ActiveMessenger::PendingSendType ActiveMessenger::broadcastMsg(
                                                    ^
/lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/messaging/active.impl.h(319): note: this candidate was rejected because there is a type mismatch after argument substitution
  ActiveMessenger::PendingSendType ActiveMessenger::broadcastMsg(
                                                    ^
/lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/messaging/active.impl.h(261): note: this candidate was rejected because there is a type mismatch after argument substitution
  ActiveMessenger::PendingSendType ActiveMessenger::broadcastMsg(
                                                    ^
          detected during:
            instantiation of "void vt::vrt::collection::CollectionManager::distConstruct(SysMsgT *) [with SysMsgT=vt::vrt::collection::CollectionCreateMsg<vt::vrt::collection::CollectionInfo<IterCol, vt::Index1D>, std::tuple<>, IterCol, vt::Index1D>]" at line 2163
            instantiation of "vt::vrt::collection::CollectionManager::CollectionProxyWrapType<ColT, ColT::IndexType> vt::vrt::collection::CollectionManager::constructMap<ColT,Args...>(ColT::IndexType, const vt::HandlerType={int64_t={long}} &, Args &&...) [with ColT=IterCol, Args=<>]" at line 2110
            instantiation of "vt::vrt::collection::CollectionManager::CollectionProxyWrapType<ColT, ColT::IndexType> vt::vrt::collection::CollectionManager::construct<ColT,Args...>(ColT::IndexType, Args &&...) [with ColT=IterCol, Args=<>]" at line 73 of "/lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/examples/lb_iter.cc"

compilation aborted for /lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/examples/lb_iter.cc (code 2)
```

PhilMiller:
It turns out that MSVC also draws the same conclusion about lacking trivial copyability of structs with const members, per Godbolt.

PhilMiller:
```c++
#include <type_traits>

struct Int {
    int i = 1;
};

struct ConstInt{
    const int i = 2;
};

int main(int argc, char* argv[]) {
    static_assert(std::is_trivially_copyable<Int>::value, "Int should be triv");
    static_assert(std::is_trivially_copyable<ConstInt>::value, "ConstInt should be triv");

    return 0;
}
```

lifflander:
Maybe an EDG problem?

PhilMiller:
https://www.reddit.com/r/cpp/comments/bdt8ep/does_msvc_still_use_edg_as_c_frontend/el0l0zu

Intellisense uses EDG, but the MS compiler itself does not.

Recall they had to rewrite their template instantiation stuff a few years ago for full C++11 conformance, to move away from the much older token stream thing they used to do.

lifflander:
> https://www.reddit.com/r/cpp/comments/bdt8ep/does_msvc_still_use_edg_as_c_frontend/el0l0zu
> 
> Intellisense uses EDG, but the MS compiler itself does not.
> 
> Recall they had to rewrite their template instantiation stuff a few years ago for full C++11 conformance, to move away from the much older token stream thing they used to do.

Interesting... So should we detect ICC and relax the trivially copyable constraint to work around the bugs for now? Note that Intel 19 has the same problem (plus others). We could also change that example, but this will affect user code also.

PhilMiller:
If it's only an assertion, and not a change in overload selection or other behavior, yeah, I would support working around the issue for ICC by loosening the checks.

This also seems a clear-cut enough case that the test I posted could be sent through the proper channels as a bug report with pressure to fix.

lifflander:
@PhilMiller Sounds good. @pnstickne Can you implement this? Figure out what ICC defines and relax trivially copyable check. It's dispatching correctly, so it should execute properly.

I'm still stuck on lb_iter. This is the reproducer:

```C++
#include "vt/transport.h"

struct IterCol : vt::Collection<IterCol,vt::Index1D> { };

  vt::theCollection()->construct<IterCol>(vt::Index1D(1));
  return 0;
}
```

```
/lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/vrt/collection/manager.impl.h(650): error: no instance of overloaded function "vt::messaging::ActiveMessenger::broadcastMsg" matches the argument list
            argument types are: (vt::vrt::collection::CollectionConsMsg *)
            object type is: vt::messaging::ActiveMessenger
      theMsg()->broadcastMsg<CollectionConsMsg,collectionFinishedHan>(
                ^
/lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/messaging/active.impl.h(418): note: this candidate was rejected because there is a type mismatch after argument substitution
  ActiveMessenger::PendingSendType ActiveMessenger::broadcastMsg(
                                                    ^
/lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/messaging/active.impl.h(429): note: this candidate was rejected because there is a type mismatch after argument substitution
  ActiveMessenger::PendingSendType ActiveMessenger::broadcastMsg(
                                                    ^
/lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/messaging/active.impl.h(342): note: this candidate was rejected because there is a type mismatch after argument substitution
  ActiveMessenger::PendingSendType ActiveMessenger::broadcastMsg(
                                                    ^
/lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/messaging/active.impl.h(319): note: this candidate was rejected because there is a type mismatch after argument substitution
  ActiveMessenger::PendingSendType ActiveMessenger::broadcastMsg(
                                                    ^
/lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/src/vt/messaging/active.impl.h(261): note: this candidate was rejected because there is a type mismatch after argument substitution
  ActiveMessenger::PendingSendType ActiveMessenger::broadcastMsg(
                                                    ^
          detected during:
            instantiation of "void vt::vrt::collection::CollectionManager::distConstruct(SysMsgT *) [with SysMsgT=vt::vrt::collection::CollectionCreateMsg<vt::vrt::collection::CollectionInfo<IterCol, vt::Index1D>, std::tuple<>, IterCol, vt::Index1D>]" at line 2163
            instantiation of "vt::vrt::collection::CollectionManager::CollectionProxyWrapType<ColT, ColT::IndexType> vt::vrt::collection::CollectionManager::constructMap<ColT,Args...>(ColT::IndexType, const vt::HandlerType={int64_t={long}} &, Args &&...) [with ColT=IterCol, Args=<>]" at line 2110
            instantiation of "vt::vrt::collection::CollectionManager::CollectionProxyWrapType<ColT, ColT::IndexType> vt::vrt::collection::CollectionManager::construct<ColT,Args...>(ColT::IndexType, Args &&...) [with ColT=IterCol, Args=<>]" at line 50 of "/lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-test-message-2/vt/vt/examples/lb_iter.cc"
```

pnstickne:
Okay, I'll add cmake checks + usages for the const (and will also add one for the eager-template instantiation case in GCC).

cmake/check_compiler.cmake
```diff
@@ -0,0 +1,64 @@
...
+    if(NOT check_cc_result AND VT_DEBUG_CMAKE)
+      message(
+        STATUS
+        ${check_cc_quirk_output}
+      )
+    endif()
+
+    # quirk'ed is negative check
+    if(check_cc_result)
+      set(${VAR} FALSE CACHE INTERNAL "TRY_COMPILE RESULT FOR ${CC_FILE}")
```
pnstickne: Docstring could be improved.

cmake/check_compiler.cmake
```diff
@@ -0,0 +1,64 @@
+#
+# Quirk-detection between different compilers.
+#
+
+function(check_cc_quirk VAR CC_FILE)
+  set(check_cc_quirk_output "")
```
pnstickne: Move to same location as assignment of check_cc_result.

cmake/check_compiler.cmake
```diff
@@ -0,0 +1,64 @@
...
+      OUTPUT_VARIABLE check_cc_quirk_output
+      CXX_STANDARD 11
+      CXX_STANDARD_REQUIRED ON
+      CXX_EXTENSIONS OFF
+    )
+
+    if(NOT check_cc_result AND VT_DEBUG_CMAKE)
+      message(
+        STATUS
+        ${check_cc_quirk_output}
```
pnstickne: Add prefix.

cmake/check_compiler.cmake
```diff
@@ -0,0 +1,64 @@
...
+      )
+    endif()
+
+    # quirk'ed is negative check
```
pnstickne: Negative-case properties ensures that, in absence due to mismatched define name (if it were to occur), the default behavior would be to not-quirk with `#ifndef vt_quirked_xyz`.

src/vt/messaging/message/message_serialize.h
```diff
@@ -216,16 +209,22 @@ enum struct SerializationMode : int
...
-// It also implicitly prohibits types with virtual members, which is arguably
-// an artificial limitation of is_trivially_copyable.
+// Trivial test to ensure that the type is eligible to be byte-copied,
+// regardless of if the message supports explicit serialization.
+// N.B. This will return TRUE on certain compilers even though the guarantee
+// cannot be validated due to implementation quirks of trivially copyable.
+#ifndef vt_quirked_trivially_copyable_on_msg
 template <typename T>
-struct byte_copyable {
+struct byte_copyable_t {
```
pnstickne: Rename `is..`

pnstickne: Done.

CMakeLists.txt
```diff
@@ -50,11 +50,22 @@ set(PROJECT_BASE_DIR    ${CMAKE_CURRENT_SOURCE_DIR})
 set(PROJECT_LIB_DIR     ${CMAKE_CURRENT_SOURCE_DIR}/lib)
 set(PROJECT_EXAMPLE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/examples)
 
+# Request C++14, allow fallback (to C++11).. default is to fallback.
```
pnstickne: Not previously set in VT..?

"[the cmake default] is to [allow] fallback"

pnstickne:
@lifflander Should address [most] issues in ICC, as well as using detection for the better static checks (in compatible compilers).

lifflander:
> @lifflander Should address [most] issues in ICC, as well as using detection for the better static checks (in compatible compilers).

Ok, I will try to rebuild with ICC. Were you able to try compiling yourself on Mutrino/ICC platform?

pnstickne:
@lifflander @PhilMiller Compiles on ICC18.

That last commit is.. special. It addresses a slew of ICC18 errors while calling a method that should have the same template arguments (and if the right template variation was called), has an identical body.

src/vt/vrt/collection/manager.impl.h
```diff
@@ -647,7 +647,12 @@ template <typename>
...
     auto new_msg = makeMessage<CollectionConsMsg>(*msg);
     theMsg()->markAsCollectionMessage(new_msg);
-    theMsg()->broadcastMsg<CollectionConsMsg,collectionFinishedHan>(
+    //TODO/QUIRK: on ICC18, using broadcastMsg here results in a huge range
+    // of seemingly unrelated errors relating to template value substitution failure.
+    // Using broadcastMsgAuto (which has an overload of the same signature and
+    // body) somehow avoids triggering all of this.
+    // Is there a very-incompatible overload in broadcastMsg?
+    theMsg()->broadcastMsgAuto<CollectionConsMsg,collectionFinishedHan>(
       new_msg.get()
```
pnstickne: In both cases, it's expected to call a different-named method with the SAME template signature and an IDENTICAL body.

pnstickne: Removing the `template <ActiveFnType* f, typename MessageT>` broadcastMsg template form does not address the problem. There is only one additional overload between the two (on a non-applicable template).

Issue #659, opened 2020-02-03
Warnings from Clang 7


PhilMiller:
When building with Clang 7, the following warnings appear:

```
-- Generating done
-- Build files have been written to: /home/mbetten/Trilinos/EMPIRE/all
[305/1119] Building CXX object TPL/vt/examples/CMakeFiles/collection_insert.dir/collection_insert.cc.o
../TPL/vt/examples/collection_insert.cc:108:32: warning: lambda capture 'proxy' is not used [-Wunused-lambda-capture]
      proxy.finishedInserting([proxy]{
                               ^~~~~
1 warning generated.
[319/1119] Building CXX object TPL/vt/examples/CMakeFiles/simple_collection_reduce.dir/simple_collection_reduce.cc.o
../TPL/vt/examples/simple_collection_reduce.cc:84:11: warning: private field 'counter_' is not used [-Wunused-private-field]
  int32_t counter_ = 0;
          ^
1 warning generated.
[321/1119] Building CXX object TPL/vt/examples/CMakeFiles/reduce.dir/reduce.cc.o
../TPL/vt/examples/reduce.cc:99:10: warning: private field 'quadValue' is not used [-Wunused-private-field]
  double quadValue = 0.0;
         ^
1 warning generated.
[344/1119] Building CXX object TPL/vt/examples/CMakeFiles/jacobi2d_vt.dir/jacobi2d_vt.cc.o
../TPL/vt/examples/jacobi2d_vt.cc:97:10: warning: private field 'normRes_' is not used [-Wunused-private-field]
  double normRes_ = 0.0;
         ^
1 warning generated.
[1119/1119] Linking CXX executable EMPIRE-Hybrid/tests/unit/fluid_particle_task/fluid_particle_task
(DESKTOP)-dir-all [CLANG-7.0.1]$
```

PhilMiller:
Closed by merge of #660 and #661

PR #661, opened 2020-02-03
#659: Eliminate unused (member/capture) variables that Clang 7 warns about

PhilMiller:


PhilMiller:
Note I've got separate pull requests up for beta.5 and develop

Issue #668, opened 2020-02-04
gcc-9.2.0 signed vs unsigned warnings test_term_cleanup


nmm0:
**What Needs to be Done?**

Warnings with GCC 9.2

**Is your feature request related to a problem? Please describe.**

Signed vs unsigned comparison warnings in some tests

**Describe potential solution outcome**
No warnings (except for macro warnings)

**Describe alternatives you've considered**

**Additional context**

PR #669, opened 2020-02-04
#668: signed vs unsigned warnings with gcc 9.2.0 in test_term_cleanup

nmm0:
This can also be applied to 1.0.0-beta-5

### DIFF
--- a/tests/unit/termination/test_term_cleanup.cc
+++ b/tests/unit/termination/test_term_cleanup.cc
@@ -89,20 +89,20 @@ TEST_F(TestTermCleanup, test_termination_cleanup_1) {
     theTerm()->addAction(epoch, [&]{ done = true; });
     do vt::runScheduler(); while (not done);
 
-    EXPECT_LT(theTerm()->getEpochState().size(), 2);
-    EXPECT_EQ(theTerm()->getEpochWaitSet().size(), 0);
-    EXPECT_EQ(theTerm()->getDSTermMap().size(), 0);
-    EXPECT_LT(theTerm()->getEpochReadySet().size(), 2);
+    EXPECT_LT(theTerm()->getEpochState().size(), std::size_t{2});
+    EXPECT_EQ(theTerm()->getEpochWaitSet().size(), std::size_t{0});
+    EXPECT_EQ(theTerm()->getDSTermMap().size(), std::size_t{0});
+    EXPECT_LT(theTerm()->getEpochReadySet().size(), std::size_t{2});
   }
 
   while (not vt::rt->isTerminated() or not vt::theSched()->isIdle()) {
     vt::runScheduler();
   }
 
-  EXPECT_LT(theTerm()->getEpochState().size(), 2);
-  EXPECT_EQ(theTerm()->getEpochWaitSet().size(), 0);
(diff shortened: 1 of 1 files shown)


tests/unit/termination/test_term_cleanup.cc
```diff
@@ -89,20 +89,20 @@ TEST_F(TestTermCleanup, test_termination_cleanup_1) {
     theTerm()->addAction(epoch, [&]{ done = true; });
     do vt::runScheduler(); while (not done);
 
-    EXPECT_LT(theTerm()->getEpochState().size(), 2);
-    EXPECT_EQ(theTerm()->getEpochWaitSet().size(), 0);
-    EXPECT_EQ(theTerm()->getDSTermMap().size(), 0);
-    EXPECT_LT(theTerm()->getEpochReadySet().size(), 2);
+    EXPECT_LT(theTerm()->getEpochState().size(), std::size_t{2});
```
pnstickne: If including cstddef, can just use `size_t` - https://stackoverflow.com/a/5813736/2864740

tests/unit/termination/test_term_cleanup.cc
```diff
@@ -89,20 +89,20 @@ TEST_F(TestTermCleanup, test_termination_cleanup_1) {
     theTerm()->addAction(epoch, [&]{ done = true; });
     do vt::runScheduler(); while (not done);
 
-    EXPECT_LT(theTerm()->getEpochState().size(), 2);
-    EXPECT_EQ(theTerm()->getEpochWaitSet().size(), 0);
-    EXPECT_EQ(theTerm()->getDSTermMap().size(), 0);
-    EXPECT_LT(theTerm()->getEpochReadySet().size(), 2);
```
pnstickne: Also, are these warnings addressed if using an *unsigned* type? ie. `2U`.

nmm0: Yeah it looks a bit uglier with initializing as a size_t but the only issue I see with suffixes is that on 64 bit unsigned is typically not the same size as size_t, and other suffixes like UL and ULL assume that unsigned long or unsigned long long is the same size as size_t. I don't know if that's a problem or not

tests/unit/termination/test_term_cleanup.cc
lifflander: @nmm0 do you want to change this or should I merge this?

nmm0: I think in this case it's better to use std:: because the standard c headers stddef.h as opposed to cstddef should be avoided because it brings everything into the global namespace.

pnstickne: @nmm0 See links.

tests/unit/termination/test_term_cleanup.cc
pnstickne: Using `ull`/`ul` would be incorrect because there is no guarantee. The check is, *will forcing an unsigned type* allow the promotion without warning?

nmm0: Looks like there is no difference in compiler warnings:

https://godbolt.org/z/E-foRb

Though using unsigned does add an extra mov instruction for some reason. It doesn't really matter though

PR #670, opened 2020-02-05
653 RDMA

### LINKED ISSUE #653: Implement RDMA for FieldSolutions/Particle use cases
**What Needs to be Done?**

- [ ]

lifflander:
Fixes #653

The implementation is done for objgroup and collections. Both rank- and over-decomposed implementations. This allows the user to create an RDMA handle that that can be get/put/etc. from using the MPI one-sided calls under the hood.

The collection-based support deals with location management of migrated handles and aggregation into a single window multiple sub-handles on a single rank.

This PR also includes tests that cover all the cases.

### DIFF
--- a/src/vt/objgroup/proxy/proxy_objgroup.impl.h
+++ b/src/vt/objgroup/proxy/proxy_objgroup.impl.h
@@ -214,7 +214,6 @@ template <typename ObjT>
 template <typename T>
 void Proxy<ObjT>::destroyHandleSetRDMA(vt::rdma::HandleSet<T> set) const {
   return vt::theHandleRDMA()->deleteHandleSetCollectiveObjGroup<T>(set);
-
 }
 
 }}} /* end namespace vt::objgroup::proxy */
--- a/src/vt/objgroup/proxy/proxy_objgroup.impl.h
+++ b/src/vt/objgroup/proxy/proxy_objgroup.impl.h
@@ -164,11 +164,11 @@ ObjGroupProxyType Proxy<ObjT>::getProxy() const {
 template <typename ObjT>
 template <typename T>
 vt::rdma::Handle<T> Proxy<ObjT>::makeHandleRDMA(
-  std::size_t size, bool is_uniform
+  std::size_t count, bool is_uniform
 ) const {
   return vt::theHandleRDMA()->makeHandleCollectiveObjGroup<
     T, rdma::HandleEnum::StaticSize
-  >(*this, size, is_uniform);
+  >(*this, count, is_uniform);
 }
 
 template <typename ObjT>
--- a/src/vt/objgroup/proxy/proxy_objgroup.impl.h
+++ b/src/vt/objgroup/proxy/proxy_objgroup.impl.h
@@ -166,7 +166,7 @@ template <typename T>
 vt::rdma::Handle<T> Proxy<ObjT>::makeHandleRDMA(
   std::size_t size, bool is_uniform
 ) const {
-  return vt::theHandle()->makeHandleCollectiveObjGroup<
+  return vt::theHandleRDMA()->makeHandleCollectiveObjGroup<
     T, rdma::HandleEnum::StaticSize
   >(*this, size, is_uniform);
 }
@@ -174,7 +174,7 @@ vt::rdma::Handle<T> Proxy<ObjT>::makeHandleRDMA(
 template <typename ObjT>
 template <typename T>
 void Proxy<ObjT>::destroyHandleRDMA(vt::rdma::Handle<T> handle) const {
-  return vt::theHandle()->deleteHandleCollectiveObjGroup<
+  return vt::theHandleRDMA()->deleteHandleCollectiveObjGroup<
     T, rdma::HandleEnum::StaticSize
   >(handle);
 }
@@ -187,7 +187,7 @@ vt::rdma::HandleSet<T> Proxy<ObjT>::makeHandleSetRDMA(
   bool dense_start_at_zero,
   bool is_uniform
 ) const {
-  return vt::theHandle()->makeHandleSetCollectiveObjGroup<
+  return vt::theHandleRDMA()->makeHandleSetCollectiveObjGroup<
     T, rdma::HandleEnum::StaticSize
   >(*this, max_elm, map, dense_start_at_zero, is_uniform);
 }
@@ -195,7 +195,7 @@ vt::rdma::HandleSet<T> Proxy<ObjT>::makeHandleSetRDMA(
 template <typename ObjT>
 template <typename T>
 void Proxy<ObjT>::destroyHandleSetRDMA(vt::rdma::HandleSet<T> set) const {
-  return vt::theHandle()->deleteHandleSetCollectiveObjGroup<T>(set);
+  return vt::theHandleRDMA()->deleteHandleSetCollectiveObjGroup<T>(set);
 
 }
 
--- a/src/vt/objgroup/proxy/proxy_objgroup.impl.h
+++ b/src/vt/objgroup/proxy/proxy_objgroup.impl.h
@@ -184,9 +184,27 @@ template <typename T>
 vt::rdma::HandleSet<T> Proxy<ObjT>::makeHandleSetRDMA(
   int32_t max_elm,
   std::unordered_map<int32_t, std::size_t> const& map,
-  bool dense_start_at_zero,
   bool is_uniform
 ) const {
+  bool dense_start_at_zero = false;
+  return vt::theHandleRDMA()->makeHandleSetCollectiveObjGroup<
+    T, rdma::HandleEnum::StaticSize
+  >(*this, max_elm, map, dense_start_at_zero, is_uniform);
+}
+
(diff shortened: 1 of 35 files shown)


src/vt/configs/types/types_type.h
```diff
@@ -66,6 +66,8 @@ using WorkerIDType            = PhysicalResourceType;
 // Runtime system entity types
 /// Used to hold a handler ID which identifier a function pointer/context
 using HandlerType             = int64_t;
+/// Used to hold a RDMA handle ID which identifier a data block
+using HandleType              = int32_t;
```
pnstickne: Too close to `HandlerType` in naming. If specific to RDMA, consider: `RdmaHandleType` or similar.

src/vt/rdmahandle/handle.h
```diff
@@ -0,0 +1,300 @@
...
+#include "vt/rdmahandle/request_holder.h"
+#include "vt/rdmahandle/lock_mpi.h"
+
+#include <functional>
+#include <vector>
+#include <memory>
+
+namespace vt { namespace rdma {
+
+enum struct HandleEnum {
```
pnstickne: Would consider more refined name, even in rdma-scope.

src/vt/rdmahandle/handle.h
```diff
@@ -0,0 +1,300 @@
...
+    isNodeType<U>* = nullptr
+  );
+
+  template <typename U = IndexT>
+  std::size_t getSize(U, isNodeType<U>* = nullptr);
+
+public:
+
+  template <typename U = IndexT>
+  void get(
```
pnstickne: Fold in calls. For the caller to take more action.

src/vt/runtime/runtime.cc
```diff
@@ -1057,6 +1058,7 @@ void Runtime::initializeComponents() {
   theLocMan = std::make_unique<location::LocationManager>();
   theVirtualManager = std::make_unique<vrt::VirtualContextManager>();
   theCollection = std::make_unique<vrt::collection::CollectionManager>();
+  theHandle = rdma::Manager::construct().get();
```
pnstickne: Too generic a global name.

src/vt/rdmahandle/handle.h
```diff
@@ -0,0 +1,300 @@
...
+  IndexT index_ = {};
+  ObjGroupProxyType proxy_ = no_obj_group;
+};
+
+}} /* end namespace vt::rdma */
+
+namespace vt {
+
+template <typename T, rdma::HandleEnum E, typename IndexT = vt::NodeType>
+using Handle = rdma::Handle<T, E, IndexT>;
```
pnstickne: Too generic in global.

src/vt/rdmahandle/handle.h
```diff
@@ -0,0 +1,300 @@
...
+  template <typename U>
+  using isIndexType =
+    typename std::enable_if<not std::is_same<IndexT,vt::NodeType>::value, U>::type;
+
+  static constexpr HandleEnum handle_type = E;
+
+  Handle() = default;
+  Handle(Handle const&) = default;
+  Handle(Handle&&) = default;
+  Handle& operator=(Handle const&) = default;
```
pnstickne: Consider move operator for parity (it might be implicit in this case already..?) or comment why such is not desired.

examples/simple_collection_collective.cc
```diff
@@ -53,31 +55,86 @@ using namespace ::vt::mapping;
...
+
+    do vt::runScheduler(); while (not handle.ready());
+
+    handle.modifyExclusive([rank](int* val){
+      for (int i = 0; i < 20; i++) {
+        val[i] = 1000*rank + i;
+      }
+    });
+
+    MPI_Barrier(MPI_COMM_WORLD);
```
pnstickne: MPI_Barrier has a method in runtime already?

src/vt/rdmahandle/cache.h
```diff
@@ -0,0 +1,87 @@
...
+#include "vt/rdmahandle/common.h"
+#include "vt/rdmahandle/handle.h"
+#include "vt/rdmahandle/index_info.h"
+
+namespace vt { namespace rdma {
+
+template <typename IndexT>
+struct Cache {
+
+  Cache() = default;
```
pnstickne: Could perhaps unify in IndexT..?

src/vt/rdmahandle/handle.h
pnstickne: Have

```
  template <typename U = IndexT>
  void get(
    U, std::size_t len, int offset, Lock l = Lock::None,
    isNodeType<U>* = nullptr
  );
  template <typename U = IndexT>
  void get(
    U, T* ptr, std::size_t len, int offset, Lock l = Lock::None,
    isNodeType<U>* = nullptr
  );
  template <typename U = IndexT>
  void get(
    U, std::size_t len, int offset, Lock l = Lock::None,
    isIndexType<U>* = nullptr
  );
  template <typename U = IndexT>
  void get(
    U, T* ptr, std::size_t len, int offset, Lock l = Lock::None,
    isIndexType<U>* = nullptr
  );
```

src/vt/rdmahandle/handle.h
```diff
@@ -0,0 +1,300 @@
...
+  template <typename U = IndexT>
+  void modifyShared(std::function<void(T*)> fn, isNodeType<U>* = nullptr);
+  template <typename U = IndexT>
+  void access(
+    Lock l, std::function<void(T*)> fn, std::size_t size, isNodeType<U>* = nullptr
+  );
+
+  // Index-level local data accessors
+  template <typename U = IndexT>
+  void readExclusive(U, std::function<void(T const*)> fn, isIndexType<U>* = nullptr);
```
pnstickne: Probably more confusing in the future to have such type-dependent overloads.

lifflander: I need to static dispatch on this type. I can't combine it into a single method and make it compile. I think this separation is good because the Index variants have different functionality and overheads.

src/vt/configs/types/types_type.h
lifflander: Good idea

examples/simple_collection_collective.cc
lifflander: I will undo this change of the example (and yes, VT has a barrier I was being lazy). I just hacked this to quickly test as I was writing the code. Once the tests get written I will roll this back

src/vt/configs/types/types_type.h
lifflander: Addressed.

src/vt/rdmahandle/handle.h
lifflander: Added.

src/vt/rdmahandle/handle.h
pnstickne: "... because the Index variants *have different functionality* and overheads ..." exactly why it shouldn't be the same >_<

lifflander: I've split these into two specializations which can have different names. Take a look

lifflander:
Code coverage is wrong here. I'm not decreasing it. I think that it's adding more files that were never touched or something: e.g., `src/vt/objgroup/proxy/proxy_objgroup.impl.h`

src/vt/rdmahandle/cache.h
lifflander: What does this mean? Not sure I understand

src/vt/rdmahandle/handle.h
lifflander: Addressed, switched `theHandle` to `theHandleRDMA`

src/vt/runtime/runtime.cc
lifflander: Addressed

lifflander:
@PhilMiller @nlslatt Can you review/approve this? I believe I addressed all your concerns, @PhilMiller.

pnstickne:
Bump @nlslatt @PhilMiller

src/vt/objgroup/proxy/proxy_objgroup.impl.h
```diff
@@ -160,6 +161,62 @@ ObjGroupProxyType Proxy<ObjT>::getProxy() const {
...
+  return vt::theHandleRDMA()->makeHandleSetCollectiveObjGroup<
+    T, rdma::HandleEnum::StaticSize
+  >(*this, max_elm, map, dense_start_at_zero, is_uniform);
+}
+
+template <typename ObjT>
+template <typename T>
+void Proxy<ObjT>::destroyHandleSetRDMA(vt::rdma::HandleSet<T> set) const {
+  return vt::theHandleRDMA()->deleteHandleSetCollectiveObjGroup<T>(set);
+
```
pnstickne: Extra newline.

lifflander: Fixed

src/vt/rdmahandle/handle.h
lifflander: Fixed

PR #675, opened 2020-02-05
671 clang 9 warnings

nmm0:


src/vt/trace/trace_log.h
```diff
@@ -131,8 +131,8 @@ struct Log final {
       UserData() = delete;
       UserData(UserData const&) = default;
       UserData(UserData&&) = default;
-      UserData& operator=(UserData const&) = delete;
```
pnstickne: These were intentionally deleted. What GCC warnings?

src/vt/vrt/collection/balance/baselb/baselb.h
```diff
@@ -87,6 +87,13 @@ struct BaseLB {
       comm_collectives_(in_comm_collectives)
   { }
 
+  BaseLB(BaseLB const &) = delete;
```
pnstickne: What warnings indicate these changes?

src/vt/vrt/collection/balance/proc_stats.cc
```diff
@@ -250,7 +250,7 @@ void StatsRestartReader::gatherMsgs(VecMsg *msg) {
   // --- Check whether all the messages have been received
   //
   const NodeType numNodes = theContext()->getNumNodes();
-  if (ProcStats::proc_reader_->msgsReceived[phaseID] < numNodes)
+  if (ProcStats::proc_reader_->msgsReceived[phaseID] < static_cast<std::size_t>(numNodes))
```
pnstickne: Static cast feels odd. What specific warning does GCC trigger on?
It is resolved when using `size_t{numNodes}` (eg.)?

src/vt/trace/trace_log.h
```diff
@@ -154,6 +154,28 @@ struct Log final {
       }
     }
 
+    Data &operator=(Data const& other_data) {
+      clean();
+      if (other_data.user.data_type == Log::LogDataType::user) {
```
pnstickne: Can fold in code? eg. `moveFrom(X&& ..)` called from constructor and operator.

src/vt/trace/trace_log.h
```diff
@@ -154,6 +154,28 @@ struct Log final {
       }
     }
 
+    Data &operator=(Data const& other_data) {
+      clean();
```
pnstickne: Check with std lib names.. seems like `reset()` might be more common? *shrug*

src/vt/trace/trace_log.h
pnstickne: If adding these, ensure comments above reflect appropriately (eg. no longer restricted to ctors only).

src/vt/vrt/collection/balance/baselb/baselb.h
nmm0: I should have been a bit clearer with the commit. No warnings explicitly here, but I felt it was necessary that the base class has the explicit move-only behavior if the child class has the behavior.

The warnings in StatsMapLB were implicitly deleted move assign and copy assign because those were not declared.

Additionally there was a warning about virtual destructors, both BaseLB and StatsMapLB need to have virtual destructors.

src/vt/trace/trace_log.h
nmm0: Yeah I just used the existing function (the git diff is a bit funky, there was already a function named clean and I just reused it). I also would prefer reset and can change that.

src/vt/trace/trace_log.h
nmm0: It seems really weird to have copy and move construction but no copy and move assignment. I can't think of any situations in which that is desirable, so I will adjust the comment

lifflander:
@nmm0 Can you address Paul's comments and ping me when this is ready to merge?

src/vt/vrt/collection/balance/proc_stats.cc
nmm0: This triggers signed/unsigned comparison warnings because msgsReceived is size_t and numNodes is signed

I think in this case static_cast is better because it's more explicit -- `std::size_t{numNodes}` technically makes a temporary object of size_t then assigns it (but the generated code is identical). However, static_cast says explicitly what it's doing, which is casting.

As far as NodeType I think in general it should be unsigned, however I think changing that could possibly break a lot of things so I don't think it would be worth it.

src/vt/trace/trace_log.h
nmm0: Done

src/vt/trace/trace_log.h
nmm0: Removed the comment -- no special behavior now, UserData is move-only

src/vt/trace/trace_log.h
nmm0: Changed to reset()

nmm0:
@pnstickne @lifflander Hopefully all comments are addressed

PR #642, opened 2019-12-21
642 simplify trace pairs

pnstickne:

PR #679, opened 2020-02-08
#642 fix trace guard

pnstickne:
- Was incorrectly guarding 'event type' == 0, which CAN be true..
  ..even if it probably should be the case. Now guards against
  the 'entry id' itself.


lifflander:
Closing this as the issue # is wrong and will cause confusion

PR #683, opened 2020-02-10
682 fix user event trace incrementing

### LINKED ISSUE #682: Fix user event trace event incrementing
**Describe the bug**

lifflander:
Fixes #682

Another shot at fixing the user event bug. The increments were not happening in the correct sequence.

### DIFF
--- a/src/vt/trace/trace.cc
+++ b/src/vt/trace/trace.cc
@@ -663,7 +663,7 @@ TraceEventIDType Trace::logEvent(LogType&& log) {
   case TraceConstantsType::UserEvent:
   case TraceConstantsType::UserEventPair:
     log.event = cur_event_;
-    if (log.user_data().user_start) {
+    if (not log.user_data().user_start) {
       cur_event_++;
     }
     break;
--- a/src/vt/trace/trace.cc
+++ b/src/vt/trace/trace.cc
@@ -666,8 +666,7 @@ TraceEventIDType Trace::logEvent(LogType&& log) {
     break;
   case TraceConstantsType::BeginUserEventPair:
   case TraceConstantsType::EndUserEventPair:
-    cur_event_++;
-    log.event = cur_event_;
+    log.event = cur_event_++;
     break;
   default:
     vtAssert(0, "Not implemented");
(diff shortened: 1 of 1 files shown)


src/vt/trace/trace.cc
```diff
@@ -663,15 +662,11 @@ TraceEventIDType Trace::logEvent(LogType&& log) {
     break;
   case TraceConstantsType::UserEvent:
   case TraceConstantsType::UserEventPair:
-    if (not log.user_data().user_start) {
-      cur_event_++;
-    }
-    log.event = cur_event_;
+    log.event = log.user_data().user_start ? cur_event_ : cur_event_++;
```
PhilMiller: I'd favor sticking with separate read and increment statements for each of these instances, but especially this one. A ternary with side effects is going to be easily mistaken later.

lifflander: Ok, fixed.

src/vt/trace/trace.cc
```diff
@@ -663,15 +662,17 @@ TraceEventIDType Trace::logEvent(LogType&& log) {
...
-    if (not log.user_data().user_start) {
-      cur_event_++;
+    TraceEventIDType event = no_trace_event;
+    if (log.user_data().user_start) {
+      event = cur_event_;
+    } else {
+      event = cur_event_++;
     }
-    log.event = cur_event_;
+    log.event = event;
```
nlslatt: I think you're overcomplicating this.  `event = cur_event_` and `event = cur_event_++` give `event` exactly the same value.

lifflander: Agreed, I'm just splitting out the branch that increments to make it clear, per @PhilMiller comment:
> I'd favor sticking with separate read and increment statements for each of these instances, but especially this one. A ternary with side effects is going to be easily mistaken later.

nlslatt: Why not make it
```
  log.event = cur_event_;
  if (not log.user_data().user_start) {
    cur_event_++;
  }
```

lifflander: Ok, done

PR #690, opened 2020-02-13
689 New tracing enable specification file

### LINKED ISSUE #689: Add tracing spec to enable/disable tracing on phases based on user-defined pattern
**What Needs to be Done?**

Create a new spec and parser for defining when to enable tracing.

lifflander:
Fixes #689 

Adds a new tracing specification file to enabled traces based on patterns on the LB phase. Needed for large scale EMPIRE runs on Trinity

### DIFF
--- a/src/vt/configs/arguments/args.cc
+++ b/src/vt/configs/arguments/args.cc
@@ -83,7 +83,7 @@ namespace vt { namespace arguments {
 /*static*/ bool        ArgConfig::vt_trace_sys_location   = false;
 /*static*/ bool        ArgConfig::vt_trace_sys_collection = false;
 /*static*/ bool        ArgConfig::vt_trace_sys_serial_msg = false;
-/*static*/ bool        ArgConfig::vt_trace_spec_enabled   = false;
+/*static*/ bool        ArgConfig::vt_trace_spec           = false;
 /*static*/ std::string ArgConfig::vt_trace_spec_file_name = "";
 
 
@@ -244,7 +244,7 @@ namespace vt { namespace arguments {
   auto tsyscoll  = "Trace system virtual context collection events";
   auto tsyssmsg  = "Trace system serialization manager events";
   auto tspec     = "Enable trace spec file (defines which phases tracing is on)";
-  auto tspecfile = "File containing trace spec; --vt_trace_spec_enabled to enable";
+  auto tspecfile = "File containing trace spec; --vt_trace_spec to enable";
   auto n  = app.add_flag("--vt_trace",              vt_trace,           trace);
   auto nm = app.add_flag("--vt_trace_mpi",          vt_trace_mpi,       trace_mpi);
   auto o  = app.add_option("--vt_trace_file",       vt_trace_file,      tfile, "");
@@ -257,7 +257,7 @@ namespace vt { namespace arguments {
   auto qx = app.add_flag("--vt_trace_sys_location",   vt_trace_sys_location,   tsysloc);
   auto qy = app.add_flag("--vt_trace_sys_collection", vt_trace_sys_collection, tsyscoll);
   auto qz = app.add_flag("--vt_trace_sys_serial_msg", vt_trace_sys_serial_msg, tsyssmsg);
-  auto qza = app.add_flag("--vt_trace_spec_enabled",   vt_trace_spec_enabled,   tspec);
+  auto qza = app.add_flag("--vt_trace_spec",          vt_trace_spec,           tspec);
   auto qzb = app.add_option("--vt_trace_spec_file_name", vt_trace_spec_file_name, tspecfile, "");
   auto traceGroup = "Tracing Configuration";
   n->group(traceGroup);
--- a/src/vt/configs/arguments/args.cc
+++ b/src/vt/configs/arguments/args.cc
@@ -84,7 +84,7 @@ namespace vt { namespace arguments {
 /*static*/ bool        ArgConfig::vt_trace_sys_collection = false;
 /*static*/ bool        ArgConfig::vt_trace_sys_serial_msg = false;
 /*static*/ bool        ArgConfig::vt_trace_spec           = false;
-/*static*/ std::string ArgConfig::vt_trace_spec_file_name = "";
+/*static*/ std::string ArgConfig::vt_trace_spec_file      = "";
 
 
 /*static*/ bool        ArgConfig::vt_lb                 = false;
@@ -258,7 +258,7 @@ namespace vt { namespace arguments {
   auto qy = app.add_flag("--vt_trace_sys_collection", vt_trace_sys_collection, tsyscoll);
   auto qz = app.add_flag("--vt_trace_sys_serial_msg", vt_trace_sys_serial_msg, tsyssmsg);
   auto qza = app.add_flag("--vt_trace_spec",          vt_trace_spec,           tspec);
-  auto qzb = app.add_option("--vt_trace_spec_file_name", vt_trace_spec_file_name, tspecfile, "");
+  auto qzb = app.add_option("--vt_trace_spec_file",   vt_trace_spec_file,      tspecfile, "");
   auto traceGroup = "Tracing Configuration";
   n->group(traceGroup);
   nm->group(traceGroup);
--- a/src/vt/configs/arguments/args.cc
+++ b/src/vt/configs/arguments/args.cc
@@ -83,6 +83,9 @@ namespace vt { namespace arguments {
 /*static*/ bool        ArgConfig::vt_trace_sys_location   = false;
 /*static*/ bool        ArgConfig::vt_trace_sys_collection = false;
 /*static*/ bool        ArgConfig::vt_trace_sys_serial_msg = false;
+/*static*/ bool        ArgConfig::vt_trace_spec_enabled   = false;
+/*static*/ std::string ArgConfig::vt_trace_spec_file_name = "";
+
 
 /*static*/ bool        ArgConfig::vt_lb                 = false;
 /*static*/ bool        ArgConfig::vt_lb_file            = false;
@@ -240,6 +243,8 @@ namespace vt { namespace arguments {
   auto tsysloc   = "Trace system location manager events";
   auto tsyscoll  = "Trace system virtual context collection events";
   auto tsyssmsg  = "Trace system serialization manager events";
+  auto tspec     = "Enable trace spec file (defines which phases tracing is on)";
+  auto tspecfile = "File containing trace spec; --vt_trace_spec_enabled to enable";
   auto n  = app.add_flag("--vt_trace",              vt_trace,           trace);
   auto nm = app.add_flag("--vt_trace_mpi",          vt_trace_mpi,       trace_mpi);
   auto o  = app.add_option("--vt_trace_file",       vt_trace_file,      tfile, "");
@@ -252,6 +257,8 @@ namespace vt { namespace arguments {
   auto qx = app.add_flag("--vt_trace_sys_location",   vt_trace_sys_location,   tsysloc);
   auto qy = app.add_flag("--vt_trace_sys_collection", vt_trace_sys_collection, tsyscoll);
   auto qz = app.add_flag("--vt_trace_sys_serial_msg", vt_trace_sys_serial_msg, tsyssmsg);
+  auto qza = app.add_flag("--vt_trace_spec_enabled",   vt_trace_spec_enabled,   tspec);
+  auto qzb = app.add_option("--vt_trace_spec_file_name", vt_trace_spec_file_name, tspecfile, "");
   auto traceGroup = "Tracing Configuration";
   n->group(traceGroup);
   nm->group(traceGroup);
@@ -264,6 +271,8 @@ namespace vt { namespace arguments {
   qx->group(traceGroup);
   qy->group(traceGroup);
   qz->group(traceGroup);
+  qza->group(traceGroup);
+  qzb->group(traceGroup);
 
 
   /*
--- a/src/vt/runtime/runtime.cc
+++ b/src/vt/runtime/runtime.cc
@@ -1057,6 +1057,10 @@ void Runtime::setup() {
     MPI_Barrier(theContext->getComm());
   });
 
+# if backend_check_enabled(trace_enabled)
+  theTrace->parseSpec();
+# endif
+
   if (ArgType::vt_pause) {
     pauseForDebugger();
   }
--- a/src/vt/runtime/runtime.cc
+++ b/src/vt/runtime/runtime.cc
@@ -599,6 +599,29 @@ void Runtime::printStartupBanner() {
       auto f12 = opt_on("--vt_trace_sys_all", f11);
       fmt::print("{}\t{}{}", vt_pre, f12, reset);
     }
+    if (ArgType::vt_trace_spec_enabled) {
+      {
+        auto f11 = fmt::format("Using trace enable specification for phases");
+        auto f12 = opt_on("--vt_trace_spec_enabled", f11);
+        fmt::print("{}\t{}{}", vt_pre, f12, reset);
+      }
+      if (ArgType::vt_trace_spec_file_name == "") {
+        auto warn_trace_file = fmt::format(
+          "{}Warning:{} {}{}{} has no effect: not specification file given"
+          " option {}{}{} is empty{}\n", red, reset, magenta,
+          "--vt_trace_spec_enabled",
+          reset, magenta, "--vt_trace_spec_file_name", reset, reset
+        );
+        fmt::print("{}\t{}{}", vt_pre, warn_trace_file, reset);
+      } else {
+        auto f11 = fmt::format(
+          "Using trace specification file \"{}\"",
+          ArgType::vt_trace_spec_file_name
+        );
+        auto f12 = opt_inverse("--vt_trace_spec_enabled", f11);
+        fmt::print("{}\t{}{}", vt_pre, f12, reset);
+      }
+    }
   }
   #endif
 
--- a/src/vt/runtime/runtime.cc
+++ b/src/vt/runtime/runtime.cc
@@ -607,7 +607,7 @@ void Runtime::printStartupBanner() {
       }
       if (ArgType::vt_trace_spec_file_name == "") {
         auto warn_trace_file = fmt::format(
-          "{}Warning:{} {}{}{} has no effect: not specification file given"
+          "{}Warning:{} {}{}{} has no effect: no specification file given"
           " option {}{}{} is empty{}\n", red, reset, magenta,
           "--vt_trace_spec_enabled",
           reset, magenta, "--vt_trace_spec_file_name", reset, reset
--- a/src/vt/runtime/runtime.cc
+++ b/src/vt/runtime/runtime.cc
@@ -1058,7 +1058,7 @@ void Runtime::setup() {
   });
 
 # if backend_check_enabled(trace_enabled)
-  theTrace->parseSpec();
+  theTrace->loadAndBroadcastSpec();
 # endif
 
   if (ArgType::vt_pause) {
--- a/src/vt/runtime/runtime.cc
+++ b/src/vt/runtime/runtime.cc
@@ -599,17 +599,17 @@ void Runtime::printStartupBanner() {
       auto f12 = opt_on("--vt_trace_sys_all", f11);
       fmt::print("{}\t{}{}", vt_pre, f12, reset);
     }
-    if (ArgType::vt_trace_spec_enabled) {
+    if (ArgType::vt_trace_spec) {
       {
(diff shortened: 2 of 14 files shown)


src/vt/configs/arguments/args.cc
```diff
@@ -252,6 +257,8 @@ namespace vt { namespace arguments {
   auto qx = app.add_flag("--vt_trace_sys_location",   vt_trace_sys_location,   tsysloc);
   auto qy = app.add_flag("--vt_trace_sys_collection", vt_trace_sys_collection, tsyscoll);
   auto qz = app.add_flag("--vt_trace_sys_serial_msg", vt_trace_sys_serial_msg, tsyssmsg);
+  auto qza = app.add_flag("--vt_trace_spec_enabled",   vt_trace_spec_enabled,   tspec);
+  auto qzb = app.add_option("--vt_trace_spec_file_name", vt_trace_spec_file_name, tspecfile, "");
```
nlslatt: Using `--vt_trace_spec` and `--vt_trace_spec_file` seems more consistent with other options.

src/vt/runtime/runtime.cc
```diff
@@ -599,6 +599,29 @@ void Runtime::printStartupBanner() {
...
     }
+    if (ArgType::vt_trace_spec_enabled) {
+      {
+        auto f11 = fmt::format("Using trace enable specification for phases");
+        auto f12 = opt_on("--vt_trace_spec_enabled", f11);
+        fmt::print("{}\t{}{}", vt_pre, f12, reset);
+      }
+      if (ArgType::vt_trace_spec_file_name == "") {
+        auto warn_trace_file = fmt::format(
+          "{}Warning:{} {}{}{} has no effect: not specification file given"
```
nlslatt: Typo: "not" should be "no".

lifflander: Fixed

src/vt/configs/arguments/args.cc
lifflander: I agree on `--vt_trace_spec`. But, for the LB file, the option is `--vt_lb_file_name`. I'm now realizing that that is a unusal case. Other files don't have the "_name" bit.

lifflander: I will go ahead and change both of them. I prefer shorting names.

src/vt/trace/file_spec/spec.h
```diff
@@ -0,0 +1,276 @@
...
+     *
+     * \return whether tracing is enabled based on this entry
+     */
+    bool testTraceEnabledFor(SpecIndex in_idx) const {
+      if (is_mod_) {
+        // Check if mod idx_ is within range, by dividing out idx_ we get the
+        // pos/neg offset from that nod value
+        return in_idx % idx_ >= idx_ + neg_ or in_idx % idx_ <= pos_;
+      } else {
+        // Check intersection with range: [idx_-neg_, idx_-pos_]
```
nlslatt: I think this comment needs correcting.

src/vt/trace/file_spec/spec.h
```diff
@@ -0,0 +1,276 @@
...
+     * \brief Check if a given phase is enabled within this spec entry
+     *
+     * \param[in] in_idx the index to check
+     *
+     * \return whether tracing is enabled based on this entry
+     */
+    bool testTraceEnabledFor(SpecIndex in_idx) const {
+      if (is_mod_) {
+        // Check if mod idx_ is within range, by dividing out idx_ we get the
+        // pos/neg offset from that nod value
```
nlslatt: Should "nod" be "mod"?

src/vt/trace/trace.cc
```diff
@@ -601,8 +623,44 @@ bool Trace::checkDynamicRuntimeEnabled() {
...
+void Trace::setTraceEnabledCurrentPhase(PhaseType cur_phase) {
+  if (spec_proxy_ != vt::no_obj_group) {
+    // SpecIndex is signed due to negative/positive, phase is not signed
+    auto spec_index = static_cast<file_spec::Spec::SpecIndex>(cur_phase);
+    vt::objgroup::proxy::Proxy<file_spec::Spec> proxy(spec_proxy_);
+    bool ret = proxy.get()->checkTraceEnabled(spec_index);
+
+    if (trace_enabled_cur_phase_ != ret) {
+      auto time = getCurrentTime();
+      // Close and pop everything, we are disabling traces at this point
```
nlslatt: Is the check for enabling/disabling traces for a phase at a point where it's possible to have open events?  If that's the case, turning off tracing will create misleading event traces at the point where it's disabled, which we'll need to keep in mind.

src/vt/vrt/collection/balance/lb_invoke/invoke.h
```diff
@@ -104,14 +104,24 @@ struct LBManager {
   void releaseNow(PhaseType phase);
 
 public:
+  void checkTraceEnabledNextPhase(PhaseType phase) {
```
nlslatt: Because there's no return value, "check..." doesn't seem like an appropriate name.  Would `setTraceEnabledNextPhase` more accurately describe it's purpose?

src/vt/trace/file_spec/spec.h
lifflander: Fixed

src/vt/trace/file_spec/spec.h
lifflander: Fixed

src/vt/trace/trace.cc
lifflander: Yes, it's almost impossible to shut them off in a place where the execution is not mid-handler. I discovered this while testing. If I don't close them, the last event covers the entire space while disabled.

They are currently shut off during `sysLB` or `sysReleaseLB` which are small handlers that kick off the LB or releasing the LB.

nlslatt: What happens when we turn tracing on?  Will we have an `EndProcessing` without a matching `BeginProcessing`?

src/vt/configs/arguments/args.cc
lifflander: Fixed.

src/vt/vrt/collection/balance/lb_invoke/invoke.h
lifflander: Addressed.

lifflander:
Note, for this to work properly (given that trace is enabled mid-handler), we will need #642 on beta.5

src/vt/trace/trace.cc
lifflander: With  #642, the tag will be empty that is generated from `beginProcessing`. This will cause the mid-handler event to be ignored.

src/vt/trace/file_spec/spec.h
```diff
@@ -0,0 +1,277 @@
...
+ * trigger a parsing error. But having a "%100" and "100" entry is valid.
+ *
+ * Whether tracing is enabled is calculated as an OR across all specification
+ * entries. Thus, if a given phase is contained in any spec line, it is
+ * enabled. Please note that 0 % 100 = 0. Therefore, if the above example did
+ * not contain the first line, tracing would be enabled as:
+ * { [0,3], [97,103], * [195,205], [297,303], ... }
+ *
+ */
+struct Spec {
```
pnstickne: TraceSpec

src/vt/trace/file_spec/spec.h
```diff
@@ -0,0 +1,277 @@
...
+namespace vt { namespace trace { namespace file_spec {
+
+/**
+ * \struct Spec spec.h vt/trace/file_spec/spec.h
+ *
+ * \brief Parses trace spec file when available and tests when its enabled. A
+ * single node parses the specification; all others receive the spec from a
+ * broadcast.
+ *
+ * Parses the following format:
```
pnstickne: Including column headers (phase, range negative, range positive or whatnot) in the would add a good bit of clarity.

src/vt/trace/file_spec/spec.h
```diff
@@ -0,0 +1,277 @@
...
+ * { [0,10], [97,103], [195,205], [297,303], ... }
+ *
+ * The sets of mod-phase and phase-specific entries must be unique. There may be
+ * overlap across the two sets, but not within them. Having two entries that
+ * start with "%100" or two entries that start with "100" would be invalid and
+ * trigger a parsing error. But having a "%100" and "100" entry is valid.
+ *
+ * Whether tracing is enabled is calculated as an OR across all specification
+ * entries. Thus, if a given phase is contained in any spec line, it is
+ * enabled. Please note that 0 % 100 = 0. Therefore, if the above example did
```
pnstickne: No "please" :)

src/vt/trace/file_spec/spec.cc
```diff
@@ -0,0 +1,252 @@
...
+    if (std::isdigit(c) or static_cast<char>(c) == '-') {
+      file >> phase_negative_offset;
+      if (phase_negative_offset > 0) {
+        auto str = fmt::format(
+          "Parsing file \"{}\" error: found offset in negative offset position"
+          " that is > 0: value \"{}\"",
+          filename,
+          phase_negative_offset
+        );
+        vtAbort(str);
```
pnstickne: Consider `vtAbortIf` or `vtAssert` instead of a conditional (and use the string formatting provided therein) to bring the intent up-front. It's a critical failure with no recourse.

src/vt/trace/file_spec/spec.cc
```diff
@@ -0,0 +1,252 @@
...
+  if (iter != map.end()) {
+    auto str = fmt::format(
+      "Parsing file \"{}\" error: multiple lines start with the same {}:"
+      " value \"{}{}\"",
+      is_mod ? "mod phase" : "phase",
+      ArgType::vt_trace_spec_file,
+      is_mod ? "%" : "",
+      phase
+    );
+    vtAbort(str);
```
pnstickne: Ditto. IMOHO, Asserts/Aborts should be primary conditions.

src/vt/trace/file_spec/spec.cc
```diff
@@ -0,0 +1,252 @@
...
+    spec_mod_ = msg->spec_mod_;
+    spec_exact_ = msg->spec_exact_;
+    has_spec_ = true;
+  }
+}
+
+void Spec::insertSpec(
+  SpecIndex phase, SpecIndex neg, SpecIndex pos, bool is_mod, SpecMapType& map
+) {
+  auto iter = map.find(phase);
```
pnstickne: Maybe just allow (not guard) this case? The user will get the "OR" logic, so..

src/vt/runtime/runtime.cc
```diff
@@ -1034,6 +1057,10 @@ void Runtime::setup() {
     MPI_Barrier(theContext->getComm());
   });
 
+# if backend_check_enabled(trace_enabled)
+  theTrace->parseSpec();
```
pnstickne: `loadTraceSpec` or `loadAndBroadcastSpec` spec eg. Parsing is only one component.

src/vt/trace/trace.h
```diff
@@ -183,6 +183,9 @@ struct Trace {
 
   bool checkDynamicRuntimeEnabled();
 
+  void parseSpec();
```
pnstickne: Name doesn't indicate full intent. Missing documentation and/or name indicators that is performs a VT broadcast and is blocking.

src/vt/trace/trace.cc
```diff
@@ -601,8 +623,44 @@ bool Trace::checkDynamicRuntimeEnabled() {
...
+  return
+    enabled_ and
+    traceWritingEnabled(theContext()->getNode()) and
+    trace_enabled_cur_phase_;
+}
+
+void Trace::setTraceEnabledCurrentPhase(PhaseType cur_phase) {
+  if (spec_proxy_ != vt::no_obj_group) {
+    // SpecIndex is signed due to negative/positive, phase is not signed
+    auto spec_index = static_cast<file_spec::Spec::SpecIndex>(cur_phase);
```
pnstickne: Using `file_spec::Spec::SpecIndex{cur_phase}` should catch narrowing errors with a warning (using the braces is relevant). Or if narrowing conversion is expected, static_cast..

src/vt/trace/trace.cc
```diff
@@ -601,8 +623,44 @@ bool Trace::checkDynamicRuntimeEnabled() {
...
+    vt::objgroup::proxy::Proxy<file_spec::Spec> proxy(spec_proxy_);
+    bool ret = proxy.get()->checkTraceEnabled(spec_index);
```
pnstickne: I clearly still fail to understand proxies :|

src/vt/trace/file_spec/spec.h
pnstickne: Also recommend adding `\internal` to all types that are not explicitly designed to be publicly used.

src/vt/trace/file_spec/spec.h
```diff
@@ -0,0 +1,277 @@
...
+ * broadcast.
+ *
+ * Parses the following format:
+ * """
+ * 0 0 10
+ * %100 -3 3
+ * 200 -5 5
+ * """
+ *
+ * This specifies that tracing will be enabled on the following phases:
```
pnstickne: I would bring mention on phase union up here and indicate which line selects the phase.

{
   [0,10],        # phase 0 with offsets 0,+10 (subsumes [0,3] from %100 -3 3)
   [97,103]     # any phase % 100 with offset -3,+3
   [195,205]  # phase 200 with offsets -5,+5 (subsumes [197,203] from  %100 -3 3)
   [297,303]  # any phase % 100 with offset -3,+3
   [n%100-3,n%100+3] ...  # any phase % 100 with offset -3,+3

src/vt/trace/file_spec/spec.h
```diff
@@ -0,0 +1,277 @@
...
+ * The sets of mod-phase and phase-specific entries must be unique. There may be
+ * overlap across the two sets, but not within them. Having two entries that
+ * start with "%100" or two entries that start with "100" would be invalid and
+ * trigger a parsing error. But having a "%100" and "100" entry is valid.
+ *
+ * Whether tracing is enabled is calculated as an OR across all specification
+ * entries. Thus, if a given phase is contained in any spec line, it is
+ * enabled. Please note that 0 % 100 = 0. Therefore, if the above example did
+ * not contain the first line, tracing would be enabled as:
+ * { [0,3], [97,103], * [195,205], [297,303], ... }
```
pnstickne: { [0,3] # any phase mod 100 from -3,+3 }

src/vt/trace/file_spec/spec.h
```diff
@@ -0,0 +1,277 @@
...
+ * %100 -3 3
+ * 200 -5 5
+ * """
+ *
+ * This specifies that tracing will be enabled on the following phases:
+ * { [0,10], [97,103], [195,205], [297,303], ... }
+ *
+ * The sets of mod-phase and phase-specific entries must be unique. There may be
+ * overlap across the two sets, but not within them. Having two entries that
+ * start with "%100" or two entries that start with "100" would be invalid and
```
pnstickne: While silly from the end-users' view, this isn't actually a critical failure.. it would simplify the current contract and code to remove the restriction.

src/vt/trace/file_spec/spec.h
```diff
@@ -0,0 +1,277 @@
...
+#if !defined INCLUDED_VT_TRACE_FILE_SPEC_SPEC_H
+#define INCLUDED_VT_TRACE_FILE_SPEC_SPEC_H
+
+#include "vt/config.h"
+#include "vt/objgroup/proxy/proxy_objgroup.h"
+#include "vt/collective/reduce/operators/default_msg.h"
+
+#include <unordered_map>
+
+namespace vt { namespace trace { namespace file_spec {
```
pnstickne: Maybe `trace_spec` or just `spec` (it's a namespace, not a type so refinement for searching is less relevant).

src/vt/trace/file_spec/spec.h
```diff
@@ -0,0 +1,277 @@
...
+     *
+     * \param[in] in_idx the index to check
+     *
+     * \return whether tracing is enabled based on this entry
+     */
+    bool testTraceEnabledFor(SpecIndex in_idx) const {
+      if (is_mod_) {
+        // Check if mod idx_ is within range, by dividing out idx_ we get the
+        // pos/neg offset from that mod value
+        return in_idx % idx_ >= idx_ + neg_ or in_idx % idx_ <= pos_;
```
pnstickne: Fancy. It might be more clear overall to keep the neg_ index as a positive value.

Then the formula are `idx_ - neg` (instead of + assumed - neg).

src/vt/trace/file_spec/spec.h
```diff
@@ -0,0 +1,277 @@
...
+
+  /**
+   * \brief Check if spec has been received
+   */
+  bool specReceived() const { return has_spec_; };
+
+  /**
+   * \brief Clear the current specification (used for testing)
+   */
+  void clear() {
```
pnstickne: If JUST for testing, I like to use sentinels like "clear_for_tests".

Then if it shows up in non-test code I can go hunt someone down 🗡

src/vt/trace/trace.cc
```diff
@@ -141,6 +143,26 @@ void Trace::initialize() {
...
+void Trace::parseSpec() {
+  if (ArgType::vt_trace_spec) {
+    auto spec_proxy = file_spec::Spec::construct();
+    theTerm()->produce();
+    if (theContext()->getNode() == 0) {
+      auto spec_ptr = spec_proxy.get();
+      spec_ptr->parse();
+      spec_ptr->broadcastSpec();
+    }
+    while (not spec_proxy.get()->specReceived()) {
```
pnstickne: What happens if this never happens?

Does a barrier fit in here, and/or could another schedule pump idiom be used (eg. this never checks if the schedule ends itself)?

src/vt/trace/trace.cc
```diff
@@ -141,6 +143,26 @@ void Trace::initialize() {
   );
 }
 
+void Trace::parseSpec() {
```
pnstickne: Ditto notes on non-covering name.

src/vt/vrt/collection/balance/lb_invoke/invoke.cc
```diff
@@ -239,4 +239,12 @@ void LBManager::releaseNow(PhaseType phase) {
   num_invocations_ = num_release_ = 0;
 }
 
+void LBManager::setTraceEnabledNextPhase(PhaseType phase) {
+  // Check if tracing is enabled for this next phase. Do this immediately before
```
pnstickne: It doesn't "check"..

lifflander: Fixed

src/vt/trace/file_spec/spec.h
lifflander: Good point, but I'm not going to screw with it right now.

src/vt/trace/file_spec/spec.h
lifflander: I get that, but I also think it will most likely be a typo that I want to catch. There isn't a valid use case beyond laziness or a mistake where you should specify the same phase twice. I would rather catch that mistake instead of just running assuming it wasn't a mistake.

src/vt/trace/file_spec/spec.h
lifflander: Good point, but I'm not going to change this now because I would have to regenerate all the header guards, etc.

src/vt/trace/file_spec/spec.cc
lifflander: Fixed

src/vt/trace/trace.cc
lifflander: If it doesn't happen, then the broadcast is broken. It will hang. It can't terminate because I explicitly produced/consumed before and after which makes it impossible to terminate early:
```C++
    theTerm()->produce();
    if (theContext()->getNode() == 0) {
      auto spec_ptr = spec_proxy.get();
      spec_ptr->parse();
      spec_ptr->broadcastSpec();
    }
    while (not spec_proxy.get()->specReceived()) {
      vt::runScheduler();
    }
    theTerm()->consume();
```

src/vt/trace/trace.cc
```diff
@@ -601,8 +623,44 @@ bool Trace::checkDynamicRuntimeEnabled() {
...
+      auto time = getCurrentTime();
+      // Close and pop everything, we are disabling traces at this point
+      while (not open_events_.empty()) {
+        traces_.push(
+          LogType{open_events_.top(), time, TraceConstantsType::EndProcessing}
+        );
+        open_events_.pop();
+      }
+    }
+
```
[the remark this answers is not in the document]
lifflander: Added the flush

PR #693, opened 2020-02-13
691 GossipLB: re-compute PMF each assignment, loosen underloaded

### LINKED ISSUE #691: Update the PMF each assignment in GossipLB
**What Needs to be Done?**

Based on the analysis by @ppebay and testing in the LBAF, we need to update the PMF to get better convergence.

lifflander:
#Fixes #691

### DIFF
--- a/src/vt/vrt/collection/balance/gossiplb/gossiplb.h
+++ b/src/vt/vrt/collection/balance/gossiplb/gossiplb.h
@@ -81,11 +81,12 @@ protected:
   void propagateRound(EpochType epoch = no_epoch);
   void propagateIncoming(GossipMsg* msg);
   bool isUnderloaded(LoadType load) const;
+  bool isUnderloadedRelaxed(LoadType over, LoadType under) const;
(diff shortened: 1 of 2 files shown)


src/vt/vrt/collection/balance/gossiplb/gossiplb.h
```diff
@@ -81,11 +81,12 @@ struct GossipLB : BaseLB {
   void propagateRound(EpochType epoch = no_epoch);
   void propagateIncoming(GossipMsg* msg);
   bool isUnderloaded(LoadType load) const;
+  bool isUnderloadedRelaxed(LoadType over, LoadType under) const;
```
pnstickne: Is `isUnderloaded` still used?

lifflander: No, I need to remove that. Good catch.

lifflander: Actually it is used

PR #701, opened 2020-02-18
696 HierarchicalLB: Add strategy argument to LB file

### LINKED ISSUE #696: Add LB file argument to control HierarchicalLB object selection strategy
**What Needs to be Done?**

- [ ] Add argument to LB file for HierarchicalLB that maps to `HeapExtractEnum`.
- [ ] Use selection as argument for `calcLoadOver`.

**Additional context**

This will allow considering migration of largest objects or objects of any size instead of only smallest objects.

nlslatt:
Note: I based this PR off of beta.5 instead of develop.

Add a string argument called `strategy` to the LB file entries for `HierarchicalLB` to control what value of `HeapExtractEnum` is passed to `calcLoadOver`. The accepted strings are the names of `HeapExtractEnum` values that have been implemented, which are currently `LoadOverLessThan` (default), `LoadOverGreaterThan`, and `LoadOverOneEach`. This option allows deviating from the default practice of preferring to migrate the smallest objects.

I have omitted `LoadOverRandom` because it doesn't appear to be implemented (and there did not appear to be any check that this value wasn't selected).  Please correct me if I am wrong.

Please let me know if you have suggestions for better strings to use.  I can see rephrasing to use something like `Smallest` for `LoadOverLessThan` and `Largest` for `LoadOverGreaterThan`, but I'm not sure what would fit for `LoadOverOneEach` (which means take one candidate from each size bin, if I understand correctly).

Fixes #696

### DIFF
--- a/src/vt/vrt/collection/balance/hierarchicallb/hierlb.cc
+++ b/src/vt/vrt/collection/balance/hierarchicallb/hierlb.cc
@@ -73,11 +73,26 @@ void HierarchicalLB::init(objgroup::proxy::Proxy<HierarchicalLB> in_proxy) {
 }
 
 void HierarchicalLB::inputParams(balance::SpecEntry* spec) {
-  std::vector<std::string> allowed{"min", "max", "auto"};
+  std::vector<std::string> allowed{"min", "max", "auto", "strategy"};
   spec->checkAllowedKeys(allowed);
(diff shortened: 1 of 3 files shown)


lifflander:
@nlslatt I switched this over to develop and rebased it since that's my usual flow.

nlslatt:
> @nlslatt I switched this over to develop and rebased it since that's my usual flow.

@lifflander Thank you.

PR #704, opened 2020-02-20
694 Fix termination segfault caused by TermDS being erased while running

### LINKED ISSUE #694: Fix bug in "vt:*/TestTermDepSendChain.test_term_dep_send_chain/*_proc_2"
**Describe the bug**
On `develop`, `vt:*/TestTermDepSendChain.test_term_dep_send_chain/*_proc_2` has been failing occasionally. A bug must have been introduced recently.

Here's an example failure:

https://github.com/DARMA-tasking/vt/runs/451407453

lifflander:
Fixes #694

This fixes a very occasional bug in termination DS that causes a segfault which is present on develop and beta.5

### DIFF
--- a/src/vt/termination/termination.cc
+++ b/src/vt/termination/termination.cc
@@ -671,9 +671,15 @@ void TerminationDetector::cleanupEpoch(EpochType const& epoch, CallFromEnum from
 
   if (epoch != any_epoch_sentinel) {
     if (isDS(epoch)) {
-      auto ds_term_iter = term_.find(epoch);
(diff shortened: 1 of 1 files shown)


lifflander:
> This looks fine.
> 
> Is the failure something we're ever seeing in operation, or just with tests running?

I was causing valgrind errors, but most of the time not failing. I believe it impacts anything with DS epochs where the DS epoch terminates at the root without any non-local events. I'm not sure why we didn't see crashes elsewhere due to it.

PR #706, opened 2020-02-21
703 mixed idle events

### LINKED ISSUE #703: Tracing bug: begin/end idle events mixed with begin/end processing
**Describe the bug**
We are seeing in the generated log files begin/end idle events inside a `beginProcessing`/`endProcessing`. This seems to be happening both on develop and beta.5.

pnstickne:
Fixes #703

### DIFF
--- a/examples/broadcast_test.cc
+++ b/examples/broadcast_test.cc
@@ -47,28 +47,50 @@
 
 using namespace vt;
 
-#define BCAST_DEBUG 0
+#define BCAST_DEBUG 1
 
-static constexpr int32_t const num_bcasts = 4;
+static constexpr int32_t const num_bcasts = 4;   // number of intial root bcasts
+static constexpr int32_t const bcast_depth = 1; // 10; // number re-broadcasts
 static NodeType my_node = uninitialized_destination;
 static int32_t bcast_count = 0;
 
 struct Msg : vt::Message {
   NodeType broot;
+  int bcastsLeft;
 
-  Msg(NodeType const& in_broot) : Message(), broot(in_broot) { }
+  Msg(
+    NodeType const& in_broot,
+    int in_bcastsLeft
+  ) : Message(), broot(in_broot), bcastsLeft(in_bcastsLeft) { }
 };
 
 static void bcastTest(Msg* msg) {
+  auto const node = theContext()->getNode();
   auto const& root = msg->broot;
 
-  #if BCAST_DEBUG
-  fmt::print("{}: bcastTestHandler root={}\n", theContext()->getNode(), msg->broot);
-  #endif
+#if BCAST_DEBUG
+  fmt::print("{}: bcastTestHandler root={} bcastsleft={}\n", node, msg->broot, msg->bcastsLeft);
+#endif
 
   vtAssert(
     root != my_node, "Broadcast should deliver to all but this node"
   );
+
+#if BCAST_DEBUG
+  fmt::print("{}: waiting on barrier before initiating sub bcast\n", node);
+#endif
+
+  // Wait on barrier as a result of RECIEVING a bcast
+  theCollective()->barrier();
+
+  if (msg->bcastsLeft > 1) {
+    theMsg()->broadcastMsg<Msg, bcastTest>(
+      makeSharedMessage<Msg>(node, msg->bcastsLeft - 1)
+    );
+
+    // Wait on barrier ourselves - as we don't get the message
+    theCollective()->barrier();
+  }
 
   bcast_count++;
 }
@@ -97,13 +119,15 @@ int main(int argc, char** argv) {
 
   theTerm()->addAction([=]{
     fmt::print("[{}] verify: bcast_count={}, expected={}\n", my_node, bcast_count, expected);
-    vtAssertExpr(bcast_count == expected);
+    //    vtAssertExpr(bcast_count == expected);
   });
 
   if (from_node == uninitialized_destination or from_node == my_node) {
     fmt::print("[{}] broadcast_test: broadcasting {} times\n", my_node, num_bcasts);
     for (int i = 0; i < num_bcasts; i++) {
-      theMsg()->broadcastMsg<Msg, bcastTest>(makeSharedMessage<Msg>(my_node));
+      theMsg()->broadcastMsg<Msg, bcastTest>(makeSharedMessage<Msg>(my_node, bcast_depth));
+      // All receives will wait on the barrier - we should too.
+      theCollective()->barrier();
     }
   }
 
--- a/src/vt/scheduler/scheduler.cc
+++ b/src/vt/scheduler/scheduler.cc
@@ -87,6 +87,7 @@ void Scheduler::enqueue(PriorityType priority, ActionType action) {
 # endif
 }
 
+/*private*/
 bool Scheduler::runNextUnit() {
   if (not work_queue_.empty()) {
     auto elm = work_queue_.pop();
@@ -101,6 +102,7 @@ bool Scheduler::runNextUnit() {
   }
 }
 
+/*private*/
 bool Scheduler::progressImpl() {
   bool const msg_sch = theMsg()->progress();
   bool const evt_sch = theEvent()->progress();
@@ -123,6 +125,7 @@ bool Scheduler::progressImpl() {
   return scheduled_work;
 }
 
+/*private*/
 bool Scheduler::progressMsgOnlyImpl() {
   return theMsg()->progress() or theEvent()->progress();
 }
@@ -243,12 +246,6 @@ void Scheduler::registerTriggerOnce(
     event_triggers.size() >= event, "Must be large enough to hold this event"
   );
   event_triggers_once[event].push_back(trigger);
-}
-
-void Scheduler::schedulerForever() {
-  while (true) {
-    scheduler();
-  }
 }
 
 }} //end namespace vt::scheduler
--- a/src/vt/scheduler/scheduler.cc
+++ b/src/vt/scheduler/scheduler.cc
@@ -88,18 +88,18 @@ void Scheduler::enqueue(PriorityType priority, ActionType action) {
 }
 
 /*private*/
-bool Scheduler::runNextUnit() {
-  if (not work_queue_.empty()) {
-    auto elm = work_queue_.pop();
-    bool const is_term = elm.isTerm();
-    elm();
-    if (is_term) {
-      num_term_msgs_--;
-    }
-    return true;
-  } else {
-    return false;
+bool Scheduler::runWorkUnit(UnitType& work) {
+  bool const is_term = work.isTerm();
+
+  ++unit_run_depth_;
+  work();
+  --unit_run_depth_;
+
+  if (is_term) {
+    --num_term_msgs_;
   }
+
+  return unit_run_depth_ == 0;
 }
 
 /*private*/
@@ -191,16 +191,18 @@ void Scheduler::scheduler(bool msg_only) {
       is_idle_minus_term = false;
       triggerEvent(SchedulerEventType::EndIdleMinusTerm);
     }
-    if (is_idle and not work_queue_.empty()) {
+    if (is_idle) {
       is_idle = false;
       triggerEvent(SchedulerEventType::EndIdle);
     }
 
     processed_after_last_progress_++;
-    runNextUnit();
+
+    UnitType work = work_queue_.pop();
+    bool top_scheduler = runWorkUnit(work);
 
     // Enter idle state immediately after processing if relevant.
-    if (not is_idle and work_queue_.empty()) {
+    if (top_scheduler and not is_idle and work_queue_.empty()) {
       is_idle = true;
       triggerEvent(SchedulerEventType::BeginIdle);
     }
--- a/src/vt/scheduler/scheduler.cc
+++ b/src/vt/scheduler/scheduler.cc
@@ -262,7 +262,9 @@ SchedulerRunner::SchedulerRunner(
 }
 
 SchedulerRunner::~SchedulerRunner() {
+#if backend_check_enabled(trace_enabled)
   theTrace()->endProcessing(processing_event_);
+#endif
 }
 
 bool SchedulerRunner::runScheduler() {
--- a/src/vt/scheduler/scheduler.cc
+++ b/src/vt/scheduler/scheduler.cc
@@ -217,6 +217,10 @@ void Scheduler::scheduler(bool msg_only) {
   }
 }
 
+NestedSchedulerRunner Scheduler::beginNestedScheduling() {
+  return NestedSchedulerRunner{unit_run_depth_};
+}
+
 void Scheduler::triggerEvent(SchedulerEventType const& event) {
   vtAssert(
     event_triggers.size() >= event, "Must be large enough to hold this event"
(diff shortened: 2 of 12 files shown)


src/vt/trace/trace.cc
```diff
@@ -669,11 +669,13 @@ TraceEventIDType Trace::logEvent(LogType&& log) {
...
 
-  // close any idle event as soon as we encounter any other type of event
+  // Close any idle event as soon as we encounter any other type of event.
+  // Take the time of the event being emitted as our own as it has already
+  // been created (in the past).
   if (idle_begun_ and
       log.type != TraceConstantsType::BeginIdle and
       log.type != TraceConstantsType::EndIdle) {
-    endIdle();
+    endIdle(log.time);
```
pnstickne: As far as I can tell (or, as I've only been able to see), this is "the fix". It prevents the idle end event from having a larger end-time then the closing event..

src/vt/scheduler/scheduler.cc
```diff
@@ -180,44 +183,31 @@ void Scheduler::scheduler(bool msg_only) {
...
-  }
 
-  /*
-   * Run a work unit!
-   */
-  if (not work_queue_.empty()) {
     processed_after_last_progress_++;
     runNextUnit();
+
+    // Enter idle state immediately after processing if relevant.
```
pnstickne: Idle can only start immediately preceding running an action.

Not related to the fix, although ensures the idle is entered prior to wait for the next loop.

pnstickne: Okay, I don't like that suspicion: going to ensure it doesn't happen!

src/vt/scheduler/scheduler.cc
```diff
@@ -180,44 +191,32 @@ void Scheduler::scheduler(bool msg_only) {
...
-   */
-  if (not work_queue_.empty()) {
     processed_after_last_progress_++;
-    runNextUnit();
+
+    bool top_scheduler = runNextUnit();
+
+    // Enter idle state immediately after processing if relevant.
+    if (top_scheduler and not is_idle and not work_queue_.empty()) {
+      is_idle = true;
```
nlslatt: Why is this `not work_queue_.empty()`?

src/vt/trace/trace.cc
pnstickne: This is a minor bug. It has been fixed.

src/vt/scheduler/scheduler.cc
pnstickne: Added 'nested scheduler' support as well as suppressing idle events from nested scheduler calls..

src/vt/scheduler/scheduler.cc
pnstickne: Legit bug in that particular commit. Fixed in subsequent updates.

src/vt/termination/term_scope.cc
```diff
@@ -97,9 +102,14 @@ namespace vt { namespace term {
   closure();
   theMsg()->popEpoch();
   theTerm()->finishedEpoch(epoch);
-  while (!term_finished) {
-    runScheduler();
+
+  if (!term_finished) {
```
pnstickne: With these changes, outside of primary program pump/loop, `theSched()->beginNestedScheduling()` is the "appropriate" method to perform a blocking scheduler loop.

It uses RAII for emitting begin and end events. Since RAII is used the `if` (while representing some duplication) scope ensures proper lifetime management. The code _could_ be updated to defer the begin event to the runScheduler call, which would change the usage pattern to:

```
    {
      auto sched = theSched()->beginNestedScheduling();
      while (!term_finished) {
        sched.runScheduler();
      }
    }
```

src/vt/vrt/collection/balance/lb_invoke/invoke.cc
```diff
@@ -168,9 +168,16 @@ void LBManager::waitLBCollective() {
...
-    vt::runScheduler();
+
+  { // blocking scheduler loop
+    auto sched = theSched()->beginScheduling(
+      sched::SchedulerTypeID::LbWaitCollective
+    );
+    while (synced_in_lb_) {
+      sched.runScheduler();
+    }
   }
```
pnstickne: With these changes, outside of primary program pump/loop, `theSched()->beginScheduling()` is the "appropriate" method to perform a blocking scheduler loop in all context except the 'main scheduling loop' which remains unchanged for client compatibility.

If calling just `vt::runScheduler()` in such cases, NO events regarding "busy time spent waiting for the scheduler loop to complete" will be emitted because only the top-level scheduler emits idle events due to tooling limitations.

It uses RAII for emitting begin and end events.

src/vt/vrt/collection/balance/gossiplb/gossiplb.cc
```diff
@@ -457,8 +462,13 @@ void GossipLB::decide() {
 
   theTerm()->finishedEpoch(lazy_epoch);
 
-  while (not decide_done) {
-    vt::runScheduler();
+  { // blocking scheduler loop
+    auto sched = theSched()->beginScheduling(
+      sched::SchedulerTypeID::LbGossipDecide
```
pnstickne: Each blocking scheduler specifies the type of scheduling. This is written to log events (only applied when the scheduler is actually in a nested context). In addition, log events contain the _depth_ of the nesting.

src/vt/trace/trace.h
```diff
@@ -116,11 +129,22 @@ struct Trace {
...
   );
 
+  TraceProcessingTag beginProcessing(
+    TraceEntryIDType const ep, TraceMsgLenType const len,
+    TraceEventIDType const event, NodeType const from_node,
+    double const time = getCurrentTime(),
+    uint64_t const idx1 = 0, uint64_t const idx2 = 0,
+    uint64_t const idx3 = 0, uint64_t const idx4 = 0
+  ) {
+    // Call overload with unifrom trailing-time.
```
pnstickne: Unicrons..

src/vt/scheduler/scheduler.h
```diff
@@ -69,6 +70,30 @@ enum SchedulerEvent {
   SchedulerEventSize = 4
 };
 
+/**
+ * \brief The type of the blocking scheduler loop being run.
+ *
+ * Values less than 1024 are reserved for internal VT usages.
+ */
+enum struct SchedulerTypeID : uint32_t {
```
PhilMiller: I **really** don't like the coupling implied by having a static list of constants describing every nested scheduler caller across the system.

That said, this is not particularly harmful coupling, and it's in service of fixing a real, immediate bug, so if everything else is ok, this complaint should *not* delay integration.

If dynamic registration is possible, it would ultimately be preferable.

src/vt/scheduler/scheduler.h
```diff
@@ -89,17 +114,35 @@ struct Scheduler {
...
   void scheduler(bool msg_only = false);
   void runProgress(bool msg_only = false);
-  bool progressImpl();
-  void schedulerForever();
+
+  /**
+   * \brief Obtain a scheduler runner.
+   *
+   * A scheduler that can be used for a local BLOCKING poll.
+   * The returnted scheduler shares the same events as the global scheduler.
```
PhilMiller: Typo 'returnted'

src/vt/scheduler/scheduler.h
```diff
@@ -89,17 +114,35 @@ struct Scheduler {
...
+   *                        Values less than 1024 are reserved for internal VT usages.
+   */
+  SchedulerRunner beginScheduling(SchedulerTypeID scheduler_tag);
+
   void registerTrigger(SchedulerEventType const& event, TriggerType trigger);
   void registerTriggerOnce(
     SchedulerEventType const& event, TriggerType trigger
   );
   void triggerEvent(SchedulerEventType const& event);
+
```
PhilMiller: Stray newline makes for a noisy diff. Don't care much here, but please clean up diffs as much as possible before posting them for review in the future.

src/vt/trace/trace.h
```diff
@@ -116,11 +129,22 @@ struct Trace {
   TraceProcessingTag beginProcessing(
     TraceEntryIDType const ep, TraceMsgLenType const len,
     TraceEventIDType const event, NodeType const from_node,
-    double const time = getCurrentTime(),
-    uint64_t const idx1 = 0, uint64_t const idx2 = 0, uint64_t const idx3 = 0,
-    uint64_t const idx4 = 0
+    uint64_t const idx1, uint64_t const idx2,
+    uint64_t const idx3, uint64_t const idx4,
+    double const time = getCurrentTime()
```
PhilMiller: @lifflander Not new to this PR, but are we really putting function calls in our default argument expressions? It may be valid, but it's kinda horrifying.

examples/broadcast_test.cc
```diff
@@ -47,29 +47,51 @@
...
   );
 
+#if BCAST_DEBUG
+  fmt::print("{}: waiting on barrier before initiating sub bcast\n", node);
+#endif
+
+  // Wait on barrier as a result of RECIEVING a bcast
+  theCollective()->barrier();
+
+  if (msg->bcastsLeft > 1) {
```
lifflander: I don't understand this change at all. Why are we broadcasting more form this handler now? Why do we need barriers here?

examples/broadcast_test.cc
```diff
@@ -97,13 +119,15 @@ int main(int argc, char** argv) {
...
+    //    vtAssertExpr(bcast_count == expected);
   });
 
   if (from_node == uninitialized_destination or from_node == my_node) {
     fmt::print("[{}] broadcast_test: broadcasting {} times\n", my_node, num_bcasts);
     for (int i = 0; i < num_bcasts; i++) {
-      theMsg()->broadcastMsg<Msg, bcastTest>(makeSharedMessage<Msg>(my_node));
+      theMsg()->broadcastMsg<Msg, bcastTest>(makeSharedMessage<Msg>(my_node, bcast_depth));
+      // All receives will wait on the barrier - we should too.
+      theCollective()->barrier();
```
lifflander: We don't need a barrier here. The termination loop at the end guarantees that all these are done before termination triggers.

src/vt/scheduler/scheduler.cc
```diff
@@ -258,13 +256,43 @@ void Scheduler::registerTriggerOnce(
...
+#endif
+}
+
+bool SchedulerRunner::runScheduler() {
+
+#if backend_check_enabled(trace_enabled)
+  // It might be more ideal to have a dedicated pairing event.
+  // n.b. static here.
+  static trace::TraceEntryIDType ep
+    = trace::TraceRegistry::registerEventHashed("Scheduler", "RunScheduler");
```
lifflander: This should have a namespace on it to differentiate it from other events. At the very least, `::vt::sched::Scheduler`

src/vt/scheduler/scheduler.h
lifflander: I complete agree. This needs to be a dynamic list. Apps may need to do this, so this needs to be a dynamic list.

src/vt/scheduler/scheduler.h
```diff
@@ -140,7 +197,38 @@ struct Scheduler {
...
 
-}} //end namespace vt::scheduler
+/**
+  * \brief Facilitates helping with nested scheduling contexts.
+  *
+  * This is a Resource Acquisition type that ensures pre/post
+  * state cleanup when invoking a scheduler in a nested context.
+  * It should be destructed immediately after scheduling.
+  */
+struct SchedulerRunner {
```
lifflander: What if we have an overload that lets the user to attach a polling function and test function. Take them in as `typename Callable` or something. Then in the destructor this can spin until it finishes?

lifflander: There could also be a simpler overload that just attached by reference to a boolean.

src/vt/vrt/collection/balance/lb_invoke/invoke.cc
lifflander: I'm worried about this. The processor is idle even if it's blocked and not doing work. All the idle analysis that the tool does will come back wrong. The user will have to add idle + runScheduler time to see when the processor is actually idle and not busy.

examples/broadcast_test.cc
lifflander: I think I get this now. You were testing. Can you roll this back for prod?

lifflander:
Here's a *after* screenshot with this PR. Note that the RED is the new `Scheduler::runScheduler`. 
Screen Shot 2020-02-24 at 3 19 30 PM

src/vt/trace/trace.h
lifflander: Yes, quite horrifying. Fortunately, this just maps down to `MPI_Wtime()`. We should just remove these overloads.

lifflander:
@PhilMiller @nlslatt I want to collect feedback from others, but I think we need to close the outer event when a recursive scheduler is invoked instead of creating a new event that doesn't display at idle. We can start creating that event, but if the sub-scheduler doesn't make immediate progress I think we need to emit an idle event. Otherwise, this implies that we will never have idle events when using a sub-scheduler.

lifflander:
The other pattern that emerges is:

```C++
int test() {
  auto ep = theTerm()->makeCollectiveEpoch();
  theMsg()->pushEpoch(ep);
  // send some messages
  theMsg()->popEpoch(ep);
  theTerm()->finishedEpoch(ep);
  bool is_done = false;
  theTerm()->addAction(ep, [&]{ is_done = true; });
  while (not is_done) {
    vt::runScheduler();
  }
}
```

It would be nice if we can translate this to:

```C++
int test() {
  auto ep = theTerm()->makeCollectiveEpoch();
  theMsg()->pushEpoch(ep);
  // send some messages
  theMsg()->popEpoch(ep);
  theTerm()->finishedEpoch(ep);
  theSched()->waitForEpoch(ep);
}
```

PhilMiller:
Better,

vt::runToTermination([](EpochType epoch){
  // Send some messages, etc.
});

That captures the common pattern of making an epoch, pushing it, popping
it, finishing it, and running until it's done.

It should also have an overload that takes a function that doesn't take an
epoch parameter.

Not for this PR though

src/vt/scheduler/scheduler.h
pnstickne: It is an intentional new newline to separate the methods.

src/vt/vrt/collection/balance/lb_invoke/invoke.cc
pnstickne: Idle **is** the appropriate event. However, if the limitation is that "idle cannot overlap with nested events", then there are very limited events.

- Fix or change tooling that cannot handle idle inside of nested events
- Don't use idle events (chosen here)
- Close all open events, enter idle, re-open all events

examples/broadcast_test.cc
pnstickne: Whoops. Meant to keep that off.

pnstickne: It is off. This was not in my latest push. The rebase/force picked up earlier commits.

pnstickne: Compare with https://github.com/DARMA-tasking/vt/pull/707

pnstickne:
Will open a new PR. Unnecessary deviations.

PR #709, opened 2020-02-26
703 mixed idle events pop

pnstickne:


src/vt/trace/trace.h
```diff
@@ -91,7 +91,9 @@ struct Trace {
   using TraceConstantsType  = eTraceConstants;
   using TimeIntegerType     = int64_t;
   using TraceContainerType  = std::queue<LogType>;
-  using TraceStackType      = std::stack<LogType>;
+  // Although should be used mostly as a stack, vector is exposed to enable
+  // the use of a synthetic pop-push to maintain the stack around idle.
+  using TraceStackType      = std::vector<LogType>;
```
pnstickne: Could also use `deque`. A vector doesn't haven't to maintain extra internal nodes.

src/vt/collective/barrier/barrier.cc
```diff
@@ -132,12 +132,13 @@ void Barrier::waitBarrier(
 
   barrierUp(is_named, is_wait, barrier, skip_term);
 
-  while (not barrier_state.released) {
-    vt::runScheduler();
-    if (poll_action) {
+  theSched()->runSchedulerWhile([&poll_action, &barrier_state]{
+    bool cond = not barrier_state.released;
```
pnstickne: A back of a wonky case.. should be okay, mostly.

src/vt/scheduler/scheduler.cc
```diff
@@ -179,52 +183,77 @@ void Scheduler::scheduler(bool msg_only) {
...
-   */
-  if (num_term_msgs_ == work_queue_.size()) {
-    if (not is_idle_minus_term) {
-      is_idle_minus_term = true;
-      triggerEvent(SchedulerEventType::BeginIdleMinusTerm);
+  if (not work_queue_.empty()) {
+    processed_after_last_progress_++;
+
+    // Leave idle states are before any potential processing.
+    // True-idle MUST be the outer state to enter/leave.
```
pnstickne: Avoids issues with events being generated in the epoch idle.

src/vt/scheduler/scheduler.cc
```diff
@@ -179,52 +183,77 @@ void Scheduler::scheduler(bool msg_only) {
...
-  }
 
-  /*
-   * Run a work unit!
-   */
-  if (not work_queue_.empty()) {
-    processed_after_last_progress_++;
-    runNextUnit();
+  } else {
+    // n.b. Work queue is empty - however, if a trace event was written WITHOUT
```
pnstickne: This case should go away in 1.1; it's currently not allowed in nested schedulers.

src/vt/scheduler/scheduler.h
```diff
@@ -89,17 +92,25 @@ struct Scheduler {
...
+   * in order to ensure proper event unwinding and idle time tracking.
+   */
+  void runSchedulerWhile(std::function<bool()> cond);
+
   void registerTrigger(SchedulerEventType const& event, TriggerType trigger);
   void registerTriggerOnce(
     SchedulerEventType const& event, TriggerType trigger
   );
   void triggerEvent(SchedulerEventType const& event);
+
```
pnstickne: Newline consciously inserted and self-reviewed.

src/vt/scheduler/scheduler.h
```diff
@@ -127,6 +150,8 @@ struct Scheduler {
   bool has_executed_      = false;
   bool is_idle            = true;
   bool is_idle_minus_term = true;
+  // The depth of work action currently executing.
+  unsigned int unit_run_depth_ = 0;
```
pnstickne: `sched_level_` or so, perhaps.

src/vt/trace/trace.cc
```diff
@@ -114,48 +112,43 @@ Trace::Trace(std::string const& in_prog_name, std::string const& in_trace_name)
...
-}
 
-/*static*/ void Trace::traceEndIdleTrigger() {
-  #if backend_check_enabled(trace_enabled)
-    if (theTrace()->inIdleEvent()) {
-      theTrace()->endIdle();
-    }
-  #endif
+  // The first (implied) scheduler always starts with an empty event stack.
+  sched_depths_.push_back(0);
```
pnstickne: "depth" applies to the open_stack_, not the scheduler nesting level (which is the size of the collection).

src/vt/trace/trace.cc
```diff
@@ -484,25 +477,32 @@ void Trace::endProcessing(
...
-    LogType{open_events_.top(), time, TraceConstantsType::EndProcessing}
+    LogType{open_events_.back(), time, TraceConstantsType::EndProcessing}
   );
-  open_events_.pop();
+  open_events_.pop_back();
 
-  if (not open_events_.empty()) {
+  if (open_events_.size() > sched_depths_.size()) {
     // Emit a '[re]start' event for the reactivated stack item.
+    // The item MUST be in the current scheduler depth.
```
pnstickne: ..current scheduler event depth..

src/vt/trace/trace.h
```diff
@@ -91,11 +91,12 @@ struct Trace {
...
+  using SchedulerDepthStackType = std::vector<int>;
```
pnstickne: Could use a stack with a vector backing.. seems fine (and consistent with TraceStackType) to keep as a raw vector though, especially as it is private.

src/vt/trace/trace.h
```diff
@@ -136,6 +137,9 @@ struct Trace {
     uint64_t const idx4 = 0
   );
 
+  void beginSchedulerLoop(double const time = getCurrentTime());
```
pnstickne: The usage of defaults can be cleaned up 'for all' events later.

src/vt/trace/trace_log.h
```diff
@@ -299,6 +299,7 @@ struct Log final {
   }
 
   inline Data::SysData const& sys_data() const {
+    vtAssert(data.sys.data_type == LogDataType::sys, "must be sys");
```
pnstickne: Oops. Remove this.

pnstickne:
image

pnstickne:
image

And here is the view with the implicit processing event opened. The yellow events represents "vt::sched::Loop": the nested barrier loops in the contrived test are busy-spinning. They are not idle because there ARE events; just no events to fulfill the runSchedulerWhile condition.

src/vt/trace/trace.h
```diff
@@ -119,6 +120,16 @@ struct Trace {
...
+  ) {
+    return beginProcessing(ep, len, event, from_node, idx1, idx2, idx3, idx4, time);
+  }
+
+  TraceProcessingTag beginProcessing(
+     TraceEntryIDType const ep, TraceMsgLenType const len,
+     TraceEventIDType const event, NodeType const from_node,
+     uint64_t const idx1, uint64_t const idx2,
+     uint64_t const idx3, uint64_t const idx4,
+     double const time = getCurrentTime()
```
pnstickne: Future consideration / vt1.1

src/vt/scheduler/scheduler.h
```diff
@@ -89,17 +92,25 @@ struct Scheduler {
...
+
+  /**
+   * \brief Runs the scheduler until a condition is met.
+   *
+   * Runs the scheduler until a condition is met.
+   * This form SHOULD be used instead of "while (..) { runScheduler(..) }"
+   * in all cases of nested scheduler loops, such as during a barrier,
+   * in order to ensure proper event unwinding and idle time tracking.
+   */
+  void runSchedulerWhile(std::function<bool()> cond);
```
pnstickne: Using std::function vs template as [virtual] function call should be easily dominated by any relevant processing.

src/vt/scheduler/scheduler.h
```diff
@@ -127,6 +150,8 @@ struct Scheduler {
   bool has_executed_      = false;
   bool is_idle            = true;
   bool is_idle_minus_term = true;
+  // The depth of work action currently executing.
```
pnstickne: Not strictly scheduler depth - at least until removing all external usages of calling `vt::runScheduler` directly.

src/vt/scheduler/scheduler.cc
```diff
@@ -64,8 +64,10 @@ namespace vt { namespace sched {
 }
 
 Scheduler::Scheduler() {
-  event_triggers.resize(SchedulerEventType::SchedulerEventSize + 1);
-  event_triggers_once.resize(SchedulerEventType::SchedulerEventSize + 1);
+  auto event_count = SchedulerEventType::LastSchedulerEvent + 1;
+  event_triggers.resize(event_count);
+  event_triggers_once.resize(event_count);
```
PhilMiller: I recognize that this isn't introduced by this PR, but I think it's worth pointing out that this smells quadratic, if only in a hopefully infrequent way with relatively small N. Probably worth cleaning up, though. It looks like it would be appropriate to change it from `vector` to `unordered_map` and just remove this code.

PhilMiller: Oh, I misread what this was doing. This isn't happening in each nested call, this is just happening once at initialization. It being written as `resize` threw me way off. These should be initializers to the intended size, probably where they're declared, since that's a `constexpr` size.

PhilMiller:
Just a general note on how Projections views the world vis-a-vis idle time and what you're calling  time not truly accounted for, which Projections calls 'overhead'. That overhead time is useful to distinguish from idle time, because it's time the system is spending on doing something useful in support of the application, but outside discretely recorded events. Idle time is time in which the process has nothing useful to do, and is truly just waiting. So the equation we actually want, and will always have, is `time==processing+idle+overhead` - the hard bit is accurately distinguishing among those three. Without the nested scheduler fixes, some idle and overhead time would be accounted as processing time from whatever outer event made the nested scheduler call, or would maybe lead to double counting some time.

src/vt/scheduler/scheduler.cc
pnstickne: Added to additional/future work notes.

pnstickne:
Closing this branch. I was overzealous. Will re-create a min-change branch only for overlap and idle bugs for beta6. Future improvements as vt1.1.

PR #711, opened 2020-02-26
710 Termination: use DS-epoch by default in dep send chain

### LINKED ISSUE #710: Collection chain set still uses non-DS rooted epoch
**Describe the bug**
The `makeEpochRooted` call in collection_chain_set.h was fixed to default to a DS-epoch. However, the call in `dependent_send_chain.h` was missed.

lifflander:
Fixes #710 

Unfortunately, the last fix to switch `nextStep` to using DS epochs by default missed a call in dependent send chain. This is a major problem for EMPIRE performance.

### DIFF
--- a/src/vt/messaging/dependent_send_chain.h
+++ b/src/vt/messaging/dependent_send_chain.h
@@ -160,7 +160,10 @@ class DependentSendChain final {
     // Set up a 'closed' epoch so that we keep an invariant of always
     // having an epoch to call addAction on, rather than edge cases of
     // whether we've made one yet or not
-    last_epoch_ = theTerm()->makeEpochRooted();
+
+    // The parameter `true` here tells VT to use an efficient rooted DS-epoch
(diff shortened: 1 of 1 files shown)


src/vt/messaging/dependent_send_chain.h
```diff
@@ -160,7 +160,10 @@ class DependentSendChain final {
     // Set up a 'closed' epoch so that we keep an invariant of always
     // having an epoch to call addAction on, rather than edge cases of
     // whether we've made one yet or not
-    last_epoch_ = theTerm()->makeEpochRooted();
+
+    // The parameter `true` here tells VT to use an efficient rooted DS-epoch
+    // by default. This can still be overridden by command-line flags
+    last_epoch_ = theTerm()->makeEpochRooted(true);
```
pnstickne: The curse of default values..

PR #718, opened 2020-03-02
595 include gtest

### LINKED ISSUE #595: Deploy Google Test 1.10 with VT and ensure warnings "go away"
GoogleTest deprecated some old API. Clean up the test to use the new method.

```
/Users/pnstick/code/vt/tests/unit/termination/test_termination_action_nested_collect.cc:72:1: warning: 'InstantiateTestCase_P_IsDeprecated' is deprecated: INSTANTIATE_TEST_CASE_P is deprecated, please use
      INSTANTIATE_TEST_SUITE_P [-Wdeprecated-declarations]
INSTANTIATE_TEST_CASE_P /* NOLINT */(
^
/Users/pnstick/opt-clang/googletest/include/gtest/gtest-param-test.h:507:38: note: expanded from macro 'INSTANTIATE_TEST_CASE_P'
  static_assert(::testing::internal::InstantiateTestCase_P_IsDeprecated(), \
                                     ^
/Users/pnstick/opt-clang/googletest/include/gtest/internal/gtest-internal.h:1220:1: note: 'InstantiateTestCase_P_IsDeprecated' has been explicitly marked deprecated here
GTEST_INTERNAL_DEPRECATED(
^
/Users/pnstick/opt-clang/googletest/include/gtest/internal/gtest-port.h:2203:59: note: expanded from macro 'GTEST_INTERNAL_DEPRECATED'
#define GTEST_INTERNAL_DEPRECATED(message) __attribute__((deprecated(message)))
                                                          ^
/Users/pnstick/code/vt/tests/unit/termination/test_termination_action_nested_collect.cc:72:1: error: expected expression
INSTANTIATE_TEST_CASE_P /* NOLINT */(
^
```

The new API is (most likely) already in 1.8.1; so it's probably a clean forward-migration.

Also might be useful to pin a higher googletest required version (as well as to update the CMake for a few small let-me-use-my-debug-libs).

pnstickne:
- Includes a [stripped] version of googletest in tests/extern, including fetch script. 
- The dependency (and header location) is added to test targets, and only test targets.
- Updates tests for gtest 1.10 compatibility and hides trailing macro commas.
- Moves cmake example/tests logic to primary CMakeList.
- Switches to positive-case VT_BUILD_TESTS, honors BUILD_TESTING, etc.
- Builds examples in CI (so that a passing test suite implies all examples ran successfully as well)

Fixes #595


pnstickne:
- GTests added to `tests/lib` (source) so it can't be accidentally referenced or copied from primary library or builds.
- Doesn't generate libgtest in `tests/lib` (output); incorrect CMake/expectation.
- Should probably exclude `tests/lib` (source) from Codacy.

lifflander:
My vote is for what you currently have---snapshot of gtest in tests/lib.

pnstickne:
@lifflander Feels finished - minus some possible subsequent cleanup.

However, the examples 'tests' are _not_ running under CI. This is not new behavior: only about 1/3rd the tests are running on CI vs locally.

lifflander:
What do you mean by the examples "tests"? The examples folder aren't the real tests? And they compose a small subset of the binaries that can be compiled?

pnstickne:
@lifflander Also, Codacy would be much happier if excluding tests/extern :|

pnstickne:
> What do you mean by the examples "tests"? The examples folder aren't the real tests? And they compose a small subset of the binaries that can be compiled?

The examples are run 'as tests' locally, where success = "no severe failure". They are not done so on the build server.

pnstickne:
> What do you mean by the examples "tests"? The examples folder aren't the real tests? And they compose a small subset of the binaries that can be compiled?

Ahh. Mis-diagnosis.. the 'example tests' aren't running locally either. It appears CI is only running for 2 nodes.

lifflander:
> > What do you mean by the examples "tests"? The examples folder aren't the real tests? And they compose a small subset of the binaries that can be compiled?
> 
> Ahh. Mis-diagnosis.. the 'example tests' aren't running locally either. It appears CI is only running for 2 nodes (1/3rd of 2..4..8 nodes).
> 
> Now the 'examples as tests' should be running everywhere. Adds about 50 'tests'. CI still on 2 nodes only: maybe 2 + 7 nodes or some such..?

The reason I disabled example tests is because they were causing timeouts on Travis CI. The additionally build/run time was too expensive. It's possible I could include them on Github actions, but it would increase the build+run time.

pnstickne:
@lifflander Good to merge!

The example build/tests (at least with only 2 procs only; some of those 8 proc runs..) seem to add non-relevant additional time.

tests/extern/fetch-googletest.sh
```diff
@@ -0,0 +1,17 @@
+#! /bin/sh
+
+GTEST_BRANCHTAG=release-1.10.0
+
+# Fetch googletest from Guthub.
```
PhilMiller: Typo here 'Guthub'

PhilMiller:
And yeah, please add the exclusions of gtest itself for Codacy, and codecov.

tests/extern/fetch-googletest.sh
lifflander: Fixed

PhilMiller:
Paul's note mentioned exclusion from Codacy as well, if that's possible

lifflander:
> Paul's note mentioned exclusion from Codacy as well, if that's possible

Done

PR #719, opened 2020-03-03
#703 fix hold-at bugs in tracing

pnstickne:
- Fixed bug where size was being used instead of
  hold stack value.

- Trace disabling triggered from phases honors the
  same logic.


pnstickne:
This should prevent phase enable/disable from crashing (regression) and fixes the case where multiple events are pushed in each nesting, leading to better trace:

image

pnstickne:
The regression fix is to nip a beta-6 regression before it's out, not strictly anything else that might be occurring on beta-5.

I have a slight suspicion that there could be an begin/end tracing issues (there definitely is for the _standard_ enable/disable trace calls which will be fixed vt1.1) that could affect the phase tracing. I'll inspect that later and create a separate ticket/etc if I find anything relevant specifically wrt the phase-based tracing.

src/vt/trace/trace.cc
```diff
@@ -498,7 +498,8 @@ void Trace::endProcessing(
   );
   open_events_.pop_back();
 
-  if (open_events_.size() > event_holds_.size()) {
+  int hold_at = event_holds_.back();
+  if (open_events_.size() > hold_at) {
```
nlslatt: `error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]`

src/vt/trace/trace.cc
```diff
@@ -705,7 +706,8 @@ void Trace::setTraceEnabledCurrentPhase(PhaseType cur_phase) {
     if (trace_enabled_cur_phase_ != ret) {
       auto time = getCurrentTime();
       // Close and pop everything, we are disabling traces at this point
-      while (not open_events_.empty()) {
+      int hold_at = event_holds_.back();
+      while (open_events_.size() > hold_at) {
```
nlslatt: `error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]`

src/vt/trace/trace.cc
```diff
@@ -750,7 +752,8 @@ TraceEventIDType Trace::logEvent(LogType&& log) {
   switch (log.type) {
   case TraceConstantsType::BeginProcessing: {
 
-    if (open_events_.size() > event_holds_.size()) {
+    int hold_at = event_holds_.back();
+    if (open_events_.size() > hold_at) {
```
nlslatt: `error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]`

src/vt/trace/trace.cc
pnstickne: Uhg, I need to crank up Clang warnings .. seems to be a GCC 9 change?

I'm not aware of a really good solution to dealing with unsigned-size_t and an int (of known valid size bounds). Using a signed int is _consciously_ done as it avoids very easy loop-condition overflows. I suppose a static-cast at each site from size_t out?

pnstickne: In particular, `unsigned_expr - 1` overflows :-/

nlslatt: I was using GNU 7.2.0.

pnstickne:
@nlslatt @PhilMiller @lifflander 
Requested changes wrt. sign compare in. Also another.. go me 🙄 update for a slight regression in idle/processing interactions.

pnstickne:
image

Now pretty little details like this show up correctly.

src/vt/trace/trace.cc
pnstickne: I added the warning flag to VT and fixed all related warnings. I wonder what other flags might be good to add..

src/vt/trace/trace.cc
```diff
@@ -538,9 +532,9 @@ void Trace::beginSchedulerLoop() {
   double const time = getCurrentTime();
 
   // Synth-pop any residue from outside scheduler.
-  int sched_top = event_holds_.back();
-  int size = open_events_.size();
-  for (int i = size - 1; i >= sched_top; i--) {
+  std::size_t parent_hold_at = event_holds_.back();
+  std::size_t size = open_events_.size();
+  for (std::size_t i = size - 1; i >= parent_hold_at and i < SIZE_MAX; i--) {
```
lifflander: Please use `std::numeric_limits<std::size_t>::max()` here instead of `SIZE_MAX`

src/vt/trace/trace.cc
nlslatt: EMPIRE uses `-Wall -Wshadow -Wunused-variable -pedantic -Wno-unknown-pragmas -Wno-narrowing -Wno-format`, which is the criteria we need to meet.  I guess one of those flags effectively includes `-Wsign-compare` on some compilers but not others.

src/vt/trace/trace.cc
pnstickne: SIZE_MAX is defined as (std::size_t)-1 :|

pnstickne: How about just a `(std::size_t)-1`?

lifflander: Sounds good to me!

pnstickne: Resolved -> `i not_eq (std::size_t)-1`

pnstickne:
@lifflander Updates in to allow end-events to unwind regardless of when event tracing is stopped.

Also switched [back] to user event from the begin/end processing for the loops. Even though such "looks better" in terms of accounted for time, it was burying what caused the loop to run..

Some small UB and edge-case fixes too, along with a no-op scoped event.

Here is a current trace with the latest changes - the user events at top cover (but don't mask) the begin/end events contained within:

image

pnstickne:
Closing this branch. I was overzealous. Will re-create a min-change branch only for overlap and idle bugs for beta6.

Issue #708, opened 2020-02-25
Balance workloads within synchronized subphases (vector loads rather than scalar)


PhilMiller:
**What Needs to be Done?**

1. [x] Add an API for objects to indicate which subphase their work is part of
1. [x] Extend LB instrumentation to record a subphase-indexed vector of loads per object per phase, rather than the current scalar(s)
1. [x] Add API for load balancing strategies to query vector loads
1. [ ] <strike>Modify LB strategy migration criterion to disallow candidates that would increase vector imbalance</strike>
1. [x] Output load vectors to LB stats vom file

**Is your feature request related to a problem? Please describe.**

In EMPIRE, each step consists of a number of subphases. Those subphases are separated by global synchronization events. Objects loads across subphases are not directly proportional. Balancing by the objects' aggregate load during each entire step leaves substantial imbalances within each subphase. This limits performance and scalability.

**Describe potential solution outcome**

**Describe alternatives you've considered**

**Additional context**

PhilMiller:
@lifflander @ppebay will want to track this

PhilMiller:
@ppebay What do you want for the format of subphase object times in the stats file?

lifflander:
I think it make sense for the sub-phases to be sorted in order as the phases are in the output. We could put it at the end to maintain some sort of positional compatibility or just add it after the phase, which probably makes more sense: i.e., in the 2nd position for each line, for both comp and comm.

PhilMiller:
Here are a few improvement criteria that could be used in a vector strategy.

Suppose we have 

```c++
TimeType getObjLoad(ObjectID, Phase, Subphase);
TimeType getProcLoad(ProcID, Phase, Subphase);

vector<TimeType> subphase_maximums(num_subphases, 0);
vector<TimeType> subphase_averages(num_subphases, 0);

for (int j = 0; j < num_subphases; j++) {
  for (int i = 0; i < num_procs; ++i) {
    subphase_maximum[j] = max(subphase_maximum[j], getProcLoad(i, phase, j));
    subphase_averages[j] += getProcLoad(i, phase, j);
  }
  subphase_averages[j] /= num_procs;
}
```

The strongest criterion would be that a proposed migration not create any overload in any subphase:

```c++
bool MigrationCreatesNoOverload(ObjectID obj, ProcID src_proc, ProcID dst_proc) {
  for (int i = 0; i < num_subphases; ++i) {
[...]

PR #720, opened 2020-03-04
#708 vector balancing, part 1

PhilMiller:
This is sufficient API to see how it fits together with EMPIRE, I think.

To validate the changes all the way through, I'm going to add code on one side or the other to force consideration of only a single subphase's load when making LB decisions.

### DIFF
--- a/src/vt/vrt/collection/balance/elm_stats.h
+++ b/src/vt/vrt/collection/balance/elm_stats.h
@@ -89,6 +89,7 @@ struct ElementStats {
   PhaseType getPhase() const;
   TimeType getLoad(PhaseType const& phase) const;
   CommMapType const& getComm(PhaseType const& phase);
+  void setSubPhase(int subphase);
 
   template <typename Serializer>
   void serialize(Serializer& s);
@@ -106,6 +107,9 @@ protected:
   PhaseType cur_phase_ = fst_lb_phase;
   std::vector<TimeType> phase_timings_ = {};
   std::vector<CommMapType> comm_ = {};
+
+  int cur_subphase_ = 0;
+  std::vector<std::vector<TimeType>> subphase_timings_ = {};
 };
 
 }}}} /* end namespace vt::vrt::collection::balance */
--- a/src/vt/vrt/collection/balance/elm_stats.h
+++ b/src/vt/vrt/collection/balance/elm_stats.h
@@ -94,6 +94,9 @@ struct ElementStats {
   CommMapType const& getComm(PhaseType const& phase);
   void setSubPhase(SubphaseType subphase);
 
+  static const constexpr SubphaseType no_subphase = std::numeric_limits<SubphaseType>::max();
+  static void setFocusedSubPhase(SubphaseType subphase);
+
   template <typename Serializer>
   void serialize(Serializer& s);
 
@@ -113,6 +116,8 @@ protected:
 
   SubphaseType cur_subphase_ = 0;
   std::vector<std::vector<TimeType>> subphase_timings_ = {};
+
+  static SubphaseType focused_subphase_;
 };
 
 }}}} /* end namespace vt::vrt::collection::balance */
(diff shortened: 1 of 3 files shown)


src/vt/vrt/collection/balance/elm_stats.h
```diff
@@ -106,6 +107,9 @@ struct ElementStats {
   PhaseType cur_phase_ = fst_lb_phase;
   std::vector<TimeType> phase_timings_ = {};
   std::vector<CommMapType> comm_ = {};
+
+  int cur_subphase_ = 0;
```
lifflander: Make sure you update the `serialize` for this in the imp.h file

src/vt/vrt/collection/balance/elm_stats.h
```diff
@@ -106,6 +107,9 @@ struct ElementStats {
...
+  std::vector<std::vector<TimeType>> subphase_timings_ = {};
```
PhilMiller: These need to be serialized.

src/vt/vrt/collection/balance/elm_stats.h
```diff
@@ -106,6 +113,11 @@ struct ElementStats {
   PhaseType cur_phase_ = fst_lb_phase;
   std::vector<TimeType> phase_timings_ = {};
   std::vector<CommMapType> comm_ = {};
+
+  SubphaseType cur_subphase_ = 0;
+  std::vector<std::vector<TimeType>> subphase_timings_ = {};
+
+  static SubphaseType focused_subphase_;
```
lifflander, line 120: This static worries me because it would apply to all collection elements and all collections.

PhilMiller: Ooh, yeah, good call. I hadn't thought that through beyond a single instance of a single collection

PhilMiller: It'll test OK in EMPIRE, but I should fix that to be maybe a static map from collection ID to focus. I don't want to make it an instance variable, because it applies to the whole collection

lifflander: The correct way is probably to make it an instance variable, and then have the collection set it on all local elements. Should that be called from the proxy?

src/vt/vrt/collection/balance/elm_stats.impl.h
```diff
@@ -86,7 +88,7 @@ template <typename ColT>
   auto const& cur_phase = msg->getPhase();
   auto const& proxy = col->getCollectionProxy();
   auto const& untyped_proxy = col->getProxy();
-  auto const& total_load = stats.getLoad(cur_phase);
+  auto const& total_load = stats.getLoad(cur_phase, getFocusedSubPhase(proxy.getProxy()));
```
PhilMiller, line 91: Oops, realized I could use `untyped_proxy` here

src/vt/vrt/collection/balance/elm_stats.h
lifflander: The fix looks good

lifflander:
@PhilMiller Want to take this out of draft status?

lifflander:
This looks good to merge onto beta.7!

src/vt/vrt/collection/balance/elm_stats.h
```diff
@@ -88,7 +90,14 @@ struct ElementStats {
   void updatePhase(PhaseType const& inc = 1);
   PhaseType getPhase() const;
   TimeType getLoad(PhaseType const& phase) const;
+  TimeType getLoad(PhaseType phase, SubphaseType subphase) const;
+
   CommMapType const& getComm(PhaseType const& phase);
+  void setSubPhase(SubphaseType subphase);
+
+  static const constexpr SubphaseType no_subphase = std::numeric_limits<SubphaseType>::max();
+  static void setFocusedSubPhase(VirtualProxyType collection, SubphaseType subphase);
```
PhilMiller, line 99: To improve the API a little bit, we should probably add a non-static member to `Migratable` or whatever has `ElmStats` as a member variable to call`setFocusedSubPhase(this->getProxy()->getProxy(), subphase)`

PhilMiller: There were compilation anomalies in getting this done, so  merged as is, and this will be resolved later

PR #722, opened 2020-03-04
721 Trace flush bug fixes

### LINKED ISSUE #721: Flush only at LB (stop unsync'ed flushes)
**Describe the bug**
We are observing unsynchronized flushes, and not all LB paths lead to flushing when enabled.

lifflander:
Fixes #721

### DIFF
--- a/src/vt/trace/trace.cc
+++ b/src/vt/trace/trace.cc
@@ -772,13 +772,6 @@ TraceEventIDType Trace::logEvent(LogType&& log) {
   TraceEventIDType event = log.event;
   traces_.push(std::move(log));
 
-  // If auto-flush, can flush immediately.
-  // TODO: log time of flushing; unify with group-end.
-  if (ArgType::vt_trace_flush_size not_eq 0
-      and traces_.size() >= static_cast<std::size_t>(ArgType::vt_trace_flush_size)) {
-    writeTracesFile(incremental_flush_mode);
-  }
-
   return event;
 }
 
--- a/src/vt/trace/trace.cc
+++ b/src/vt/trace/trace.cc
@@ -861,16 +861,12 @@ void Trace::writeTracesFile(int flush, bool is_incremental_flush) {
       );
     }
 
-    if (is_incremental_flush) {
-      vt::trace::TraceScopedEvent scope(flush_event_);
-      outputTraces(
-        log_file_.get(), traces_, start_time_, flush
-      );
-    } else {
-      outputTraces(
-        log_file_.get(), traces_, start_time_, flush
-      );
-    }
+    vt::trace::TraceScopedEvent scope(
+      is_incremental_flush ? flush_event_ : no_user_event_id
+    );
+    outputTraces(
+      log_file_.get(), traces_, start_time_, flush
+    );
 
     trace_write_count_ += to_write;
   }
--- a/src/vt/trace/trace.cc
+++ b/src/vt/trace/trace.cc
@@ -673,8 +673,8 @@ void Trace::setTraceEnabledCurrentPhase(PhaseType cur_phase) {
 
       // Go ahead and perform a trace flush when tracing is disabled (and was
       // previously enabled) to reduce memory footprint.
-      if (not ret) {
-        flushTracesFile(false);
+      if (not ret and ArgType::vt_trace_flush_size != 0) {
+        writeTracesFile(incremental_flush_mode, true);
       }
     }
 
--- a/src/vt/trace/trace.cc
+++ b/src/vt/trace/trace.cc
@@ -145,7 +145,7 @@ void Trace::initialize() {
   );
 
   // Register a trace user event to demarcate flushes that occur
-  flush_event = vt::trace::registerEventCollective("trace_flush");
+  flush_event_ = vt::trace::registerEventCollective("trace_flush");
 }
 
 void Trace::loadAndBroadcastSpec() {
@@ -862,7 +862,7 @@ void Trace::writeTracesFile(int flush, bool is_incremental_flush) {
     }
 
     if (is_incremental_flush) {
-      vt::trace::TraceScopedEvent scope(flush_event);
+      vt::trace::TraceScopedEvent scope(flush_event_);
       outputTraces(
         log_file_.get(), traces_, start_time_, flush
       );
--- a/src/vt/trace/trace.cc
+++ b/src/vt/trace/trace.cc
@@ -142,6 +142,9 @@ void Trace::initialize() {
   theSched()->registerTrigger(
     sched::SchedulerEvent::EndIdle, traceEndIdleTrigger
   );
+
+  // Register a trace user event to demarcate flushes that occur
+  flush_event = vt::trace::registerEventCollective("trace_flush");
 }
 
 void Trace::loadAndBroadcastSpec() {
@@ -805,7 +808,7 @@ void Trace::cleanupTracesFile() {
   disableTracing();
 
   //--- Dump everything into an output file and close.
-  writeTracesFile(Z_FINISH);
+  writeTracesFile(Z_FINISH, false);
 
   assert(log_file_ && "Trace file must be open"); // opened in writeTracesFile
   outputFooter(log_file_.get(), node, start_time_);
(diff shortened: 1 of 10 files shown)


src/vt/trace/trace.cc
```diff
@@ -141,6 +142,9 @@ void Trace::initialize() {
   theSched()->registerTrigger(
     sched::SchedulerEvent::EndIdle, traceEndIdleTrigger
   );
+
+  // Register a trace user event to demarcate flushes that occur
+  flush_event = vt::trace::registerEventCollective("trace_flush");
```
pnstickne: `flush_event_`.

src/vt/vrt/collection/balance/lb_invoke/invoke.cc
```diff
@@ -242,7 +242,8 @@ void LBManager::releaseNow(PhaseType phase) {
 void LBManager::setTraceEnabledNextPhase(PhaseType phase) {
   // Set if tracing is enabled for this next phase. Do this immediately before
   // LB runs so LB is always instrumented as the beginning of the next phase
-# if backend_check_enabled(trace_enabled)
+#if backend_check_enabled(trace_enabled)
+  theTrace()->flushTracesFile(false);
   theTrace()->setTraceEnabledCurrentPhase(phase + 1);
```
pnstickne: Have flushTracesFile in setTraceEnabledCurrentPhase? Seems like internal detail.

src/vt/trace/trace.h
```diff
@@ -219,7 +219,7 @@ struct Trace {
     vt_gzFile* file, NodeType const node, double const start
   );
 
-  void writeTracesFile(int flush);
+  void writeTracesFile(int flush, bool is_incremental_flush);
```
pnstickne: `is_incremental` sufficient intent.

src/vt/trace/trace.cc
```diff
@@ -839,9 +836,25 @@ void Trace::writeTracesFile(int flush) {
...
+      vt_print(
+        trace,
+        "writeTracesFile: to_write={}, already_written={}\n",
+        traces_.size(), trace_write_count_
+      );
+    }
+
+    if (is_incremental_flush) {
+      vt::trace::TraceScopedEvent scope(flush_event);
+      outputTraces(
```
pnstickne: Duplicate feels odd. Maybe always have an event? (incremental or full).

src/vt/trace/trace.cc
```diff
@@ -755,13 +759,6 @@ TraceEventIDType Trace::logEvent(LogType&& log) {
   TraceEventIDType event = log.event;
   traces_.push(std::move(log));
 
-  // If auto-flush, can flush immediately.
```
pnstickne: Implies tracing only flushed in during phase shift.

Seems fine, but this might be the last usage of vt_trace_flush_size. If leaving in vt_trace_flush_size (defaults to 0/off, IIRC), should be able to leave such in as well.

src/vt/trace/trace.cc
```diff
@@ -653,8 +657,8 @@ void Trace::setTraceEnabledCurrentPhase(PhaseType cur_phase) {
 
       // Go ahead and perform a trace flush when tracing is disabled (and was
```
pnstickne: Not in this change, but I still feel something is off with the enable/disable transition above. At the very least the "..we are disabling traces.." is a misleading statement when tracing is disabled -> enabled.

I was thinking about the problem and I think it might be good to, when tracing is disabled, still allow end processing events. Then the stack is not event unwound and is guaranteed free from corruption.

Here is a case where corruptions might be able to occur:

     event = theTrace()->beginProcessing(..);
     setTraceEnabledCurrentPhase(); // manipulates stack here
     theTrace->endProcessing(event);

pnstickne: As far as trace flushing, perhaps on BeginIdle might be a good location to 'always flush'.

src/vt/trace/trace.cc
pnstickne: As far as trace flushing, perhaps on BeginIdle might be a good location to 'always flush'.

If that was a reasonable place, it might be possible to eliminate the flush_size setting entirely, eg.

src/vt/trace/trace.cc
pnstickne: `bool isTraceEnabled(eventType)` -> EndProcessing/EndIdle are always allowed if tracing is enabled via node/configuration.

src/vt/trace/trace.cc
lifflander: Unfortunately, BeginIdle is not a good location for flushing. For flushing to not impact performance, it must be synchronized across nodes. A short idle period will not cover the latency/bandwidth to disk.

src/vt/trace/trace.cc
lifflander: I agree the duplicate feels odd. However, I can't have an event when doing the final (non-incremental) flush because it will add an event after the trace gets finalized.

src/vt/trace/trace.cc
lifflander: Addressed.

src/vt/trace/trace.cc
lifflander: I sort of understand your point, but not completely. There isn't a bug because in the disable -> enable transition the `event` returned from `beginProcessing` is essentially nil. In the enable -> disable transition, `Trace::setTraceEnabledCurrentPhase` will actually forcibly unwind the stack and close all events at that point before it's disabled.

src/vt/vrt/collection/balance/lb_invoke/invoke.cc
lifflander: Good point. I've split this into a separate method. I acknowledge and a double flush might occur. Should be harmless.

src/vt/trace/trace.cc
pnstickne: Scenario:

- Given tracing is enabled.
- A begin processing event is written, and the processing tag returned. _This is done at the start of a nested scheduler loop in #703_, and from some other message processing locations.
- Tracing is disabled (via disable, or via phase) when the previous processing event has not been end'ed by the taker of the tag.
- The code unwinds the open events.
- The code that called begin processing calls end processing with the tag it acquired.

So if there is ever a way that this could _could_ run as part of a nested scheduler or as a message that (maybe in the future) added begin end processing tags, then kapow.

pnstickne: As far as the comment:

    if (trace_enabled_cur_phase_ != ret) {
        if (tracingWasEnabled) {
            // Close and pop everything, we are disabling traces at this point
            ..
         }

src/vt/trace/trace.cc
pnstickne: On another note, I was thinking it could spread cost and possibly be more efficient overall of writing the traces directly to a byte-buffer (snprintf, eg) and then in the flush simply sending the prewritten buffers to gzip.

- CPU time of printf is spread to trace sites
- No need to keep around traces_ structure (would keep around around a deque of char[1MB] or whatever instead) or copy into such
- Flush sites effectively limited to writing stream to gzip+IO.

pnstickne: Benchmarks like https://www.boost.org/doc/libs/1_55_0/libs/spirit/doc/html/spirit/karma/performance_measurements/numeric_performance/format_performance.html are all based around FP numbers with additional formatting specifiers.. :|

src/vt/trace/trace.cc
pnstickne: Can a TraceScopedEvent be a no-op?

      auto scopedEvent = is_incremental_flush
        ? vt::trace::TraceScopedEvent(flush_event_)
        : vt::trace::TraceScopedEvent(null);

src/vt/trace/trace.cc
lifflander: Ah, I get the problem now. This is a definitely a real problem interacting with #703

lifflander:
> I would check interactions with #703 which may inadvertently affect the open event unwinding. (I plan to do so later.)
> 
> Even when unwinding, I would ensure to only do so for the enabled -> disabled case (even if the stack is presumed to be empty going in the other direction) for clarity.
> 
> I can update #703 to ensure that close processing/idle events are allowed (trace statically enabled, dynamically disabled), which would remove the extra stack unwinding entirely with the very minor caveat that an open event, if any, wouldn't actually be written to the trace until the subsequent flush.

So will you address this in #703? Should I merge this?

pnstickne:
Changes are on #703 to beta.6 PR now. I need to git (get?!) those changes back on another branch to merge down..

src/vt/trace/trace.cc
lifflander: No, that is isn't supported. I could add that support

src/vt/trace/trace.cc
lifflander: I have an idea. Let's add an opt-in flag for flush at idle. I don't think in general we can tolerate that many writes to disk that often in performance-critical settings.

pnstickne:
Rebase onto current origin/develop and force pushed. Slight method reorderings.

lifflander:
@pnstickne This is ready to go once it is rebased?

PR #724, opened 2020-03-06
723 Add debugging epoch labels

### LINKED ISSUE #723: Add labels to epochs for debugging (e.g., output to DOT file)
**What Needs to be Done?**

Add strings to epoch construction to label epochs for debugging purposes.

lifflander:
Fixes #723 

Add a new optional parameter when creating a new epoch to specify a label that is dumped in debugging output (DOT files). In order to make this change correctly, a breaking change is made on non-default overloads of `TerminationDetector::makeEpochCollective` and `TerminationDetector::makeEpochRooted` due to char* to bool overload problems.

Thus, a tag class is used to specify/capture if the successor epoch should be the current epoch (or an manual override by the user) or whether a DS epoch should be used. This has the added benefit of making the interface more explicit and less error prone.

[outfile-global.pdf](https://github.com/DARMA-tasking/vt/files/4296367/outfile-global.pdf)

This has potential for beta.6, depending on feedback.

### DIFF
--- a/src/vt/messaging/collection_chain_set.h
+++ b/src/vt/messaging/collection_chain_set.h
@@ -121,7 +121,7 @@ class CollectionChainSet final {
 
       // The parameter `true` here tells VT to use an efficient rooted DS-epoch
       // by default. This can still be overridden by command-line flags
-      EpochType new_epoch = theTerm()->makeEpochRooted(true);
+      EpochType new_epoch = theTerm()->makeEpochRooted(term::UseDS{true});
       vt::theMsg()->pushEpoch(new_epoch);
 
       chain.add(new_epoch, step_action(idx));
--- a/src/vt/messaging/collection_chain_set.h
+++ b/src/vt/messaging/collection_chain_set.h
@@ -111,17 +111,18 @@ class CollectionChainSet final {
    * epoch for this step to contain/track completion of all the causally related
    * messages.
    *
+   * \param[in] label Label for the epoch created for debugging
    * \param[in] step_action The action to perform as a function that returns a
    * \c PendingSend
    */
-  void nextStep(std::function<PendingSend(Index)> step_action) {
+  void nextStep(std::string label, std::function<PendingSend(Index)> step_action) {
     for (auto &entry : chains_) {
       auto& idx = entry.first;
       auto& chain = entry.second;
 
       // The parameter `true` here tells VT to use an efficient rooted DS-epoch
       // by default. This can still be overridden by command-line flags
-      EpochType new_epoch = theTerm()->makeEpochRooted(term::UseDS{true});
+      EpochType new_epoch = theTerm()->makeEpochRooted(label, term::UseDS{true});
       vt::theMsg()->pushEpoch(new_epoch);
 
       chain.add(new_epoch, step_action(idx));
@@ -129,6 +130,17 @@ class CollectionChainSet final {
       vt::theMsg()->popEpoch(new_epoch);
       theTerm()->finishedEpoch(new_epoch);
     }
+  }
+
+  /**
+   * \brief The next step to execute on all the chains resident in this
+   * collection chain set
+   *
+   * \param[in] step_action The action to perform as a function that returns a
+   * \c PendingSend
+   */
+  void nextStep(std::function<PendingSend(Index)> step_action) {
+    return nextStep("", step_action);
   }
 
 #if 0
@@ -160,10 +172,13 @@ class CollectionChainSet final {
    * to track all the casually related messages and collectively wait for
    * termination of all of the recursive sends..
    *
+   * \param[in] label Label for the epoch created for debugging
    * \param[in] step_action the next step to execute, returning a \c PendingSend
    */
-  void nextStepCollective(std::function<PendingSend(Index)> step_action) {
-    auto epoch = theTerm()->makeEpochCollective();
+  void nextStepCollective(
+    std::string label, std::function<PendingSend(Index)> step_action
+  ) {
+    auto epoch = theTerm()->makeEpochCollective(label);
     vt::theMsg()->pushEpoch(epoch);
 
     for (auto &entry : chains_) {
@@ -174,6 +189,16 @@ class CollectionChainSet final {
 
     vt::theMsg()->popEpoch(epoch);
     theTerm()->finishedEpoch(epoch);
+  }
+
+  /**
+   * \brief The next collective step to execute across all resident elements
+   * across all nodes.
+   *
+   * \param[in] step_action the next step to execute, returning a \c PendingSend
+   */
+  void nextStepCollective(std::function<PendingSend(Index)> step_action) {
+    return nextStepCollective("", step_action);
   }
 
   /**
--- a/src/vt/messaging/collection_chain_set.h
+++ b/src/vt/messaging/collection_chain_set.h
@@ -115,7 +115,9 @@ class CollectionChainSet final {
    * \param[in] step_action The action to perform as a function that returns a
    * \c PendingSend
    */
-  void nextStep(std::string label, std::function<PendingSend(Index)> step_action) {
+  void nextStep(
+    std::string const& label, std::function<PendingSend(Index)> step_action
+  ) {
     for (auto &entry : chains_) {
       auto& idx = entry.first;
       auto& chain = entry.second;
@@ -176,7 +178,7 @@ class CollectionChainSet final {
    * \param[in] step_action the next step to execute, returning a \c PendingSend
    */
   void nextStepCollective(
-    std::string label, std::function<PendingSend(Index)> step_action
+    std::string const& label, std::function<PendingSend(Index)> step_action
   ) {
     auto epoch = theTerm()->makeEpochCollective(label);
     vt::theMsg()->pushEpoch(epoch);
--- a/src/vt/termination/epoch_label.h
+++ b/src/vt/termination/epoch_label.h
@@ -50,7 +50,6 @@
 namespace vt { namespace term {
 
 struct EpochLabel {
-  bool hasLabel() const;
   std::string getLabel() const;
   void setLabel(std::string const& label);
 
--- a/src/vt/termination/epoch_label.h
+++ b/src/vt/termination/epoch_label.h
@@ -2,7 +2,7 @@
 //@HEADER
 // *****************************************************************************
 //
-//                                term_scope.cc
+//                                epoch_label.h
 //                           DARMA Toolkit v. 1.0.0
 //                       DARMA/vt => Virtual Transport
 //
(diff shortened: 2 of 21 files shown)


src/vt/termination/graph/epoch_graph.cc
```diff
@@ -92,49 +92,64 @@ void EpochGraph::detectCycles() {
...
   if (epoch == term::any_epoch_sentinel) {
     return std::make_tuple(epoch, static_cast<NodeType>(-1), true, "Global");
   } else {
+    std::string label_format = "";
+    if (label != "") {
+      label_format = fmt::format(
+        "{}{}{}",
+        label != "" ? "\\\""  : "",
+        label != "" ? label : "",
+        label != "" ? "\\\"\\n"  : ""
```
PhilMiller: All of these ternaries are redundant with the enclosing `if`

lifflander: Addressed

tests/unit/termination/test_termination_action_common.impl.h
```diff
@@ -56,12 +56,12 @@ std::vector<vt::EpochType> generateEpochs(int nb, bool rooted, bool useDS) {
   if (rooted) {
     vtAssert(channel::node == channel::root, "Node should be root");
     // create rooted epoch sequence
-    sequence[0] = vt::theTerm()->makeEpochRooted(useDS);
+    sequence[0] = vt::theTerm()->makeEpochRooted(term::UseDS{true});
```
PhilMiller: Shouldn't these be `term::UseDS{useDS}`? You changed the previously conditional code to a constant.

src/vt/termination/epoch_tags.h
```diff
@@ -0,0 +1,72 @@
...
+
+#if !defined INCLUDED_VT_TERMINATION_EPOCH_TAGS_H
+#define INCLUDED_VT_TERMINATION_EPOCH_TAGS_H
+
+#include "vt/config.h"
+
+namespace vt { namespace term {
+
+struct SuccessorEpochCapture {
+  explicit SuccessorEpochCapture(EpochType epoch)
```
PhilMiller: Does this need to be `explicit`? In the places where it's used, if the caller passes an `EpochType` argument, we know exactly what they're asking for.

Are you worried about integer conversions? In the long run, we should change the definition of `EpochType` from `using EpochType = uint64_t;` or whatever to `struct EpochType { uint64_t e; };` so that confusion becomes impossible

src/vt/messaging/collection_chain_set.h
```diff
@@ -111,17 +111,18 @@ class CollectionChainSet final {
    * epoch for this step to contain/track completion of all the causally related
    * messages.
    *
+   * \param[in] label Label for the epoch created for debugging
    * \param[in] step_action The action to perform as a function that returns a
    * \c PendingSend
    */
-  void nextStep(std::function<PendingSend(Index)> step_action) {
+  void nextStep(std::string label, std::function<PendingSend(Index)> step_action) {
```
pnstickne: Use `std::string const&` to avoid extra copies.

src/vt/messaging/collection_chain_set.h
```diff
@@ -160,10 +172,13 @@ class CollectionChainSet final {
    * to track all the casually related messages and collectively wait for
    * termination of all of the recursive sends..
    *
+   * \param[in] label Label for the epoch created for debugging
    * \param[in] step_action the next step to execute, returning a \c PendingSend
    */
-  void nextStepCollective(std::function<PendingSend(Index)> step_action) {
-    auto epoch = theTerm()->makeEpochCollective();
+  void nextStepCollective(
+    std::string label, std::function<PendingSend(Index)> step_action
```
pnstickne: Ditto.

src/vt/termination/epoch_tags.cc
```diff
@@ -0,0 +1,60 @@
...
+
+#include "vt/config.h"
+#include "vt/termination/epoch_tags.h"
+#include "vt/messaging/active.h"
+
+namespace vt { namespace term {
+
+SuccessorEpochCapture::SuccessorEpochCapture()
+  : epoch_(
+      (theMsg()->getEpoch() != no_epoch and
```
pnstickne: Consider TU-local method accepting `theMsg()->getEpoch() -> maybe_epoch`.

src/vt/termination/epoch_tags.h
```diff
@@ -0,0 +1,72 @@
...
+    : epoch_(epoch)
+  { }
+  SuccessorEpochCapture();
+  operator EpochType() const { return epoch_; }
+  bool valid() const { return epoch_ != no_epoch; }
+  EpochType epoch_ = no_epoch;
+};
+
+struct UseDS {
+  explicit UseDS(bool use_it)
```
pnstickne: `use_ds_`?

Regarless, `UseDS{true/false}` feels a bit odd as it's a singular binary - precluding a larger (closed) set: `UseDS`, `NeverUseDS`, `UseSomethingElse`.

src/vt/termination/epoch_tags.h
```diff
@@ -0,0 +1,72 @@
...
+namespace vt { namespace term {
+
+struct SuccessorEpochCapture {
+  explicit SuccessorEpochCapture(EpochType epoch)
+    : epoch_(epoch)
+  { }
+  SuccessorEpochCapture();
+  operator EpochType() const { return epoch_; }
+  bool valid() const { return epoch_ != no_epoch; }
+  EpochType epoch_ = no_epoch;
```
pnstickne: `epoch_` can probably be private as it's exposed via EpochType().

src/vt/termination/graph/epoch_graph.cc
```diff
@@ -92,49 +92,59 @@ void EpochGraph::detectCycles() {
...
-EpochGraph::EpFormat EpochGraph::formatDOTEpoch(EpochType epoch) {
+EpochGraph::EpFormat EpochGraph::formatDOTEpoch(
+  EpochType epoch, std::string label
+) {
   if (epoch == term::any_epoch_sentinel) {
     return std::make_tuple(epoch, static_cast<NodeType>(-1), true, "Global");
   } else {
+    std::string label_format = "";
+    if (label != "") {
+      label_format = fmt::format("{}{}{}", "\\\"", label, "\\\"\\n");
```
pnstickne: `\"My Label\"`? Or is this really applied as a format specifier itself?

src/vt/termination/graph/epoch_graph.cc
```diff
@@ -92,49 +92,59 @@ void EpochGraph::detectCycles() {
...
 }
 
 std::string EpochGraph::outputDOT(bool verbose) {
   std::unordered_map<EpochType, EpFormat> eps;
-  std::set<std::tuple<EpochType, EpochType>> links;
+  std::set<std::tuple<EpochType, std::string, EpochType, std::string>> links;
   std::string builder = "";
   outputImpl(links);
 
   for (auto&& elm : links) {
```
pnstickne: I would find it more clear with the tuple values assigned to temporaries.

pnstickne: auto epoch = std::get<0>(elm);
    auto& epoch_label = std::get<1>(elm);
    auto succ_epoch = std::get<2>(elm);
    auto& succ_epoch_label = std::get<3>(elm);

src/vt/termination/epoch_label.h
```diff
@@ -0,0 +1,63 @@
...
+#if !defined INCLUDED_VT_TERMINATION_EPOCH_LABEL_H
+#define INCLUDED_VT_TERMINATION_EPOCH_LABEL_H
+
+#include "vt/config.h"
+
+namespace vt { namespace term {
+
+struct EpochLabel {
+  bool hasLabel() const;
+  std::string getLabel() const;
```
pnstickne: Since the label is checked for `op ""` elsewhere, `hasLabel()` might not add to much value.

src/vt/termination/term_scope.cc
```diff
@@ -75,7 +75,7 @@ namespace vt { namespace term {
   bool small, ActionType closure, ActionType action
 ) {
   bool const use_dijkstra_scholten = small == true;
-  auto const epoch = theTerm()->makeEpochRooted(use_dijkstra_scholten);
+  auto const epoch = theTerm()->makeEpochRooted(UseDS{use_dijkstra_scholten});
```
pnstickne: Throwing out stuff:

     makeEpochRooted(use_dijkstra_scholten ? EpochAlg.DS : EpoechAlg.Default)
     makeEpochRooted(use_dijkstra_scholten ? EpochAlg.DS : EpoechAlg.NameOfTheOtherOne)

     makeEpochRooted(use_dijkstra_scholten ? EpochDS.Force : EpochDS.Default)

src/vt/termination/termination.h
```diff
@@ -135,15 +136,78 @@ struct TerminationDetector :
...
+   *
+   * \param[in] label epoch label for debugging purposes
+   * \param[in] use_ds whether to use the Dijkstra-Scholten algorithm
+   * \param[in] successor successor epoch that waits for this new epoch
+   *
+   * \return the new epoch
    */
   EpochType makeEpochRooted(
-    bool useDS = false, bool child = true, EpochType parent = no_epoch
+    std::string label,
```
pnstickne: Ditto const-ref.

tests/unit/termination/test_termination_action_common.impl.h
lifflander: Yes, that is a bug.

src/vt/messaging/collection_chain_set.h
lifflander: I'm started with `const&` but I'm not sure if I'm allowed to pass a literal if the argument is by const&. The primary use case will be a char* literal, which will be upgraded to a std::string.

src/vt/termination/epoch_tags.cc
lifflander: Not sure what this means.

src/vt/termination/epoch_tags.h
lifflander: It is fairly binary. Either you want it to use the DS algorithm or not. The default is based on how UseDS is constructed when the parameter is left to the default value.

Forcing one or the other is only done typically for testing. There is possibly another level of must have DS or must have Wave. But, in that case, you call down to the underlying function `makeEpochRooted{Wave,DS}`.

src/vt/termination/epoch_tags.h
lifflander: Will do.

src/vt/termination/graph/epoch_graph.cc
lifflander: Yes, it's applied to a sort of format specifier (not exactly, but sort of) itself, which is a string going into an underlying file which is parsed by Graphviz. This is valid code that is working.

I don't fully understand what you think is wrong with it.

src/vt/termination/termination.h
lifflander: Again, I'm worried about inline usage, and thus removed all const& from those sites. Do you know if it's correct to specify inline with const&?

src/vt/termination/epoch_tags.h
lifflander: Yes, I think it needs to be explicit. At least, `UseDS` must be explicit. I might be able to make this one automatic, but it would lead to non-uniformity across the interface.

I agree that strong types for VT's core types would be nice. `NodeType`, `EpochType`, etc. However, it would lead to a lot of code needing to change. Something to consider in 1.1.

src/vt/termination/epoch_label.h
lifflander: Removed

src/vt/termination/graph/epoch_graph.cc
lifflander: Done

src/vt/messaging/collection_chain_set.h
lifflander: From what I read, it is actually legal to pass a default with const&. Updating code

PR #727, opened 2020-03-07
726 Trace memory usage

### LINKED ISSUE #726: Trace memory usage
**What Needs to be Done?**

Opt-in for outputting memory usage traces.

lifflander:
There isn't a good general way to get memory usage across all platforms. Thus, I've added a host of methods for reporting memory usage that are protected by cmake system checks before they get enabled. The user can specify a precedence order for the memory reporters with the flag `--vt_memory_reporters`. The default order is `mstats,machinfo,selfstat,sbrk,mallinfo,getrusage,ps`.

Additionally, there is a flag for enabling memory usage prints after each phase 
`--vt_print_memory_each_phase` along with which node the user wants to be informed about 
`--vt_print_memory_node`. Those phase prints look something like this:

Screen Shot 2020-03-06 at 7 49 57 PM

With all this in place, I now have a new option to trace memory usage also, enabled with the flag: 
`--vt_trace_memory_usage`. The highest priority working memory reporter return value gets output to the trace files every begin and end processing. We eventually might want to reduce the frequency, but this was how we were using it in the EMPIRE hacks.

At startup, VT give some info about the reporters and precedence order requested and what was found to be working on the given architecture, along with initial values for memory usage at startup:

Screen Shot 2020-03-06 at 7 52 49 PM

Fixes #726

### DIFF
--- a/src/vt/trace/trace.h
+++ b/src/vt/trace/trace.h
@@ -161,6 +161,11 @@ struct Trace {
     TraceEventIDType const event = no_trace_event
   );
 
+  void addMemoryEvent(
+    std::size_t memory,
+    double const time = getCurrentTime()
+  );
+
   TraceEventIDType messageCreation(
     TraceEntryIDType const ep, TraceMsgLenType const len,
     double const time = getCurrentTime()
@@ -230,6 +235,10 @@ private:
   /// or no_trace_event if not accepted (eg. no tracing on node).
   /// The log object is invalidated after the call.
   TraceEventIDType logEvent(LogType&& log);
+
+  std::size_t getTracesSize() const {
+    return traces_.size() * sizeof(Log);
+  }
 
 private:
   /*
--- a/src/vt/trace/trace_log.h
+++ b/src/vt/trace/trace_log.h
@@ -293,6 +293,14 @@ struct Log final {
   {
   }
 
+  // Memory usage event, could create another union type here, but hardly seems
+  // worth it; re-use SysData
+  Log(
+    double in_time, TraceConstantsType in_type, std::size_t in_memory_bytes
+  ) : time(in_time), type(in_type),
+      data(Data::SysData{in_memory_bytes})
+  { }
+
   inline Data::UserData const& user_data() const {
     assert(data.user.data_type == LogDataType::user && "Expecting user data-type");
     return data.user;
--- a/src/vt/utils/memory/memory_units.cc
+++ b/src/vt/utils/memory/memory_units.cc
@@ -45,6 +45,8 @@
 #include "vt/config.h"
 #include "vt/utils/memory/memory_units.h"
 
+#include <unordered_map>
+
 namespace vt { namespace util { namespace memory {
 
 std::unordered_map<MemoryUnitEnum, std::string> memory_unit_names = {
@@ -53,5 +55,9 @@ std::unordered_map<MemoryUnitEnum, std::string> memory_unit_names = {
   {MemoryUnitEnum::Megabytes, std::string{"MB"}},
   {MemoryUnitEnum::Gigabytes, std::string{"GB"}}
 };
+
+std::string getMemoryUnitName(MemoryUnitEnum unit) {
+  return memory_unit_names[unit];
+}
 
 }}} /* end namespace vt::util::memory */
--- a/src/vt/utils/memory/memory_units.cc
+++ b/src/vt/utils/memory/memory_units.cc
@@ -51,9 +51,9 @@ namespace vt { namespace util { namespace memory {
 
 std::unordered_map<MemoryUnitEnum, std::string> memory_unit_names = {
   {MemoryUnitEnum::Bytes,     std::string{"bytes"}},
-  {MemoryUnitEnum::Kilobytes, std::string{"kB"}},
-  {MemoryUnitEnum::Megabytes, std::string{"MB"}},
-  {MemoryUnitEnum::Gigabytes, std::string{"GB"}}
+  {MemoryUnitEnum::Kilobytes, std::string{"KiB"}},
+  {MemoryUnitEnum::Megabytes, std::string{"MiB"}},
+  {MemoryUnitEnum::Gigabytes, std::string{"GiB"}}
 };
 
 std::string getMemoryUnitName(MemoryUnitEnum unit) {
--- a/src/vt/utils/memory/memory_units.cc
+++ b/src/vt/utils/memory/memory_units.cc
@@ -1,0 +1,57 @@
+/*
+//@HEADER
+// *****************************************************************************
+//
+//                               memory_units.cc
+//                           DARMA Toolkit v. 1.0.0
+//                       DARMA/vt => Virtual Transport
+//
+// Copyright 2019 National Technology & Engineering Solutions of Sandia, LLC
+// (NTESS). Under the terms of Contract DE-NA0003525 with NTESS, the U.S.
+// Government retains certain rights in this software.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+// * Redistributions of source code must retain the above copyright notice,
+//   this list of conditions and the following disclaimer.
+//
+// * Redistributions in binary form must reproduce the above copyright notice,
+//   this list of conditions and the following disclaimer in the documentation
+//   and/or other materials provided with the distribution.
+//
+// * Neither the name of the copyright holder nor the names of its
+//   contributors may be used to endorse or promote products derived from this
+//   software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+//
+// Questions? Contact darma@sandia.gov
+//
+// *****************************************************************************
+//@HEADER
+*/
+
+#include "vt/config.h"
+#include "vt/utils/memory/memory_units.h"
+
+namespace vt { namespace util { namespace memory {
+
+std::unordered_map<MemoryUnitEnum, std::string> memory_unit_names = {
+  {MemoryUnitEnum::Bytes,     std::string{"bytes"}},
+  {MemoryUnitEnum::Kilobytes, std::string{"kB"}},
+  {MemoryUnitEnum::Megabytes, std::string{"MB"}},
+  {MemoryUnitEnum::Gigabytes, std::string{"GB"}}
+};
+
+}}} /* end namespace vt::util::memory */
--- a/src/vt/utils/memory/memory_units.h
+++ b/src/vt/utils/memory/memory_units.h
@@ -48,7 +48,6 @@
 #include "vt/config.h"
 
 #include <string>
-#include <unordered_map>
 
 namespace vt { namespace util { namespace memory {
 
@@ -59,7 +58,7 @@ enum struct MemoryUnitEnum : int8_t {
   Gigabytes = 3
 };
 
-extern std::unordered_map<MemoryUnitEnum, std::string> memory_unit_names;
+std::string getMemoryUnitName(MemoryUnitEnum unit);
 
 }}} /* end namespace vt::util::memory */
 
--- a/src/vt/utils/memory/memory_units.h
+++ b/src/vt/utils/memory/memory_units.h
@@ -1,0 +1,66 @@
+/*
+//@HEADER
+// *****************************************************************************
+//
+//                                memory_units.h
+//                           DARMA Toolkit v. 1.0.0
+//                       DARMA/vt => Virtual Transport
+//
+// Copyright 2019 National Technology & Engineering Solutions of Sandia, LLC
+// (NTESS). Under the terms of Contract DE-NA0003525 with NTESS, the U.S.
+// Government retains certain rights in this software.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+// * Redistributions of source code must retain the above copyright notice,
+//   this list of conditions and the following disclaimer.
+//
+// * Redistributions in binary form must reproduce the above copyright notice,
+//   this list of conditions and the following disclaimer in the documentation
+//   and/or other materials provided with the distribution.
+//
+// * Neither the name of the copyright holder nor the names of its
+//   contributors may be used to endorse or promote products derived from this
+//   software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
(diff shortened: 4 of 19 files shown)


src/vt/utils/memory/memory_usage.cc
```diff
@@ -0,0 +1,355 @@
...
+
+std::string Mallinfo::getName() {
+  return "mallinfo";
+}
+
+std::size_t Getrusage::getUsage() {
+# if defined(vt_has_getrusage)
+    struct rusage usage;
+    getrusage(RUSAGE_SELF, &usage);
+    return usage.ru_maxrss;
```
PhilMiller: If [what I'm reading](https://unix.stackexchange.com/questions/30940/getrusage-system-call-what-is-maximum-resident-set-size) is correct, that's a high-water mark, where others are all ostensibly current value. That distinction could be misleading in traces

lifflander: I noticed this also. Is there a more accurate call with rusage? If not, should we decrease the precedence of this by default?

src/vt/utils/memory/memory_usage.cc
```diff
@@ -0,0 +1,355 @@
...
+      return 0;
+    } else {
+      return static_cast<std::size_t>(info.resident_size);
+    }
+# else
+    return 0;
+# endif
+}
+
+  std::string MachTaskInfo::getName() {
```
lifflander: Indentation is broken here.

src/vt/utils/memory/memory_usage.cc
```diff
@@ -0,0 +1,355 @@
...
+}
+
+std::string Sbrk::getName() {
+  return "sbrk";
+}
+
+std::size_t PS::getUsage() {
+# if defined(vt_has_popen) && defined(vt_has_pclose) && defined(vt_has_getpid)
+    auto cmd = fmt::format("/bin/ps -o rss= -p {}", getpid());
+    FILE* p = popen(cmd.c_str(), "r");
```
PhilMiller: Given the intended usage, the prospect of forking and executing a shell to call this for potentially *every single handler traced* is pretty out-there. I can potentially see this actually causing some havoc on compute nodes, and getting us booted off systems.

It's fine to have this method available, but it definitely needs to not be enabled by default, or used in some super high-frequency call path.

lifflander: Good point, PS has the lower precedence of all the utilities, for it to be used for tracing all other would have to fail. What do you suggest? Should we add a new frequency flag to the trace memory?

src/vt/utils/memory/memory_usage.cc
PhilMiller: I don't think there's a more accurate call, and a quick search didn't see any means of resetting it, either.

src/vt/utils/memory/memory_usage.cc
lifflander: This is the default precedence order from highest to lowest: `mstats,machinfo,selfstat,sbrk,mallinfo,getrusage,ps`.

src/vt/utils/memory/memory_usage.cc
PhilMiller: > `long int ru_idrss`

>    An integral value expressed the same way, which is the amount of unshared memory used for data.

Might be closer to what we want.

src/vt/utils/memory/memory_usage.cc
lifflander: All the working memory reporters are queried at phase boundary if the flag
 `--vt_print_memory_each_phase` is enabled (disabled by default)

PhilMiller: Just to be super-cautious, I'd add an `enabled` flag to the `PS` class, defaulting to `false`, that would have it fail if unset, and some separate means of setting it.

lifflander: Good idea, I will make PS an explicit opt-in.

Currently, another bad case would be if the user set these as so (each phase every node would query PS right now):
 `--vt_print_memory_each_phase=true`
`--vt_print_memory_node=all`

src/vt/utils/memory/memory_usage.cc
lifflander: Fixed.

src/vt/utils/memory/memory_usage.cc
lifflander: Added new option `--vt_allow_memory_report_with_ps` with a warning in the --help  about each argument

src/vt/utils/memory/memory_usage.cc
lifflander: On my Mac, `ru_idrss` returns 0.

lifflander: Also, I'm noticing the docs say that these are unmaintained:
```
       ru_ixrss (unmaintained)
              This field is currently unused on Linux.

       ru_idrss (unmaintained)
              This field is currently unused on Linux.

       ru_isrss (unmaintained)
              This field is currently unused on Linux.
```

lifflander:
Just confirmed that this builds on Mutrino. We probably want to check this across multiple platforms to make sure I didn't screw up any of the memory checks or cmake checks.

lifflander:
Mutrino output on 4 nodes for lb_iter. I'm suspecting that the units are off!

```
vt: 	Working memory reporters: selfstat,mallinfo,getrusage,ps
vt: 	Initial memory usage: selfstat=106.18 mallinfo=1.75922e+13 getrusage=0.00774384 ps=3.57031 (MB)
vt: 	Option: flag --vt_print_memory_each_phase on: Printing memory usage each phase
vt: 	Option: flag --vt_print_memory_node on: Printing memory usage from node: 0
lb_iter: elms=64, weight=1.0, num_iter=8
0: iterWork: idx=idx(0)
iteration: iter=0,time=0.04848313331604004
vt: [0] general: Memory Usage: phase=0: selfstat=108.18 mallinfo=1.75922e+13 getrusage=0.00774384 ps=4.32422 (MB)
vt: [0] lb: LBManaager::releaseNow: finished LB, phase=0, invocations=0
0: iterWork: idx=idx(0)
iteration: iter=1,time=0.01743912696838379
vt: [0] general: Memory Usage: phase=1: selfstat=108.18 mallinfo=1.75922e+13 getrusage=0.00774384 ps=4.38672 (MB)
vt: [0] lb: LBManaager::releaseNow: finished LB, phase=1, invocations=0
0: iterWork: idx=idx(0)
```

src/vt/trace/trace_log.h
```diff
@@ -293,6 +293,14 @@ struct Log final {
   {
   }
 
+  // Memory usage event, could create another union type here, but hardly seems
```
pnstickne: Feels like system data.

src/vt/utils/memory/memory_units.h
```diff
@@ -0,0 +1,66 @@
...
+namespace vt { namespace util { namespace memory {
+
+enum struct MemoryUnitEnum : int8_t {
+  Bytes     = 0,
+  Kilobytes = 1,
+  Megabytes = 2,
+  Gigabytes = 3
+};
+
+extern std::unordered_map<MemoryUnitEnum, std::string> memory_unit_names;
```
pnstickne: Might also just expose operation, not data.

src/vt/trace/trace.h
```diff
@@ -231,6 +236,10 @@ struct Trace {
   /// The log object is invalidated after the call.
   TraceEventIDType logEvent(LogType&& log);
 
+  std::size_t getTracesSize() const {
```
lifflander: This is not used; got pulled in from a cherry-pick. Question: should I subtract this off? We were using this to find memory leaks...but it doesn't make sense as an absolute value.

src/vt/utils/memory/memory_usage.cc
```diff
@@ -0,0 +1,365 @@
...
+    return 0;
+# endif
+}
+
+std::string Mstats::getName() {
+  return "mstats";
+}
+
+std::size_t Sbrk::getUsage() {
+# if defined(vt_has_sbrk)
```
pnstickne: Differences between '#if' formatting.

src/vt/utils/memory/memory_usage.cc
```diff
@@ -0,0 +1,365 @@
...
+# endif
+}
+
+std::string Sbrk::getName() {
+  return "sbrk";
+}
+
+std::size_t PS::getUsage() {
+# if defined(vt_has_popen) && defined(vt_has_pclose) && defined(vt_has_getpid)
+    if (arguments::ArgConfig::vt_allow_memory_report_with_ps) {
```
pnstickne: I really want to tackle all these sets-as-single-options D:

      --vt_report_memory_strategy=sbrk,ps,..

src/vt/utils/memory/memory_usage.cc
```diff
@@ -0,0 +1,365 @@
...
+  for (auto&& r : all_reporters) {
+    name_map[r->getName()] = cur;
+    cur++;
+  }
+
+  for (auto&& r : results) {
+    auto iter = name_map.find(r);
+    if (iter != name_map.end()) {
+      reporters_.emplace_back(std::move(all_reporters[iter->second]));
+    }
```
lifflander: Might want to error check unmatching names here. Probably a bug in user input if they specify a memory reporter that doesn't exist!

src/vt/utils/memory/memory_usage.cc
lifflander: It's useful to report from multiple sources at phase boundaries. Because some may be more accurate than others (having that issue currently in EMPIRE).

PS is special because it's potentially not scalable to do a popen/pclose often on a supercomputer. Thus, it has a special case to enable it.

src/vt/utils/memory/memory_usage.h
```diff
@@ -0,0 +1,132 @@
...
+  std::size_t getUsage() override;
+  std::string getName() override;
+};
+
+struct Stat : Reporter {
+  std::size_t getUsage() override;
+  std::string getName() override;
+
+private:
+  bool failed = false;
```
pnstickne: `failed` to base Reporter? Also, perhaps `failed_`.

src/vt/utils/memory/memory_usage.h
```diff
@@ -0,0 +1,132 @@
...
+  static void initialize();
+
+  static void finalize();
+
+private:
+  static std::unique_ptr<MemoryUsage> impl_;
+
+  std::vector<std::unique_ptr<Reporter>> reporters_;
+  int first_valid_reporter_ = -1;
+  bool has_working_ = false;
```
pnstickne: `has_working_` implied by `not reports_.empty()`.

src/vt/utils/memory/memory_usage.cc
lifflander: The suggested format for VT is indenting after the pound which is standard compliant.

src/vt/utils/memory/memory_usage.cc
lifflander: It is a single option. Default value for `--vt_memory_reporters` is `mstats,machinfo,selfstat,sbrk,mallinfo,getrusage,ps`

src/vt/utils/memory/memory_usage.h
lifflander: No, `has_working_` implies three things: `impl_ != nullptr`, reporters are not empty, and a reporter returned non-zero.

lifflander: Actually, nevermind. I see what you're saying. I probably could remove it

lifflander:
> Mutrino output on 4 nodes for lb_iter. I'm suspecting that the units are off!
> 

Just checked---units are correct. For some reason, mallinfo is just returning a seemingly bogus value.

src/vt/utils/memory/memory_usage.cc
```diff
@@ -0,0 +1,365 @@
...
+        pretty ? debug::reset() : "",
+        pretty ? debug::magenta() : "",
+        dusage,
+        pretty ? debug::reset() : ""
+      );
+      num_valid++;
+    }
+  }
+  if (num_valid > 0) {
+    builder += fmt::format("({})", memory_unit_names[unit]);
```
pnstickne: eg. `getMemoryUnitName(unit)`.

src/vt/utils/memory/memory_usage.cc
```diff
@@ -0,0 +1,365 @@
...
+  }
+  return builder;
+}
+
+std::vector<std::string> MemoryUsage::getWorkingReporters() {
+  std::vector<std::string> working;
```
pnstickne: If `reporters_` are guarnateed to only be working set at initialization..

lifflander: There's a slight issue that a reporter might start "working" (reporting non-zero) later. For example, the first time that sbrk runs, it might look like it's not working until a later call. This is actually an issue in current code also.

src/vt/utils/memory/memory_usage.cc
lifflander: Done

src/vt/utils/memory/memory_units.h
lifflander: Fixed

src/vt/vrt/collection/balance/lb_invoke/invoke.cc
```diff
@@ -268,4 +269,22 @@ void LBManager::unregisterListenerAfterLB(int element) {
   listeners_[element] = nullptr;
 }
 
+void LBManager::printMemoryUsage(PhaseType phase) {
+  if (arguments::ArgConfig::vt_print_memory_each_phase) {
+    auto this_node = theContext()->getNode();
+    if (
+      "all" == arguments::ArgConfig::vt_print_memory_node or
+      std::to_string(this_node) == arguments::ArgConfig::vt_print_memory_node
```
pnstickne: to_string feels odd here. Not necessarily this commit, but it would be nice to separate parameters and parameter-derived state.

src/vt/utils/memory/memory_usage.cc
```diff
@@ -0,0 +1,365 @@
...
+    }
+  }
+}
+
+std::size_t MemoryUsage::getAverageUsage() {
+  std::size_t usage = 0;
+  std::size_t num_valid = 0;
+  for (auto&& r : reporters_) {
+    auto rusage = r->getUsage();
+    if (rusage != 0) {
```
pnstickne: No check needed if `valid_reporters_` 🗡

src/vt/utils/memory/memory_usage.cc
lifflander: Done.

src/vt/utils/memory/memory_units.cc
```diff
@@ -0,0 +1,57 @@
...
+*/
+
+#include "vt/config.h"
+#include "vt/utils/memory/memory_units.h"
+
+namespace vt { namespace util { namespace memory {
+
+std::unordered_map<MemoryUnitEnum, std::string> memory_unit_names = {
+  {MemoryUnitEnum::Bytes,     std::string{"bytes"}},
+  {MemoryUnitEnum::Kilobytes, std::string{"kB"}},
```
pnstickne: Are these 1000/1024 or undetermined? If guaranteed to be 1024, I prefer "MiB", etc.

src/vt/utils/memory/memory_usage.h
lifflander: Made the private ones have an underscore

src/vt/utils/memory/memory_units.cc
lifflander: I fetch everything in bytes (except for PS which I convert from Kilobytes to bytes first). I multiply by 1024 there. Not sure if that's correct

src/vt/trace/trace.h
lifflander: I'm going to leave this in for now, since it might be useful and doesn't hurt.

src/vt/utils/memory/memory_units.cc
lifflander: Fixed

src/vt/vrt/collection/balance/lb_invoke/invoke.cc
lifflander: Agreed on this point. We need to refactor/create a new interface to parsed argument values.

src/vt/utils/memory/memory_usage.cc
lifflander: Simplified it out.

lifflander:
Just a future note. One SUN/AIX/Solaris we might want to implement this overload:

```C++
#elif (defined(_AIX) || defined(__TOS__AIX__)) || (defined(__sun__) || defined(__sun) || defined(sun) && (defined(__SVR4) || defined(__svr4__)))
    /* AIX and Solaris ------------------------------------------ */
    struct psinfo psinfo;
    int fd = -1;
    if ( (fd = open( "/proc/self/psinfo", O_RDONLY )) == -1 )
        return (size_t)0L;      /* Can't open? */
    if ( read( fd, &psinfo, sizeof(psinfo) ) != sizeof(psinfo) )
    {
        close( fd );
        return (size_t)0L;      /* Can't read? */
    }
    close( fd );
    return (size_t)(psinfo.pr_rssize * 1024L);
```

lifflander:
Here's the latest code on Mutrino:

```
vt: 	Working memory reporters: selfstat,selfstatm,mallinfo,getrusage
vt: 	Initial memory usage: selfstat=106.301 selfstatm=3.53906 mallinfo=1.75922e+13 getrusage=7.92578 (MiB)
vt: 	Option: flag --vt_print_memory_each_phase on: Printing memory usage each phase
vt: 	Option: flag --vt_print_memory_node on: Printing memory usage from node: all
lb_iter: elms=64, weight=1.0, num_iter=8
0: iterWork: idx=idx(0)
iteration: iter=0,time=0.044148921966552734
vt: [0] general: Memory Usage: phase=0: selfstat=108.301 selfstatm=4.30859 mallinfo=1.75922e+13 getrusage=7.92578 (MiB)
```

Everything looks good except for mallinfo!

lifflander:
The latest changes make it report correctly on Mutrino:

```
vt: 	Initial memory usage: selfstat=106.309 selfstatm=25.5391 mallinfo=3063.29 getrusage=7.92578 ps=106.309 (MiB)
```

3GB is correct due to the memory pool.

lifflander:
I've added a new memory checker. mallinfo is the most accurate on Mutrino. However, it unfortunately does not work over 4GB of memory because the fields are defined as int!

https://bugzilla.redhat.com/show_bug.cgi?id=173813

pnstickne:
> I've added a new memory checker. mallinfo is the most accurate on Mutrino. However, it unfortunately does not work over 4GB of memory because the fields are defined as int!
> 
> https://bugzilla.redhat.com/show_bug.cgi?id=173813

I would expect PS RSS (same as selfstat here) to be accurate in terms of actual usage, not reserved.

PR #739, opened 2020-03-11
738 Print memory usage

### LINKED ISSUE #738: Print memory usage at at threshold increment from scheduler
**What Needs to be Done?**

Add some opt-in informational prints about memory usage to track when it starts to blow up in EMPIRE.

lifflander:
Fixes #738 

In EMPIRE, @nlslatt and I realized that it would help to be able to see when memory usage crosses some threshold increment to track if memory usage is blowing up.

Add three new opt-in arguments:
* `--vt_print_memory_at_threshold` : enable prints at a memory threshold
* `--vt_print_memory_sched_poll=100` : check memory usage every 100 scheduler polls (certain memory reporters can be expensive so this gives control to the user)
* `--vt_print_memory_threshold="1.2 GiB"` : whenever the first memory usage reporter crosses an increment of 1.2 GiB (up or down), VT will print to screen the current usage

Prints look like this (run for ever 1 GiB):
```
vt: [0] general: Memory usage (+) mstats=1028.92 machinfo=1047.39 getrusage=1047.39 (MiB)
vt: [0] general: Memory usage (-) mstats=900.919 machinfo=2071.4 getrusage=2071.4 (MiB)
```

### DIFF
--- a/src/vt/configs/arguments/args.cc
+++ b/src/vt/configs/arguments/args.cc
@@ -71,6 +71,9 @@ namespace vt { namespace arguments {
 /*static*/ bool        ArgConfig::vt_print_memory_each_phase = false;
 /*static*/ std::string ArgConfig::vt_print_memory_node  = "0";
 /*static*/ bool        ArgConfig::vt_allow_memory_report_with_ps = false;
+/*static*/ bool        ArgConfig::vt_print_memory_at_threshold = false;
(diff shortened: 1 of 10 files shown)


nlslatt:
But remember that `--vt_print_memory_threshold="1.2 GiB"` is not accepted through EMPIRE due to command-line parsing issues.

PR #745, opened 2020-03-26
744 Fix location bug

### LINKED ISSUE #744: After LB, with a large # msgs, location manager fails to deliver
**Describe the bug**
Location cache entries are allowed to expire in the cache. However, for elements off-node, the home entry should never expire; otherwise, messages may never arrive at their destinations since the home doesn't know how to deliver them.

This was showing up in EMPIRE as a hang, either detected by VT TD or not.

Minimal reproducer:

```C++
#include "vt/transport.h"

using namespace ::vt;
using namespace ::vt::collective;
using namespace ::vt::mapping;

static constexpr int const default_num_elms = 64;
static int num_elms = default_num_elms;

struct TestColl : Collection<TestColl,vt::Index2D> {
  TestColl() = default;

  struct TestMsg : CollectionMessage<TestColl> { };

  struct NeighborMsg : vt::CollectionMessage<TestColl> {
    NeighborMsg() = default;
    explicit NeighborMsg(vt::Index2D in_idx)
      : idx_(in_idx)
    { }
    vt::Index2D idx_;
  };

  int wrap(int dim, int maxd) {
    while (dim < 0) {
      dim += maxd;
    }
    if (dim >= maxd) {
      return dim % maxd;
    }
    return dim;
  }

  void doWork(TestMsg* msg) {
    auto proxy = this->getCollectionProxy();
    auto idx = this->getIndex();
    auto x = idx.x();
    auto y = idx.y();
    auto num_nodes = vt::theContext()->getNumNodes();

    for (int x1 = x - 35; x1 < x + 35; x1++) {
      for (int y1 = y - 35; y1 < y + 35; y1++) {
        proxy(wrap(x1,num_nodes),wrap(y1,num_elms)).template send<
          NeighborMsg,&TestColl::informNeighbor
        >(idx);
      }
    }
  }

  void informNeighbor(NeighborMsg* m) {
    auto proxy = this->getCollectionProxy();
    auto sidx = m->idx_;
    proxy(sidx).template send<TestMsg,&TestColl::getBack>();
  }

  void getBack(TestMsg* m) {

  }

  void dolb(TestMsg* m) {
    auto idx = this->getIndex();
    auto proxy = getCollectionProxy();
    proxy[idx].LB<TestMsg,&TestColl::cont>();
  }

  void cont(TestMsg* m) {

  }

    Collection<TestColl,Index2D>::serialize(s);
    s | vec;
  }

  std::vector<double> vec;
};

  CollectiveOps::initialize(argc, argv);

  auto num_nodes = theContext()->getNumNodes();
  auto this_node = theContext()->getNode();

  if (argc > 1) {
    num_elms = atoi(argv[1]);
  }

  auto const& range = vt::Index2D((int)num_nodes, (int)num_elms);
  auto proxy = theCollection()->constructCollective<TestColl>(
    range, [](vt::Index2D idx){
      return std::make_unique<TestColl>();
    }
  );

  for (int i = 0; i < 1000; i++) {
    {
      auto ep = vt::theTerm()->makeEpochCollective();
      vt::theMsg()->pushEpoch(ep);
      if (this_node == 0) {
        auto msg = makeSharedMessage<TestColl::TestMsg>();
        proxy.broadcast<TestColl::TestMsg,&TestColl::doWork>(msg);
      }
      vt::theMsg()->popEpoch(ep);
      vt::theTerm()->finishedEpoch(ep);
      bool done = false;
      vt::theTerm()->addAction(ep, [&done]{ done = true; });
      do vt::runScheduler(); while (!done);
    }
    {
      auto ep = vt::theTerm()->makeEpochCollective();
      vt::theMsg()->pushEpoch(ep);
      if (this_node == 0) {
        auto msg = makeSharedMessage<TestColl::TestMsg>();
        proxy.broadcast<TestColl::TestMsg,&TestColl::dolb>(msg);
      }
      vt::theMsg()->popEpoch(ep);
      vt::theTerm()->finishedEpoch(ep);
      bool done = false;
      vt::theTerm()->addAction(ep, [&done]{ done = true; });
      do vt::runScheduler(); while (!done);
    }
  }

  while (!rt->isTerminated()) {
    runScheduler();
  }
  CollectiveOps::finalize();
  return 0;
}
```

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

lifflander:
Fixes #744 

This fixes and very difficult to find bug in EMPIRE due to cache entires expiring that are not allowed to expire given the algorithm.

### DIFF
--- a/src/vt/topos/location/directory/directory.h
+++ b/src/vt/topos/location/directory/directory.h
@@ -53,17 +53,20 @@ namespace vt { namespace location {
 
 template <typename KeyT, typename ValueT>
 struct Directory {
+  using DirectoryMapType = std::unordered_map<KeyT, ValueT>;
 
   Directory() = default;
 
   bool exists(KeyT const& key) const;
   std::size_t getSize() const;
   ValueT const& get(KeyT const& key);
+  typename DirectoryMapType::iterator getIter(KeyT const& key);
+  typename DirectoryMapType::iterator getIterEnd();
   void remove(KeyT const& key);
   void insert(KeyT const& key, ValueT const& value);
 
 private:
-  std::unordered_map<KeyT, ValueT> dir_;
+  DirectoryMapType dir_;
 };
 
 }} /* end namespace vt::location */
--- a/src/vt/topos/location/directory/directory.h
+++ b/src/vt/topos/location/directory/directory.h
@@ -1,0 +1,73 @@
+/*
+//@HEADER
+// *****************************************************************************
+//
+//                                 directory.h
+//                           DARMA Toolkit v. 1.0.0
+//                       DARMA/vt => Virtual Transport
+//
+// Copyright 2019 National Technology & Engineering Solutions of Sandia, LLC
+// (NTESS). Under the terms of Contract DE-NA0003525 with NTESS, the U.S.
+// Government retains certain rights in this software.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+// * Redistributions of source code must retain the above copyright notice,
+//   this list of conditions and the following disclaimer.
+//
+// * Redistributions in binary form must reproduce the above copyright notice,
+//   this list of conditions and the following disclaimer in the documentation
+//   and/or other materials provided with the distribution.
+//
+// * Neither the name of the copyright holder nor the names of its
+//   contributors may be used to endorse or promote products derived from this
+//   software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+//
+// Questions? Contact darma@sandia.gov
+//
+// *****************************************************************************
+//@HEADER
+*/
+
+#if !defined INCLUDED_VT_TOPOS_LOCATION_DIRECTORY_DIRECTORY_H
+#define INCLUDED_VT_TOPOS_LOCATION_DIRECTORY_DIRECTORY_H
+
+#include "vt/config.h"
+
+#include <unordered_map>
+
+namespace vt { namespace location {
+
+template <typename KeyT, typename ValueT>
+struct Directory {
+
+  Directory() = default;
+
+  bool exists(KeyT const& key) const;
+  std::size_t getSize() const;
+  ValueT const& get(KeyT const& key);
+  void remove(KeyT const& key);
+  void insert(KeyT const& key, ValueT const& value);
+
+private:
+  std::unordered_map<KeyT, ValueT> dir_;
+};
+
+}} /* end namespace vt::location */
+
+#include "vt/topos/location/directory/directory.impl.h"
+
+#endif /*INCLUDED_VT_TOPOS_LOCATION_DIRECTORY_DIRECTORY_H*/
--- a/src/vt/topos/location/location.h
+++ b/src/vt/topos/location/location.h
@@ -176,6 +176,7 @@ struct EntityLocationCoord : LocationCoord {
   void printCurrentCache() const;
 
   bool isCached(EntityID const& id) const;
+  void clearCache();
 
   template <typename MessageT>
   bool useEagerProtocol(MsgSharedPtr<MessageT> msg) const;
--- a/src/vt/topos/location/location.h
+++ b/src/vt/topos/location/location.h
@@ -52,7 +52,7 @@
 #include "vt/topos/location/utility/entity.h"
 #include "vt/topos/location/utility/coord.h"
 #include "vt/topos/location/message/msg.h"
-#include "vt/topos/location/cache/cache.h"
+#include "vt/topos/location/lookup/lookup.h"
 #include "vt/topos/location/record/record.h"
 #include "vt/topos/location/record/state.h"
 #include "vt/context/context.h"
@@ -73,7 +73,7 @@ struct collection_lm_tag_t {};
 template <typename EntityID>
 struct EntityLocationCoord : LocationCoord {
   using LocRecType = LocRecord<EntityID>;
-  using LocCacheType = LocationCache<EntityID, LocRecType>;
+  using LocCacheType = LocLookup<EntityID, LocRecType>;
   using LocEntityMsg = LocEntity<EntityID>;
   using LocalRegisteredContType = std::unordered_set<EntityID>;
   using LocalRegisteredMsgContType = std::unordered_map<EntityID, LocEntityMsg>;
@@ -115,10 +115,11 @@ struct EntityLocationCoord : LocationCoord {
    *
    *   1) Node 0: registerEntity(my_id, ...);
    *   2) Node 0: entityMigrated(my_id, 1);
-   *   3) Node 1: registerEntityMigrated(my_id, 0, ...);
+   *   3) Node 1: registerEntityMigrated(my_id, <home>, 0, ...);
    */
   void registerEntityMigrated(
-    EntityID const& id, NodeType const& __attribute__((unused)) from,
+    EntityID const& id, NodeType const& home_node,
+    NodeType const& __attribute__((unused)) from_node,
     LocMsgActionType msg_action = nullptr
   );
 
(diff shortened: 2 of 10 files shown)


src/vt/topos/location/lookup/lookup.impl.h
```diff
@@ -0,0 +1,111 @@
...
+
+template <typename KeyT, typename ValueT>
+bool LocLookup<KeyT, ValueT>::exists(KeyT const& key) const {
+  return directory_.exists(key) or cache_.exists(key);
+}
+
+template <typename KeyT, typename ValueT>
+ValueT const& LocLookup<KeyT, ValueT>::get(KeyT const& key) {
+  if (directory_.exists(key)) {
+    return directory_.get(key);
```
PhilMiller: Isn't this going to do a repeat search in the map? This should be getting and testing an iterator

src/vt/topos/location/location.impl.h
```diff
@@ -221,23 +223,29 @@ void EntityLocationCoord<EntityID>::entityMigrated(
...
+  recs_.update(id, LocRecType{id, eLocState::Remote, new_node});
 }
 
 template <typename EntityID>
 void EntityLocationCoord<EntityID>::registerEntityMigrated(
-  EntityID const& id, NodeType const& from, LocMsgActionType msg_action
+  EntityID const& id, NodeType const& home_node, NodeType const& from,
+  LocMsgActionType msg_action
 ) {
   // @todo: currently `from' is unused, but is passed to this method in case we
```
pnstickne: Commend outdated?

src/vt/topos/location/lookup/lookup.impl.h
pnstickne: Equivalent of try-get-value might be useful.

    ValueT v;
    if (directory_.tryGet(key, v)) return v;

src/vt/topos/location/lookup/lookup.impl.h
```diff
@@ -0,0 +1,111 @@
...
+template <typename KeyT, typename ValueT>
+void LocLookup<KeyT, ValueT>::clearCache() {
+  cache_ = LocationCache<KeyT, ValueT>{max_cache_size_};
+}
+
+template <typename KeyT, typename ValueT>
+void LocLookup<KeyT, ValueT>::printCache() const {
+  cache_.printCache();
+}
+
```
pnstickne: Extra newline.

src/vt/vrt/collection/manager.impl.h
```diff
@@ -2825,7 +2825,15 @@ MigrateStatus CollectionManager::migrateIn(
   vrt_elm_ptr->temp_elm_id_ = balance::ProcStats::getNextElm();
 
   bool const is_static = ColT::isStaticSized();
-  auto const home_node = uninitialized_destination;
+
+  using BaseIdxType = vt::index::BaseIndex;
+  auto idx_copy = idx;
```
pnstickne: For trivial types, consider using actual type.

src/vt/topos/location/location.impl.h
lifflander: Actually, the comment is still in play. Just added another parameter `home_node`

src/vt/topos/location/lookup/lookup.impl.h
lifflander: Fixed

src/vt/vrt/collection/manager.impl.h
lifflander: Addressed

src/vt/topos/location/lookup/lookup.impl.h
lifflander: Ok, I've switched to using an iterator.

tests/unit/location/test_location.cc
```diff
@@ -198,6 +200,48 @@ TEST_F(TestLocation, test_migrate_entity) /* NOLINT */ {
...
 
+TEST_F(TestLocation, test_migrate_entity_expire_cache) /* NOLINT */ {
+
+  auto const nb_nodes = vt::theContext()->getNumNodes();
+
+  // cannot perform entity migration if less than 3 nodes
+  if (nb_nodes > 2) {
+    auto const my_node  = vt::theContext()->getNode();
+    auto const entity   = location::default_entity;
+    auto const old_home = 0;
```
PhilMiller: The terminology `old_home` here is really problematic. An object's 'home' should never change. Please pick something else to call it. I've sometimes used 'host', or just 'current process'

lifflander: Yeah this was copied from another test. All the names in the location test suite are confusing and wrong.

lifflander: Ok, I've changed all the names to be coherent in the new test.

tests/unit/location/test_location.cc
```diff
@@ -198,6 +200,48 @@ TEST_F(TestLocation, test_migrate_entity) /* NOLINT */ {
...
+    // Clear the cache, to see if we can correctly fetch location after all
+    // cached entries are removed. Without the permanent directory, the home
+    // node will not know how to fetch the location.
+    vt::theLocMan()->virtual_loc->clearCache();
+    vt::theCollective()->barrier();
+
+    vt::theLocMan()->virtual_loc->getLocation(
+      entity, old_home, [=](vt::NodeType node) {
+        // With a clear cache, the result should always be correct
+        EXPECT_TRUE(node == new_home);
```
PhilMiller: Just a general vt testing question, that this brings up - are the tests required to exhibit global termination? Is there something enforcing that this `EXPECT_TRUE` actually gets evaluated? Or do you need to explicitly check for termination after this `getLocation` call?

src/vt/topos/location/directory/directory.h
```diff
@@ -0,0 +1,76 @@
...
+  bool exists(KeyT const& key) const;
+  std::size_t getSize() const;
+  ValueT const& get(KeyT const& key);
+  typename DirectoryMapType::iterator getIter(KeyT const& key);
+  typename DirectoryMapType::iterator getIterEnd();
+  void remove(KeyT const& key);
+  void insert(KeyT const& key, ValueT const& value);
+
+private:
+  DirectoryMapType dir_;
```
PhilMiller: This is rather precisely wrapping `DirectoryMapType`. It adds no functionality, it imposes no additional invariants. I don't see what purpose it serves over a simple typedef. Is there some expectation of expanding on it that I'm missing?

src/vt/topos/location/location.h
```diff
@@ -115,10 +115,11 @@ struct EntityLocationCoord : LocationCoord {
    *
    *   1) Node 0: registerEntity(my_id, ...);
    *   2) Node 0: entityMigrated(my_id, 1);
-   *   3) Node 1: registerEntityMigrated(my_id, 0, ...);
+   *   3) Node 1: registerEntityMigrated(my_id, <home>, 0, ...);
    */
   void registerEntityMigrated(
```
PhilMiller: Could we name these various functions with `Immigrated` (e.g. `entityImmigratedFrom`) and `Emigrated` (`entityMigratedTo`), so it's clearer which end each one gets called on?

src/vt/topos/location/lookup/lookup.h
```diff
@@ -0,0 +1,84 @@
...
+  LocLookup(LocationSizeType const& in_max_cache_size, NodeType in_this_node)
+    : max_cache_size_(in_max_cache_size),
+      cache_(in_max_cache_size),
+      this_node_(in_this_node)
+  { }
+
+  bool exists(KeyT const& key) const;
+  LocationSizeType getCacheSize() const;
+  ValueT const& get(KeyT const& key);
+  void remove(KeyT const& key);
```
PhilMiller: `remove` should only get called for an object that's deleted, right? Emigration of an object from a process should involve moving its entry from the directory to the cache, I think. In that case, there should be a dedicated method for such a transition.

src/vt/topos/location/lookup/lookup.impl.h
```diff
@@ -0,0 +1,111 @@
...
+template <typename KeyT, typename ValueT>
+void LocLookup<KeyT, ValueT>::remove(KeyT const& key) {
+  directory_.remove(key);
+  cache_.remove(key);
+}
+
+template <typename KeyT, typename ValueT>
+void LocLookup<KeyT, ValueT>::insert(
+  KeyT const& key, NodeType const home, ValueT const& value
```
PhilMiller: Does this mean 'home' in the sense of 'the process that will always know where to find object `key`',  or 'the process on which object `key` resides'? Both of those need to do the insertion in the directory, but for different reasons.

tests/unit/location/test_location.cc
lifflander: No, that's a good point. If we have lambda's/functions that might not be executed, we should assert that they do in addition to the actual assertion inside.

lifflander: Fixed

src/vt/topos/location/location.h
lifflander: Ok, I've created a issue to this renaming https://github.com/DARMA-tasking/vt/issues/746

src/vt/topos/location/lookup/lookup.h
lifflander: Correct, the only time its called is `LocationManager::unregisterEntity` which is only called during entity deletion. So there is a dedicated method of that in the location manager.

src/vt/topos/location/directory/directory.h
lifflander: Right now, there isn't any expansion beyond that. In the future, I plan to add some special debug printing, etc. which might make the encapsulation useful. Since it doesn't really hurt, I'm intending to keep it.

src/vt/topos/location/lookup/lookup.impl.h
lifflander: 'Home' means the node for which the map function returned for a given index which is the ultimate mediator for the entity.

lifflander: I need clarification on what 'the process on which object `key` resides' means.

src/vt/topos/location/lookup/lookup.impl.h
```diff
@@ -0,0 +1,111 @@
...
+    directory_.insert(key, value);
+  } else {
+    cache_.insert(key, value);
+  }
+}
+
+template <typename KeyT, typename ValueT>
+void LocLookup<KeyT, ValueT>::update(KeyT const& key, ValueT const& value) {
+  if (directory_.exists(key)) {
+    directory_.insert(key, value);
```
PhilMiller, line 93: This is going to do a double-search in `directory_` - if you find the iterator, you should be able to update it in place, I think?

lifflander: Fixed

Issue #647, opened 2019-12-30
Warnings in trace with Intel 18 compilers


lifflander:
**Describe the bug**
```
/lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-new-1b/vt/vt/src/vt/trace/trace.cc(801): warning #181: argument of type "vt::trace::TraceEntrySeqType={uint32_t={unsigned int}}" is incompatible with format "%lu", expecting argument of type "unsigned long"
          event_seq_id,
          ^

/lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-new-1b/vt/vt/src/vt/trace/trace.cc(805): warning #181: argument of type "vt::trace::TraceMsgLenType={size_t={unsigned long}}" is incompatible with format "%d", expecting argument of type "int"
          log->msg_len,
          ^

/lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-new-1b/vt/vt/src/vt/trace/trace.cc(818): warning #181: argument of type "vt::trace::TraceEntrySeqType={uint32_t={unsigned int}}" is incompatible with format "%lu", expecting argument of type "unsigned long"
          event_seq_id,
          ^

/lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-new-1b/vt/vt/src/vt/trace/trace.cc(822): warning #181: argument of type "vt::trace::TraceMsgLenType={size_t={unsigned long}}" is incompatible with format "%d", expecting argument of type "int"
          log->msg_len,
          ^

/lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-new-1b/vt/vt/src/vt/trace/trace.cc(853): warning #181: argument of type "vt::trace::TraceEntrySeqType={uint32_t={unsigned int}}" is incompatible with format "%lu", expecting argument of type "unsigned long"
          event_seq_id,
          ^

/lscratch1/jliffla/vt-clean/vt-auto-build/vt-release-new-1b/vt/vt/src/vt/trace/trace.cc(857): warning #181: argument of type "vt::trace::TraceMsgLenType={size_t={unsigned long}}" is incompatible with format "%d", expecting argument of type "int"
          log->msg_len,
          ^

[...]

PR #747, opened 2020-03-27
#647 fix format-string type warnings

pnstickne:
- Ensure types are fully compatible with the format specifier.
  Some compilers check and generate warnings.

No warnings in GCC7/Clang5/ICC18 for VT or example builds.

### DIFF
--- a/src/vt/vrt/collection/balance/proc_stats.cc
+++ b/src/vt/vrt/collection/balance/proc_stats.cc
@@ -144,8 +144,8 @@ void StatsRestartReader::inputStatsFile(
   fpos_t pos;
   bool finished = false;
   while (!finished) {
-    if (fscanf(pFile, "%zu %c %llu %c %lf", &phaseID, &separator, &elmID,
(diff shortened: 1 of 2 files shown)


src/vt/vrt/collection/balance/proc_stats.cc
```diff
@@ -144,8 +144,8 @@ void StatsRestartReader::inputStatsFile(
   fpos_t pos;
   bool finished = false;
   while (!finished) {
-    if (fscanf(pFile, "%zu %c %llu %c %lf", &phaseID, &separator, &elmID,
-               &separator, &tval) > 0) {
+    if (fscanf(pFile, "%zu %c %lu %c %lf",
```
pnstickne: Or change ElementIDType to be large enough for llu.

pnstickne: Actually, in this case it might be better to capture to a unsigned long local and convert before use to minimize changes..

PR #748, opened 2020-03-28
665 - Unify Message Lifetimes / MsgPtr

pnstickne:
Unifies message ownership lifetime, including detection of invalid message creation.

The primary purpose of these changes is to simplify the message tracking and work to a consistent/unified usage of 'MsgPtr' aka SharedMsgPtr.

Following change will be to accept MsgPtr uniformly at sendMsg level and deprecate/switch code to use makeMessage.

(There were additional changes relating to ensuring the buffers to MPI_Isend were not re-used too early. However, I've pulled them out of this PR as there were additional complexity and it only solved half the issue.)

### DIFF
--- a/src/vt/event/event_record.cc
+++ b/src/vt/event/event_record.cc
@@ -91,7 +91,7 @@ bool EventRecord::testMPIEventReady() {
 
   bool const mpiready = flag == 1;
 
-  if (mpiready and msg_ != nullptr and isSharedMessage(msg_.get())) {
+  if (mpiready and msg_ != nullptr) {
     debug_print_verbose(
       active, node,
       "testMPIEventRead: deref: msg={}\n",
--- a/src/vt/group/collective/group_info_collective.cc
+++ b/src/vt/group/collective/group_info_collective.cc
@@ -210,10 +210,10 @@ void InfoColl::setupCollective() {
   if (collective_->getInitialChildren() == 0) {
     auto const& size = static_cast<NodeType>(is_in_group ? 1 : 0);
     auto const& child = theContext()->getNode();
-    auto msg = makeSharedMessage<GroupCollectiveMsg>(
+    auto msg = makeMessage<GroupCollectiveMsg>(
       group_, up_tree_cont_, in_group, size, child
     );
-    theMsg()->sendMsg<GroupCollectiveMsg,upHan>(parent, msg);
+    theMsg()->sendMsg<GroupCollectiveMsg,upHan>(parent, msg.get());
   }
 }
 
@@ -316,10 +316,10 @@ void InfoColl::upTree() {
       auto const& extra = static_cast<GroupCollectiveMsg::CountType>(
         msg_in_group.size()-1
       );
-      auto msg = makeSharedMessage<GroupCollectiveMsg>(
+      auto msg = makeMessage<GroupCollectiveMsg>(
         group,new_root_cont_,true,subtree_zero,root_node,0,extra
       );
-      theMsg()->sendMsg<GroupCollectiveMsg,newRootHan>(root_node, msg);
+      theMsg()->sendMsg<GroupCollectiveMsg,newRootHan>(root_node, msg.get());
 
       for (std::size_t i = 1; i < msg_list.size(); i++) {
         debug_print(
@@ -356,10 +356,10 @@ void InfoColl::upTree() {
       subtree_, total_subtree
     );
 
-    auto cmsg = makeSharedMessage<GroupCollectiveMsg>(
+    auto cmsg = makeMessage<GroupCollectiveMsg>(
       group,op,is_in_group,total_subtree,child,level
     );
-    theMsg()->sendMsg<GroupCollectiveMsg,upHan>(p, cmsg);
+    theMsg()->sendMsg<GroupCollectiveMsg,upHan>(p, cmsg.get());
 
     for (auto&& msg : msg_in_group) {
       span_children_.push_back(msg->getChild());
@@ -385,10 +385,10 @@ void InfoColl::upTree() {
       subtree_
     );
 
-    auto msg = makeSharedMessage<GroupCollectiveMsg>(
+    auto msg = makeMessage<GroupCollectiveMsg>(
       group,op,is_in_group,static_cast<NodeType>(subtree_),child,0,extra
     );
-    theMsg()->sendMsg<GroupCollectiveMsg,upHan>(p, msg);
+    theMsg()->sendMsg<GroupCollectiveMsg,upHan>(p, msg.get());
     /*
      *  Forward all the children messages up the tree (up to 2 of them)
      */
@@ -415,10 +415,10 @@ void InfoColl::upTree() {
       subtree_, total_subtree
     );
 
-    auto msg = makeSharedMessage<GroupCollectiveMsg>(
+    auto msg = makeMessage<GroupCollectiveMsg>(
       group,op,is_in_group,total_subtree,child,0,extra
     );
-    theMsg()->sendMsg<GroupCollectiveMsg,upHan>(p, msg);
+    theMsg()->sendMsg<GroupCollectiveMsg,upHan>(p, msg.get());
     theMsg()->sendMsg<GroupCollectiveMsg,upHan>(p, msg_in_group[0].get());
   } else {
     vtAssertExpr(msg_in_group.size() > 2);
@@ -453,10 +453,10 @@ void InfoColl::upTree() {
       subtree_, total_subtree
     );
 
-    auto msg = makeSharedMessage<GroupCollectiveMsg>(
+    auto msg = makeMessage<GroupCollectiveMsg>(
       group,op,is_in_group,total_subtree,child,0,extra
     );
-    theMsg()->sendMsg<GroupCollectiveMsg,upHan>(p, msg);
+    theMsg()->sendMsg<GroupCollectiveMsg,upHan>(p, msg.get());
 
     debug_print(
       group, node,
@@ -611,14 +611,14 @@ void InfoColl::downTree(GroupCollectiveMsg* msg) {
   } else {
     auto const& num = collective_->span_children_.size();
     auto const& child = collective_->span_children_[msg->getChild() % num];
-    auto nmsg = makeSharedMessage<GroupCollectiveMsg>(*msg);
-    theMsg()->sendMsg<GroupCollectiveMsg,downHan>(child,nmsg);
+    auto nmsg = makeMessage<GroupCollectiveMsg>(*msg);
+    theMsg()->sendMsg<GroupCollectiveMsg,downHan>(child,nmsg.get());
     ++send_down_;
   }
 
   auto const& group_ = getGroupID();
-  auto nmsg = makeSharedMessage<GroupOnlyMsg>(group_,down_tree_fin_cont_);
-  theMsg()->sendMsg<GroupOnlyMsg,downFinishedHan>(from,nmsg);
+  auto nmsg = makeMessage<GroupOnlyMsg>(group_,down_tree_fin_cont_);
+  theMsg()->sendMsg<GroupOnlyMsg,downFinishedHan>(from,nmsg.get());
 }
 
 void InfoColl::newTree(NodeType const& parent) {
@@ -654,8 +654,8 @@ void InfoColl::sendDownNewTree() {
       "InfoColl::sendDownNewTree: group={:x}, sending to child={}\n",
       group_, c
     );
-    auto msg = makeSharedMessage<GroupOnlyMsg>(group_,new_tree_cont_);
-    theMsg()->sendMsg<GroupOnlyMsg,newTreeHan>(c,msg);
+    auto msg = makeMessage<GroupOnlyMsg>(group_,new_tree_cont_);
+    theMsg()->sendMsg<GroupOnlyMsg,newTreeHan>(c,msg.get());
   }
 }
 
@@ -693,10 +693,10 @@ void InfoColl::finalize() {
         group_, c
       );
 
-      auto msg = makeSharedMessage<GroupOnlyMsg>(
+      auto msg = makeMessage<GroupOnlyMsg>(
         group_,finalize_cont_,known_root_node_,is_default_group_
       );
-      theMsg()->sendMsg<GroupOnlyMsg,finalizeHan>(c,msg);
+      theMsg()->sendMsg<GroupOnlyMsg,finalizeHan>(c,msg.get());
     }
 
     if (!is_in_group) {
--- a/src/vt/messaging/active.cc
+++ b/src/vt/messaging/active.cc
@@ -185,7 +185,6 @@ EventType ActiveMessenger::sendMsgBytes(
 
   auto const epoch = envelopeIsEpochType(msg->env) ?
     envelopeGetEpoch(msg->env) : term::any_epoch_sentinel;
-  auto const is_shared = isSharedMessage(msg);
   auto const is_term = envelopeIsTerm(msg->env);
   auto const is_bcast = envelopeIsBcast(msg->env);
 
@@ -200,9 +199,7 @@ EventType ActiveMessenger::sendMsgBytes(
     );
   }
 
-  if (is_shared) {
-    mpi_event->setManagedMessage(base.to<ShortMessage>());
-  }
+  mpi_event->setManagedMessage(base.to<ShortMessage>());
 
   vtWarnIf(
     !(dest != theContext()->getNode() || is_bcast),
@@ -846,7 +843,7 @@ bool ActiveMessenger::tryProcessIncomingActiveMsg() {
 }
 
 void ActiveMessenger::finishPendingActiveMsgAsyncRecv(InProgressIRecv* irecv) {
-  auto buf = irecv->buf;
+  char* buf = irecv->buf;
   auto num_probe_bytes = irecv->probe_bytes;
   auto sender = irecv->sender;
 
@@ -858,8 +855,12 @@ void ActiveMessenger::finishPendingActiveMsgAsyncRecv(InProgressIRecv* irecv) {
 # endif
 
   MessageType* msg = reinterpret_cast<MessageType*>(buf);
-  messageConvertToShared(msg);
-  auto base = promoteMsgOwner(msg);
+  // Reset envelope to avoid ref-count=not_shared_msg if the data
+  // comes from a 'new RecvMsgType(..)'.
+  // The MsgPtr will then acquire a ref-count itself.
+  // TODO: Maybe envelope ref-count should be 0 on initialize?
+  envelopeSetRef(msg->env, 0);
+  MsgPtr<MessageType> base = MsgPtr<MessageType>{msg};
 
   auto const is_term = envelopeIsTerm(msg->env);
   auto const is_put = envelopeIsPut(msg->env);
--- a/src/vt/messaging/active.h
+++ b/src/vt/messaging/active.h
@@ -437,8 +437,8 @@ struct ActiveMessenger : runtime::component::PollableComponent<ActiveMessenger> 
    *     // collective invocation across nodes
    *     HandlerType const han = registerNewHandler(my_handler);
    *
-   *     MyMsg* msg = makeSharedMessage<MyMsg>(156);
-   *     theMsg()->sendMsg(29, han, msg);
+   *     auto msg = makeMessage<MyMsg>(156);
+   *     theMsg()->sendMsg(29, han, msg.get());
    *   }
    * \endcode
    * @{
--- a/src/vt/messaging/active.h
+++ b/src/vt/messaging/active.h
@@ -179,6 +179,7 @@ struct BufferedActiveMsg {
 };
 
 /**
+ * \internal
  * \struct ActiveMessenger active.h vt/messaging/active.h
  *
  * \brief Core component of VT used to send messages.
@@ -230,7 +231,6 @@ struct ActiveMessenger : runtime::component::PollableComponent<ActiveMessenger> 
   std::string name() override { return "ActiveMessenger"; }
 
   /**
-   * \internal
    * \brief Mark a message as a termination message.
    *
    * Used to ignore certain messages for the sake of termination detection
@@ -239,13 +239,13 @@ struct ActiveMessenger : runtime::component::PollableComponent<ActiveMessenger> 
    *
    * \param[in] msg the message to mark as a termination message
    */
-  template <typename MsgT>
-  void markAsTermMessage(MsgT* msg);
+  template <typename MsgPtrT>
+  void markAsTermMessage(MsgPtrT const msg);
 
   /**
    * \brief Mark a message as a location message
    *
-   * \param[in,out] msg  the message to mark as a location message
+   * \param[in] msg the message to mark as a location message
    */
   template <typename MsgPtrT>
   void markAsLocationMessage(MsgPtrT const msg);
@@ -253,7 +253,7 @@ struct ActiveMessenger : runtime::component::PollableComponent<ActiveMessenger> 
   /**
    * \brief Mark a message as a serialization control message
    *
-   * \param[in,out] msg  the message to mark as a serialization control message
+   * \param[in] msg the message to mark as a serialization control message
    */
   template <typename MsgPtrT>
   void markAsSerialMsgMessage(MsgPtrT const msg);
@@ -261,13 +261,12 @@ struct ActiveMessenger : runtime::component::PollableComponent<ActiveMessenger> 
   /**
    * \brief Mark a message as a collection message
    *
-   * \param[in,out] msg  the message to mark as a collection message
+   * \param[in] msg the message to mark as a collection message
    */
   template <typename MsgPtrT>
   void markAsCollectionMessage(MsgPtrT const msg);
 
   /**
-   * \internal
    * \brief Set the epoch in the envelope of a message
    *
    * \param[in] msg the message to mark the epoch on (envelope must be able to hold)
@@ -286,13 +285,10 @@ struct ActiveMessenger : runtime::component::PollableComponent<ActiveMessenger> 
   template <typename MsgT>
   void setTagMessage(MsgT* msg, TagType tag);
 
-  template <typename MessageT>
-  ActiveMessenger::PendingSendType sendMsgCopyableImpl(
-    NodeType dest,
-    HandlerType han,
-    MsgSharedPtr<MessageT>& msg,
-    ByteType msg_size,
-    TagType tag
+  template <typename MsgPtrT>
+  trace::TraceEventIDType makeTraceCreationSend(
+    MsgPtrT msg, HandlerType const handler, auto_registry::RegistryTypeEnum type,
+    MsgSizeType msg_size, bool is_bcast
   );
 
   // With serialization, the correct method is resolved via SFINAE.
@@ -411,6 +407,15 @@ struct ActiveMessenger : runtime::component::PollableComponent<ActiveMessenger> 
 #endif
     return sendMsgCopyableImpl<MessageT>(dest, han, msg, msg_size, tag);
   }
+
+  template <typename MessageT>
+  ActiveMessenger::PendingSendType sendMsgCopyableImpl(
+    NodeType dest,
+    HandlerType han,
+    MsgSharedPtr<MessageT>& msg,
+    ByteType msg_size,
+    TagType tag
+  );
 
   /**
    * \defgroup preregister Basic Active Message Send with Pre-Registered Handler
@@ -1483,12 +1488,6 @@ struct ActiveMessenger : runtime::component::PollableComponent<ActiveMessenger> 
   void clearListeners() {
     send_listen_.clear();
   }
-
-  template <typename MsgPtrT>
-  trace::TraceEventIDType makeTraceCreationSend(
-    MsgPtrT msg, HandlerType const handler, auto_registry::RegistryTypeEnum type,
-    MsgSizeType msg_size, bool is_bcast
-  );
 
 private:
   bool testPendingActiveMsgAsyncRecv();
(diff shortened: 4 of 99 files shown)


src/vt/event/event_record.cc
```diff
@@ -87,7 +87,13 @@ bool EventRecord::testMPIEventReady() {
 
   bool const mpiready = flag == 1;
 
-  if (mpiready and msg_ != nullptr and isSharedMessage(msg_.get())) {
+  if (mpiready and msg_ != nullptr) {
+    // Ensure valid-ref message.
+    vtAssertInfo(
+      envelopeGetRef(msg_.get()->env) >= 0, "Bad Message Ref",
```
pnstickne: When can it be 0?

src/vt/group/collective/group_info_collective.cc
```diff
@@ -301,26 +301,32 @@ void InfoColl::upTree() {
...
+          msg_in_group.size()-1
+        );
+        auto msg = makeSharedMessage<GroupCollectiveMsg>(
+          group, new_root_cont_, true, subtree_zero, root_node, 0, extra
+        );
+        theMsg()->sendMsg<GroupCollectiveMsg,newRootHan>(root_node, msg);
+      }
 
       for (std::size_t i = 1; i < msg_list.size(); i++) {
+        MsgPtr<GroupCollectiveMsg> pmsg = copyAndResetMessage(msg_list[i]);
```
pnstickne: Messages cannot be re-sent. This includes received messages.

src/vt/messaging/active.cc
```diff
@@ -178,9 +178,14 @@ EventType ActiveMessenger::sendMsgBytes(
 ) {
   auto const& msg = base.get();
 
+  // Ensure valid-ref message.
```
pnstickne: When can this be 0?

examples/trace_ring.cc
```diff
@@ -89,7 +89,7 @@ static void ring(RingMsg* msg) {
 }
 
 static void sendToNext() {
-  RingMsg* msg = new RingMsg(my_node);
```
pnstickne: Direct creation of messages will result in runtime assertion failures.

src/vt/messaging/active.cc
```diff
@@ -195,9 +200,7 @@ EventType ActiveMessenger::sendMsgBytes(
     );
   }
 
-  if (is_shared) {
```
pnstickne: All messages are always shared (or share-able).

src/vt/messaging/active.h
```diff
@@ -280,13 +279,10 @@ struct ActiveMessenger {
   template <typename MsgT>
   void setTagMessage(MsgT* msg, TagType tag);
 
-  template <typename MessageT>
```
pnstickne: Moving around methods for consistency. No logic change here.

src/vt/messaging/active.h
```diff
@@ -233,35 +233,34 @@ struct ActiveMessenger {
    *
    * \param[in] msg the message to mark as a termination message
    */
-  template <typename MsgT>
```
pnstickne: Ideally, all take in a MsgT*; this change moves to ensure consistency as the other forms still accept a Message* or a MsgPtr<Message>.

src/vt/messaging/envelope/envelope_ref.impl.h
```diff
@@ -52,12 +52,38 @@ namespace vt {
...
+    "'Not shared message' encountered on message ref-increment. "
+    "This is can be caused by explicitly using 'new Message(..)' "
+    "instead of a MsgPtr/makeMessage/makeSharedMessage construct."
+  );
+  vtAssertInfo( // "wow", likely UB already
+    envp->ref >= 0,
+    "Bad ref-count on message ref-increment.",
+    static_cast<RefType>(envp->ref)
+  );
+  vtAssert(
```
pnstickne: Combine with above?

src/vt/messaging/envelope/envelope_ref.impl.h
```diff
@@ -52,12 +52,38 @@ namespace vt {
...
+
+  envp->ref++;
 }
 
 template <typename Env>
 inline RefType envelopeDeref(Env& env) {
-  return --(reinterpret_cast<Envelope*>(&env))->ref;
+  Envelope* envp = reinterpret_cast<Envelope*>(&env);
+
+  vtAssertInfo(
```
pnstickne: Ref-count checks in core, not MsgPtr.

src/vt/messaging/message/message.h
```diff
@@ -203,7 +207,6 @@ struct ActiveMsg : BaseMsg {
   void serialize(SerializerT& s) {
     base_serialize_called_ = true;
     // n.b. not actually used, as extracted during transmission.
-    s | has_owner_;
```
pnstickne: Feels like some changes were lost. Wasn't there work required to proxy through `env` for serialization?

src/vt/messaging/message/refs.h
```diff
@@ -57,9 +57,6 @@ void messageRef(MsgT* msg);
 template <typename MsgT>
 void messageDeref(MsgT* msg);
 
-template <typename MsgT>
```
pnstickne: No external API usage.

src/vt/messaging/message/shared_message.h
```diff
@@ -52,30 +52,35 @@
...
 
+/// Create a new wrapped message using a forwarded constructor call.
 template <typename MsgT, typename... Args>
-MsgSharedPtr<MsgT> makeMessageSz(std::size_t size, Args&&... args);
+MsgPtr<MsgT> makeMessageSz(std::size_t size, Args&&... args);
 
-///[obsolete] Use makeMessage instead.
 template <typename MsgT, typename... Args>
-MsgSharedPtr<MsgT> makeMsg(Args&&... args);
+[[deprecated("Use makeMessage instead")]]
```
pnstickne: Remove. No API usages.

src/vt/messaging/message/shared_message.h
```diff
@@ -52,30 +52,35 @@
...
+MsgPtr<MsgT> makeMsg(Args&&... args);
 
+/**
+ * \internal
+ * \brief Make a copy of a message with a reset envelope.
```
pnstickne: Expand comment.

src/vt/messaging/message/shared_message.impl.h
```diff
@@ -50,72 +50,50 @@
 #include "vt/messaging/message/smart_ptr.h"
 #include "vt/pool/pool.h"
 
-#if HAS_SERIALIZATION_LIBRARY
-  #include "serialization_library_headers.h"
-#endif
-
 namespace vt {
 
-template <typename MsgT>
```
pnstickne: Message compatibility is checked in sendMsg/broadcastMsg flows, which are more complete.

src/vt/messaging/message/shared_message.h
```diff
@@ -52,30 +52,35 @@
 
 namespace vt {
 
+/// Create a new 'raw' message, using a forwarded constructor call.
```
pnstickne: Improve comments. When using makeSharedMessage/Sz, the resulting object MUST be supplied to promoteMsg or there _will_ be memory leaks. The ref-count starts as 0 and is only incremented to one on the first acquisition of a MsgPtr.

src/vt/messaging/message/message.h
```diff
@@ -85,14 +87,15 @@ struct ActiveMsg : BaseMsg {
    * this field may be accessed in contexts where the EnvelopeType is not yet
    * checked/determined
    */
-
```
pnstickne: All messages must be tracked.

src/vt/messaging/message/smart_ptr.h
```diff
@@ -97,17 +97,17 @@ struct MsgSharedPtr final {
 
   MsgSharedPtr(std::nullptr_t) {}
 
-  MsgSharedPtr(T* in, bool takeRef) {
-    init(in, takeRef, &statics::Holder::TypedMsgPtrImpls<T>);
+  MsgSharedPtr(T* in) {
```
pnstickne: Could allow special overload to NOT take a ref just for the case of makeMessage to about the 1..2..1.. ref count when returning a copy from the method.

src/vt/messaging/message/smart_ptr.h
```diff
@@ -167,136 +167,118 @@ struct MsgSharedPtr final {
...
+ */
 template <typename T>
 using MsgVirtualPtr = messaging::MsgSharedPtr<T>;
 
-// For historic reasons;
-// Functionality is now part of MsgSharedPtr
+/**
+ * \internal
+ * \obsolete Use \c MsgPtr<ShortMessage>, or as appropriate.
+ */
```
pnstickne: Not used externally.

src/vt/messaging/message/smart_ptr.h
pnstickne: Just kidding, https://en.cppreference.com/w/cpp/language/copy_elision requires RVO.

src/vt/messaging/message/shared_message.impl.h
```diff
@@ -50,72 +50,50 @@
...
   return msg;
 }
 
 template <typename MsgT, typename... Args>
-MsgSharedPtr<MsgT> makeMessage(Args&&... args) {
-  messageTypeChecks<MsgT>();
-  MsgT* msg = makeSharedMessage<MsgT>(std::forward<Args>(args)...);
-  return promoteMsgOwner<MsgT>(msg);
+MsgPtr<MsgT> makeMessage(Args&&... args) {
+  // Message ref due to copy will be 1..2..1..
```
pnstickne: Comments are not accurate as RVO is required in this case - https://en.cppreference.com/w/cpp/language/copy_elision

src/vt/messaging/message/message.h
```diff
@@ -73,8 +73,10 @@ struct BaseMsg { };
  * uses the default envelope.
  *
  * This is templated over the envelope to allow longer or shorter messages to be
- * used depending on the target and needed envelope bits.
+ * used depending on the required state associated with the message..
  *
+ * All messages must be copy-constructable and invoke the base copy-constructor.
+ * Messages should generally not be copy-assignable.
```
pnstickne: Why NOT copy-assignable?

src/vt/group/collective/group_info_collective.cc
```diff
@@ -301,26 +301,32 @@ void InfoColl::upTree() {
...
-      auto const& subtree_zero = static_cast<NodeType>(0);
-      auto const& extra = static_cast<GroupCollectiveMsg::CountType>(
-        msg_in_group.size()-1
-      );
-      auto msg = makeSharedMessage<GroupCollectiveMsg>(
-        group,new_root_cont_,true,subtree_zero,root_node,0,extra
-      );
-      theMsg()->sendMsg<GroupCollectiveMsg,newRootHan>(root_node, msg);
+      {
+        // Send root
```
pnstickne: Tidy/blocked. No logic change.

src/vt/messaging/active.impl.h
```diff
@@ -171,6 +171,16 @@ ActiveMessenger::PendingSendType ActiveMessenger::sendMsgCopyableImpl(
 
   MessageT* rawMsg = msg.get();
 
+  // Semantic guard, which triggers for any message with an already initialized
+  // destination. This guard is valid even for raw MsgT* API calls.
+  // (Probably best type-guard would be to required MsgPtr&& on public API,
```
pnstickne: Not true. Public APIs can accept `MsgPtr` easily as even with a copy it's size is only "two pointers".

src/vt/messaging/envelope/envelope_ref.impl.h
```diff
@@ -52,12 +52,35 @@ namespace vt {
 
 template <typename Env>
 inline void envelopeRef(Env& env) {
-  (reinterpret_cast<Envelope*>(&env))->ref++;
+  Envelope* envp = reinterpret_cast<Envelope*>(&env);
+
+  vtAssert(
+    envp->ref != not_shared_message,
```
pnstickne: Helpful assert when attempting to use a message directly created..

src/vt/messaging/message/shared_message.impl.h
```diff
@@ -50,72 +50,50 @@
...
-  );
-}
-
 template <typename MsgT, typename... Args>
 MsgT* makeSharedMessage(Args&&... args) {
-  messageTypeChecks<MsgT>();
   MsgT* msg = new MsgT{std::forward<Args>(args)...};
-  envelopeSetRef(msg->env, 1);
-  msg->has_owner_ = false;
+  // n.b. do NOT actually take a ref here.
```
pnstickne: Returned pointer is a 'leak' until promoted.. just as it always has.

src/vt/messaging/message/shared_message.impl.h
```diff
@@ -50,72 +50,50 @@
...
-void messageSetUnmanaged(MsgT* msg) {
-  envelopeSetRef(msg->env, not_shared_message);
-}
-
-template <typename MsgT>
-void messageResetDeserdes(MsgSharedPtr<MsgT> const& msg) {
-  envelopeSetRef(msg->env, 1);
-  msg->has_owner_ = true;
+MsgPtr<MsgT> copyAndResetMessage(MsgT* msg) {
+  MsgT* copy = new MsgT{*msg}; // should be copy-ctor (loses "sz" alloc..)
```
pnstickne: This probably never runs into the case where an 'sz' alloc was required.. unknown behavior in such a case.

src/vt/serialization/messaging/serialized_messenger.impl.h
```diff
@@ -62,6 +63,15 @@
 
 namespace vt { namespace serialization {
 
+template <typename MsgT>
+static MsgPtr<MsgT> deserializeFullMessage(SerialByteType* source, size_t len) {
+  auto msg = makeSharedMessage<MsgT>();
+  deserializeInPlace<MsgT>(source, len, msg);
+  // Reset ref-count to 0 (don't accept any deserialized value)
+  envelopeSetRef(msg->env, 0);
+  return promoteMsg(msg);
```
pnstickne: Change to return MsgPtr{..} for RVO.

src/vt/messaging/message/message.h
lifflander: I don't think anything was lost here, but I don't remember

src/vt/group/collective/group_info_collective.cc
lifflander: Should we consider the case when a message is done sending and can be reused? `vt::theEvent()` can tell when a message is done and triggers an event.

src/vt/messaging/active.cc
lifflander: Great!

examples/trace_ring.cc
```diff
@@ -89,7 +89,7 @@ static void ring(RingMsg* msg) {
...
+  auto msg = makeSharedMessage<RingMsg>(my_node);
```
lifflander: Can we add a deprecated attribute to `makeSharedMessage`?

src/vt/group/collective/group_info_collective.cc
lifflander: We should start thinking about read-only messages that might inhibit copies and increase efficiency. Like a broadcast with multiple endpoints on the same node can deliver the same data to each one if the data is read-only.

pnstickne: Waiting for the event may prevent parallel dispatching in MPI (to different nodes). Probably depends heavily on if the MPI implementation uses buffered mode implicitly.

The mutability issue is the envelope is packed into the data, and such is currently required in some cases on a receive (possibly bugs-in-hiding) - although it's not permissible to re-use an MPI_Isend buffer until the send is complete.

If it was guaranteed that an envelop can be "local-state-zero'd" (that is, reset all state not set before a send is initiated) on each receive, then I think it might be OK to re-use the buffer.

I'll see if I can't split out/off the local mutability.

src/vt/messaging/message/message.h
pnstickne: I definitely made a change.. and probably reset it myself. 🥇

src/vt/messaging/message/shared_message.impl.h
pnstickne: This is used in serialization. However, serialization will supply a message *not* requiring serialization down the lower calls which currently use `copyAndResetMessage`.

examples/trace_ring.cc
pnstickne: Ahah, so many test/example updates 💃 

And, yes. Would probably consider a follow-on branch to update all test/example code not use such API.

src/vt/group/collective/group_info_collective.cc
pnstickne: The user would still be responsible to make a copy when changing any payload and/or when altering transmittable envelope data.

pnstickne: @lifflander Nope. Not possible to separate out due to manipulating state for both local and remote needs. The current usages of MPI_Isend are _technically undefined_ when the same message is re-sent with a modified envelope.

Options:

- Copy on individual problematic sites (as done, and roughly checked with a guard); cost is only paid when needed, such as in a broadcast. It might be possible to reduce copies if handlers not changed. Handlers appear the real problematic bit.

- Copy generally internally on send (memcpy). Does not use same buffer pool, penalties on all sends.

- Wait for buffer release before sending - each call to MPI_Isend serialized, so likely no better than MPI_send. This also makes the calls themselves async wrt. message access and likely requires a good bit of rework. Saves on some memory copies, perhaps.

- Use MPI_send - blocking until the buffer is available again (buffered or matching receive, implicit. dependent).

pnstickne: Another option might be to (at least in the code that may double-send), send two messages: one with a newly allocated buffer containing JUST the envelope, and then using the message body. It would double number of sent messages in certain cases, and require a little bit more logic to put back together (relying on non-overtaking and last-seen).

pnstickne: Although, going the internal buffer copy route would also simplify/enable things like enabling "rendezvous" more generally. Around Pending Send the downstream data would always be an envelope and buffer pointer with release capability. Then there would be one copy from that into the MPI send buffer.

pnstickne: @lifflander Copies will now be made automatically on an 'as needed' basis. The core issue (of using a message-as-a-buffer-with-envelope) prohibits other options until willing to accept multiple messages or send buffers that are separate from messages.

src/vt/messaging/active.impl.h
```diff
@@ -169,8 +169,26 @@ ActiveMessenger::PendingSendType ActiveMessenger::sendMsgCopyableImpl(
...
+  // lost in PendingSend; it could be pushed down.
+  if (envelopeIsLocked(msg->env))
+  {
+    debug_print(
+      active, node,
+      "Made message copy to avoid mutating shared buffer: ptr={}, type={}\n",
+      print_ptr(rawMsg), typeid(MessageT).name()
+    );
+
+    msg = copyAndResetMessage<MessageT>(rawMsg);
```
lifflander, line 187: I was thinking about whether all messages are copyable. I suppose they have to be if they are non-serialized.

src/vt/messaging/active.h
lifflander: This is great, we should do this along more of the interface calls if we are going to allow both forms.

src/vt/messaging/message/shared_message.h
```diff
@@ -52,30 +52,80 @@
...
+ * when the returned MsgPtr (and all copies of such) are destroyed.
+ */
 template <typename MsgT, typename... Args>
-MsgSharedPtr<MsgT> makeMessageSz(std::size_t size, Args&&... args);
+MsgPtr<MsgT> makeMessageSz(std::size_t size, Args&&... args);
 
-///[obsolete] Use makeMessage instead.
 template <typename MsgT, typename... Args>
-MsgSharedPtr<MsgT> makeMsg(Args&&... args);
+[[deprecated("Use makeMessage instead")]]
```
lifflander, line 112: Isn't this just an alias for `makeMessage`

src/vt/messaging/message/shared_message.h
```diff
@@ -52,30 +52,80 @@
...
+ * \warning
+ * The returned pointer represents a leaked object until 'promoted' to a MsgPtr.
+ * While \c theMsg send API will automatically perform a promotion, automatic
+ * message promotion should generally not be relied upon.
+ */
 template <typename MsgT, typename... Args>
 MsgT* makeSharedMessageSz(std::size_t size, Args&&... args);
 
+/**
+ * \deprecated Use \c makeMesssage.
```
lifflander, line 86: This isn't deprecated (copy-paste?)

src/vt/messaging/message/shared_message.h
```diff
@@ -52,30 +52,80 @@
...
+ *
+ * The lifetime of the message is controlled by MsgPtr and will be destroyed
+ * when the returned MsgPtr (and all copies of such) are destroyed.
+ */
 template <typename MsgT, typename... Args>
-MsgSharedPtr<MsgT> makeMessage(Args&&... args);
+MsgPtr<MsgT> makeMessage(Args&&... args);
 
+/**
+ * \deprecated Use \c makeMesssage.
```
lifflander, line 99: This isn't deprecated (copy-paste?)

src/vt/messaging/message/shared_message.impl.h
```diff
@@ -50,72 +50,50 @@
...
+  envelopeSetRef(msg->env, 0);
   return msg;
 }
 
 template <typename MsgT, typename... Args>
-MsgSharedPtr<MsgT> makeMessage(Args&&... args) {
-  messageTypeChecks<MsgT>();
-  MsgT* msg = makeSharedMessage<MsgT>(std::forward<Args>(args)...);
-  return promoteMsgOwner<MsgT>(msg);
+MsgPtr<MsgT> makeMessage(Args&&... args) {
```
lifflander, line 74: I don't know if there's a point---but should be make these explicitly inline?

src/vt/group/collective/group_info_collective.cc
lifflander: @pnstickne Are there more copies being made now than before? I'm not sure I completely understand the change. I do see the "copy-on-demand" code you've added.

AFAIK, the current code does not modify the envelope between k broadcast sends. It does do multiple sends on the same buffer without it changing. Have you changed that?

examples/trace_ring.cc
lifflander: Also, why don't we just change this to `makeMessage` since that's the direction we are headed anyway?

src/vt/messaging/message/smart_ptr.h
```diff
@@ -167,136 +167,124 @@ struct MsgSharedPtr final {
...
 
+
+// Expose public/common types in vt:: namespace.
 namespace vt {
 
-// For historic reasons;
-// Functionality is now part of MsgSharedPtr
+/**
+ * \internal
+ * \obsolete Use \c MsgPtr<T>, for which is is an alias.
```
lifflander, line 232: Are we allowed to add deprecated to type aliases? I would like to do that if possible for `MsgVirtualPtr`

src/vt/group/collective/group_info_collective.cc
pnstickne: @lifflander I made no changes to the broadcast code. However, I can guarantee that changes to the envelope _are_ being made.

src/vt/messaging/active.impl.h
pnstickne: At this point all the messages are copyable as the serialization generated the eager and/or rendezvous.

Anyway, I'm going to pull this code out of this request. There are additional guards that need to be done in order to fully reduce edge-cases of modifying the same buffer.

src/vt/group/collective/group_info_collective.cc
pnstickne: @lifflander I pulled out changes relating to the envelope as it was insufficient to cover some edge-cases. This PR is merely about lifetimes now.

lifflander:
@PhilMiller Please review this so we can get it moving forward on develop.

src/vt/messaging/message/smart_ptr.h
pnstickne: Yes. Steps through.. one can get a [rebased] version of changes like this, it's easier to layer on cleanup work focusing internal external usages.

src/vt/messaging/envelope/envelope_extended.h
```diff
@@ -53,6 +53,9 @@
 
 namespace vt { namespace messaging {
 
+// Envelope type requirements:
+// - All envelops must have an 'ActiveEnvelope env' field as the FIRST field.
```
PhilMiller, line 57: Typo 'envelopes'

src/vt/messaging/message/shared_message.h
```diff
@@ -52,30 +52,65 @@
...
+ * \brief Create a new message, of a size.
+ *
+ * Create a new message already wrapped in a MsgPtr.
+ * The arguments are forwarded down to the underlying message's constructor.
+ *
+ * The lifetime of the message is controlled by MsgPtr and will be destroyed
+ * when the returned MsgPtr (and all copies of such) are destroyed.
+ */
```
PhilMiller, line 107: Please document what the `size` parameter means.

src/vt/messaging/message/message.h
PhilMiller: An earlier revision of the branch had `envelopeSetHasBeenSerialized(env, true);` here. I don't think I've seen that elsewhere, so it may actually be missing.

PhilMiller: Oops, it just moved up above the comment.

PhilMiller:
Code all looks good to me.

lifflander:
For some reason, I can't approve this PR...but it's approved

lifflander:


```diff

Clones removed
==============
+ src/vt/objgroup/proxy/proxy_objgroup.impl.h  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5263287&bid=17243143)

lifflander:


```diff

Clones removed
==============
+ src/vt/objgroup/proxy/proxy_objgroup.impl.h  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5263287&bid=17243143)

src/vt/messaging/message/shared_message.h
lifflander: done

src/vt/messaging/envelope/envelope_extended.h
lifflander: Fixed

src/vt/messaging/message/smart_ptr.h
lifflander: I've switched all these uses over now

lifflander:


```diff

Clones added
============
- src/vt/messaging/message/shared_message.impl.h  2
- src/vt/objgroup/manager.h  1
- src/vt/objgroup/manager.fwd.h  1
- tests/unit/memory/test_memory_lifetime.cc  2
- tests/unit/pipe/test_callback_send.cc  2
- tutorial/tutorial_1g.h  3
         

Clones removed
==============
+ src/vt/objgroup/proxy/proxy_objgroup.impl.h  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5263287&bid=17243143)

src/vt/pipe/interface/remote_container.impl.h
```diff
@@ -100,7 +100,7 @@ RemoteContainer<MsgT,TupleT>::triggerDirect(CallbackT cb, MsgU* data) {
     "RemoteContainer: (typed) invoke trigger: pipe={:x}, multi={}, ptr={}\n",
     pid, multi_callback, print_ptr(data)
   );
-  MsgT* cur_msg = data;
+  MsgSharedPtr<MsgT> cur_msg = promoteMsg(data);
```
pnstickne, line 103: This feels at-odds with the if clause below. It also seems like it should own (or take ownership of) the incoming pointer.

src/vt/objgroup/manager.cc
```diff
@@ -56,7 +56,7 @@ ObjGroupProxyType ObjGroupManager::getProxy(ObjGroupProxyType proxy) {
   return proxy;
 }
 
-void ObjGroupManager::dispatch(MsgVirtualPtrAny msg, HandlerType han) {
+void ObjGroupManager::dispatch(MsgSharedPtr<ShortMessage> msg, HandlerType han) {
```
pnstickne, line 59: Code setup to use `MsgPtr<T>` instead of `MsgSharedPtr<T>`, for lazier fingers.

src/vt/pipe/interface/remote_container.impl.h
pnstickne: Also ideal to remove `promoteMsg`, can be replaced with `MsgPtr<T>{msg_ptr}` - should check usage counts/locations.

lifflander: Good point, I think this should just call the copy ctor on the data for safety. There are a lot of possible things that can be done in a callback that could modify data

```C++
  /*
   *  Make a copy of the message for each callback trigger because the callback
   *  trigger may send/bcast messages, thus a copy must be made
   */
  if (multi_callback) {
    auto cur_msg = makeMessage<MsgT>(*data);
    cb.trigger(cur_msg.get(),pid);
  } else {
    auto cur_msg = data;
    cb.trigger(cur_msg,pid);
  }
```

lifflander:


```diff

Clones added
============
- tests/unit/memory/test_memory_lifetime.cc  2
- tutorial/tutorial_1g.h  3
- tests/unit/pipe/test_callback_send.cc  2
- src/vt/messaging/message/shared_message.impl.h  2
- src/vt/objgroup/manager.fwd.h  1
- src/vt/objgroup/manager.h  1
         

Clones removed
==============
+ src/vt/objgroup/proxy/proxy_objgroup.impl.h  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5263287&bid=17243143)

PR #751, opened 2020-04-03
750 CI build system rewrite & *fix memory errors*

### LINKED ISSUE #750: Rewrite CI build system configuration
**What Needs to be Done?**

- [ ] Do it

lifflander:
Fixes #750

- Re-write the entire CI for VT using docker compose with volumes and proper composition container techniques
- Volumes are used to mount persist `ccache` data for each platform that can be cached across multiple GitHub actions or through docker when running locally.
- Add capabilities for building with address sanitizer in VT and through a docker config environment variable `VT_ASAN=1`
- Expand testing to include ubuntu/clang configurations by building mpich in base image with non-default (clang) compiler
- Add support for interactive building/debugging with docker and valgrind to check for memory errors
- Adds a new test target build on Mac OS X with apple clang!

- Add **address sanitizer** as default CI target
- **Fix a bunch of memory leaks and invalid reads/writes to build cleanly with ASAN**


.github/workflows/dockerimage-clang-3.9-ubuntu-mpich.yml
```diff
@@ -0,0 +1,77 @@
+name: PR tests (clang-3.9, ubuntu, mpich)
```
pnstickne, line 1: Is there a way to create/generate-update these files?

Same with other files that are mostly the same.

CMakeLists.txt
```diff
@@ -83,6 +83,20 @@ if (${vt_mimalloc_enabled})
...
 
+option(vt_asan_enabled "Build VT with address sanitizer" OFF)
+if (${vt_asan_enabled})
+  if(CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang" OR
+     CMAKE_CXX_COMPILER_ID STREQUAL "Clang" OR
+     (CMAKE_CXX_COMPILER_ID
+       STREQUAL
+       "GNU"
+       AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER "4.8"))
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=address -DADDRESS_SANITIZER")
```
pnstickne, line 94: Isn't there "set option if available"?

cmake/define_build_types.cmake
```diff
@@ -145,7 +146,15 @@ set (vt_feature_cmake_mpi_rdma "0")
 set (vt_feature_cmake_print_term_msgs "0")
 set (vt_feature_cmake_default_threading "1")
 set (vt_feature_cmake_no_pool_alloc_env "0")
-set (vt_feature_cmake_memory_pool "1")
+
+if (${vt_pool_enabled})
```
pnstickne, line 150: Should be `if (vt_pool_enabled)` unless explicitly wishing (usually not needed) the variable to be substituted first.

if (x)      # if variable contains a true-ish value
if (${x}) # expand variable x, and see if the resulting condition is true-ish

examples/serialize_example.cc
```diff
@@ -151,10 +157,13 @@ void testSerializeByteUserClass() {
 
   fmt::print("ptr={}, size={}\n", buf, buf_size);
 
-  auto* tptr1 = deserialize<Type>(buf, buf_size);
+  auto buffer = new Type;
```
pnstickne, line 160: unique_ptr?

tests/unit/collectives/test_collectives_reduce.cc
```diff
@@ -152,36 +152,36 @@ TEST_F(TestReduce, test_reduce_op) {
   auto const my_node = theContext()->getNode();
   auto const root = 0;
 
-  auto msg = makeSharedMessage<MyReduceMsg>(my_node);
+  auto msg = makeMessage<MyReduceMsg>(my_node);
   debug_print(reduce, node, "msg->num={}\n", msg->num);
-  theCollective()->reduce<MyReduceMsg, reducePlus>(root, msg);
+  theCollective()->reduce<MyReduceMsg, reducePlus>(root, msg.get());
```
pnstickne, line 157: Public-API ops can feed through `MsgPtr<T>` (many examples of API).

Main reason why this is easier to stage separately.

src/vt/messaging/message/smart_ptr.h
```diff
@@ -62,6 +62,11 @@ namespace vt { namespace messaging { namespace statics {
   /// Static objects with global lifetime and elimination of allocation.
   /// They are stateless and have a very high/constant reuse pattern.
   struct Holder {
+    template <typename T>
+    static MsgPtrImplTyped<T>* getImpl() {
```
pnstickne, line 66: Proxy name loses intent clarity. Renamed class could restore such. Holder..getImpl..of what? :|

     TypeMsgPtr::getImpl<T>()

     TypeMsgPtrImpl::get<T>()

Still not great, but contains more context.

PR #761, opened 2020-04-09
760 Tutorial: work around bug in collection buffering

### LINKED ISSUE #760: Work around for tutorial breakage due to reduction race
**Describe the bug**
There is a proper fix #757 for the reduction race described in #713, which is too heavy to go on the release branch. So this issue is for a temporary work around for the bug described in #713.

lifflander:
Fixes #760

### DIFF
--- a/tutorial/tutorial_2b.h
+++ b/tutorial/tutorial_2b.h
@@ -113,13 +113,21 @@ static inline void collectionReduce() {
    * This is an example of reducing over a virtual context collection
    */
 
-  if (this_node == 0) {
-    // Range of 32 elements for the collection
-    auto range = ::vt::Index1D(32);
-    // Construct the collection: invoked by one node. By default, the elements
-    // will be block mapped to the nodes
-    auto proxy = theCollection()->construct<ReduceCol>(range);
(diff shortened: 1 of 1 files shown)


tutorial/tutorial_2b.h
```diff
@@ -113,13 +113,21 @@ static inline void collectionReduce() {
...
+
+  // Construct the collection collectively. By default, the elements will be
+  // block mapped to the nodes
+  auto proxy = theCollection()->constructCollective<ReduceCol>(
+    range, [](vt::Index1D idx){
+      return std::make_unique<ReduceCol>();
+    }
+  );
 
+  // Barrier to wait for all nodes to construct the collection
```
PhilMiller, line 127: Is having made the `constructCollective` call enough, or is there something that needs to be completed afterward?

PhilMiller: Assuming that's actually fine, this looks good to me.

lifflander: Collective construction is enough. Currently it waits for the group to be ready correctly but not for construction. This ensures that all nodes have constructed and know about the collection.

PR #768, opened 2020-04-14
767 irecv holder fix

### LINKED ISSUE #767: Fix Irecv holder problems
**Describe the bug**
Irecvs building up in holder.

lifflander:
Fixes #767

### DIFF
--- a/src/vt/messaging/irecv_holder.h
+++ b/src/vt/messaging/irecv_holder.h
@@ -98,7 +98,10 @@ struct IRecvHolder {
       MPI_Status stat;
       MPI_Test(&e.req, &flag, &stat);
 
-      if (flag == 1) {
+      if (flag == 0) {
+        ++i;
+        continue;
+      } else {
         #if backend_check_enabled(trace_enabled)
           if (ArgType::vt_trace_mpi) {
             auto tr_note = fmt::format(
@@ -113,14 +116,12 @@ struct IRecvHolder {
         e.valid = false;
 
         if (i < holder_.size() - 1) {
-          holder_[i] = std::move(holder_[holder_.size() - 1]);
+          holder_[i] = std::move(holder_[holder_.back()]);
         } else {
           ++i;
         }
 
         holder_.pop_back();
-      } else {
-        ++i;
       }
     }
 
--- a/src/vt/messaging/irecv_holder.h
+++ b/src/vt/messaging/irecv_holder.h
@@ -112,9 +112,9 @@ struct IRecvHolder {
         progress_made = true;
         e.valid = false;
 
-        if (i < holder_.size()) {
+        if (i < holder_.size() - 1) {
           holder_[i] = std::move(holder_[holder_.size() - 1]);
-          holder_.resize(holder_.size() - 1);
+          holder_.pop_back();
         } else {
           ++i;
         }
--- a/src/vt/messaging/irecv_holder.h
+++ b/src/vt/messaging/irecv_holder.h
@@ -101,28 +101,28 @@ struct IRecvHolder {
       if (flag == 0) {
         ++i;
         continue;
-      } else {
-        #if backend_check_enabled(trace_enabled)
-          if (ArgType::vt_trace_mpi) {
-            auto tr_note = fmt::format(
-              "Irecv completed: from={}", stat.MPI_SOURCE
-            );
-            trace::addUserNote(tr_note);
-          }
-        #endif
-
-        c(&e);
-        progress_made = true;
-        e.valid = false;
-
-        if (i < holder_.size() - 1) {
-          holder_[i] = std::move(holder_.back());
-        } else {
-          ++i;
-        }
-
-        holder_.pop_back();
       }
+
+      #if backend_check_enabled(trace_enabled)
+        if (ArgType::vt_trace_mpi) {
+          auto tr_note = fmt::format(
+            "Irecv completed: from={}", stat.MPI_SOURCE
+          );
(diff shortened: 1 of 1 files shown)


src/vt/messaging/irecv_holder.h
```diff
@@ -102,54 +90,44 @@ struct IRecvHolder {
...
-          progress_made = true;
-          e.valid = false;
-          holes_.push_back(i);
+      vtAssert(e.valid, "Must be valid");
+
+      int flag = 0;
+      MPI_Status stat;
+      MPI_Test(&e.req, &flag, &stat);
+
+      if (flag == 1) {
```
PhilMiller, line 101: Could you reverse this, to

```
if (flag == 0) {
  ++i;
  continue;
}
```

And then don't nest the rest of the logic?

src/vt/messaging/irecv_holder.h
```diff
@@ -102,54 +90,44 @@ struct IRecvHolder {
...
+            trace::addUserNote(tr_note);
+          }
+        #endif
+
+        c(&e);
+        progress_made = true;
+        e.valid = false;
+
+        if (i < holder_.size() - 1) {
+          holder_[i] = std::move(holder_[holder_.size() - 1]);
```
PhilMiller, line 116: This can be written as `holder_[i] = std::move(holder_.back());`

src/vt/messaging/irecv_holder.h
lifflander: Addessed

src/vt/messaging/irecv_holder.h
lifflander: Fixed

src/vt/messaging/irecv_holder.h
```diff
@@ -102,54 +90,46 @@ struct IRecvHolder {
...
+        #endif
+
+        c(&e);
+        progress_made = true;
+        e.valid = false;
+
+        if (i < holder_.size() - 1) {
+          holder_[i] = std::move(holder_.back());
+        } else {
+          ++i;
```
PhilMiller, line 121: I don't think you need the `else { ++i; }` at all - if the condition is false, then the `pop_back();` will make the `for` loop's condition false

src/vt/messaging/irecv_holder.h
```diff
@@ -102,54 +90,46 @@ struct IRecvHolder {
...
+            trace::addUserNote(tr_note);
+          }
+        #endif
+
+        c(&e);
+        progress_made = true;
+        e.valid = false;
+
+        if (i < holder_.size() - 1) {
+          holder_[i] = std::move(holder_.back());
```
PhilMiller, line 119: And I'm prettty sure the self-assignment from an r-value reference it totally fine.

Really, I think this can just be 

```c++
holder_[i] = std::move(holder_.back());
holder_.pop_back();
```

No if or else around it, at all.

src/vt/messaging/irecv_holder.h
```diff
@@ -102,54 +90,46 @@ struct IRecvHolder {
...
-          holes_.push_back(i);
+      vtAssert(e.valid, "Must be valid");
+
+      int flag = 0;
+      MPI_Status stat;
+      MPI_Test(&e.req, &flag, &stat);
+
+      if (flag == 0) {
+        ++i;
+        continue;
```
pnstickne, line 103: Since `continue`, might be more clear to omit the `else`.

src/vt/messaging/irecv_holder.h
```diff
@@ -102,54 +90,46 @@ struct IRecvHolder {
...
-
-          c(&e);
-          progress_made = true;
-          e.valid = false;
-          holes_.push_back(i);
+      vtAssert(e.valid, "Must be valid");
+
+      int flag = 0;
+      MPI_Status stat;
+      MPI_Test(&e.req, &flag, &stat);
```
pnstickne, line 99: An aside: would using `MPI_Testsome` be more efficient? Or, at least delegate such as an implementation detail?

src/vt/messaging/irecv_holder.h
lifflander: Fixed

src/vt/messaging/irecv_holder.h
lifflander: Fixed

src/vt/messaging/irecv_holder.h
lifflander: Yes, `MPI_Testany` is probably the right way to do it. However, since we still aren't storing the `MPI_Request`s contiguously in a buffer, we can't use that form. Eventually we can make this SOA to get that benefit.

src/vt/messaging/irecv_holder.h
lifflander: Addressed

src/vt/messaging/irecv_holder.h
```diff
@@ -102,54 +90,46 @@ struct IRecvHolder {
...
+      #endif
+
+      c(&e);
+      progress_made = true;
+      e.valid = false;
+
+      if (i < holder_.size() - 1) {
+        holder_[i] = std::move(holder_.back());
+      } else {
+        ++i;
```
PhilMiller, line 122: On further thought, this increment **really** can never be relevant. It only happens if we're already looking at the last element, and we're popping that last element off in the following statement. The loop ends in this case, no matter what. No need to redundantly push it off the cliff

lifflander: Addressed

PR #775, opened 2020-04-16
774 irecv test right away

### LINKED ISSUE #774: Test Irecvs right away after posting
**What Needs to be Done?**

Do an MPI_Test immediately after posting.

lifflander:
Fixes #774


PhilMiller:
That looks good to me. @nlslatt did you see any impact on performance before and after this change?

nlslatt:
> That looks good to me. @nlslatt did you see any impact on performance before and after this change?

I didn't analyze the performance impact of this change because the code was not running correctly at the time.  Would you like me to check it now?

nlslatt:
> > That looks good to me. @nlslatt did you see any impact on performance before and after this change?
> 
> I didn't analyze the performance impact of this change because the code was not running correctly at the time. Would you like me to check it now?

Also, I had only implemented the immediate test on data msgs, not active msgs.

PR #779, opened 2020-04-16
778 always flush

### LINKED ISSUE #778: Add flag for flush by default
**What Needs to be Done?**

Add a new flag to tell VT to always flush stdout for debug/normal VT prints.

lifflander:
Fixes #778

Runtime implementation based on needs in EMPIRE.

### DIFF
--- a/src/vt/configs/debug/debug_print.h
+++ b/src/vt/configs/debug/debug_print.h
@@ -160,6 +160,10 @@ namespace vt { namespace debug {
 NodeType preNode();
 }} /* end namespace vt::debug */
 
+namespace vt { namespace arguments {
+bool alwaysFlush();
+}} /* end namespace vt::arguments */
+
 namespace vt { namespace config {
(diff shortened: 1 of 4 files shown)


src/vt/configs/debug/debug_print.h
```diff
@@ -177,7 +181,7 @@ static inline void debugPrintImpl(NodeType node, Arg&& arg, Args&&... args) {
       vt_print_colorize_impl(::vt::debug::green(), PrettyPrintCat<cat>::print(), ":"),
       user
     );
-    if (vt_option_check_enabled(mod, ModeEnum::flush)) {
+    if (vt_option_check_enabled(mod, ModeEnum::flush) or arguments::alwaysFlush()) {
```
pnstickne, line 184: I'd enable always-flush, regardless of any flag if EMPIRE is going to (always) run with this flag.

PR #783, opened 2020-04-20
764 abstract components

### LINKED ISSUE #764: Abstract out runtime component pieces to enable better diagnostics
**What Needs to be Done?**

Make general interfaces for components so diagnostics can be collected.

lifflander:
Fixes #764

- Create a new abstract interface for VT components, universalizing `initialize` and `finalize`.
- Add new type-level dependency management between components to ensure valid startup/finalization order
  - Use topological sort to determine valid order
  - Note: the order is not necessary deterministic across platforms/compilers, but is always the same across all nodes for a given binary and always respects declared dependences
- Create abstraction for `PollableComponent<T>` inheriting from `Progressable` so scheduler knows which components to poll automatically
- Stub abstraction for diagnostics (eventually with have a universal interface for dumping state and calculating diagnostics values at shutdown, e.g., largest or average message size)
  - **Diagnostics will arrive in another PR**
- Stub abstraction for universal buffering for VT components
  - **Universal buffering will arrive in another PR**
- Add base component overload for revealing a name `std::string` for each component for use in debugging and later diagnostics

TODO:
  - [x] Add doxygen
  - [x] Cleanup runtime
  - [x] Hookup `Progressable` to scheduler component

### DIFF
--- a/src/vt/messaging/active.cc
+++ b/src/vt/messaging/active.cc
@@ -905,7 +905,7 @@ bool ActiveMessenger::testPendingDataMsgAsyncRecv() {
   );
 }
 
-bool ActiveMessenger::progress() {
+int ActiveMessenger::progress() {
   bool const started_irecv_active_msg = tryProcessIncomingActiveMsg();
   bool const started_irecv_data_msg = tryProcessDataMsgRecv();
   processMaybeReadyHanTag();
--- a/src/vt/objgroup/manager.h
+++ b/src/vt/objgroup/manager.h
@@ -46,6 +46,7 @@
 #define INCLUDED_VT_OBJGROUP_MANAGER_H
 
 #include "vt/config.h"
+#include "vt/runtime/component/component_pack.h"
 #include "vt/objgroup/common.h"
 #include "vt/objgroup/manager.fwd.h"
 #include "vt/objgroup/proxy/proxy_objgroup.h"
@@ -65,7 +66,7 @@
 
 namespace vt { namespace objgroup {
 
-struct ObjGroupManager {
+struct ObjGroupManager : runtime::component::PollableComponent<ObjGroupManager> {
   template <typename ObjT>
   using ProxyType           = proxy::Proxy<ObjT>;
   template <typename ObjT>
@@ -170,7 +171,7 @@ struct ObjGroupManager {
   /*
    * Run the progress function to push along postponed events (such as self sends)
    */
-  bool progress();
+  int progress() override;
 
   /*
    * Untyped calls for broadcasting or sending msgs to an obj group
--- a/src/vt/objgroup/manager.h
+++ b/src/vt/objgroup/manager.h
@@ -82,6 +82,13 @@ struct ObjGroupManager : runtime::component::Component<ObjGroupManager> {
 
   ObjGroupManager() = default;
 
+  /**
+   * \brief Return the name of the component
+   *
+   * \return the name
+   */
+  std::string name() override { return "ObjGroupManager"; }
+
   /*
    * Creation of a new object group across the distributed system. For now,
    * these use the default group which includes all the nodes in the
--- a/src/vt/objgroup/manager.h
+++ b/src/vt/objgroup/manager.h
@@ -66,7 +66,7 @@
 
 namespace vt { namespace objgroup {
 
-struct ObjGroupManager : runtime::component::PollableComponent<ObjGroupManager> {
+struct ObjGroupManager : runtime::component::Component<ObjGroupManager> {
   template <typename ObjT>
   using ProxyType           = proxy::Proxy<ObjT>;
   template <typename ObjT>
@@ -167,11 +167,6 @@ struct ObjGroupManager : runtime::component::PollableComponent<ObjGroupManager> 
    * Dispatch to a live obj group pointer with a handler
    */
   void dispatch(MsgVirtualPtrAny msg, HandlerType han);
-
-  /*
-   * Run the progress function to push along postponed events (such as self sends)
-   */
-  int progress() override;
 
   /*
    * Untyped calls for broadcasting or sending msgs to an obj group
--- a/src/vt/objgroup/manager.h
+++ b/src/vt/objgroup/manager.h
@@ -82,11 +82,6 @@ struct ObjGroupManager : runtime::component::Component<ObjGroupManager> {
 
   ObjGroupManager() = default;
 
-  /**
-   * \brief Return the name of the component
-   *
-   * \return the name
-   */
   std::string name() override { return "ObjGroupManager"; }
 
   /*
--- a/src/vt/parameterization/parameterization.h
+++ b/src/vt/parameterization/parameterization.h
@@ -130,6 +130,13 @@ static void dataMessageHandler(DataMsg<Tuple>* msg) {
 
 struct Param : runtime::component::Component<Param> {
 
+  /**
+   * \brief Return the name of the component
+   *
+   * \return the name
+   */
+  std::string name() override { return "Param"; }
+
   template <typename... Args>
   void sendDataTuple(
     NodeType const& dest, HandlerType const& han, std::tuple<Args...>&& tup
--- a/src/vt/parameterization/parameterization.h
+++ b/src/vt/parameterization/parameterization.h
@@ -130,11 +130,6 @@ static void dataMessageHandler(DataMsg<Tuple>* msg) {
 
 struct Param : runtime::component::Component<Param> {
 
-  /**
-   * \brief Return the name of the component
-   *
-   * \return the name
-   */
   std::string name() override { return "Param"; }
 
   template <typename... Args>
--- a/src/vt/parameterization/parameterization.h
+++ b/src/vt/parameterization/parameterization.h
@@ -51,6 +51,7 @@
 #include "vt/registry/auto/auto_registry_interface.h"
 #include "vt/utils/static_checks/all_true.h"
 #include "vt/parameterization/param_meta.h"
+#include "vt/runtime/component/component_pack.h"
 
 #include <tuple>
 #include <utility>
@@ -127,7 +128,7 @@ static void dataMessageHandler(DataMsg<Tuple>* msg) {
 #endif
 }
 
-struct Param {
+struct Param : runtime::component::Component<Param> {
 
   template <typename... Args>
   void sendDataTuple(
--- a/src/vt/runtime/component/component.h
+++ b/src/vt/runtime/component/component.h
@@ -177,7 +177,7 @@ struct PollableComponent : Component<T> {
    * \return number of units processed---zero
    */
   virtual int progress() override {
-    vtAbort("PollableComponent should have a progress function");
+    vtAbort("PollableComponent should override the empty progress function");
     return 0;
   }
 
--- a/src/vt/runtime/component/component.h
+++ b/src/vt/runtime/component/component.h
@@ -55,56 +55,127 @@
 
 namespace vt { namespace runtime { namespace component {
 
+/**
+ * \struct Component component.h vt/runtime/component/component.h
+ *
+ * \brief \c Component class for a generic VT runtime module, CRTP'ed over the
+ * component's actual type
+ */
 template <typename T>
 struct Component : BaseComponent {
 
   Component() = default;
 
+  /**
+   * \brief Constructor for a new component with the appropriate typed
+   * dependencies
+   *
+   * \param[in] DepsPack a pack of typed dependencies for the given component
+   * that it depends on
+   *
+   * \return the new component
+   */
   template <typename... Deps>
   Component(DepsPack<Deps...>) {
     ComponentRegistry::dependsOn<T, Deps...>();
   }
 
-  /// Traits for objgroup components which have a specialized static construct
+  /// Trait for components that have a specialized static construct method (used
+  /// for objgroups)
   template <typename U>
   using hasCons = typename std::enable_if<ComponentTraits<U>::hasConstruct, T>::type;
+
+  /// Trait for components that do not have a specialized static construct method
   template <typename U>
   using hasNoCons = typename std::enable_if<not ComponentTraits<U>::hasConstruct, T>::type;
 
+  /**
+   * \brief Construct the component with the specialized construct method
+   *
+   * \param[in] args the arguments to forward to the component's constructor
+   *
+   * \return unique pointer to the the component
+   */
   template <typename... Args, typename U = T>
   static std::unique_ptr<T> staticInit(Args&&... args, hasCons<U>* = nullptr) {
     return T::construct(std::forward<Args>(args)...);
   }
 
+  /**
+   * \brief Construct the component with normal component constructor
+   *
+   * \param[in] args the arguments to forward to the component's constructor
+   *
+   * \return unique pointer to the the component
+   */
   template <typename... Args, typename U = T>
   static std::unique_ptr<T> staticInit(Args&&... args, hasNoCons<U>* = nullptr) {
     return std::make_unique<T>(std::forward<Args>(args)...);
   }
 
+  /**
+   * \brief Whether the component requires the scheduler to poll
+   *
+   * \return default component is not pollable
+   */
   bool pollable() override {
     return false;
   }
 
+  /**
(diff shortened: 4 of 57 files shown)


lifflander:


```diff

Clones added
============
- src/vt/messaging/active.cc  2
- tests/unit/memory/test_memory_active.cc  2
- src/vt/worker/worker_group.h  1
- src/vt/trace/trace.cc  2
- src/vt/worker/worker_group_omp.h  1
- src/vt/sequence/sequencer.impl.h  3
- src/vt/pool/pool.cc  2
- src/vt/messaging/active.h  4
- src/vt/parameterization/parameterization.h  2
- src/vt/runtime/runtime.cc  2
- src/vt/rdma/rdma.h  6
- src/vt/trace/trace.h  2
- tests/unit/memory/test_memory_lifetime.cc  11
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5390178&bid=17670967)

lifflander:


```diff

Clones added
============
- src/vt/messaging/active.h  4
- src/vt/trace/trace.cc  2
- tests/unit/memory/test_memory_active.cc  2
- src/vt/worker/worker_group_omp.h  1
- tests/unit/memory/test_memory_lifetime.cc  11
- src/vt/worker/worker_group.h  1
- src/vt/parameterization/parameterization.h  2
- src/vt/trace/trace.h  2
- src/vt/runtime/runtime.cc  2
- src/vt/messaging/active.cc  2
- src/vt/pool/pool.cc  2
- src/vt/sequence/sequencer.impl.h  3
- src/vt/rdma/rdma.h  6
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5390178&bid=17670967)

lifflander:
@PhilMiller @nlslatt @pnstickne  Please review!

lifflander:


```diff

Clones added
============
- tests/unit/memory/test_memory_lifetime.cc  11
- src/vt/trace/trace.h  2
- src/vt/pool/pool.cc  2
- src/vt/worker/worker_group_omp.h  1
- src/vt/sequence/sequencer.impl.h  3
- src/vt/rdma/rdma.h  6
- src/vt/worker/worker_group.h  1
- src/vt/messaging/active.h  4
- src/vt/parameterization/parameterization.h  2
- src/vt/messaging/active.cc  2
- src/vt/runtime/runtime.cc  2
- tests/unit/memory/test_memory_active.cc  2
- src/vt/trace/trace.cc  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5390178&bid=17670967)

src/vt/runtime/component/component.h
```diff
@@ -0,0 +1,188 @@
...
+   * \brief Empty default overridden progress method
+   *
+   * \return no units processed
+   */
+  virtual int progress() override { return 0; }
+
+  /**
+   * \brief Empty default diagnostic dump state
+   */
+  void dumpState() override { }
```
pnstickne, line 150: `virtual`?

src/vt/runtime/component/component_pack.impl.h
```diff
@@ -0,0 +1,98 @@
...
+namespace vt { namespace runtime { namespace component {
+
+template <typename T, typename... Deps, typename... Cons>
+registry::AutoHandlerType ComponentPack::registerComponent(
+  T** ref, typename BaseComponent::DepsPack<Deps...>, Cons&&... cons
+) {
+  ComponentRegistry::dependsOn<T, Deps...>();
+  auto idx = registry::makeIdx<T>();
+  registered_components_.push_back(idx);
+  construct_components_.emplace(
```
pnstickne, line 59: Would separate out [nested] creation and adding to components.

src/vt/runtime/runtime_component_fwd.h
```diff
@@ -97,6 +97,9 @@ struct PipeManager;
 namespace objgroup {
 struct ObjGroupManager;
 }
+namespace util { namespace memory {
```
pnstickne, line 100: New line for consistency.

src/vt/trace/trace.cc
```diff
@@ -172,52 +171,48 @@ bool Trace::inIdleEvent() const {
...
-  std::string const& in_prog_name, std::string const& in_trace_name,
-  std::string const& in_dir_name
-) {
-  auto const node = theContext()->getNode();
-
-  prog_name_ = in_prog_name;
-  trace_name_ = in_trace_name;
-  start_time_ = getCurrentTime();
+void Trace::setupNames(std::string const& in_prog_name) {
+  if (ArgType::vt_trace) {
```
pnstickne, line 175: Might be cleaner to if-not-return.

src/vt/objgroup/manager.h
```diff
@@ -81,6 +82,13 @@ struct ObjGroupManager {
 
   ObjGroupManager() = default;
 
+  /**
```
pnstickne, line 85: I'd be fine with removing these comments, as they are not applied consistently.
Then unified comments can be added to the base type(s).

src/vt/parameterization/parameterization.h
```diff
@@ -116,7 +117,14 @@ static void dataMessageHandler(DataMsg<Tuple>* msg) {
 #endif
 }
 
-struct Param {
+struct Param : runtime::component::Component<Param> {
+
+  /**
```
pnstickne, line 122: Ditto.

src/vt/runtime/component/component.h
```diff
@@ -0,0 +1,188 @@
...
+  static std::unique_ptr<T> staticInit(Args&&... args, hasNoCons<U>* = nullptr) {
+    return std::make_unique<T>(std::forward<Args>(args)...);
+  }
+
+  /**
+   * \brief Whether the component requires the scheduler to poll
+   *
+   * \return default component is not pollable
+   */
+  bool pollable() override {
```
pnstickne, line 121: Add unified comments to base/virtual and allow inherit?

src/vt/runtime/component/component_pack.cc
```diff
@@ -0,0 +1,164 @@
...
+*/
+
+#if !defined INCLUDED_VT_RUNTIME_COMPONENT_COMPONENT_PACK_CC
+#define INCLUDED_VT_RUNTIME_COMPONENT_COMPONENT_PACK_CC
+
+#include "vt/runtime/component/component_pack.h"
+
+namespace vt { namespace runtime { namespace component {
+
+void ComponentPack::construct() {
```
pnstickne, line 52: While neat, might be overkill as there is a fixed set of VT deps, and VT is itself not a dependency management library to clients.

src/vt/runtime/component/component.h
lifflander: Fixed

src/vt/trace/trace.cc
lifflander: Done

src/vt/objgroup/manager.h
lifflander: Fixed

src/vt/runtime/component/component_pack.cc
lifflander: I think as we add more components to VT it might be useful. Also, there was a lot of tricky code before to get the startup order correct. This solves it in a uniform way

lifflander:
@PhilMiller Please review!

PhilMiller:
In progress

src/vt/runtime/component/progressable.h
```diff
@@ -0,0 +1,58 @@
...
+
+#if !defined INCLUDED_VT_RUNTIME_COMPONENT_PROGRESSABLE_H
+#define INCLUDED_VT_RUNTIME_COMPONENT_PROGRESSABLE_H
+
+#include "vt/config.h"
+
+namespace vt { namespace runtime { namespace component {
+
+struct Progressable {
+  virtual int progress() = 0;
```
PhilMiller, line 57: This needs a comment to document the meaning of the return value.

src/vt/runtime/runtime.cc
```diff
@@ -1190,62 +1189,221 @@ void Runtime::finalizeContext() {
...
+  p.add<param::Param>();
+  p.add<seq::Sequencer>();
+  p.add<seq::SequencerVirtual>();
+  p.add<location::LocationManager>();
+  p.add<vrt::VirtualContextManager>();
+  p.add<vrt::collection::CollectionManager>();
+  p.add<rdma::Manager>();
+  p.add<registry::Registry>();
+  p.add<event::AsyncEvent>();
+  p.add<pool::Pool>();
```
PhilMiller: Why are there separate `registerComponent` and `add` calls? These seem redundant.

Also, there's at least one inconsistency between `rdma::Manager` and `rdma::RDMAManager`

src/vt/runtime/runtime.cc
```diff
@@ -1012,9 +1012,7 @@ bool Runtime::finalize(bool const force_now) {
     fflush(stdout);
     fflush(stderr);
     sync();
-    finalizeComponents();
-    finalizeOptionalComponents();
-    finalizeTrace();
+    p_ = nullptr;
```
PhilMiller, line 1020: Stylistically, I think it's slightly preferable to say `p_.reset(nullptr)`, but that's a nitpick

src/vt/runtime/runtime.h
```diff
@@ -180,7 +180,7 @@ struct Runtime {
   bool runtime_active_ = false;
   bool is_interop_ = false;
   WorkerCountType num_workers_ = no_workers;
-  MPI_Comm* communicator_ = nullptr;
```
PhilMiller, line 187: I don't think there's actually a need to handle an `MPI_Comm` as a pointer - they're typically already handles to an MPI-internal data structure, and so it makes sense to pass them around by value.

If you were doing some RAII thing of duplicating on construction, and freeing on destruction, this would be a bit more sensible. I don't think there's a need for that in this circumstance.

PhilMiller: There's a sentinel value `MPI_COMM_NULL` defined in the standard.

src/vt/runtime/runtime.h
```diff
@@ -81,6 +81,7 @@ struct Runtime {
 
   virtual ~Runtime();
 
+  bool live() const { return p_ and p_->live(); }
```
PhilMiller, line 84: This should be named `isLive()`

lifflander:
> It would be wise, I think, for the component interface to have a method `std::string getComponentName()`, so that debug prints during startup and shutdown can use that automatically, and less needs to be baked into each individually

I don't think I understand. How do we automatically get the name without having an overload per component? typeid?

src/vt/runtime/component/progressable.h
lifflander: Added

src/vt/runtime/runtime.cc
lifflander: `registerComponent` does static registration with dependencies, but does not add it to the construct set. When a component is added, it will be constructed along with all its dependencies. 

I realize that the distinction isn't needed right now, but thought it might be useful in the future. Say you want a customized runtime with just collection functionality, you could just add collection and get that plus everything it needs to run properly.

Regarding `rdma::Manager` vs `rdma::RDMAManager` that's an unfortunate results of having two very different rdma components in the system right now. Eventually, that will get resolved.

src/vt/runtime/runtime.cc
lifflander: Done

src/vt/runtime/runtime.h
lifflander: Done

src/vt/runtime/runtime.h
lifflander: I can get rid of the unique_ptr there. The reason it was done is because it gets potentially modified by the context, but I can just pass the memory address.

lifflander: Addressed

src/vt/messaging/active.cc
```diff
@@ -905,7 +905,7 @@ bool ActiveMessenger::testPendingDataMsgAsyncRecv() {
   );
 }
 
-bool ActiveMessenger::progress() {
+int ActiveMessenger::progress() {
```
PhilMiller, line 908: This is probably OK, but it isn't strictly satisfying the contract of a "count" of units of progress made.

src/vt/worker/worker_group_omp.cc
```diff
@@ -81,8 +81,9 @@ bool WorkerGroupOMP::commScheduler() {
   return WorkerGroupComm::schedulerComm(finished_fn_);
 }
 
-void WorkerGroupOMP::progress() {
+int WorkerGroupOMP::progress() {
   WorkerGroupCounter::progress();
+  return 0;
```
PhilMiller, line 86: Is it safe for a component to uniformly return 'no progress'?

src/vt/vrt/collection/manager.h
```diff
@@ -90,7 +91,9 @@
 
 namespace vt { namespace vrt { namespace collection {
 
-struct CollectionManager {
+struct CollectionManager
+  : runtime::component::PollableComponent<CollectionManager>
```
PhilMiller, line 95: Should this be a non-pollable `Component`, since `progress` always returns 0?

src/vt/trace/trace.cc
```diff
@@ -172,17 +171,15 @@ bool Trace::inIdleEvent() const {
...
+  }
 
-  prog_name_ = in_prog_name;
-  trace_name_ = in_trace_name;
-  start_time_ = getCurrentTime();
+  auto const node = theContext()->getNode();
 
-  std::string dir_name = (in_dir_name.empty()) ? prog_name_ + "_trace" : in_dir_name;
+  trace_name_ = prog_name_ + "." + std::to_string(node) + ".log.gz";
+  auto dir_name = prog_name_ + "_trace";
```
PhilMiller, line 182: I'm not sure why this change of logic is coming with this PR, but it doesn't seem objectionable as such.

src/vt/sequence/sequencer_fwd.h
```diff
@@ -50,15 +50,13 @@
 
 namespace vt { namespace seq {
 
-template <typename SeqTag, template <typename> class SeqTrigger>
-struct TaggedSequencer;
+struct Sequencer;
 
-using Sequencer = TaggedSequencer<SeqType, SeqMigratableTriggerType>;
+//using Sequencer = TaggedSequencer<SeqType, SeqMigratableTriggerType>;
```
PhilMiller, line 55: Why not just delete these, rather than comment them out?

PhilMiller:
> > It would be wise, I think, for the component interface to have a method `std::string getComponentName()`, so that debug prints during startup and shutdown can use that automatically, and less needs to be baked into each individually
> 
> I don't think I understand. How do we automatically get the name without having an overload per component? typeid?

Never mind, I see that a later commit actually added a `name` method. That's all I was asking for.

My original comment was made as I was reviewing commit-by-commit, since they seemed to form a decent progression.

PhilMiller:
Assuming none of my last few comments indicate a real issue to you, this has my +1

src/vt/trace/trace.cc
lifflander: It used to be in the runtime and I had to move it over to Trace where it should be now that Runtime avoids component-specific logic.

src/vt/messaging/active.cc
lifflander: Yes, right now the return value is not used. Eventually it might be, when that happens we'll make all these return the correct values.

src/vt/worker/worker_group_omp.cc
lifflander: Yes, it is. Eventually we can make these return correctly if needed.

src/vt/vrt/collection/manager.h
lifflander: Done, fixed

src/vt/sequence/sequencer_fwd.h
lifflander: My mistake

PR #784, opened 2020-04-20
Add subphase setting API for application callers

PhilMiller:
The final include cleanup was necessary to make this API possible. The earlier ones got some distractions out of the way so I could find the solution.

### DIFF
--- a/src/vt/collective/reduce/operators/default_op.impl.h
+++ b/src/vt/collective/reduce/operators/default_op.impl.h
@@ -47,7 +47,6 @@
 
 #include "vt/config.h"
 #include "vt/collective/reduce/operators/default_op.h"
-#include "vt/pipe/pipe_headers.h"
 
 namespace vt { namespace collective { namespace reduce { namespace operators {
 
--- a/src/vt/termination/termination.cc
+++ b/src/vt/termination/termination.cc
@@ -57,6 +57,7 @@
 #include "vt/configs/arguments/args.h"
 #include "vt/configs/debug/debug_colorize.h"
 #include "vt/collective/collective_alg.h"
+#include "vt/pipe/pipe_headers.h"
 
 #include <memory>
 
--- a/src/vt/vrt/proxy/collection_proxy.h
+++ b/src/vt/vrt/proxy/collection_proxy.h
@@ -109,6 +109,12 @@ struct CollectionProxy : ProxyCollectionTraits<ColT, IndexT> {
   >
   ElmProxyType operator()(IndexU const& idx) const;
 
+  /**
+   * Sets a subphase from which load records should be considered, in
+   * place of the summed load over the entire phase.
+   *
+   * This must be called on every process
+   */
   void setFocusedSubPhase(balance::ElementStats::SubphaseType subphase);
 };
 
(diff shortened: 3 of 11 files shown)


PhilMiller:
I want this in the next beta, so that experiments in EMPIRE can use it.

src/vt/vrt/proxy/collection_proxy.h
```diff
@@ -107,6 +108,8 @@ struct CollectionProxy : ProxyCollectionTraits<ColT, IndexT> {
     >::type
   >
   ElmProxyType operator()(IndexU const& idx) const;
+
+  void setFocusedSubPhase(balance::ElementStats::SubphaseType subphase);
```
PhilMiller, line 112: This is probably breaking some abstraction/layering boundary putting it here, but it's not clear that there's a better place for it that doesn't make dependency structure even worse

PhilMiller, line 112: I need a comment on this noting that it should be called on *every* process

PhilMiller:
@lifflander review please?

src/vt/collective/reduce/operators/default_op.impl.h
```diff
@@ -47,7 +47,6 @@
 
 #include "vt/config.h"
 #include "vt/collective/reduce/operators/default_op.h"
-#include "vt/pipe/pipe_headers.h"
```
lifflander, line 50: The reason this was here was because of the `cb.send` if a reduce contains a callback. Should be fine to push down to uses in VT that invoke a callback.

src/vt/termination/termination.cc
```diff
@@ -57,6 +57,7 @@
 #include "vt/configs/arguments/args.h"
 #include "vt/configs/debug/debug_colorize.h"
 #include "vt/collective/collective_alg.h"
+#include "vt/pipe/pipe_headers.h"
```
lifflander, line 60: Terminated started using callbacks recently... makes sense

src/vt/vrt/proxy/collection_proxy.h
lifflander: Yes, typically these are put in place with traits classes for the proxy.  But this is acceptable given how horrid the dependencies can be.

I have some ideas to completely refactor all the dependencies in VT across callbacks/reduce/proxy etc once I get some time in the future.

lifflander:
This isn't going to drop into beta.8. 1.0.0 has different header file includes that we'll have to change a little to get this change working.

PhilMiller:
Once this is good for `develop`, I'll work out the corresponding header changes necessary on beta.8

PhilMiller:
Uhhh, it looks like this actually got merged to beta.7 and shipped. It's not on develop, though.

PhilMiller:
Dunno how this ought to be tagged, at this point

lifflander:
Just keep it untagged at this point. Please create the PRs for develop and mark them for release. I will do the rebasing without an PR into proposed-proposed-update to test it. Otherwise, the flow gets really complex.

lifflander:
@PhilMiller Can you rebase this and merge?

PR #795, opened 2020-04-29
793 Safe MPI collectives

### LINKED ISSUE #793: Add ability to schedule collective MPI calls correctly from within VT
**What Needs to be Done?**

- [x] Implement an API for this
- [x] Add ability to scheduler for consensus
- [ ] Use the API where MPI collectives are used in VT (VT groups, etc.)

lifflander:
Fixes #793 

Implement a new call to enqueue an MPI collective and run it safely **while** VT is running.

```C++
auto tag = vt::theCollective()->mpiCollective([]{
   MPI_Barrier(MPI_COMM_WORLD);
});

vt::theCollective()->waitCollective(tag);
```

### DIFF
--- a/src/vt/collective/collective_alg.cc
+++ b/src/vt/collective/collective_alg.cc
@@ -144,7 +144,8 @@ CollectiveAlg::CollectiveAlg()
 
 
 TagType CollectiveAlg::mpiCollectiveAsync(ActionType action) {
-  auto tag = next_tag_;
+  auto tag = next_tag_++;
+
   CollectiveInfo info(tag, action);
 
   planned_collective_.emplace(
@@ -171,7 +172,6 @@ TagType CollectiveAlg::mpiCollectiveAsync(ActionType action) {
     collective_root, msg.get(), cb, tag, no_seq_id, 1, ident
   );
 
-  next_tag_++;
   return tag;
 }
 
--- a/src/vt/collective/collective_alg.cc
+++ b/src/vt/collective/collective_alg.cc
@@ -147,6 +147,7 @@ CollectiveAlg::CollectiveAlg()
 TagType CollectiveAlg::mpiCollectiveAsync(ActionType action) {
   auto tag = next_tag_++;
 
+  // Create a new collective action with the next tag
   CollectiveInfo info(tag, action);
 
   planned_collective_.emplace(
@@ -161,6 +162,11 @@ TagType CollectiveAlg::mpiCollectiveAsync(ActionType action) {
     tag
   );
 
+  // Do a reduction followed by a broadcast to trigger a collective
+  // operation. Note that in VT reductions and broadcasts can be executed out of
+  // order. This implies that runCollective might be called with different tags
+  // on different nodes. Thus, in runCollective, we will use a consensus
+  // protocol to agree on a consistent tag across all the nodes.
   NodeType collective_root = 0;
 
   auto cb = theCB()->makeBcast<CollectiveMsg,&runCollective>();
--- a/src/vt/collective/collective_alg.cc
+++ b/src/vt/collective/collective_alg.cc
@@ -52,4 +52,141 @@ CollectiveAlg::CollectiveAlg()
     barrier::Barrier()
 { }
 
+/*static*/ void CollectiveAlg::runCollective(CollectiveMsg* msg) {
+  // We need a reentrancy counter to ensure that this is only on the scheduler
+  // stack once!
+  static int reenter_counter_ = 0;
+
+  if (reenter_counter_ == 0) {
+    reenter_counter_++;
+
+    auto const this_node = theContext()->getNode();
+    TagType consensus_tag = no_tag;
+    ActionType action = no_action;
+
+    auto& planned = theCollective()->planned_collective_;
+
+    // The root decides the next tag and tells the remaining nodes
+    if (msg->root_ == this_node) {
+      debug_print(
+        gen, node,
+        "mpiCollective: running collective associated with tag={}\n",
+        msg->tag_
+      );
+
+      auto iter = planned.find(msg->tag_);
+      vtAssert(iter != planned.end(), "Planned collective does not exist");
+      consensus_tag = msg->tag_;
+      action = iter->second.action_;
+    }
+
+    // We can not use a VT broadcast here because it involves the scheduler and
+    // MPI progress and there's no safe way to guarantee that the message has been
+    // received and processed and doesn't require progress from another node
+
+    int const root = msg->root_;
+    auto comm = theContext()->getComm();
+    MPI_Request req;
+
+    // Do a async broadcast of the tag we intend to execute collectively
+    MPI_Ibcast(&consensus_tag, 1, MPI_INT32_T, root, comm, &req);
+
+    int flag = 0;
+    while (not flag) {
+      MPI_Test(&req, &flag, MPI_STATUS_IGNORE);
+      runScheduler();
+    }
+
+    // We need a barrier here so the root doesn't finish the broadcast first and
+    // then enter the collective. We could replace the Ibcast/Ibarrier with a
+    // Iallreduce, but I think this is cheaper
+    MPI_Ibarrier(comm, &req);
+
+    flag = 0;
+    while (not flag) {
+      MPI_Test(&req, &flag, MPI_STATUS_IGNORE);
+      runScheduler();
+    }
+
+    debug_print(
+      gen, node,
+      "mpiCollective: consensus_tag={}, msg->tag_={}\n",
+      consensus_tag, msg->tag_
+    );
+
+    // The root had a different collective in mind... let's do that one
+    if (msg->root_ != this_node) {
+      auto iter = planned.find(consensus_tag);
+      vtAssert(iter != planned.end(), "Planned collective does not exist");
+      action = iter->second.action_;
+    }
+
+    // Run the collective safely
+    action();
+
+    // Erase the tag that was actually executed
+    planned.erase(planned.find(consensus_tag));
+
+    reenter_counter_--;
+
+    // Recurse, running a postponed collective
+    auto& postponed = theCollective()->postponed_collectives_;
+    if (postponed.size() > 0) {
+      auto next_msg = postponed.back();
+      postponed.pop_back();
+      runCollective(next_msg.get());
+    }
+  } else {
+    // Postpone any collective while the current one resolves
+    theCollective()->postponed_collectives_.emplace_back(promoteMsg(msg));
+  }
+}
+
+
+TagType CollectiveAlg::mpiCollective(ActionType action) {
+  auto tag = next_tag_;
+  CollectiveInfo info(tag, action);
+
+  planned_collective_.emplace(
+    std::piecewise_construct,
+    std::forward_as_tuple(tag),
+    std::forward_as_tuple(info)
+  );
+
+  debug_print(
+    gen, node,
+    "mpiCollective: new MPI collective with tag={}\n",
+    tag
+  );
+
+  NodeType collective_root = 0;
+
+  auto cb = theCB()->makeBcast<CollectiveMsg,&runCollective>();
+
+  // Put this in a separate namespace
+  // @todo: broader issue: need to implement a better way to scope reductions
+  auto ident = 0xEFFFFFFFFFFFFFFF;
+  auto msg = makeMessage<CollectiveMsg>(tag, collective_root);
+  this->reduce<collective::None>(
+    collective_root, msg.get(), cb, tag, no_seq_id, 1, ident
+  );
+
+  next_tag_++;
+  return tag;
+}
+
+bool CollectiveAlg::isCollectiveDone(TagType tag) {
+  return planned_collective_.find(tag) == planned_collective_.end();
+}
+
+void CollectiveAlg::waitCollective(TagType tag) {
+  while (not isCollectiveDone(tag)) {
+    runScheduler();
+  }
+}
+
+void CollectiveAlg::mpiCollectiveWait(ActionType action) {
+  waitCollective(mpiCollective(action));
+}
+
 }}  // end namespace vt::collective
--- a/src/vt/collective/collective_alg.cc
+++ b/src/vt/collective/collective_alg.cc
@@ -143,7 +143,7 @@ CollectiveAlg::CollectiveAlg()
 }
 
 
-TagType CollectiveAlg::mpiCollective(ActionType action) {
+TagType CollectiveAlg::mpiCollectiveAsync(ActionType action) {
   auto tag = next_tag_;
   CollectiveInfo info(tag, action);
 
@@ -155,7 +155,7 @@ TagType CollectiveAlg::mpiCollective(ActionType action) {
 
   debug_print(
     gen, node,
-    "mpiCollective: new MPI collective with tag={}\n",
+    "mpiCollectiveAsync: new MPI collective with tag={}\n",
     tag
   );
 
@@ -186,7 +186,7 @@ void CollectiveAlg::waitCollective(TagType tag) {
 }
 
 void CollectiveAlg::mpiCollectiveWait(ActionType action) {
-  waitCollective(mpiCollective(action));
+  waitCollective(mpiCollectiveAsync(action));
 }
 
 }}  // end namespace vt::collective
--- a/src/vt/collective/collective_alg.cc
+++ b/src/vt/collective/collective_alg.cc
@@ -66,9 +66,7 @@ CollectiveScope CollectiveAlg::makeCollectiveScope(TagType in_scope_tag) {
   auto iter = scopes.find(scope_tag);
   vtAssert(iter == scopes.end(), "Scope must not already exist");
   // Must construct in call to make_unique, since constructor is private
-  auto ptr = std::make_unique<detail::ScopeImpl>(
-    detail::ScopeImpl{is_user_tag, scope_tag}
-  );
+  auto ptr = std::make_unique<detail::ScopeImpl>(detail::ScopeImpl{});
   scopes[scope_tag] = std::move(ptr);
 
   return CollectiveScope(is_user_tag, scope_tag);
--- a/src/vt/collective/collective_alg.cc
+++ b/src/vt/collective/collective_alg.cc
@@ -67,7 +67,6 @@ CollectiveAlg::CollectiveAlg()
 
   auto const this_node = theContext()->getNode();
   TagType consensus_tag = no_tag;
-  ActionType action = no_action;
 
   auto& planned = theCollective()->planned_collective_;
 
@@ -82,7 +81,6 @@ CollectiveAlg::CollectiveAlg()
     auto iter = planned.find(msg->tag_);
     vtAssert(iter != planned.end(), "Planned collective does not exist");
     consensus_tag = msg->tag_;
-    action = iter->second.action_;
   }
 
   // We can not use a VT broadcast here because it involves the scheduler and
@@ -102,6 +100,8 @@ CollectiveAlg::CollectiveAlg()
     runScheduler();
   }
 
+  vtAssert(consensus_tag != no_tag, "Selected tag must be valid");
+
   // We need a barrier here so the root doesn't finish the broadcast first and
   // then enter the collective. We could replace the Ibcast/Ibarrier with a
   // Iallreduce, but I think this is cheaper
@@ -119,18 +119,16 @@ CollectiveAlg::CollectiveAlg()
     consensus_tag, msg->tag_
   );
 
-  // The root had a different collective in mind... let's do that one
-  if (msg->root_ != this_node) {
-    auto iter = planned.find(consensus_tag);
-    vtAssert(iter != planned.end(), "Planned collective does not exist");
-    action = iter->second.action_;
-  }
+  // Execute the action that the root selected
+  auto iter = planned.find(consensus_tag);
+  vtAssert(iter != planned.end(), "Planned collective does not exist");
+  auto action = iter->second.action_;
 
   // Run the collective safely
   action();
 
   // Erase the tag that was actually executed
-  planned.erase(planned.find(consensus_tag));
+  planned.erase(iter);
 
   reenter_counter_--;
 
--- a/src/vt/collective/collective_alg.cc
+++ b/src/vt/collective/collective_alg.cc
@@ -57,88 +57,89 @@ CollectiveAlg::CollectiveAlg()
   // stack once!
   static int reenter_counter_ = 0;
 
-  if (reenter_counter_ == 0) {
-    reenter_counter_++;
-
-    auto const this_node = theContext()->getNode();
-    TagType consensus_tag = no_tag;
-    ActionType action = no_action;
-
-    auto& planned = theCollective()->planned_collective_;
-
-    // The root decides the next tag and tells the remaining nodes
-    if (msg->root_ == this_node) {
-      debug_print(
-        gen, node,
-        "mpiCollective: running collective associated with tag={}\n",
-        msg->tag_
-      );
-
-      auto iter = planned.find(msg->tag_);
-      vtAssert(iter != planned.end(), "Planned collective does not exist");
-      consensus_tag = msg->tag_;
-      action = iter->second.action_;
-    }
-
-    // We can not use a VT broadcast here because it involves the scheduler and
-    // MPI progress and there's no safe way to guarantee that the message has been
-    // received and processed and doesn't require progress from another node
-
-    int const root = msg->root_;
-    auto comm = theContext()->getComm();
-    MPI_Request req;
-
-    // Do a async broadcast of the tag we intend to execute collectively
-    MPI_Ibcast(&consensus_tag, 1, MPI_INT32_T, root, comm, &req);
-
-    int flag = 0;
-    while (not flag) {
-      MPI_Test(&req, &flag, MPI_STATUS_IGNORE);
-      runScheduler();
-    }
-
-    // We need a barrier here so the root doesn't finish the broadcast first and
-    // then enter the collective. We could replace the Ibcast/Ibarrier with a
-    // Iallreduce, but I think this is cheaper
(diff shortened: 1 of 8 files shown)


tests/unit/collectives/test_mpi_collective.cc
```diff
@@ -0,0 +1,140 @@
...
+  while (not vt::rt->isTerminated()) {
+    runScheduler();
+  }
+}
+
+TEST_F(TestMPICollective, test_mpi_collective_3) {
+  int done = 0;
+
+  // These three collective can execute in any order, but it will always be
+  // consistent across all the nodes
```
PhilMiller, line 114: I think this comment was incorrectly copied from above

tests/unit/collectives/test_mpi_collective.cc
```diff
@@ -0,0 +1,140 @@
...
+  auto tag = theCollective()->mpiCollective([&done]{
+    auto comm = theContext()->getComm();
+    int val = 0;
```
PhilMiller, line 117: For a stronger test, initialize `val` to the present rank, and then assert after the `waitCollective` that they all end up with `val == 0`

tests/unit/collectives/test_mpi_collective.cc
```diff
@@ -0,0 +1,140 @@
...
+  });
+
+  theCollective()->waitCollective(tag);
+
+  EXPECT_EQ(done, 1);
+
+  theCollective()->mpiCollectiveWait([&done]{
+    auto comm = theContext()->getComm();
+    int val_in = 1;
+    int val_out = 0;
```
PhilMiller, line 130: Similarly, it would be good to have an assertion that this comes out to equal `Comm_size`

src/vt/collective/collective_alg.h
```diff
@@ -83,6 +83,69 @@ struct CollectiveAlg :
   //----------------------------------------------------------------------------
   //----------------------------------------------------------------------------
 
+public:
+  /**
+   * \brief Enqueue a lambda with an embedded MPI collective invocation to
```
PhilMiller, line 88: Is it strictly
* One MPI collective?
* Just collectives? 

I would think the actual protocol should be equally fine with any closed set of MPI operations. I.e. as long as the function isn't leaving any incomplete requests around, anything should go.

For the Zoltan use case, this is kind of an important distinction, since the call to Zoltan is not in itself 'an MPI collective' - it makes some number of MPI calls, collective, p2p, or otherwise.

src/vt/collective/collective_alg.h
```diff
@@ -83,6 +83,69 @@ struct CollectiveAlg :
...
+   * execute in the future. Returns immediately, enqueuing the action for the
+   * future.
```
PhilMiller, line 90: There's also a potential need to document the buffer semantics across this sort of thing. Essentially, any buffer referenced inside `action` needs to be treated as part of an active MPI request until the `tag` is waited upon or `isCollectiveDone` returns true  for it.

src/vt/collective/collective_alg.h
```diff
@@ -83,6 +83,69 @@ struct CollectiveAlg :
...
+   *
+   * \param[in] action the action containing an MPI collective
+   *
+   * \return tag representing the collective
+   */
+  TagType mpiCollective(ActionType action);
```
PhilMiller, line 96: I think it would be preferable to name this `mpiCollectiveAsync` to be explicit about it.

PhilMiller, line 96: It would be nice if we could declare this `[[nodiscard]]`, but that's C++17 and the compiler support isn't there yet. If you have a macro for it in VT, it should be applied here.

PhilMiller, line 96: How are callers running from asynchronous handlers supposed to ensure that they're calling this in the same sequence everywhere? I assume that's necessary to ensure that each action is associated with the same tag on all processes

PhilMiller: Regardless, if that is a requirement, it needs to be documented

src/vt/collective/collective_alg.cc
```diff
@@ -52,4 +52,141 @@ CollectiveAlg::CollectiveAlg()
...
+  next_tag_++;
+  return tag;
+}
+
+bool CollectiveAlg::isCollectiveDone(TagType tag) {
+  return planned_collective_.find(tag) == planned_collective_.end();
+}
+
+void CollectiveAlg::waitCollective(TagType tag) {
+  while (not isCollectiveDone(tag)) {
```
PhilMiller, line 183: Does this very soon need to be the nested scheduler call? Just flagging an upcoming semantic merge conflict

src/vt/collective/collective_alg.cc
```diff
@@ -52,4 +52,141 @@ CollectiveAlg::CollectiveAlg()
     barrier::Barrier()
 { }
 
+/*static*/ void CollectiveAlg::runCollective(CollectiveMsg* msg) {
+  // We need a reentrancy counter to ensure that this is only on the scheduler
+  // stack once!
+  static int reenter_counter_ = 0;
+
+  if (reenter_counter_ == 0) {
```
PhilMiller, line 60: As with many other instances, could you put the `else` case up front, ending it with a return, and un-nest the rest of the function?

src/vt/collective/collective_alg.cc
```diff
@@ -52,4 +52,141 @@ CollectiveAlg::CollectiveAlg()
...
+
+  debug_print(
+    gen, node,
+    "mpiCollective: new MPI collective with tag={}\n",
+    tag
+  );
+
+  NodeType collective_root = 0;
+
+  auto cb = theCB()->makeBcast<CollectiveMsg,&runCollective>();
```
PhilMiller, line 164: I forget - do vt broadcasts guarantee ordered delivery? If so, it would be helpful to put a comment to that effect here, since it seems like this code relies on that being the case.

src/vt/collective/collective_alg.cc
```diff
@@ -52,4 +52,141 @@ CollectiveAlg::CollectiveAlg()
...
+
+  // Put this in a separate namespace
+  // @todo: broader issue: need to implement a better way to scope reductions
+  auto ident = 0xEFFFFFFFFFFFFFFF;
+  auto msg = makeMessage<CollectiveMsg>(tag, collective_root);
+  this->reduce<collective::None>(
+    collective_root, msg.get(), cb, tag, no_seq_id, 1, ident
+  );
+
+  next_tag_++;
```
PhilMiller, line 174: Why not put this increment up next to the read from `next_tag_`? When I saw that above, I wondered "where's the increment? Is it missing?"

src/vt/collective/collective_alg.cc
```diff
@@ -52,4 +52,141 @@ CollectiveAlg::CollectiveAlg()
...
+
+    // Run the collective safely
+    action();
+
+    // Erase the tag that was actually executed
+    planned.erase(planned.find(consensus_tag));
+
+    reenter_counter_--;
+
+    // Recurse, running a postponed collective
```
PhilMiller, line 132: If you have ordered vt broadcast, LIFO ordering on `postponed_collectives_` can get you mis-matched collectives, since you're not guaranteed that every process has received the same last broadcast.

Even if this gets switched to FIFO, ordered delivery doesn't ensure that every process has received the broadcast carrying the 'next' collective yet. I think this needs to do an `MPI_Allreduce(&postponed.front->tag, &min_tag, MPI_MIN)`, everyone checks `bool matching = min_tag == my_front_tag;`, and then `MPI_Allreduce(matching, MPI_LAND)` to make sure every process agrees

PhilMiller: OK, I see the logic a little better now - we can only reach this point because of a broadcast, that's the callback from a reduction, in which every process contributed (ostensibly) the same tag. So, we're guaranteed up front that every process has that tag and the corresponding action.

src/vt/collective/collective_alg.cc
```diff
@@ -52,4 +52,141 @@ CollectiveAlg::CollectiveAlg()
...
+
+    int flag = 0;
+    while (not flag) {
+      MPI_Test(&req, &flag, MPI_STATUS_IGNORE);
+      runScheduler();
+    }
+
+    // We need a barrier here so the root doesn't finish the broadcast first and
+    // then enter the collective. We could replace the Ibcast/Ibarrier with a
+    // Iallreduce, but I think this is cheaper
```
PhilMiller, line 102: It's not really going to be cheaper, but it is easier to reason out the logic that everyone has the same value for `consensus_tag`, if it's coming from a broadcast

src/vt/collective/collective_alg.cc
PhilMiller: Ok, so it doesn't seem like the code actually relies on ordering, since it goes with whatever the root is running. This just needs prose ahead of the code describing the reasoning involved.

src/vt/collective/collective_alg.cc
```diff
@@ -52,4 +52,141 @@ CollectiveAlg::CollectiveAlg()
...
+      debug_print(
+        gen, node,
+        "mpiCollective: running collective associated with tag={}\n",
+        msg->tag_
+      );
+
+      auto iter = planned.find(msg->tag_);
+      vtAssert(iter != planned.end(), "Planned collective does not exist");
+      consensus_tag = msg->tag_;
+      action = iter->second.action_;
```
PhilMiller, line 80: You can omit checking `planned` and pulling out `action` here, and then just make the code below do it unconditionally, rather than only on non-root. Everything up to that point is only concerned with determining `consensus_tag`

src/vt/collective/collective_alg.h
lifflander: I agree it's any closed set that is allowed that doesn't bleed out.

src/vt/collective/collective_alg.h
lifflander: Yes, that is a requirement. It must be called in the same order everywhere. I wanted to relax this requirement, but couldn't figure out a cheap/safe and relatively easy way.

I started with a `std::string` parameter to match them up. But, I couldn't figure out how to make that API performant and correct. I hashed the `std::string`, which is fine, but collisions quickly become an issue. I can detect a collisions easily enough on a given node, but resolving that collision is a problem. If I allow any arbitrary order with `std::string` tags, I can't ever hold anything back because that could deadlock due to a bad ordering with a colliding string

src/vt/collective/collective_alg.cc
lifflander: Yes, good point

src/vt/collective/collective_alg.cc
lifflander: Good point

src/vt/collective/collective_alg.h
PhilMiller: You could require the callers to provide a unique integer tag, instead of assigning it internally. They're much better positioned to be able to generate them properly. Either there's actually one stream of events that generate these calls, or there are multiple sources, and they need to partition the space.

There's an intermediate, of offering a 'sequence' facility, in which distinct call sources each pass a different value indicating what 'sequence' the call is part of, and then you just have an incrementing counter for each.

tests/unit/collectives/test_mpi_collective.cc
lifflander: Fixed

tests/unit/collectives/test_mpi_collective.cc
lifflander: I've strengthened the test based on this idea

tests/unit/collectives/test_mpi_collective.cc
lifflander: Done

lifflander:


```diff

Clones added
============
- tests/unit/collectives/test_mpi_collective.cc  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5442758&bid=17847806)

src/vt/collective/collective_alg.h
lifflander: Ok, I've renamed this `mpiCollectiveAsync`

lifflander: You are allowed to discard by calling this in an epoch and waiting for termination of that epoch.

I was thinking that maybe I missed a produce/consume in postponed collectives, but since I recurse on the stack it shouldn't be able to complete the epoch until the message buffer is empty.

lifflander:


```diff

Clones added
============
- tests/unit/collectives/test_mpi_collective.cc  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5442758&bid=17847806)

src/vt/collective/collective_alg.cc
lifflander: Ok, I've moved it up

src/vt/collective/collective_alg.cc
lifflander: Yes, I agree. The Iallreduce logic (which I started with) is definitely tricker.

src/vt/collective/collective_alg.cc
lifflander: Done

src/vt/collective/collective_alg.cc
lifflander: Ok, I've added a comment to explain this.

src/vt/collective/collective_alg.h
lifflander: I've added some documentation for this

src/vt/collective/collective_alg.h
lifflander: I've improved the documentation to state a closed set of MPI collectives

lifflander:


```diff

Clones added
============
- tests/unit/collectives/test_mpi_collective.cc  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5442758&bid=17847806)

src/vt/collective/collective_alg.h
```diff
@@ -83,6 +83,77 @@ struct CollectiveAlg :
   //----------------------------------------------------------------------------
   //----------------------------------------------------------------------------
 
+public:
+  /**
+   * \brief Enqueue a lambda with an embedded closed set of MPI collectives
```
PhilMiller, line 88: The operation isn't limited to MPI collectives, either. Just MPI operations.

src/vt/collective/collective_alg.h
```diff
@@ -83,6 +83,77 @@ struct CollectiveAlg :
...
+   * invocations to execute in the future. Returns immediately, enqueuing the
+   * action for the future.
+   *
+   * The set of operations specified in the lambda must be closed, meaning that
+   * MPI requests must not escape the lambda. After the lambda finishes, the set
+   * of MPI collective calls should be complete.
+   *
+   * Any buffers captured in the lambda to use with MPI collective operations
+   * need to exist until \c isCollectiveDone returns \c true or \c
+   * waitCollective returns on the returned \c TagType
```
PhilMiller, line 98: It's not just that the buffers need to exist, it's that as far as MPI operation is concerned, they need to be considered 'in use' (I don't remember if that's the phrasing from the standard) - other code can't read or write to them, without screwing up results.

src/vt/collective/collective_alg.h
```diff
@@ -83,6 +83,77 @@ struct CollectiveAlg :
...
+   * \brief Wait on an MPI collective set to complete
+   *
+   * \param[in] tag MPI collective set identifier
+   */
+  void waitCollective(TagType tag);
+
+  /**
+   * \brief Enqueue a lambda with an embedded closed set of MPI
+   * collectives. Spin in the VT scheduler until it terminates. All other nodes
+   * must not depend on this node to enqueue a matching collective.
```
PhilMiller, line 125: Given the scheduler spinning, is the rule against cross-node dependencies actually necessary?

src/vt/collective/collective_alg.cc
```diff
@@ -52,4 +52,148 @@ CollectiveAlg::CollectiveAlg()
...
+    runScheduler();
+  }
+
+  debug_print(
+    gen, node,
+    "mpiCollective: consensus_tag={}, msg->tag_={}\n",
+    consensus_tag, msg->tag_
+  );
+
+  // The root had a different collective in mind... let's do that one
```
PhilMiller, line 122: Repeating my earlier comment - The code can just declare and look up `action` here, unconditionally. No need for the root to pick it out earlier, or to conditionalize this.

PhilMiller: By making that change, you also then don't need to repeat `planned.find(consensus_tag)` below in the `erase` call

PhilMiller:
OK, protocol-wise, I'm satisfied that this works as intended. I think there may actually be a short workshop paper in this, since the design should be useful for Charm++, HPX, etc. "Embedding distributed MPI operations in asynchronous tasking runtime applications" or something

src/vt/collective/collective_alg.h
lifflander: I think you're right, it isn't necessary. As long as the user has messages in flight that might enable the collective elsewhere, it should be fine.

src/vt/collective/collective_alg.h
lifflander: Ok, fixed

src/vt/collective/collective_alg.h
lifflander: Changed to operations

src/vt/collective/collective_alg.cc
lifflander: Ok, done

lifflander:


```diff

Clones added
============
- tests/unit/collectives/test_mpi_collective.cc  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5442758&bid=17847806)

src/vt/collective/collective_alg.h
lifflander: What do you think about this:
```C++
auto scope = theCollective()->createCollectiveScope(/*integer tag*/ tag);
scope.mpiCollectiveAsync([]{}); //ordered within scope
scope.mpiCollectiveAsync([]{}); //ordered within scope
```

Where `scope` is a move-only object that the user can hold on (e.g., inside an objgroup). Within scope all operations are collectively ordered. Each scope has its own sequence.

PhilMiller:
That all looks great.

I'm 90% convinced that the protocol as implemented guarantees that every call made does eventually get completed. The vt reductions and broadcasts are guaranteed to make progress, resulting in every process seeing as many handler calls as there were operation requests. For any given request, on any given process, either the process will be at `reenter == 0`, and go into the consensus protocol, or it will be non-zero, and will re-start the consensus protocol for it when the current operation is complete. So, every process eventually executes as many rounds of the consensus protocol, and hence that many actions, as calls were made.

Also, this will end up having interesting interactions with a hypothetical PMPI tracing into vt's log files implementation.

lifflander:
Travis CI seems to be screwed up---those build failures are spurious.

src/vt/collective/collective_alg.h
PhilMiller: Yes, that's pretty much what I had in mind. Much easier to uniquely identify the scopes, and then sequence on each one.

Are you going to just make the keys `pair<scope_tag, sequence_in_scope>` then?

lifflander:
@PhilMiller I think I've found a theoretical bug. If two mpiCollectiveAsync operations are enqueued in different epochs, and inside those operations messages are sent from VT, they might get the wrong epoch because I don't push/pop the epoch when recursing on the stack.

PhilMiller:
> @PhilMiller I think I've found a theoretical bug. If two mpiCollectiveAsync operations are enqueued in different epochs, and inside those operations messages are sent from VT, they might get the wrong epoch because I don't push/pop the epoch when recursing on the stack.

I think you may be better off disallowing vt message sends during these operations, or set things up in a way that they have to push/pop their own epoch explicitly, with a failure if they don't

lifflander:
Should I implement the scoping as discussed in that comment in this PR? Shouldn't be too difficult to do.

PhilMiller:
May as well get the scoping in now, sure

lifflander:


```diff

Clones added
============
- tests/unit/collectives/test_mpi_collective.cc  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5442758&bid=17847806)

lifflander:
@PhilMiller I've adding in scopes. Ended up being more complex than I thought to get the desired behavior. Not sure it's worth the complexity, let me know what you think.

If we keep this I need to write more tests.

You can call `auto scope = theCollective()->makeCollectiveScope()` and each time get a new distinct system-tagged scope.

You can call it with an explicit tag `auto scope = theCollective()->makeCollectiveScope(5)`, and get user-tagged scope that won't conflict with generated system scopes.

When a `scope` goes out of scope (move-only), you can't add any more operations. It doesn't wait on everything (considered that design but I think its too tricky), but it does cleanup when it finishes.

lifflander:


```diff

Clones added
============
- tests/unit/collectives/test_mpi_collective.cc  12
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5442758&bid=17847806)

lifflander:


```diff

Clones added
============
- tests/unit/collectives/test_mpi_collective.cc  12
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5442758&bid=17847806)

src/vt/collective/collective_scope.h
```diff
@@ -0,0 +1,174 @@
...
+  /**
+   * \internal \brief Get the scope state
+   *
+   * \return the scope state
+   */
+  detail::ScopeImpl* getScope();
+
+private:
+  bool is_user_tag_ = false;
+  TagType scope_ = no_tag;
```
PhilMiller, line 169: Why duplicate the tag between the public interface and the underlying implementation?

tests/unit/collectives/test_mpi_collective.cc
```diff
@@ -151,5 +151,57 @@ TEST_F(TestMPICollective, test_mpi_collective_3) {
...
+    EXPECT_EQ(bcast_val, 29);
+    EXPECT_EQ(reduce_val_out, num_nodes);
+  });
+
+  while (not vt::rt->isTerminated()) {
+    runScheduler();
+  }
+}
+
+
```
PhilMiller: Could this test be adapted to actually affirm that the MPI operations were run in the same order everywhere?

PhilMiller:
I think the scopes look good. In the tests, it may be worthwhile to spell out `vt::collective::CollectiveScope`, since actual users will need to declare member variables of that type.

lifflander:
> vt::collective::CollectiveScope

Do you think I should alias `CollectiveScope` in the `vt` namespace directly?

src/vt/collective/collective_scope.h
lifflander: Good point. I don't need it in `ScopeImpl`, just here to look up the scope from the `CollectiveScope`

lifflander: Fixed

lifflander:
I have spelled it out in the tests now.

tests/unit/collectives/test_mpi_collective.cc
lifflander: Yes, I just added that!

lifflander:


```diff

Clones added
============
- tests/unit/collectives/test_mpi_collective.cc  6
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5442758&bid=17847806)

lifflander:
@PhilMiller

- I left `CollectiveScope` in `collective::` for now without an alias.
- I spelled it out in the tests
- I added a consistent order check across nodes for the last test
- Removed duplicate/unneeded scoping info in `ScopeImpl`

I think this is ready to go!

The only last thing would be to start using the interface across the system, but I think that should go in a seperate PR. Maybe starting with `group/collective/group_info_collective.cc`.

lifflander:
Looks like there might be non-deterministic failure on Travis that is real.

lifflander:


```diff

Clones added
============
- tests/unit/collectives/test_mpi_collective.cc  6
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5442758&bid=17847806)

lifflander:
Removing that barrier caused a bunch of other failures:

```
The following tests FAILED:
1673	  7 - vt:*/TestActiveBroadcastPut.test_type_safe_active_fn_bcast2/*_proc_2 (Timeout)
1674	  8 - vt:*/TestActiveBroadcast.test_type_safe_active_fn_bcast2/*_proc_2 (Timeout)
1675	 66 - vt:*/TestSequencerExtensive.test_wait_1/*_proc_2 (Timeout)
1676	 73 - vt:*/TestSequencerParallelParam.test_seq_parallel_param/*_proc_2 (Timeout)
1677	 91 - vt:*/TestTermGlobal.test_term_detect_routed/*_proc_2 (Timeout)
```

lifflander:
I think I have the problem solved now. We need the collective barrier, but not the MPI one after.

PhilMiller:
Could you explain why, and what went wrong?

lifflander:


```diff

Clones added
============
- tests/unit/collectives/test_mpi_collective.cc  6
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5442758&bid=17847806)

lifflander:
> Could you explain why, and what went wrong?
> […](#)

I changed this line to:

```C++
theCollective->barrierThen([]{
  MPI_Barrier(comm);
});
```

Just a regular, blocking VT barrier.

```C++
theCollective->barrier();
```

The regular barrier is needed and safe. When VT re-inits the runtime, which is used in some tests, it calls setup to synchronize.

lifflander:
@PhilMiller

PR #804, opened 2020-05-07
803 Conform to new checkpoint API, make required

### LINKED ISSUE #803: Adapt VT's checkpoint usage to conform to adjusted API
**What Needs to be Done?**

Per https://github.com/DARMA-tasking/checkpoint/pull/92 and https://github.com/DARMA-tasking/checkpoint/pull/90, the API to serialization has changed namespaces and other minor changes.

lifflander:
Fixes #803

- Conform to 1.0.0-beta.8 changes in checkpoint, refactoring the
  namespaces and checkpoint API

- Checkpoint is considered a optional dependency, but VT has been
  failing to compile without checkpoint for a long time. Thus, make it
  a required dependency.

- Remove all the conditional logic on `HAS_SERIALIZTION_LIBRARY`

- Remove the "mock" checkpoint-replacement serializer that handled all
  the basic types, `std::vector`, `std::tuple` of basic arthimetic types
  when checkpoint was not present.


lifflander:
I am completely at a loss as to what is wrong with the alpine-clang build after spending an hour trying to debug why the fuck cmake isn't including the proper directories. The generated targets file looks fine... but for some reason on that platform, with a late enough version of cmake, it doesn't include the interface directories from the imported target.

nlslatt:
> I am completely at a loss as to what is wrong with the alpine-clang build after spending an hour trying to debug why the fuck cmake isn't including the proper directories. The generated targets file looks fine... but for some reason on that platform (with what should be a sufficient version of cmake), it doesn't include the interface directories from the imported target.

Do you need to add a `target_include_directories` line like we do for CLI11 and FMT?

lifflander:
> > I am completely at a loss as to what is wrong with the alpine-clang build after spending an hour trying to debug why the fuck cmake isn't including the proper directories. The generated targets file looks fine... but for some reason on that platform (with what should be a sufficient version of cmake), it doesn't include the interface directories from the imported target.
> 
> Do you need to add a `target_include_directories` line like we do for CLI11 and FMT?

I tried that and it didn't solve the problem.

krcb:
@lifflander I just spent a day dealing with a similar horror. Can you try setting policy CM0060 to true?

lifflander:
> CM0060

I tried setting this:

```cmake
if (POLICY CMP0060)
  cmake_policy(SET CM0060 OLD)
endif()
```

But, I'm getting this:

```
CMake Error at CMakeLists.txt:5 (cmake_policy):
  Policy "CM0060" is not known to this version of CMake.
```

lifflander:
The problem doesn't happen unless checkpoint is linked in, instead of being in the build tree.

The generated checkpointTargets.cmake has the appropriate `INTERFACE_LINK_LIBRARIES`, with the path needed, assuming `${_IMPORT_PREFIX}` is correct.

lifflander:
I found the bug! In the alpine image, I had accidentally installed checkpoint globally `/usr/local/cmake`, and find_package was finding the global installation with the old config. I shouldn't have installed it globally, but the way I use find_package should use the local installation, so that seems like a problem.

PR #805, opened 2020-05-11
802 Wrap tag around when hits MPI max tag (attribute MPI_TAG_UB)

### LINKED ISSUE #802: Use MPI constant for max allowed tag
**Describe the bug**
Currently we don't use the proper MPI constant and tags monotonically increase.

lifflander:
Fixes #802

### DIFF
--- a/src/vt/messaging/active.cc
+++ b/src/vt/messaging/active.cc
@@ -325,7 +325,7 @@ ActiveMessenger::SendDataRetType ActiveMessenger::sendData(
   auto const& num_bytes = std::get<1>(ptr);
 
   int send_tag = 0;
-  if (tag == no_tag) {
+  if (tag != no_tag) {
     send_tag = tag;
   } else {
     auto const max_tag = util::MPI_Attr::getMaxTag();
--- a/src/vt/messaging/active.cc
+++ b/src/vt/messaging/active.cc
@@ -329,12 +329,11 @@ ActiveMessenger::SendDataRetType ActiveMessenger::sendData(
     send_tag = tag;
   } else {
     auto const max_tag = util::MPI_Attr::getMaxTag();
-    send_tag = cur_direct_buffer_tag_++;
 
-    // If max, wrap around back to the starting tag
-    if (cur_direct_buffer_tag_ >= max_tag) {
+    if (cur_direct_buffer_tag_ == max_tag) {
       cur_direct_buffer_tag_ = starting_direct_buffer_tag;
     }
+    send_tag = cur_direct_buffer_tag_++;
   }
 
   auto const event_id = theEvent()->createMPIEvent(this_node_);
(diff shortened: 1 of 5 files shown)


lifflander:
These tests will all fail until checkpoint changes get merged!

src/vt/messaging/active.cc
```diff
@@ -322,7 +323,19 @@ ActiveMessenger::SendDataRetType ActiveMessenger::sendData(
 ) {
   auto const& data_ptr = std::get<0>(ptr);
   auto const& num_bytes = std::get<1>(ptr);
-  auto const send_tag = tag == no_tag ? cur_direct_buffer_tag_++ : tag;
+
+  int send_tag = 0;
+  if (tag == no_tag) {
```
PhilMiller, line 328: You reversed the sense of the comparison here - should be `tag != no_tag`

src/vt/utils/mpi_limits/mpi_max_tag.h
```diff
@@ -53,30 +53,40 @@
...
 
 /**
- * \struct MPI_MaxTag
+ * \struct MPI_Attr
  *
- * \brief Get the MPI maximum value for a tag. This can vary across MPI
- * implementations, but the MPI standard dictates it be at least 32767.
+ * \brief Get MPI attributes, including the MPI maximum value for a tag. This
+ * can vary across MPI implementations, but the MPI standard dictates it be at
+ * least 32767.
```
PhilMiller, line 60: This comment is partially out of place now. The bits about the tag should be on `getMaxTag`

src/vt/messaging/active.cc
```diff
@@ -322,7 +323,19 @@ ActiveMessenger::SendDataRetType ActiveMessenger::sendData(
...
+    send_tag = tag;
+  } else {
+    auto const max_tag = util::MPI_Attr::getMaxTag();
+    send_tag = cur_direct_buffer_tag_++;
+
+    // If max, wrap around back to the starting tag
+    if (cur_direct_buffer_tag_ >= max_tag) {
+      cur_direct_buffer_tag_ = starting_direct_buffer_tag;
+    }
+  }
```
PhilMiller, line 338: I think we may need to be a bit more cautious here about incrementing through `INT_MAX` if `MPI_TAG_UB` is equal to it.

I think what we want is rather

```
auto const max_tag = util::MPI_Attr::getMaxTag();
if (cur_direct_buffer_tag_ == max_tag) {
      cur_direct_buffer_tag_ = starting_direct_buffer_tag;
    }
send_tag = cur_direct_buffer_tag_++;
```

lifflander: Ok, fixed.

src/vt/utils/mpi_limits/mpi_max_tag.h
lifflander: Fixed

src/vt/messaging/active.cc
lifflander: Fixed

PR #807, opened 2020-05-11
806 Use RDMA for large contiguous chunks or serialized messages

### LINKED ISSUE #806: Use new RDMA handles for sending large messages or contiguous chunks
**What Needs to be Done?**

- [ ] Now that checkpoint allows custom traversal, contiguous data of certain types can be RDMA'ed using a traverser: https://github.com/DARMA-tasking/checkpoint/pull/94
- [ ] Hook into `theMsg()->sendData` for sending chunks (if copies are used)
- [ ] Or use traversals to find `std::vector` or `Kokkos::View`

lifflander:
Fixes #806

Still it its infancy...

### DIFF
--- a/src/vt/messaging/rdma/msg_rdma.cc
+++ b/src/vt/messaging/rdma/msg_rdma.cc
@@ -1,0 +1,58 @@
+/*
+//@HEADER
+// *****************************************************************************
+//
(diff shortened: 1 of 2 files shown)


lifflander:
not useful

PR #812, opened 2020-05-12
789 - Fix function name in Trace having last character removed

pnstickne:
Fixes and off-by-one bug and adds tests.


pnstickne:
@lifflander Bump.

PR #820, opened 2020-05-21
819 Collective insertable collections

### LINKED ISSUE #819: Implement collective-epoch-based insertable collections
**What Needs to be Done?**

This is needed for NimbleSM. Instead of the rooted insert, implement a collective variant that will be compatible with collection chain set. Make `theCollection()->constructCollective<T>(..)` work with insertable collections (or remove assertion if it works).

lifflander:
Fixes #819

- Allow insertable collections to be created with 
  `theCollection()->constructCollective<ColT>`.
- Add new functionality for collective insertion epochs instead of the rooted ones currently implemented which are not as suitable for NimbleSM's use case.

- [x] Write tests for new functionality

### DIFF
--- a/docs/md/insertable-collections.md
+++ b/docs/md/insertable-collections.md
@@ -1,0 +1,14 @@
+\page insertable-collections Insertable Collections
+\brief Dynamically inserted collections
+
+The virtual context collection component
+`vt::vrt::collection::CollectionManager` supports sparse collections that can be
+dynamically inserted at runtime. To declare an insertable collection, inherit
+from `vt::InsertableCollection` instead of the dense collection
+`vt::Collection`. After the collection is constructed, elements will not be
+created automatically for the index range passed to `theCollection()->construct`
+or `theCollection()->constructCollective()`. Instead, the user is responsible
+for inserting the indices that exist after the collection is constructed.
+
+\section insertable-example-program Insertable Collection Example
+\snippet examples/collection/collective_insertable.cc Collective insertable example
--- a/examples/collection/collective_insertable.cc
+++ b/examples/collection/collective_insertable.cc
@@ -42,6 +42,8 @@
 //@HEADER
 */
 
+/// [Collective insertable example]
+
 #include <vt/transport.h>
 
 static constexpr int32_t const default_num_elms = 64;
@@ -109,3 +111,5 @@ int main(int argc, char** argv) {
 
   return 0;
 }
+
+/// [Collective insertable example]
--- a/examples/collection/collective_insertable.cc
+++ b/examples/collection/collective_insertable.cc
@@ -1,0 +1,109 @@
+/*
+//@HEADER
+// *****************************************************************************
+//
+//                          collective_insertable.cc
+//                           DARMA Toolkit v. 1.0.0
+//                       DARMA/vt => Virtual Transport
+//
+// Copyright 2019 National Technology & Engineering Solutions of Sandia, LLC
+// (NTESS). Under the terms of Contract DE-NA0003525 with NTESS, the U.S.
+// Government retains certain rights in this software.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+// * Redistributions of source code must retain the above copyright notice,
+//   this list of conditions and the following disclaimer.
+//
+// * Redistributions in binary form must reproduce the above copyright notice,
+//   this list of conditions and the following disclaimer in the documentation
+//   and/or other materials provided with the distribution.
+//
+// * Neither the name of the copyright holder nor the names of its
+//   contributors may be used to endorse or promote products derived from this
+//   software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+//
+// Questions? Contact darma@sandia.gov
+//
+// *****************************************************************************
+//@HEADER
+*/
+
+#include <vt/transport.h>
+
+static constexpr int32_t const default_num_elms = 64;
+
+struct InsertCol : vt::InsertableCollection<InsertCol, vt::Index1D> {
+  InsertCol() {
+    vt::NodeType this_node = vt::theContext()->getNode();
+    ::fmt::print("{}: constructing: idx={}\n", this_node, getIndex().x());
+  }
+
+  ~InsertCol() {
+    vtAssert(reduce_done_, "Reduce must hit all live element");
+  }
+
+  using Msg = vt::CollectionMessage<InsertCol>;
+  using ReduceMsg = vt::collective::ReduceNoneMsg;
+
+  void done(ReduceMsg* msg) {
+    //fmt::print("{}: done\n", this->getIndex());
+    reduce_done_ = true;
+  }
+
+  void doWork(Msg* msg) {
+    auto proxy = getCollectionProxy();
+    auto red_msg = vt::makeMessage<ReduceMsg>();
+    auto cb = vt::theCB()->makeBcast<InsertCol,ReduceMsg,&InsertCol::done>(proxy);
+    proxy.reduce(red_msg.get(),cb);
+  }
+
+private:
+  bool reduce_done_ = false;
+};
+
+int main(int argc, char** argv) {
+  vt::initialize(argc, argv);
+
+  vt::NodeType this_node = vt::theContext()->getNode();
+  vt::NodeType num_nodes = vt::theContext()->getNumNodes();
+
+  int32_t num_elms = default_num_elms;
+  if (argc > 1) {
+    num_elms = atoi(argv[1]);
+  }
+
+  auto range = vt::Index1D(num_elms);
+  auto proxy = vt::theCollection()->constructCollective<InsertCol>(range);
+
+  proxy.startInsertCollective();
+
+  for (int i = 0; i < 32; i++) {
+    if (i < range.x() and i % num_nodes == this_node) {
+      proxy(i).insert();
+    }
+  }
+
+  proxy.finishInsertCollective();
+
+  if (this_node == 0) {
+    proxy.broadcast<InsertCol::Msg,&InsertCol::doWork>();
+  }
+
+  vt::finalize();
+
+  return 0;
+}
--- a/examples/collection/collective_insertable.cc
+++ b/examples/collection/collective_insertable.cc
@@ -86,22 +86,24 @@ int main(int argc, char** argv) {
     num_elms = atoi(argv[1]);
   }
 
-  auto range = vt::Index1D(num_elms);
-  auto proxy = vt::theCollection()->constructCollective<InsertCol>(range);
+  vt::runInEpochCollective([&]{
+    auto range = vt::Index1D(num_elms);
+    auto proxy = vt::theCollection()->constructCollective<InsertCol>(range);
 
-  proxy.startInsertCollective();
+    proxy.startInsertCollective();
 
-  for (int i = 0; i < 32; i++) {
-    if (i < range.x() and i % num_nodes == this_node) {
-      proxy(i).insert();
+    for (int i = 0; i < 32; i++) {
+      if (i < range.x() and i % num_nodes == this_node) {
+        proxy(i).insert();
+      }
     }
-  }
 
-  proxy.finishInsertCollective();
+    proxy.finishInsertCollective();
 
-  if (this_node == 0) {
-    proxy.broadcast<InsertCol::Msg,&InsertCol::doWork>();
-  }
+    if (this_node == 0) {
+      proxy.broadcast<InsertCol::Msg,&InsertCol::doWork>();
+    }
+  });
 
   vt::finalize();
 
--- a/src/vt/vrt/collection/insert/insert_finished.impl.h
+++ b/src/vt/vrt/collection/insert/insert_finished.impl.h
@@ -63,19 +63,19 @@ void InsertFinished<ColT,IndexT,BaseProxyT>::finishedInserting(
   ActionType action
 ) const {
   auto const col_proxy = this->getProxy();
-  theCollection()->finishedInserting<ColT,IndexT>(col_proxy,action);
+  theCollection()->finishedInserting<ColT>(col_proxy,action);
 }
 
 template <typename ColT, typename IndexT, typename BaseProxyT>
 void InsertFinished<ColT,IndexT,BaseProxyT>::startInsertCollective() const {
   auto const col_proxy = this->getProxy();
-  theCollection()->startInsertCollective<ColT,IndexT>(col_proxy);
+  theCollection()->startInsertCollective<ColT>(col_proxy);
 }
 
 template <typename ColT, typename IndexT, typename BaseProxyT>
 void InsertFinished<ColT,IndexT,BaseProxyT>::finishInsertCollective() const {
   auto const col_proxy = this->getProxy();
-  theCollection()->finishInsertCollective<ColT,IndexT>(col_proxy);
+  theCollection()->finishInsertCollective<ColT>(col_proxy);
 }
 
 }}} /* end namespace vt::vrt::collection */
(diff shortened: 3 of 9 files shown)


lifflander:


```diff

Clones added
============
- examples/collection/collective_insertable.cc  4
- src/vt/vrt/collection/manager.impl.h  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5566989&bid=18256282)

src/vt/vrt/collection/manager.impl.h
```diff
@@ -1763,73 +1757,75 @@ CollectionManager::constructCollectiveMap(
...
   auto fn = auto_registry::getHandlerMap(map_han);
 
-  // Walk through the index range with the mapping function and invoke the
-  // construct function (parameter to this fn) for local elements, populating
-  // the holder
-  range.foreach([&](IndexT cur_idx) mutable {
-    using BaseIdxType      = vt::index::BaseIndex;
-    using VirtualElmPtr    = VirtualPtrType<ColT,IndexT>;
-    using IdxContextHolder = InsertContextHolder<IndexT>;
+  if (is_static) {
```
PhilMiller, line 1760: This is a lot of code to have under a single `if`; maybe pull the body of the `foreach` out into a separate function?

src/vt/vrt/collection/manager.impl.h
```diff
@@ -1712,18 +1731,12 @@ CollectionManager::constructCollectiveMap(
...
-  vtAssertInfo(
-    is_static, "Distributed collection construct must be statically sized",
-    is_static, tag, map_han
-  );
-
   // Invoke getCollectionLM() to create a new location manager instance for this
   // collection
   theLocMan()->getCollectionLM<ColT, IndexT>(proxy);
 
+  auto const is_static = ColT::isStaticSized();
```
pnstickne, line 1738: I recommend using the actual type (eg `bool`) for primitives and trivial aliases, as such leaves useful text context clues.

src/vt/vrt/collection/manager.impl.h
```diff
@@ -1744,73 +1757,75 @@ CollectionManager::constructCollectiveMap(
...
-  // the holder
-  range.foreach([&](IndexT cur_idx) mutable {
-    using BaseIdxType      = vt::index::BaseIndex;
-    using VirtualElmPtr    = VirtualPtrType<ColT,IndexT>;
-    using IdxContextHolder = InsertContextHolder<IndexT>;
+  if (is_static) {
+    // Walk through the index range with the mapping function and invoke the
+    // construct function (parameter to this fn) for local elements, populating
+    // the holder
+    range.foreach([&](IndexT cur_idx) mutable {
```
pnstickne, line 1764: Can update `[&]` with capture list?

src/vt/vrt/collection/insert/insert_finished.impl.h
```diff
@@ -66,6 +66,18 @@ void InsertFinished<ColT,IndexT,BaseProxyT>::finishedInserting(
   theCollection()->finishedInserting<ColT,IndexT>(col_proxy,action);
 }
 
+template <typename ColT, typename IndexT, typename BaseProxyT>
```
pnstickne, line 69: Can be made independent of BaseProxyT? (Not necessarily specific is PR.)

src/vt/vrt/collection/manager.h
```diff
@@ -197,6 +197,14 @@ struct CollectionManager
    *  function for every index in the system based on the where each index is
    *  mapped with the MapFnT.
    */
+  template <
+    typename ColT,  mapping::ActiveMapTypedFnType<typename ColT::IndexType> fn
```
pnstickne, line 201: Extra space after comma. Optionally could be multiple lines.

src/vt/vrt/collection/manager.h
```diff
@@ -785,6 +793,16 @@ struct CollectionManager
     ActionType insert_action = nullptr
   );
 
+  template <typename ColT, typename IndexT = typename ColT::IndexType>
```
pnstickne, line 796: Is there a case where IndexT is expected/allowed to be different that ColT::IndexType?

If not, I would pull this out of the template to prevent such in the future.

src/vt/vrt/collection/manager.h
lifflander: Fixed

src/vt/vrt/collection/manager.h
lifflander: Removed

src/vt/vrt/collection/insert/insert_finished.impl.h
lifflander: No, currently the way this is written it needs `BaseProxyT` because the base class the composed in later.

src/vt/vrt/collection/manager.impl.h
lifflander: Fixed

src/vt/vrt/collection/manager.impl.h
lifflander: Done

src/vt/vrt/collection/manager.impl.h
lifflander: Done

lifflander:


```diff

Clones added
============
- examples/collection/collective_insertable.cc  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5566989&bid=18256282)

lifflander:
@PhilMiller @pnstickne Please review changes and approve

src/vt/vrt/collection/manager.h
```diff
@@ -219,6 +227,15 @@ struct CollectionManager
     HandlerType const& map_han, TagType const& tag
   );
 
+private:
+  template <typename ColT>
+  void constructElement(
+    VirtualProxyType proxy, typename ColT::IndexType cur_idx,
```
lifflander, line 233: I should pass idx and range by const& here.

lifflander:


```diff

Clones added
============
- examples/collection/collective_insertable.cc  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5566989&bid=18256282)

lifflander:
@PhilMiller Please re-review

src/vt/vrt/collection/manager.h
PhilMiller: Given recent discussions, I agree

examples/collection/collective_insertable.cc
```diff
@@ -0,0 +1,109 @@
...
+    }
+  }
+
+  proxy.finishInsertCollective();
+
+  if (this_node == 0) {
+    proxy.broadcast<InsertCol::Msg,&InsertCol::doWork>();
+  }
+
+  vt::finalize();
```
PhilMiller, line 106: This repeats the earlier anti-pattern of using `finalize` to mean running everything that's been launched. Please wrap the relevant bits in a `runInEpochCollective` or whatever.

lifflander: Fixed

lifflander:


```diff

Clones added
============
- examples/collection/collective_insertable.cc  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5566989&bid=18256282)

src/vt/vrt/collection/manager.impl.h
```diff
@@ -1530,79 +1618,19 @@ CollectionManager::constructCollectiveMap(
...
-      vrt_coll, node,
-      "construct (dist): foreach: cur_idx={}, mapped_node={}\n",
-      cur_idx.toString(), mapped_node
+  if (is_static) {
+    // Walk through the index range with the mapping function and invoke the
+    // construct function (parameter to this fn) for local elements, populating
+    // the holder
+    range.foreach(
+      [&proxy,&range,&map_han,&fn,&user_construct_fn,this](IndexT cur_idx) mutable {
+        constructElement<ColT>(proxy, cur_idx, range, map_han, fn, user_construct_fn);
```
PhilMiller, line 1630: I think the refactoring might have been done a bit too aggressively. I think it would make sense for a function named `constructElement` to unequivocally *construct an element*, not "think about constructing an element". That suggests doing the mapping and test of `mapped_node == this_node` here, and conditionally making the call based on it.

In doing that, the `range` argument can just become `num_elems`, which is being pointlessly recomputed in each call - it should be calculated outside this `foreach`, and the `map_fn` doesn't need to be passed as an argument.

src/vt/vrt/collection/manager.impl.h
```diff
@@ -2258,17 +2288,65 @@ template <typename ColT, typename IndexT>
...
+
+  // Now, we need to create a new group because broadcasts/reductions might need
+  // to hit new nodes based on the current insert set!
+  groupConstruction<ColT>(untyped_proxy, true /* do it now! */);
+
+  // Spin until the group is ready. Don't let the user proceed. In theory this
+  // can be relaxed by buffering operations, but this is simpler for now
+  theSched()->runSchedulerWhile([&elm_holder]{
+    return not elm_holder->groupReady();
+  });
```
PhilMiller, line 2340: I don't know if the eventual call to `setGroupReady(true)` is the result of consensus.

If it isn't then I suspect it needs to be, or other nodes that depend on this one to finish group setup might not complete it.

If it is, then why is that consensus in there, when we can just call `groupConstruction` in `runInEpochCollective`?

src/vt/vrt/collection/manager.impl.h
```diff
@@ -2258,17 +2288,65 @@ template <typename ColT, typename IndexT>
...
+
+  // Get the current insert epoch
+  auto insert_epoch = UniversalIndexHolder<>::insertGetEpoch(untyped_proxy);
+  vtAssert(insert_epoch != no_epoch, "Epoch should be valid");
+
+  // Wait on the insert epoch, ensuring that insertions are complete
+  bool done = false;
+  theTerm()->addAction(insert_epoch, [&done] { done = true; });
+  theTerm()->finishedEpoch(insert_epoch);
+  theSched()->runSchedulerWhile([&done] { return not done; });
```
PhilMiller, line 2320: Trivially, this can just be done as `runThrough(insert_epoch)` without the flag and action.

More broadly, I'm not a fan of the embedded scheduler loops in here. The first one could be eliminated by making the whole continuation an action on `insert_epoch`, and then waiting on the condition of the second. I'd philosophically prefer to just return control to the calling code, with a suitable epoch, and let it decide whether to block on this. That might not be the kindest API choice, though, since I think I agree with the implicit assumption here that almost all callers will want to immediately wait for the collection to be fully re-activated

src/vt/vrt/collection/manager.impl.h
lifflander: Yes, this is a result on consensus. The reason it's not in `runInEpochCollective` is that it eventually will return control back to the user and buffer. Also, it was implemented before that existed. So maybe I should switch it for now.

lifflander:


```diff

Clones added
============
- tests/unit/collection/test_collective_insertable.extended.cc  2
- examples/collection/collective_insertable.cc  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5566989&bid=18256282)

src/vt/vrt/collection/manager.impl.h
lifflander: This is addressed.

lifflander:


```diff

Clones added
============
- examples/collection/collective_insertable.cc  4
- tests/unit/collection/test_collective_insertable.extended.cc  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5566989&bid=18256282)

src/vt/vrt/collection/manager.impl.h
lifflander: This is fixed, using `runSchedulerThrough` now.

src/vt/vrt/collection/manager.impl.h
PhilMiller: If your "eventually it will return control back to the user and buffer" describes a plan for the evolution of this interface, I'd think the way to accomplish that would be to return an appropriate epoch for them to test and/or wait upon. That would seem to be consistent with shifting to use of TD now.

docs/md/insertable-collections.md
```diff
@@ -0,0 +1,14 @@
+\page insertable-collections Insertable Collections
+\brief Dynamically inserted collections
+
+The virtual context collection component
+`vt::vrt::collection::CollectionManager` supports sparse collections that can be
+dynamically inserted at runtime. To declare an insertable collection, inherit
+from `vt::InsertableCollection` instead of the dense collection
```
PhilMiller, line 7: The name `InsertableCollection` may be a red herring. Which of these is the defining trait for users?
- the sparsity,
- the dynamic insertion,
- either under different circumstances,
- both

tests/unit/collection/test_collective_insertable.extended.cc
```diff
@@ -0,0 +1,137 @@
...
+
+  auto range = vt::Index1D(num_elms);
+
+  // Create the collection
+  vt::runInEpochCollective([&]{
+    proxy = vt::theCollection()->constructCollective<MyCol, &myMap>(range);
+  });
+
+  // No elements should be created at this point; this must be in a collective
+  // epoch so it doesn't race with future insert messages
```
PhilMiller, line 84: This usage, in combination with the documentation, makes me kind of uneasy with the interface design here. The meaning of `constructCollective` changes depending on the type of the collection being constructed. I'm inclined toward saying that the relevant construction method should have a different name, so the semantics of the call don't have to depend on the inheritance of a template argument.

docs/md/insertable-collections.md
lifflander: `InsertableCollection` are always dynamically inserted and may be sparse depending on which indices the user actually inserts. I think that dynamic insertion is the defining trait. Being sparse is just a result of what ends up being inserted.

Also, insertable collections support element deletions, which are not supported for up-front constructed collections.

tests/unit/collection/test_collective_insertable.extended.cc
lifflander: In the implementation, this call isn't actually type dependent on the insertable trait. What really happens is that insertable collection just returns a boolean `isStaticSized()` that is `false` when it is insertable.

lifflander: VT does a query on that at runtime to see if its "insertable" and determines based on that whether to construct all the elements up front.

However, I agree that we should have a distinct construct for insertable. Maybe `constructCollectiveInsertable`?

PhilMiller: Ok, so given that it's really insertion *and* deletion *and* potential divergence from complete-axis-aligned-prism membership, I think `DynamicCollection` would be a less confining name.

That said, there's still an orthogonality issue here. Users might legitimately want (my guess of the most likely hypothetical) a collection in which the initial elements are created by bulk insertion of a given range, but they then have the freedom to later insert or delete elements.

This starts to get into constructing a [Parameter Object](https://wiki.c2.com/?ParameterObject), to separately specify desired dynamism, initial construction, and index range bound. There's also a design tension between the *caller* of the collection construction routine specifying these things, and the collection class itself specifying them. We should think a little bit about what we want the relevant design to look like for applications. Right now, I think you're kind of mixing the two possibilities. On general principal, that mixture is probably not ideal.

PhilMiller: Here's a vague outline of what would seemingly go into such a parameter objects

```c++
struct ArrayParameters {
  std::optional<IndexT> bulk_insertion_range_;
  std::optional<IndexT> index_bounds_; // invariant: bulk_insertion_range_ is a subset of index_bounds_
  bool dynamic_membership_;
};
```

Looked at like this, is there an actual need for the `bool dynamic_membership_`? I haven't checked calls to `isStaticSized()`, but is the system going to do anything special for collections on which that parameter would be set to `false` after construction time, distinct from which elements, if any, it constructs?

PhilMiller: Even given a bulk insertion range, we could still allow users the flexibility of providing an action to do the insertions entirely by themselves, and just error-check afterwards that exactly that set of objects was inserted *somewhere*, if the user doesn't want them initially inserted wherever the mapping function would place them by default.

lifflander: We would probably want to include the initial map function in the ParameterObject. And I've been planning to expand map function to stateful functors (maybe using an objgroup? passing a proxy to one that matches some concept?).

PhilMiller: That sounds good, yeah

lifflander: The only difference (beyond up-front element creation) would have been the initial creation of an insert epoch, which is simplified/removed by this PR. In general, tracking of insert epochs is added, but that's a simple add-on. I initially designed the API as type-specific intended to have different types on the proxies as well. I moved away from that design over time. Right now, every proxy can call `proxy.insert(..)` but it will throw an assertion if the collection does not support that.

PhilMiller: Amending my earlier remark: If the user says they want to end up with a fixed range of indices having elements, but do it themselves, I think we should just structure that exactly the same way we would dynamic insertion. The runtime provides the caller with an epoch, the caller does all the insertions in that epoch, and once that epoch has terminated, the runtime does a bit of checking of the result. No need for a separate interface mechanism from dynamic insertion.

As for the checking, I believe there's an efficient algorithm that doesn't require looping over the index range and evaluating the map for each index:
* Do a reduction to compute the total number of elements constructed, comparing that to the volume of the range
* Every node loops over the elements it has been informed of as their home node, and checks that their indices are in range and the map attributes them to this node

There's also a necessary error check against duplicate insertion, that during an insertion epoch, we should never see a location informing message describing an element for which we already have a location

PhilMiller: Improving on that algorithm a bit:

*  (as above) Every node loops over the elements it has been informed of as their home node, and checks that their indices are in range and the map attributes them to this node
* Do a sum reduction over the number of elements each node has been informed of as the home node, and compare that sum to the volume of the range

That removes the need for the separate check against duplicate insertion for ensuring coverage of the  insertion range, but we should still do it for its own sake. I think the following would (for the dynamic bulk insertion case) satisfy it:

* Do a sum reduction over the number of elements present on each node, and compare that sum to the volume of the range

I think that ends up establishing a bijection between indices in the range and elements actually constructed.

lifflander:
This is implemented by another PR

PR #826, opened 2020-06-01
#708: Pass subphase timings through to ProcStats, and write them out

PhilMiller:
Fixes: #708


PhilMiller:
I think we should close #708 with this, since further work on models and/or strategies that make effective use of the subphase timing data is a pretty open-ended project.

PhilMiller:
@ppebay Have a look at the format in which I'm writing out the subphase times. Will that pose a problem for the analysis tools?

src/vt/vrt/collection/balance/proc_stats.cc
```diff
@@ -202,16 +208,29 @@ void ProcStats::outputStatsFile() {
...
+
+      auto obj_str = fmt::format("{},{},{},{},[", i, id, time, subphases);
+      for (size_t s = 0; s < subphases; s++) {
+        obj_str += std::to_string(subphase_times[s]);
+        if (s != subphases - 1)
+          obj_str += ",";
+      }
+
+      obj_str += "]\n";
+
```
PhilMiller, line 230: @ppebay Here's where I've modified the output

lifflander:
@PhilMiller What happened here? Why is MPI intercept code on this branch?

Can you write a Doxygen comment documenting the new format (and current one also)?

PhilMiller:
Rebased to clear up the integration issues.

Wrote up a full description of the stats file output in the comments.

lifflander:
This seems to be failing Travis CI sometimes

PR #860, opened 2020-06-11
791 Fix bug in collection group re-creation after LB

### LINKED ISSUE #791: Does not rebuild group for collection after LB
**Describe the bug**
After LB, if the set of nodes the collection is mapped to shrinks, the group must be re-created for reductions to complete properly. This is a latent bug that requires multiple collections or comm-aware LB to show up.

lifflander:
Fixes #791 

This addresses a bug when a collection hits only some of the nodes. Initially, the code works correctly because a group is created. However, after LB, if elements are migrated, the group is not re-created. Thus, reductions will fail to complete.

Does not contain the fix yet. Just a test that fails.

### DIFF
--- a/tests/unit/collection/test_collection_group_recreate.cc
+++ b/tests/unit/collection/test_collection_group_recreate.cc
@@ -1,0 +1,136 @@
+/*
+//@HEADER
+// *****************************************************************************
+//
+//                      test_collection_group_recreate.cc
+//                           DARMA Toolkit v. 1.0.0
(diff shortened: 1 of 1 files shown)


lifflander:


```diff

Clones added
============
- tests/unit/collection/test_collection_group_recreate.cc  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5685176&bid=18608051)

lifflander:
Fixed in a different PR.

PR #862, opened 2020-06-13
#845 Log events 'between' top-level schedulers; can cover when VT is not running.

pnstickne:
All code that (consistently) uses `runSchedulerWhile` will emit events "between" such top-level loops. This also includes time between the last `runSchedulerWhile` and when the scheduler is finalized.

The red processing events on node 0 are time spent between the calls to `runSchedulerWhile` which is called via `runInEpoch*` (an artificial delay has been added). Node 1 jumps directly to vt::finalize and thus does not have (much) time "between" top-level schedulers.

image

These changes should be matched in all client usages to never call `runScheduler` directly. This is because a 'between' event may remain on the open event stack until the next call to `runSchedulerWhile` or VT finalization.

--

Using user events (older commits):

image

Fixes #845

### DIFF
--- a/src/vt/scheduler/scheduler.cc
+++ b/src/vt/scheduler/scheduler.cc
@@ -290,9 +290,7 @@ void Scheduler::runSchedulerWhile(std::function<bool()> cond) {
     "Nested schedulers never expected from idle context"
   );
 
-  if (action_depth_ == 0) {
-    between_sched_event_ = nullptr;
-  }
+  between_sched_event_ = nullptr;
 
   triggerEvent(SchedulerEventType::BeginSchedulerLoop);
 
@@ -369,6 +367,8 @@ void runSchedulerThrough(EpochType epoch) {
 }
 
 void runInEpochRooted(ActionType&& fn) {
+  theSched()->between_sched_event_ = nullptr; // loop will be entered
+
   auto ep = theTerm()->makeEpochRooted();
   theMsg()->pushEpoch(ep);
   fn();
@@ -378,6 +378,8 @@ void runInEpochRooted(ActionType&& fn) {
 }
 
 void runInEpochCollective(ActionType&& fn) {
+  theSched()->between_sched_event_ = nullptr; // loop will be entered
+
   auto ep = theTerm()->makeEpochCollective();
   theMsg()->pushEpoch(ep);
   fn();
--- a/src/vt/scheduler/scheduler.cc
+++ b/src/vt/scheduler/scheduler.cc
@@ -265,10 +265,6 @@ void Scheduler::scheduler(bool msg_only) {
       is_idle = true;
       triggerEvent(SchedulerEventType::BeginIdle);
     }
-
-  } else {
-    // n.b. Work queue is empty - however, if a trace event was written WITHOUT
-    // new work being queued then the trace state may be in a non-idle state.
   }
 
   if (not msg_only) {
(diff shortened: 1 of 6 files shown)


src/vt/scheduler/scheduler.cc
```diff
@@ -287,6 +300,16 @@ void Scheduler::runSchedulerWhile(std::function<bool()> cond) {
...
   triggerEvent(SchedulerEventType::EndSchedulerLoop);
+
+  if (action_depth_ == 0) {
+    // Start an event representing time outside of top-level scheduler.
+    if (outside_scheduler_event_type_ == trace::no_user_event_id) {
+      outside_scheduler_event_type_ = trace::registerEventHashed(
+        "Between Top-Level VT Schedulers"
+      );
+    }
+    outside_scheduler_event_ = trace::TraceScopedEvent{outside_scheduler_event_type_};
```
lifflander, line 311: Can we make this a {begin,end}Processing instead of a scoped event? That way it won't show up as idle/overhead

pnstickne: The first draft was entirely outdated - just a placeholder to add notes before I pushed again :|

pnstickne: The idle control already occurs when exiting a top-level event per the rules of "when no more messages are queued". I think it's useful to keep separate, as it could be the case that one is 'between loops' while VT is 'not idle' as messages remain.

src/vt/scheduler/scheduler.h
```diff
@@ -169,6 +187,10 @@ struct Scheduler : runtime::component::Component<Scheduler> {
   std::size_t last_threshold_memory_usage_ = 0;
   std::size_t threshold_memory_usage_ = 0;
   std::size_t last_memory_usage_poll_ = 0;
+
+  // Access to between_sched_event_
+  friend void vt::runInEpochRooted(ActionType&& fn);
```
pnstickne, line 192: A bit awkward.

Could move / proxy the runInEpoch* calls into the scheduler, or expose a method to 'indicate entering a loop is imminent', although neither seems necessarily better..

src/vt/scheduler/scheduler.cc
pnstickne: Oh, hmm. It is 'not idle' in this case..

src/vt/scheduler/scheduler.cc
```diff
@@ -265,28 +277,42 @@ void Scheduler::scheduler(bool msg_only) {
 }
 
 void Scheduler::runSchedulerWhile(std::function<bool()> cond) {
+  // This loop construct can run either in a top-level or nested context.
+  // 1. In a top-level context the idle time will encompass the time not
+  //    processing any actions when the work queue is empty. Leaving starts
+  //    a 'between scheduler' event which will only complete on the next call.
+  // 2. In a nested context, idle is always existed at the end
```
pnstickne, line 284: Typo 'existed'

src/vt/scheduler/scheduler.cc
pnstickne: @lifflander Updated to use begin/end event, and also the contract:

When a top-level runSchedulerWhile ends, idle will be exited, regardless if there are messages remaining and will only be entered again on the next call to runSchedulerWhile.

PhilMiller:
The substantive changes all look fine.

examples/collection/migrate_collection.cc
```diff
@@ -86,6 +86,13 @@ static void migrateToNext(ColMsg* msg, Hello* col) {
   col->migrate(next_node);
 }
 
+void waitToShowEventBetweenLoops(int millis) {
+  // Wait for a small duration so that the trace event issued
+  // between runInEpochRooted is noticable.
+  double end = MPI_Wtime() + (millis / 1000.0);
+  while (MPI_Wtime() <= end) { /* loop */ }
+}
+
```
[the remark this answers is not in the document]
pnstickne: Removed changes.

pnstickne:
Updated to remove an extra/unused declaration..

lifflander:
@PhilMiller Can you approve the changes so we can merge this?

PR #863, opened 2020-06-15
#853 Tests only: fix race condition in scheduler loop tests

### LINKED ISSUE #853: Fix race condition causing hang in test_scheduler_loop_nesting
**Describe the bug**

Occurs  intermittently on the macosx CI target. TD hang (timeout) amidst `412/478 Test #412: vt:TestSchedulerLoop.test_scheduler_loop_nesting_1_proc_4 ...................***Timeout  60.12 sec
843
` Look at this failure.

https://github.com/DARMA-tasking/vt/runs/758943913?check_suite_focus=true

pnstickne:
Just test ack to avoid entering an out-of-order wait.

Fixes #853

### DIFF
--- a/tests/unit/scheduler/test_scheduler_loop.cc
+++ b/tests/unit/scheduler/test_scheduler_loop.cc
@@ -55,10 +55,11 @@ struct TestMsg : vt::Message {
   int action_;
 };
 
-static int last_action_completed = 0;
 static bool done = false;
 
-static std::string actions[] = {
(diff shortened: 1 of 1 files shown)


lifflander:
I confirmed that this fixes the test, ran many times without failure.

lifflander:
@pnstickne Looks like we ran out of time on the Travis CI. It finished running all the tests and just failed at the end.

PR #864, opened 2020-06-15
Fix nvcc warning about sign mismatch

PhilMiller:
> EMPIRE/TPL/vt/src/vt/trace/trace_common.h(59): warning: integer conversion resulted in a change of sign


PhilMiller:
I could swear I've fixed this exact thing previously, but it's not in develop or the stable branch, so here we go again.

PR #888, opened 2020-06-23
854 - Outside Scheduler Loop Events

pnstickne:
Ported to beta.9 branch.

**Note**: this _may_ fail in code that uses manual scheduler loops (not using `runSchedulerWhile`) when tracing is enabled if begin/end processing events are intertwined. All client code should be updated to use `runSchedulerWhile` for 'guaranteed' pairings.

It might be useful to accept the other scheduler-loop constructs into beta9, if such makes implementation easier.

### DIFF
--- a/src/vt/trace/trace.cc
+++ b/src/vt/trace/trace.cc
@@ -116,41 +116,48 @@ Trace::Trace(std::string const& in_prog_name, std::string const& in_trace_name)
    */
 
   incremental_flush_mode = Z_SYNC_FLUSH;
-}
 
-Trace::Trace() { }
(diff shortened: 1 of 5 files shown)


src/vt/trace/trace.cc
```diff
@@ -229,13 +229,19 @@ void Trace::setupNames(
 }
 
 /*virtual*/ Trace::~Trace() {
+  // Always end any between-loop event left open.
```
pnstickne, line 232: Here instead of `cleanupTracesFile` to avoid potential of recursion.

PR #894, opened 2020-06-25
893 Add GitHub Actions for code quality checks

### LINKED ISSUE #893: Add hooks for Github checks for commits
**What Needs to be Done?**

- [ ] Trailing whitespace
- [ ] git --check
- [ ] Whatever else I find that might be useful

lifflander:
Fixes #893

### DIFF
--- a/.github/workflows/check-commit-format.yml
+++ b/.github/workflows/check-commit-format.yml
@@ -8,6 +8,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
       - name: Fetch base_ref HEAD
         run: git fetch --depth=1 origin +refs/heads/${{github.base_ref}}:refs/remotes/origin/${{github.base_ref}}
       - name: Display base sha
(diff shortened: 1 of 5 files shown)


lifflander:
This found some real errors in the code base:
```
./src/vt/vrt/collection/manager.h:289
./src/vt/vrt/collection/manager.h:1315
./src/vt/vrt/collection/manager.h:1316
./src/vt/vrt/context/context_vrtmessage.h:65
```

However, I need to exclude lib somehow.

lifflander:
> Good to go once you've got the exclusions squared away

Exclusions are working now. I now have two checks, one is for trailing space the other runs `git --check` on the range of PR commits.

  - I was looking for a spell checker for comments, but can't find anything easy to use. Might be more annoying that its worth also.
  - At some point we should add clang-tidy/cppcheck, but maybe not in this PR. I already have a container that runs those checks locally. I just need to automate it.

What other checks could I add that would be useful?

lifflander:
Ok, I've added a new check for commit format! All working now.

PhilMiller:
Probably for a separate pr, or several, but clang static analyzer would be good.

If we go that route, we may also end up needing something like I think SonarQube does, to track which reported issues are new, vs preexisting

PR #898, opened 2020-06-26
869 make command-line arguments in VT non-static 

### LINKED ISSUE #869: Make command-line arguments in VT non-static
**What Needs to be Done?**

Currently, if you look at `vt/src/configs/arguments/args.h` and `vt/src/configs/arguments/args.cc` the command-line arguments, parsed using the CLI11 library, are static member of  the `vt::arguments::ArgConfig` class. After CLI runs, it populates these static variables. Then, as VT needs, it reads these directly throughout the system. This design limits the flexibility of how arguments get input into VT.

The work is to make these non-static instances and reveal these to the VT system in a way that is runtime-instance specific and groups them all together with a instance of `ArgConfig`. Then, in all the places where `ArgConfig` is used directly, convert these to access the instantiated class.

There are several designs possible with non-static members. Let's discuss possibilities here before you implement. For example, we could create a singleton, like `vt::theArg()` that holds them that is accessible throughout the runtime.

@pnstickne

cz4rs:
draft for further discussion regarding #869 

- promote ArgConfig to a component
- use default member initialization
- remove unnecessary `include`s from headers when possible

Fixes #869


lifflander:


```diff

Clones added
============
- src/vt/configs/error/error.impl.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5765956&bid=18871868)

lifflander:


```diff

Clones added
============
- src/vt/configs/error/error.impl.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5765956&bid=18871868)

src/vt/configs/arguments/args.h
```diff
@@ -45,183 +45,203 @@
...
 #include <functional>
+#include <memory>
 #include <string>
 #include <vector>
 #include <tuple>
 
 namespace vt { namespace arguments {
 
-struct ArgConfig {
+struct ArgConfig : runtime::component::Component<ArgConfig> {
```
pnstickne, line 63: Random note: ideally it'd be good to separate the command-line arguments (and artifact of input/combining and argparse) from the internal structures used in VT to apply the options.

src/vt/event/event.cc
```diff
@@ -62,7 +62,7 @@ namespace vt { namespace event {
 
 void AsyncEvent::initialize() {
 # if backend_check_enabled(trace_enabled)
-  if (vt::arguments::ArgConfig::vt_trace_event_polling) {
+  if (vt::theArgConfig()->vt_trace_event_polling) {
```
pnstickne, line 65: It might be nice to 'inject' theArgConfig into components/types as applicable. I'm a bit undecided on this.

lifflander:


```diff

Clones added
============
- src/vt/configs/error/error.impl.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5765956&bid=18871868)

src/vt/event/event.cc
lifflander: I agree in the future we might might want to inject into components. We could do the same for other component dependencies also. For now, I think this design will suffice.

lifflander:


```diff

Clones added
============
- src/vt/configs/error/error.impl.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5765956&bid=18871868)

cz4rs:
commit d34db06d54a82531c610b6fbf9f8ff6a254274dd removes all invocations of `vt::theArgConfig` that happen before `Runtime::initializeComponents()` is called, so at least some tests / examples can run without seg faults (this gives me 98% of tests passing on my machine and only 30% in CI, but that's for later analysis)

- most of the calls come from `vt::runtime::Runtime` constructor, with `arguments::ArgParse::parse` being prime example
- `debug_print` uses debug family arguments, so it would have to be avoided early or have specialized version with default options
- `Context` depends on `user_argc_` and `user_argv_`, so it cannot be registered without getting these

@lifflander  @pnstickne do you guys have any advice on this? especially whether moving `initializeComponents()` to the beginning of `Runtime` ctor even makes sense?

lifflander:
> @lifflander @pnstickne do you guys have any advice on this? does it even make sense to move `initializeComponents()` to the beginning of `Runtime` constructor?

Ah, I see the problem that you are encountering. Basically, the arg config being a component means it starts up with the pack, but the `vt::Runtime` uses arguments pre-startup to configure the startup and output options. I think you should construct the arg config outside of the components pack and then move it into place for the component. So construct it early in each runtime. Let me know if you have questions on how to achieve this.

lifflander:


```diff

Clones added
============
- src/vt/configs/error/error.impl.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5765956&bid=18871868)

src/vt/configs/arguments/args.cc
```diff
@@ -710,22 +582,22 @@ std::tuple<int, std::string> parseArguments(CLI::App& app, int& argc, char**& ar
...
   }
 
-  ArgConfig::prog_name = prog_name;
-  ArgConfig::argv_prog_name = argv[0];
-  ArgConfig::mpi_init_args = mpi_args;
-  ArgConfig::passthru_args = passthru_args;
+  prog_name = clean_prog_name;
+  argv_prog_name = argv[0];
+  mpi_init_args = mpi_args;
+  passthru_args = passthru_args;
```
lifflander, line 600: Codacy found an issue: [Redundant assignment of 'passthru_args' to itself.](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5765956)

lifflander:


```diff

Issues
======
- Added 1
           

Clones added
============
- src/vt/configs/error/error.impl.h  1
         

Clones removed
==============
+ src/vt/runtime/runtime.cc  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5765956&bid=18871868)

lifflander:


```diff

Issues
======
- Added 1
           

Clones added
============
- src/vt/configs/error/error.impl.h  1
         

Clones removed
==============
+ src/vt/runtime/runtime.cc  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5765956&bid=18871868)

src/vt/configs/arguments/args.cc
lifflander, line 600: Codacy found an issue: [Redundant assignment of 'passthru_args' to itself.](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5765956)

lifflander:


```diff

Issues
======
- Added 1
           

Clones added
============
- src/vt/configs/error/error.impl.h  1
         

Clones removed
==============
+ src/vt/runtime/runtime.cc  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5765956&bid=18871868)

src/vt/configs/arguments/args.cc
```diff
@@ -710,22 +581,22 @@ std::tuple<int, std::string> parseArguments(CLI::App& app, int& argc, char**& ar
...
```
lifflander, line 599: Codacy found an issue: [Redundant assignment of 'passthru_args' to itself.](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5765956)

lifflander:


```diff

Issues
======
- Added 1
           

Clones added
============
- src/vt/configs/error/error.impl.h  1
         

Clones removed
==============
+ src/vt/runtime/runtime.cc  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5765956&bid=18871868)

src/vt/configs/arguments/args.cc
lifflander, line 599: Codacy found an issue: [Redundant assignment of 'passthru_args' to itself.](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5765956)

lifflander:


```diff

Clones added
============
- src/vt/configs/error/error.impl.h  1
         

Clones removed
==============
+ src/vt/runtime/runtime.cc  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5765956&bid=18871868)

lifflander:


```diff

Clones added
============
- src/vt/configs/error/error.impl.h  1
         

Clones removed
==============
+ src/vt/runtime/runtime.cc  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5765956&bid=18871868)

lifflander:


```diff

Clones added
============
- src/vt/configs/error/error.impl.h  1
         

Clones removed
==============
+ src/vt/runtime/runtime.cc  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5765956&bid=18871868)

lifflander:


```diff

Clones added
============
- src/vt/configs/error/error.impl.h  1
         

Clones removed
==============
+ src/vt/runtime/runtime.cc  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5765956&bid=18871868)

lifflander:


```diff

Issues
======
- Added 1
           

Clones added
============
- src/vt/configs/error/error.impl.h  1
         

Clones removed
==============
+ src/vt/runtime/runtime.cc  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5765956&bid=18871868)

src/vt/runtime/runtime.cc
```diff
@@ -556,15 +557,23 @@ void Runtime::finalizeMPI() {
...
 
 void Runtime::initializeComponents() {
-  vt_debug_print(runtime, node, "begin: initializeComponents\n");
+  //vt_debug_print(runtime, node, "begin: initializeComponents\n");
 
   using component::ComponentPack;
   using component::Deps;
 
   p_ = std::make_unique<ComponentPack>();
+  std::string const prog_name = argConfig_->prog_name;
```
lifflander, line 566: Codacy found an issue: [Variable 'prog_name' is assigned a value that is never used.](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5765956)

lifflander:


```diff

Clones added
============
- src/vt/configs/error/error.impl.h  1
         

Clones removed
==============
+ src/vt/runtime/runtime.cc  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5765956&bid=18871868)

lifflander:


```diff

Clones added
============
- src/vt/configs/error/error.impl.h  1
         

Clones removed
==============
+ src/vt/runtime/runtime.cc  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5765956&bid=18871868)

lifflander:


```diff

Clones added
============
- src/vt/configs/error/error.impl.h  1
         

Clones removed
==============
+ src/vt/runtime/runtime.cc  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5765956&bid=18871868)

lifflander:


```diff

Clones added
============
- src/vt/configs/error/error.impl.h  1
         

Clones removed
==============
+ src/vt/runtime/runtime.cc  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5765956&bid=18871868)

cz4rs:
a couple of builds are failing when dereferencing `MPI_Comm* comm` during construction of `theContext`
```gdb
Program received signal SIGSEGV, Segmentation fault.
0x00005555556fedda in vt::ctx::Context::Context (this=0x55555587b6d0, argc=<optimized out>, 
    argv=<optimized out>, is_interop=<optimized out>, comm=0x18)
77	  if (comm != nullptr and *comm != MPI_COMM_NULL) {
(gdb) bt
#0  0x00005555556fedda in vt::ctx::Context::Context (this=0x55555587b6d0, argc=<optimized out>, 
    argv=<optimized out>, is_interop=<optimized out>, comm=0x18)
#1  0x0000555555692ca3 in std::make_unique<vt::ctx::Context, int&, char**, bool&, int*> ()
#2  vt::runtime::component::ComponentConstructor<vt::ctx::Context, void, int&, char**, bool&, int*>::apply<int&, char**, bool&, int*> () 
#3  vt::runtime::component::Component<vt::ctx::Context>::staticInit<int&, char**, bool&, int*> ()
#4  vt::runtime::component::ComponentPack::registerComponent<vt::ctx::Context, vt::arguments::ArgConfig, int&, char**, bool&, int*>(vt::ctx::Context**, vt::runtime::component::BaseComponent::DepsPack<vt::arguments::ArgConfig>, int&, char**&&, bool&, int*&&)::{lambda()#1}::operator()() (this=<optimized out>)
#5  std::_Function_handler<void (), vt::runtime::component::ComponentPack::registerComponent<vt::ctx::Context, vt::arguments::ArgConfig, int&, char**, bool&, int*>(vt::ctx::Context**, vt::runtime::component::BaseComponent::DepsPack<vt::arguments::ArgConfig>, int&, char**&&, bool&, int*&&)::{lambda()#1}>::_M_invoke(std::_Any_data const&)
#6  0x00005555556b491a in std::function<void ()>::operator()() const (this=<optimized out>)
#7  vt::runtime::component::ComponentPack::construct (this=0x55555587ad60)
#8  0x000055555568bb44 in vt::runtime::Runtime::initializeComponents (this=this@entry=0x555555872040)
(...)
```
this doesn't happen when using Debug builds and even in the CI all builds apart from 4 gcc configurations are passing

I have tracked down the change that introduces this - capturing by reference when registering components:
```diff
@@ registry::AutoHandlerType ComponentPack::registerComponent(
-      [ref,this,cons...]() mutable {
+      [ref,this,&cons...]() mutable {
```
when we pass the MPI_Comm to `registerComponent` before the construction it has valid value, but somehow it gets changed into something like `0x18` (as in the stacktrace above)

lifflander:
@cz4rs I have a proper fix for the problem that supports move-only parameters along with l-val refs.

https://github.com/DARMA-tasking/vt/pull/972

You will want to remove the ComponentConstructor changes (for detecting arguments) from your branch as they will be merged as part of this #972 now.

lifflander:


```diff

Clones added
============
- src/vt/configs/error/error.impl.h  1
         

Clones removed
==============
+ src/vt/runtime/runtime.cc  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5765956&bid=18871868)

lifflander:


```diff

Clones added
============
- src/vt/configs/error/error.impl.h  1
         

Clones removed
==============
+ src/vt/runtime/runtime.cc  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5765956&bid=18871868)

lifflander:


```diff

Clones added
============
- src/vt/configs/error/error.impl.h  1
         

Clones removed
==============
+ src/vt/runtime/runtime.cc  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5765956&bid=18871868)

lifflander:


```diff

Clones added
============
- src/vt/configs/error/error.impl.h  1
         

Clones removed
==============
+ src/vt/runtime/runtime.cc  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5765956&bid=18871868)

lifflander:


```diff

Clones added
============
- src/vt/configs/error/error.impl.h  1
         

Clones removed
==============
+ src/vt/runtime/runtime.cc  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5765956&bid=18871868)

lifflander:


```diff

Clones added
============
- src/vt/configs/error/error.impl.h  1
         

Clones removed
==============
+ src/vt/runtime/runtime.cc  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5765956&bid=18871868)

lifflander:


```diff

Clones added
============
- src/vt/configs/error/error.impl.h  1
         

Clones removed
==============
+ src/vt/runtime/runtime.cc  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5765956&bid=18871868)

lifflander:


```diff

Clones added
============
- src/vt/configs/error/error.impl.h  1
         

Clones removed
==============
+ src/vt/runtime/runtime.cc  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5765956&bid=18871868)

lifflander:


```diff

Clones added
============
- src/vt/configs/error/error.impl.h  1
         

Clones removed
==============
+ src/vt/runtime/runtime.cc  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5765956&bid=18871868)

cz4rs:
@lifflander 
I have rebased to include the fix for #972 (thanks for that!) and fixed the remaining issues. It looks like the PR is finally ready for review / merging. 

A couple of random remarks:

- TimeTrigger tests were failing on numerous occasions (re-running usually helped, but the tolerance might need to be increased)
`The following tests FAILED:
	178 - vt:TestTimeTrigger.test_time_trigger_2_proc_2 (Failed)`
- there are some places where we cannot use `vt_debug_print` and I have used simple `fmt` printing instead (similar to what is used in [context.cc](src/vt/context/context.cc))
- we have to print shutdown banner before we destroy components if we want to use standard `vt_debug_print` infrastructure, I didn't see any change in statistics displayed
- `ArgParse` class was removed as it didn't offer any real separation in it's current shape, although some refactoring as mentioned by @pnstickne could be considered
- in the end, `theObjGroup` part was omitted in `construct()` method (and the method simply returns moved unique_ptr)
```cpp
  auto proxy = vt::theObjGroup()->makeCollective<ArgConfig>(ptr.get());
  proxy.get()->setup(proxy); // you might not need this part
```

lifflander:
> @lifflander
> I have rebased to include the fix for #972 (thanks for that!) and fixed the remaining issues. It looks like the PR is finally ready for review / merging.
> 

Great!

> * TimeTrigger tests were failing on numerous occasions (re-running usually helped, but the tolerance might need to be increased)

Yes, I've seen this. There's another occasional bug on develop I'm trying to solve for `test_load_balancer`.

> * there are some places where we cannot use `vt_debug_print` and I have used simple `fmt` printing instead (similar to what is used in [context.cc](src/vt/context/context.cc)) [a85ee83](https://github.com/DARMA-tasking/vt/commit/a85ee830a56e319c8bceac2d6c4014460d5895c6)
> * we have to print shutdown banner before we destroy components if we want to use standard `vt_debug_print` infrastructure, I didn't see any change in statistics displayed [b63d62e](https://github.com/DARMA-tasking/vt/commit/b63d62e378a6220688f17393a26591853cff0d9b)

I see. Let me look at the best way to work around that.

> * `ArgParse` class was removed as it didn't offer any real separation in it's current shape, although some refactoring as mentioned by @pnstickne could be considered

I'll have a look at that.

> * in the end, `theObjGroup` part was omitted in `ArgConfig::construct()` method (and the method simply returns consumed `unique_ptr`)

Makes sense. You shouldn't need a objgroup for the arg config since it doesn't do communication.

src/vt/configs/arguments/args.h
```diff
@@ -45,184 +45,219 @@
 #if !defined INCLUDED_VT_CONFIGS_ARGUMENTS_ARGS_H
 #define INCLUDED_VT_CONFIGS_ARGUMENTS_ARGS_H
 
-// Do not pull in any VT dependencies here
+// Do not pull in any other VT dependencies here
+#include "vt/runtime/component/component.h"
```
pnstickne, line 49: Disregards previous comment. Does this increase the bleed of "component.h" usages? C++ is very leaky.

src/vt/configs/arguments/args.h
```diff
@@ -45,184 +45,219 @@
...
 #include <tuple>
 
+namespace CLI {
+class App;
+}
+
 namespace vt { namespace arguments {
 
-struct ArgConfig {
+struct ArgConfig : runtime::component::Component<ArgConfig> {
```
pnstickne, line 63: It feels off to combine the arguments with the type responsible to load such from the command line.

I would separate the/a component, which is useful for initialization, and the configuration that the component loads. It's possible to have `theConfig` (say populated during the component initialization), as separate from any other aspect of the components. Keeping the settings out of the "the component" would also decreases the coupling between settings and how settings are loaded.

src/vt/configs/arguments/args.h
```diff
@@ -45,184 +45,219 @@
...
+  bool colorize_output;
+
+  int32_t vt_sched_num_progress = 2;
+  int32_t vt_sched_progress_han = 0;
+  double vt_sched_progress_sec  = 0.0;
+  bool vt_no_sigint    = false;
+  bool vt_no_sigsegv   = false;
+  bool vt_no_terminate = false;
+  std::string vt_memory_reporters =
+  # if vt_check_enabled(mimalloc)
```
pnstickne, line 125: Indent inconsistency.

src/vt/configs/debug/debug_colorize.h
```diff
@@ -52,7 +52,7 @@
 namespace vt { namespace debug {
 
 inline bool colorizeOutput() {
-  return arguments::ArgConfig::colorize_output;
+  return vt::theArgConfig()->colorize_output;
```
pnstickne, line 55: Just `theConfig` might be sufficient, when referring to the configuration itself. A component might be (and might not even need to be named, depending on how it's used) called `theConfigReader` or similar.

src/vt/configs/arguments/args.h
cz4rs: fixed

lifflander:


```diff

Clones added
============
- src/vt/configs/error/error.impl.h  1
         

Clones removed
==============
+ src/vt/runtime/runtime.cc  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5765956&bid=18871868)

src/vt/configs/arguments/args.h
cz4rs: storage (and the warning about includes) moved into separate class (see `app_config.h)`, now only `Runtime` includes this header

src/vt/configs/arguments/args.h
cz4rs: configuration variables were extracted into `struct AppConfig`
I have moved all members for now - I will check if some of them should remain here

src/vt/configs/arguments/args.h
cz4rs: this leaves only `<string>` and `<vector>` as dependencies for `AppConfig`

src/vt/configs/debug/debug_colorize.h
cz4rs: renamed `theArgConfig()` to `theConfig()` and it returns simple configuration struct now (implementation in `runtime_get.cc` stands out from the rest of getters, but I don't think it's much of an issue)

I haven't renamed `ArgConfig` class since it's strictly for CLI parsing at the moment, so it's not a general `ConfigReader` yet (although it doesn't store the configuration any more, so maybe it could become the `ArgParse` now?)

cz4rs:
@pnstickne  thanks for the comments! I have updated the code and introduced some simple separation between parsing and storage

lifflander:


```diff

Clones added
============
- src/vt/configs/error/error.impl.h  1
         

Clones removed
==============
+ src/vt/runtime/runtime.cc  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5765956&bid=18871868)

src/vt/configs/arguments/args.h
lifflander: I'm mostly worried about including `component.h` here. Component includes reducers and other headers which will create a lot more dependencies between files. @pnstickne What do you think of a light component that doesn't have any of that overhead?

lifflander:
@cz4rs Can you look at what happens when component.h is included in the arg file? Does component end up included everywhere else creating more dependencies/slower compilation?

src/vt/runtime/runtime_component_fwd.h
```diff
@@ -51,6 +51,9 @@
 
 namespace vt {
 
+namespace arguments {
+struct ArgConfig;
```
lifflander, line 55: Should this be `AppConfig` instead of `ArgConfig`?

lifflander:
What in `vt_debug_print` is causing `DEBUG_VT_COMPONENTS` to be required? Can we loosen those checks to see if `theConfig()` is not null?

src/vt/configs/arguments/args.h
cz4rs: please note that in the latest version, `args.h` is included only in `args.cc`, `runtime.cc` and `runtime_get.cc` - so there's not much pollution going on here
the thing that gets included everywhere to access `theConfig()` is `app_config.h` which has really minimal set of dependencies

cz4rs:
> @cz4rs Can you look at what happens when component.h is included in the arg file? Does component end up included everywhere else creating more dependencies/slower compilation?

like mentioned in the discussion above, `args.h` gets included in only a couple of .cc files, and the header that gets included "everywhere" is `app_config.h` (which does not depend on `component.h`)

> What in vt_debug_print is causing DEBUG_VT_COMPONENTS to be required? Can we loosen those checks to see if theConfig() is not null?

```shell
$ cat src/vt/configs/debug/debug_print.h | grep -n theConfig
86:  ::vt::theConfig()->vt_debug_ ## opt
88:#define vt_debug_all_option ::vt::theConfig()->vt_debug_all
143:    if (!::vt::theConfig()->vt_quiet) {                        \
170:  if ((verb and ::vt::theConfig()->vt_debug_verbose) or not verb) {
179:    if (vt_option_check_enabled(mod, ModeEnum::flush) or ::vt::theConfig()->alwaysFlush()) {
189:    if (rt_option or vt::theConfig()->vt_debug_all) {
201:    if (rt_option or vt::theConfig()->vt_debug_all) {
```
there's a couple of calls, most importantly we probably want to respect vt_quiet
still, even when we want to check if theConfig is null, we have to first go through `CUR_RT` macro in `runtime_get.cc`, and this ends up depending on `theContext`
```cpp
#define IS_COMM_THREAD curRT->theContext->getWorker() == worker_id_comm_thread
#define CUR_RT_SAFE (IS_COMM_THREAD ? curRT : no_rt)
#define CUR_RT_TS curRT
#define CUR_RT CUR_RT_SAFE
```

src/vt/runtime/runtime_component_fwd.h
cz4rs: even with the addition of `AppConfig`, the component is still based on `ArgConfig` and forward declaration here means that we don't have to include `args.h` header in `runtime.h`
`AppConfig` needs to be known only in implementation files, where full header is required anyways

src/vt/runtime/component/component.h
```diff
@@ -187,13 +186,20 @@ struct PollableComponent : Component<T> {
...
-    vtAbort("PollableComponent should override the empty progress function");
-    return 0;
-  }
+  virtual int progress() override;
 
 };
 
 }}} /* end namespace vt::runtime::component */
 
+#include "vt/configs/error/hard_error.h"
```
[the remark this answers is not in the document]
cz4rs: this was creating circular dependency between `Component` and `ArgConfig` in earlier versions of the PR, but now with the dependencies cut down it actually builds fine without this change - should I revert it?
another option would be to extract `PollableComponent` into a separate header, which would remove the need for `hard_error.h` in `component.h`

src/vt/configs/arguments/args.h
pnstickne: +1 With the AppConfig structure separated the original comment is not applicable as it's a clean separation.

src/vt/configs/arguments/args.h
pnstickne: +1 This is a useful separation.

src/vt/configs/arguments/args.h
```diff
@@ -45,183 +45,55 @@
...
-  static char* argv_prog_name;
-
-  /// Arguments to pass to MPI Init.
-  /// Does not include argv[0]. Original char* objects.
-  static std::vector<char*> mpi_init_args;
-  /// Arguments are being ref-returend as the result of parse(..).
-  /// Does not include argv[0]. Original char* objects.
-  static std::vector<char*> passthru_args;
+public:
+  void addColorArgs(CLI::App& app);
```
pnstickne, line 77: These can likely be private to reduce scope.

src/vt/runtime/runtime.cc
```diff
@@ -238,42 +239,42 @@ void Runtime::pauseForDebugger() {
...
+  std::string dir  = vt::theConfig()->vt_stack_dir == "" ? "" : vt::theConfig()->vt_stack_dir + "/";
   std::string path = dir + file;
   FILE* f = fopen(path.c_str(), "w+");
   fprintf(f, "%s", str.c_str());
   fclose(f);
 }
 
 void Runtime::setupSignalHandler() {
-  if (!ArgType::vt_no_sigsegv) {
+  if (!argConfig_->config_.vt_no_sigsegv) {
```
pnstickne, line 264: It should be possible to use `theConfig` at this point (same for other occurrences in file).

src/vt/runtime/runtime.cc
```diff
@@ -292,11 +293,12 @@ void Runtime::setupTerminateHandler() {
 bool Runtime::tryInitialize() {
   bool const init_now = !initialized_ && !finalized_;
 
-  vt_debug_print(
-    runtime, unknown,
-    "Runtime: tryInitialize: initialized_={}, finalized_={}, init_now={}\n",
-    print_bool(initialized_), print_bool(finalized_), print_bool(init_now)
-  );
+  #if DEBUG_VT_COMPONENTS
+    fmt::print(
```
pnstickne, line 297: Even though `fmt` is bled, it's usage should not be encouraged.

I suspect this output adds little value as it is now (as even when the configuration was accessed it would be "off"), and I would remove all output that can happen BEFORE the config is basically initialized.

src/vt/runtime/runtime_get.cc
```diff
@@ -132,6 +133,8 @@ vrt::collection::balance::NodeStats* theNodeStats() { return CUR_RT->theNodeStat
 vrt::collection::balance::StatsRestartReader* theStatsReader() { return CUR_RT->theStatsReader;      }
 vrt::collection::balance::LBManager* theLBManager() { return CUR_RT->theLBManager;      }
 timetrigger::TimeTriggerManager* theTimeTrigger()   { return CUR_RT->theTimeTrigger;    }
+vt::arguments::AppConfig*   theConfig()             { return &CUR_RT->theArgConfig->config_;      }
+
```
pnstickne, line 137: Extra newline.

src/vt/configs/arguments/args.cc
```diff
@@ -696,7 +566,7 @@ std::tuple<int, std::string> parseArguments(CLI::App& app, int& argc, char**& ar
       rargs = &vt_args;
       rargs->push_back(c);
     } else {
-      passthru_args.push_back(c);
```
pnstickne, line 699: I recommend using local variables for local mutation. These values might not even need to be exposed.

src/vt/configs/arguments/args.h
pnstickne: Now `#include "vt/runtime/component/component.h"` can and should be removed.

src/vt/configs/arguments/args.cc
```diff
@@ -335,7 +208,7 @@ void addTraceArgs(CLI::App& app) {
   qze->group(traceGroup);
 }
 
-void addDebugPrintArgs(CLI::App& app) {
+void ArgConfig::addDebugPrintArgs(CLI::App& app) {
```
pnstickne, line 211: These can probably remain TU-local. If put in the class, they should be `private` to clearly indicate they are not external.

src/vt/configs/debug/debug_colorize.h
pnstickne: The name change is much more readable. Since the component has a very limited scope, the name is much less exposed. Leaving it as ArgParse is fine.

src/vt/configs/error/error.impl.h
```diff
@@ -57,6 +57,16 @@
 
 #include "fmt/format.h"
 
+namespace vt { namespace debug {
+
+std::string stringizeMessage(
```
pnstickne, line 62: Method doc would be useful.

src/vt/runtime/component/component.h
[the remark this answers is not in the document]
pnstickne: I would put it back if not needed - unless it adds value.

src/vt/runtime/component/component_pack.impl.h
```diff
@@ -91,11 +95,12 @@ registry::AutoHandlerType ComponentPack::registerComponent(
     [ref, this, tup = std::move(cons_tuple)]() mutable {
       auto ptr = tupleCons<T>(std::move(tup));
 
-      vt_debug_print(
-        runtime, node,
-        "ComponentPack: constructed component={}, pollable={}\n",
-        ptr->name(), ptr->pollable()
-      );
+      #if DEBUG_VT_COMPONENTS
```
pnstickne, line 98: The more I think about this, the more I think about this, the more it seems like it can just be eliminated.

src/vt/runtime/runtime.cc
```diff
@@ -238,42 +239,42 @@ void Runtime::pauseForDebugger() {
...
     return true;
   } else {
     return false;
   }
 }
 
 /*static*/ void Runtime::writeToFile(std::string const& str) {
-  std::string& app_name = ArgType::prog_name;
-  std::string name = ArgType::vt_stack_file == "" ? app_name : ArgType::vt_stack_file;
+  std::string& app_name = vt::theConfig()->prog_name;
```
pnstickne, line 252: (Musing)

Now that a non-static type is used, it _might_ be nice to reduce access to `theConfig` and have it stored either as a member, or even reduced locally. This does slightly reduce some text search ability though:

    // eg.
    auto* config = vt::theConfig();
    std::string dir = config->vt_stack_dir;

src/vt/runtime/runtime_component_fwd.h
pnstickne: (Confused)

src/vt/runtime/runtime.cc
cz4rs: This happens before any components are created, so 'theConfig' is not yet accessible - that's why we parse and store the configuration, and then move it in place during components creation

src/vt/runtime/component/component_pack.impl.h
cz4rs: elimination sounds good to me, although this specific line provides good overview of component creation order
any thoughts on this @lifflander ?

side note: I've tried to enable `DEBUG_VT_CONTEXT` in `context.cc` (which is there for similar purpose) and it doesn't compile:
```
/home/cz4rs/nga/vt/src/vt/context/context.cc:63:5:   required from here
../lib/fmt/fmt/core.h:896:67: error: static assertion failed: mismatch between char-types of context and argument
  896 |                              typename internal::char_t<S>::type>::value,
      |                                                                   ^~~~~
../lib/fmt/fmt/core.h:898:24: error: could not convert ‘fmt::v5::to_string_view<int>(((const int*)((int*)val)))’ from ‘fmt::v5::basic_string_view<int>’ to ‘fmt::v5::internal::init<fmt::v5::basic_format_context<std::back_insert_iterator<fmt::v5::internal::buffer<char> >, char>, fmt::v5::basic_string_view<char>, fmt::v5::internal::string_type>’
  898 |   return to_string_view(val);
      |          ~~~~~~~~~~~~~~^~~~~
      |                        |
      |                        fmt::v5::basic_string_view<int>
```
I don't know if it's worth fixing / raising an issue or this could be eliminated as well?

lifflander:


```diff

Clones removed
==============
+ src/vt/runtime/runtime.cc  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5765956&bid=18871868)

src/vt/configs/error/error.impl.h
cz4rs: fixed: removed redundant forward declaration (no longer necessary after reducing dependencies)

src/vt/configs/arguments/args.cc
cz4rs: changed to `private`

src/vt/runtime/runtime_get.cc
cz4rs: removed

src/vt/runtime/runtime.cc
cz4rs: removed

lifflander:


```diff

Clones removed
==============
+ src/vt/runtime/runtime.cc  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5765956&bid=18871868)

src/vt/runtime/component/component_pack.impl.h
cz4rs: for now, the `DEBUG_VT_COMPONENTS` and the traces were removed

src/vt/configs/arguments/args.h
lifflander: Yes, this looks good to me now

src/vt/configs/arguments/args.h
lifflander: Agreed

src/vt/runtime/runtime_component_fwd.h
lifflander: Ok, I mixed those up I think when reading the code.

src/vt/configs/arguments/args.h
lifflander: Looks like this was solved

src/vt/runtime/runtime.h
```diff
@@ -371,6 +369,7 @@ struct Runtime {
   static void termHandler();
 
 public:
+  ComponentPtrType<arguments::ArgConfig> theArgConfig;
```
lifflander, line 372: Can you make this match the singleton name: `theConfig`?

src/vt/runtime/component/component_pack.impl.h
```diff
@@ -91,12 +91,6 @@ registry::AutoHandlerType ComponentPack::registerComponent(
     [ref, this, tup = std::move(cons_tuple)]() mutable {
       auto ptr = tupleCons<T>(std::move(tup));
 
-      vt_debug_print(
```
lifflander, line 94: I think we need this back for debugging component startup

lifflander:


```diff

Clones removed
==============
+ src/vt/runtime/runtime.cc  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5765956&bid=18871868)

src/vt/runtime/runtime.h
lifflander: Nevermind.. this will cause other problems.

lifflander:


```diff

Clones removed
==============
+ src/vt/runtime/runtime.cc  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5765956&bid=18871868)

lifflander:
I'm working a little refactoring to print before the runtime starts up. This is very useful for debugging so I don't want to loose that functionality.

lifflander:


```diff

Clones removed
==============
+ src/vt/runtime/runtime.cc  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5765956&bid=18871868)

src/vt/runtime/component/component_pack.impl.h
lifflander: I've addressed this

src/vt/runtime/component/component_pack.impl.h
lifflander: I've put this back in place.. It's definitely needed to debug VT while components are not fully active. We need to be able to print after VT components are destroyed.

lifflander:


```diff

Clones removed
==============
+ src/vt/runtime/runtime.cc  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5765956&bid=18871868)

lifflander:
> there's a couple of calls, most importantly we probably want to respect vt_quiet
> still, even when we want to check if theConfig is null, we have to first go through `CUR_RT` macro in `runtime_get.cc`, and this ends up depending on `theContext`

@cz4rs I've resolved a few problems.

- Added the debug prints back in, and now `-vt_quiet` is respected even when VT is not fully initialized or it is finalized (as long as `vt::Runtime` is constructed.
- Moved the shutdown back to where it was before
- Fixed a bug you introduced in `test_signal_cleanup` (not actually cleaning up) by changing it back
- Wrote detailed comments in runtime.cc about the lifecycle of the ArgConfig
```C++
  /// =========================================================================
  /// Notes on lifecycle for the ArgConfig/AppConfig
  /// =========================================================================
  ///
  /// - After \c vt::Runtime is constructed, the ArgConfig lives in
  ///   \c arg_config_ and \c app_config_ contains a pointer to the internals.
  ///
  /// - After the pack registers the ArgConfig component, \c arg_config_ is no
  ///   longer valid. The config is in a tuple awaiting construction---thus, we
  ///   can't easily access it. app_config_ remains valid during this time
  ///
  /// - After construction, the \c arg_config_ is in the component
  ///   \c theConfig() and can be accessed normally
  ///
  /// - After \c Runtime::finalize() is called but before the pack is destroyed,
  ///   we extract the \c ArgConfig from the component and put it back in
  ///   \c arg_config_ for use after.
  ///
  /// - From then on, until the \c vt::Runtime is destroyed or VT is
  ///   re-initialized \c arg_config_ will contain the configuration.
  ///
  ///  For this to all work correctly, the \c vt_debug_print infrastructure
  ///  calls \c vt::config::getConfig() which always grabs the correct app
  ///  config, either from the component singleton or the \c vt::Runtime
  ///
  /// =========================================================================
```
- Added an extraction method that is used before the component pack is destroyed so the arg_config_ is not lost (in case you re-init the runtime) and is moved back into place
- Fixed some const issues---made access to AppConfig safer in some contexts

src/vt/runtime/runtime.cc
lifflander: Agreed, sounds like it would be useful in a follow-up PR perhaps.

cz4rs:
@lifflander Great, thanks for the fixes and improvements

PR #901, opened 2020-06-30
900 - Log user events from PMPI wrappers

pnstickne:
Logs user events for every MPI call _outside_ of VT processing a message. Per the image below, the MPI calls in VT prior to entering a VT event loop (and after exiting) are logged. Must be enabled via `-vt_trace_pmpi` (not to be confused with `-vt_trace_mpi`..)

Can be controlled via include/exclude patterns; some calls are excluded for practicality reasons.

It is currently a 'Component' to facilitate the registration of "user collective events", although such should be relatively easy to back-port to 1.0. (Remove override, base def, call startup explicitly..)

image


lifflander:


```diff

Clones removed
==============
+ src/vt/trace/trace_user_event.cc  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5778239&bid=18917384)

PhilMiller:
This looks like a good start, and may be useful even just as written.

Would it make sense to have the PMPI wrappers defined as a `Component`, and do the registration up-front in an initialization function? That would better guarantee uniformity across nodes.

In terms of which MPI routines to log, the variants of send and recv, the collectives, and `MPI_Wait*` seem like the most important to pick up.

pnstickne:
> This looks like a good start, and may be useful even just as written.
> 
> Would it make sense to have the PMPI wrappers defined as a `Component`, and do the registration up-front in an initialization function? That would better guarantee uniformity across nodes.

Yes. I was trying to dodge that initially as then it splits up the template regions although it is more appropriate.

> In terms of which MPI routines to log, the variants of send and recv, the collectives, and `MPI_Wait*` seem like the most important to pick up.

I'll add coverage for *Recv, *Send, and Wait*.

lifflander:
@PhilMiller @pnstickne I want to get a version of this on the release branch for our Vortex debugging so we can track the MPI operations (size, etc.) as we try to debug the very slow MPI operations that are causing bad performance with colors in EMPIRE.

If we do hoist this into a component, I would suggest that we have a light version before we do that that we can use pre-componentization on the release branch.

Also, since RDMA operations seem to be problematic, it would be very helpful if the MPI RDMA interfaces were included in the first pass: MPI_Get, MPI_Put, MPI_Fetch_and_op, MPI_Accumulate, MPI_Win_create, MPI_Rget, MPI_Rput, ... (partial list of them)

lifflander:


```diff

Clones removed
==============
+ src/vt/trace/trace_user_event.cc  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5778239&bid=18917384)

lifflander:


```diff

Clones removed
==============
+ src/vt/trace/trace_user_event.cc  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5778239&bid=18917384)

src/vt/runtime/runtime.h
```diff
@@ -51,9 +51,11 @@
 #include "vt/worker/worker_headers.h"
 #include "vt/configs/arguments/args.h"
 
+// Optional components
 #if backend_check_enabled(trace_enabled)
 #include "vt/trace/trace.h"
 #endif
+#include "vt/pmpi/pmpi_component.h"
```
pnstickne, line 58: `#if` guard inside of header itself, over entirety. Feels.. okay? Should be made consistent in either case.

lifflander:


```diff

Clones removed
==============
+ src/vt/trace/trace_user_event.cc  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5778239&bid=18917384)

lifflander:


```diff

Clones removed
==============
+ src/vt/trace/trace_user_event.cc  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5778239&bid=18917384)

src/vt/pmpi/generate_mpi_wrappers.pl
```diff
@@ -77,47 +88,183 @@ sub should_guard_call {
...
+
+sub register_event {
+    my ($name) = @_;
+    say "  event_${name} = theTrace()->registerUserEventColl(\"_${name}\");";
+}
+
+sub log_event {
+    my ($name) = @_;
+    say "#if vt_check_enabled(trace_enabled)";
+    say "  vt::trace::TraceScopedEvent scopedEvent{";
```
lifflander, line 104: I think this is fine, but might this induce too much overhead when trace is compile-time enabled, but runtime disabled?

src/vt/pmpi/generate_mpi_wrappers.pl
```diff
@@ -77,47 +88,183 @@ sub should_guard_call {
     } @no_guard_patterns;
 }
 
+sub declare_event {
+    my ($name) = @_;
+    say "  vt::trace::UserEventIDType event_${name} = vt::trace::no_user_event_id;";
```
lifflander, line 93: This generates a lot of global statics. I'm worried about overflowing Projections user event ID space. @PhilMiller Do you know if it's limited to `sizeof(uint8_t)`?

src/vt/pmpi/generate_mpi_wrappers.pl
```diff
@@ -77,47 +88,183 @@ sub should_guard_call {
...
+#   registerEvent? => subref,
+#   beforeCall? => subref
+# }
+my @call_actions = (
+    {
+        include => qr/.*/,
+        # MPI_Init/Finalize/Abort - outside of access
+        # MPI_Comm_rank - trivial
+        # MPI_Get_count - called by VT alot 'outside' of event loop
+        exclude => qr/MPI_Init|MPI_Finalize|MPI_Abort|MPI_Comm_rank|MPI_Get_count/,
```
lifflander, line 124: What about the trivial call `MPI_Comm_size`; does that need to excluded here also? Also, we need to exclude `MPI_Get_address(x)`, which almost directly translate to `&x`. We should probably add that to the higher level exclude list also.

src/vt/pmpi/pmpi_component.h
```diff
@@ -0,0 +1,102 @@
...
+ * \brief Component to support PMPI operations, as relevant.
+ *
+ * This provides an external entry point to some PMPI operations and
+ * and infrastructure lifetimes such as registrations of events.
+ */
+struct PMPIComponent : runtime::component::Component<PMPIComponent> {
+
+  using ArgType = vt::arguments::ArgConfig;
+
+  PMPIComponent() {
```
lifflander, line 69: `= default`?

src/vt/pmpi/pmpi_component.h
```diff
@@ -0,0 +1,102 @@
...
+
+  void startup() override {
+    registerEventHandlers();
+  }
+
+  /**
+   * \internal
+   * \brief Return true iff a PMPI call should be logged in the current context.
+   */
+  static bool shouldLogCall() {
```
lifflander, line 82: I understand why this is `static` right now, but that should probably change once we have the argument configuration as an instance. Also, should this be enabled/disabled by the user? I guess if the trace specification file disables tracing this will implicitly disable this also---so maybe this is fine.

src/vt/configs/arguments/args.cc
```diff
@@ -276,8 +277,10 @@ void addTraceArgs(CLI::App& app) {
    * Flags to control tracing output
    */
   auto trace     = "Enable tracing (must be compiled with trace_enabled)";
-  auto trace_mpi = "Enable tracing of MPI calls (must be compiled with "
-    "trace_enabled)";
+  auto trace_mpi = "Enable tracing of select internal MPI calls"
```
lifflander, line 280: I guess this is fine. I would like `--vt_trace_mpi_external`, but that would require the other to change to `--vt_trace_mpi_internal`, which would be a breaking change

src/vt/pmpi/generate_mpi_wrappers.pl
lifflander: Actually, @pnstickne if you tested this and they showed up, then it's probably fine.

lifflander: I don't think these are limited after all

src/vt/pmpi/generate_mpi_wrappers.pl
pnstickne: It'll run through the (simple) ctor and short-circuit in the finalizer.

src/vt/pmpi/generate_mpi_wrappers.pl
pnstickne: Updated.

src/vt/configs/arguments/args.cc
pnstickne: Names aren't great :|

src/vt/pmpi/pmpi_component.h
pnstickne: It will follow the current trace event settings when writing (or trying to write), including the different phases.

The startup code can shim the non-static -> static settings through cleaner when the change come. Actually, I guess I can prep that a bit..

src/vt/configs/arguments/args.cc
PhilMiller: Do we need to worry about that breakage? It affects some non-production run scripts, at worst

src/vt/pmpi/generate_mpi_wrappers.pl
```diff
@@ -14,6 +14,20 @@
     die "Incorrect usage: missing parameters";
 }
 
+# MPI definitions that should always be ignored.
```
PhilMiller, line 17: I just noticed - this script needs a license header inserted.

src/vt/pmpi/generate_mpi_wrappers.pl
```diff
@@ -51,23 +65,21 @@ sub extract_defs {
...
-# or are otherwise excluded due to minimal benefit, compatibility issues, etc.
+# MPI calls that can be 'safely' used without being guarded.
 # - Common idempotent 'get' calls
-# - MPI_Aint_(add|diff) - in OpenMPI 1.10 as macros
-# - MPI_File_(iwrite|iread)_(all|at_all) - missing in OpenMPI 1.10
 my @no_guard_patterns = qw(
     MPI_Comm_get_.*
     MPI_Comm_rank
     MPI_Comm_size
     MPI_Get.*(?<!_accumulate)
```
PhilMiller, line 82: A bare `MPI_Get` does need to be guarded.

I'd suggest amending this pattern to starting with `MPI_Get_`, and then match anything after that other than `accumulate`, as you do now.

src/vt/pmpi/generate_mpi_wrappers.pl
PhilMiller: This is kinda micro-optimization, but for code-size and branch prediction localization, would it make sense to have a composed class `MPIEvent` that contains this logic, and owns a `TraceScopedEvent`?

src/vt/pmpi/pmpi_component.h
pnstickne: And done.

src/vt/configs/arguments/args.cc
pnstickne: Proposal: remove the original internal internal MPI tracing and co-opt the current flag name (unless such is currently used/useful). It only traces two (?) call-sites which are [probably] covered by the larger trace constructs.

src/vt/pmpi/generate_mpi_wrappers.pl
pnstickne: Inserting .. I don't suppose there is a way to get away with

    // See DARMA-VT.license.txt

(or perhaps just remove the file name from the text?)

src/vt/pmpi/generate_mpi_wrappers.pl
```diff
@@ -14,6 +60,20 @@
     die "Incorrect usage: missing parameters";
 }
 
+# MPI definitions that should always be ignored.
+# The most reasonable reason for being in this list is that they do
+# not exist on all platforms or are inconsitently implemented as macros, etc.
```
PhilMiller, line 65: Typo 'inconsistently'

src/vt/pmpi/generate_mpi_wrappers.pl
pnstickne: License added.

IIRC J. has a script to update the licenses in C++ files, maybe adapt to also support pl and check in to a VT-utils dir?

src/vt/pmpi/generate_mpi_wrappers.pl
pnstickne: This pattern should do that.

src/vt/pmpi/generate_mpi_wrappers.pl
```diff
@@ -77,47 +89,183 @@ sub should_guard_call {
...
+#   registerEvent? => subref,
+#   beforeCall? => subref
+# }
+my @call_actions = (
+    {
+        include => qr/.*/,
+        # MPI_Init/Finalize/Abort - outside of access.
+        # MPI_Comm_rank/Comm_size - trivial infrastructure.
+        # MPI_Get_count - called by VT alot 'outside' of event loop.
+        exclude => qr/MPI_Init|MPI_Finalize|MPI_Abort|MPI_Comm_rank|MPI_Comm_size|MPI_Get_count/,
```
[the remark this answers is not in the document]
pnstickne: Probably. While still declarative, the rules are spread out.

src/vt/pmpi/generate_mpi_wrappers.pl
pnstickne: Oh, placement of `_`..

src/vt/pmpi/generate_mpi_wrappers.pl
pnstickne: Fixed.

src/vt/pmpi/generate_mpi_wrappers.pl
pnstickne: Wouldn't that wrapper will still run through TraceScopedEvent RAII? The TSE already contains short-circuits in the ctor (to avoid reading time) and in the finalizer (to prevent writing any events).

If a change is requested, I'd remove the RAII construct here entirely, although seems micro. Not sure the call count to these functions would be sufficient to see benefit.

src/vt/pmpi/generate_mpi_wrappers.pl
pnstickne: Numbers show it's good to 'at least' 16 bits :D

pnstickne: Not reasonably limited, no :)

src/vt/pmpi/pmpi_component.h
lifflander: Ok, looks good

src/vt/runtime/runtime.h
lifflander: I think this is fine... a little ugly but shouldn't cause any trouble.

src/vt/pmpi/generate_mpi_wrappers.pl
lifflander: Ok, I doubt this will be a substantial overhead compared on other parts.

src/vt/configs/arguments/args.cc
lifflander: The flag `--vt_trace_mpi` is currently used in EMPIRE a lot to enable the MPI events while VT is running along with some notes that are generated about the sizes. I don't think we should remove that unless we are directly replacing that functionality.

src/vt/pmpi/pmpi_component.h
pnstickne: No semantic diff here as default'ed constructor rules not in play.

src/vt/configs/arguments/args.cc
pnstickne: So where are we on the naming..?
Leave or change or add alternative form to "internal" MPI too?
    --vt_trace_mpi_external
    --vt_trace_mpi,--vt_trace_mpi_internal

lifflander: Let's change the current `--vt_trace_mpi` to `--vt_trace_mpi_internal`

pnstickne: Updated to use options. Examples:

     --vt_trace_mpi internal
     --vt_trace_mpi external
     --vt_trace_mpi internal external

PR #911, opened 2020-07-07
908 add cuda to ci serialize

pnstickne:
Fixes the upfront NVCC / CUDA 10.1 template issues. Compilation still fails much later where it looks like the checkpoint is not picking up on a serialization method.

### DIFF
--- a/src/vt/messaging/active.impl.h
+++ b/src/vt/messaging/active.impl.h
@@ -55,7 +55,7 @@
 
 namespace vt { namespace messaging {
 
-constexpr ByteType msgsize_not_specified = -1;
+constexpr ByteType msgsize_not_specified = static_cast<ByteType>(-1);
 constexpr NodeType broadcast_dest = uninitialized_destination;
 
 template <typename MsgPtrT>
--- a/src/vt/messaging/message/message_serialize.h
+++ b/src/vt/messaging/message/message_serialize.h
@@ -53,10 +53,37 @@
 
 namespace vt { namespace messaging {
 
+// C++17 port. ref. https://en.cppreference.com/w/cpp/types/conjunction
+template<class...> struct cxx14_conjunction : std::true_type { };
+template<class B1> struct cxx14_conjunction<B1> : B1 { };
+template<class B1, class... Bn>
+struct cxx14_conjunction<B1, Bn...>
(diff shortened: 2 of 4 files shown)


src/vt/messaging/message/message_serialize.h
```diff
@@ -53,11 +53,37 @@
...
 // causing templated instantiation (which will fail).
 #ifndef vt_quirked_serialize_method_detection
 
+using SerializeSizerType = ::checkpoint::Sizer;
+
+template <class T>
+struct has_any_serialize_member_t
+{
+  template <typename U>
+  static auto test(U* u) -> decltype(
```
pnstickne, line 80: NVCC is 'okay' with decltype used like this, even when the member is deleted.

src/vt/messaging/message/message_serialize.h
```diff
@@ -77,7 +104,14 @@ struct has_own_serialize_member_t<U,
 template <typename T>
 static constexpr auto const has_own_serialize =
   ::checkpoint::SerializableTraits<T>::has_serialize_noninstrusive
-  or has_own_serialize_member_t<T>::value;
+  // NVCC (CUDA 10.1) does use lazy evaluation with 'and' substituted here.
+  // The check for has_any.. is also required for NVCC which will
+  // fail compilation at the decltype usage in has_own.., although
+  // correctly not-substitute in has_any..
+  or cxx14_conjunction<
+    has_any_serialize_member_t<T>,
```
pnstickne, line 112: `has_any.. and has_own..` still fails in NVCC even though it should short-circuit on `has_any..` failing.

src/vt/messaging/active.impl.h
```diff
@@ -55,7 +55,7 @@
 
 namespace vt { namespace messaging {
 
-constexpr ByteType msgsize_not_specified = -1;
+constexpr ByteType msgsize_not_specified = static_cast<ByteType>(-1);
```
pnstickne, line 58: Add note about signage change?

src/vt/messaging/message/message_serialize.h
PhilMiller: Try using `&&` and `||` instead of the keywords.

pnstickne:
Broke GitHub PR somehow 🤷. Re-created as https://github.com/DARMA-tasking/vt/pull/919.

PR #915, opened 2020-07-08
914 Fix stacktrace symbols when dladdr fails

### LINKED ISSUE #914: Fix stack traces with gcc
**What Needs to be Done?**

Stack traces do not expose the line just the address when using gcc without "-rdynamic". We can work around this. Here is a good example of what we should implement:

https://web.stanford.edu/class/archive/cs/cs106b/cs106b.1176/lectures/27-Inheritance/code/Inheritance/lib/StanfordCPPLib/system/call_stack_gcc.cpp

lifflander:
Fixes #914

### DIFF
--- a/cmake-modules/CheckLinkerFlag.cmake
+++ b/cmake-modules/CheckLinkerFlag.cmake
@@ -7,8 +7,6 @@ endif()
 
 # This is written because cmake does not support `check_linker_flag` until 3.18
 function(check_linker_flag flag out_var)
-  set(OLD_CMAKE_REQUIRED_FLAGS "${CMAKE_REQUIRED_FLAGS}")
-  set(CMAKE_REQUIRED_FLAGS "${CMAKE_REQUIRED_FLAGS} ${flag}")
+  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} ${flag}")
   check_cxx_compiler_flag("" ${out_var})
-  set(CMAKE_REQUIRED_FLAGS ${OLD_CMAKE_REQUIRED_FLAGS})
 endfunction()
--- a/cmake-modules/CheckLinkerFlag.cmake
+++ b/cmake-modules/CheckLinkerFlag.cmake
@@ -1,0 +1,14 @@
+include(CheckCXXCompilerFlag)
+
+if(POLICY CMP0056)
+  # Honor link flags in try_compile() source-file signature.
+  cmake_policy(SET CMP0056 NEW)
+endif()
+
+# This is written because cmake does not support `check_linker_flag` until 3.18
+function(check_linker_flag flag out_var)
+  set(OLD_CMAKE_REQUIRED_FLAGS "${CMAKE_REQUIRED_FLAGS}")
+  set(CMAKE_REQUIRED_FLAGS "${CMAKE_REQUIRED_FLAGS} ${flag}")
+  check_cxx_compiler_flag("" ${out_var})
+  set(CMAKE_REQUIRED_FLAGS ${OLD_CMAKE_REQUIRED_FLAGS})
+endfunction()
--- a/cmake/set_linker_flags.cmake
+++ b/cmake/set_linker_flags.cmake
@@ -1,0 +1,11 @@
+include(CheckLinkerFlag)
+
+check_linker_flag(-rdynamic LINKER_SUPPORTS_RDYNAMIC)
+
+if (LINKER_SUPPORTS_RDYNAMIC)
+  if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.13)
+    target_link_options(${VIRTUAL_TRANSPORT_LIBRARY} INTERFACE -rdynamic)
+  else()
+    target_link_libraries(${VIRTUAL_TRANSPORT_LIBRARY} INTERFACE -rdynamic)
+  endif()
+endif()
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -268,6 +268,7 @@ if(CODE_COVERAGE AND CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
   endif()
 endif()
 
+include(../cmake/set_linker_flags.cmake)
 
 install(
   TARGETS                   ${VIRTUAL_TRANSPORT_LIBRARY}


lifflander:


```diff

Issues
======
- Added 1
           

Clones added
============
- src/vt/configs/error/stack_out.cc  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5823304&bid=19063243)

src/vt/configs/error/stack_out.cc
```diff
@@ -95,16 +97,76 @@ DumpStackType dumpStack(int skip) {
...
+            i, std::get<0>(t), std::get<1>(t), std::get<2>(t), std::get<3>(t)
+          );
+          found_sym = true;
+
+          pclose(f);
+        }
+      }
+# endif
+
+      if (not found_sym) {
```
lifflander, line 158: Codacy found an issue: [Condition '!found_sym' is always true](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5823304)

lifflander:


```diff

Issues
======
- Added 1
           

Clones added
============
- src/vt/configs/error/stack_out.cc  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5823304&bid=19063243)

src/vt/configs/error/stack_out.cc
lifflander, line 158: Codacy found an issue: [Condition '!found_sym' is always true](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5823304)

lifflander:


```diff

Issues
======
- Added 1
           

Clones added
============
- src/vt/configs/error/stack_out.cc  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5823304&bid=19063243)

src/vt/configs/error/stack_out.cc
```diff
@@ -95,16 +97,79 @@ DumpStackType dumpStack(int skip) {
...
+
+          pclose(f);
+        }
+      }
+#else
+      // use variable to avoid warnings when define is false
+      (void)found_sym;
+#endif
+
+      if (not found_sym) {
```
lifflander, line 161: Codacy found an issue: [Condition '!found_sym' is always true](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5823304)

PhilMiller:
Before going down this path, why not try out [libbacktrace](https://github.com/gcc-mirror/gcc/tree/master/libbacktrace)? It's part of gcc, but it's BSD licensed

lifflander:
> Before going down this path, why not try out [libbacktrace](https://github.com/gcc-mirror/gcc/tree/master/libbacktrace)? It's part of gcc, but it's BSD licensed

I considered that, but it's another dependency (which, by the way, doesn't have cmake). Our current stack traces work correctly except for extracting the symbols for some gcc versions. I remedy that by calling addr2line or atos depending on the platform. I don't see a strong advantage to using an external library compared to the existing simpler solution.

PhilMiller:
In a process failure setting, especially, calling `popen` seems super sketchy. That's a `fork()` and `exec()`, per frame, when we may have already blown up memory.

Given the license, libbacktrace could be vendored into our source tree as a stand-alone piece.

There's also [boost stacktrace](https://www.boost.org/doc/libs/1_73_0/doc/html/stacktrace.html), which again, could be imported into our tree if we didn't want an external dependency.

lifflander:
> In a process failure setting, especially, calling `popen` seems super sketchy. That's a `fork()` and `exec()`, per frame, when we may have already blown up memory.
> 
> Given the license, libbacktrace could be vendored into our source tree as a stand-alone piece.
> 
> There's also [boost stacktrace](https://www.boost.org/doc/libs/1_73_0/doc/html/stacktrace.html), which again, could be imported into our tree if we didn't want an external dependency.

Yes, I'm familiar. There's a proposal to put stack traces in the standard that follows the boost implementation:

http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p0881r5.html

PhilMiller:
Plus, with an external binary, there's no guarantee that its presence in the build environment means it's available in the execution environment.

lifflander:
> Plus, with an external binary, there's no guarantee that its presence in the build environment means it's available in the execution environment.

Another option would be to put `addr2line` in lib. It's a simple utility with about 400 lines of code. Do you think that libbacktrace compiles everywhere we need it?

PhilMiller:
addr2line is GPL licensed as part of GNU binutils, and depends on libbfd, which is also part of the GPL licensed GNU binutils

PhilMiller:
> Do you think that libbacktrace compiles everywhere we need it?

'Everywhere we need it' is relatively modern Linux and macOS? It has ELF and Mach-O support, so we should be good.

PhilMiller:
For a separate concern with anything using `popen` of an external binary, how do MPI,  Kokkos (including OpenMP usage as well), and CUDA feel about `fork()` calls?

lifflander:
Here is a project that adds cmake support for libbacktrace with an older version. We should use it's cmake with the latest libbacktrace code:
https://github.com/ErwanLegrand/libbacktrace

There is also a PR open in the base libbacktrace for adding cmake support, but it is incomplete and it refers to this other fork:

https://github.com/ianlancetaylor/libbacktrace/pull/19

lifflander:
The boost stacktrace library relies on a lot of boostisms and macros and might be hard to extract. It also doesn't have cmake:

https://github.com/boostorg/stacktrace

lifflander:
The other thing to note is that most these libraries depend on a low-level system function: `_Unwind_Backtrace` (and `_Unwind_GetIP` in some cases). Boost stack trace is really just a very pretty wrapper.

lifflander:
Rust supports backtraces and has recently moved away from libbacktrace due to several portability and maintainability issues outlined here: https://github.com/rust-lang/rust/pull/73441

Their new solution is rolled from the ground up and it written in rust (not super helpful): https://github.com/gimli-rs

lifflander:
@PhilMiller Actually, I might have just found a really good alternative with cmake support: https://github.com/bombela/backward-cpp

PhilMiller:
> Here is a project that adds cmake support for libbacktrace with an older version. We should use it's cmake with the latest libbacktrace code:
> https://github.com/ErwanLegrand/libbacktrace
> 
> There is also a PR open in the base libbacktrace for adding cmake support, but it is incomplete and it refers to this other fork:
> 
> [ianlancetaylor/libbacktrace#19](https://github.com/ianlancetaylor/libbacktrace/pull/19)
> 
> **edit: actually someone tried to use it with the newer libbacktrace code and was more trouble than it was worth**

Ian Lance Taylor is the upstream maintainer of libbacktrace. I suspect that if you gave that PR a nudge, possibly with suitable testing report, that it would move ahead.

lifflander:
> > Here is a project that adds cmake support for libbacktrace with an older version. We should use it's cmake with the latest libbacktrace code:
> > https://github.com/ErwanLegrand/libbacktrace
> > There is also a PR open in the base libbacktrace for adding cmake support, but it is incomplete and it refers to this other fork:
> > [ianlancetaylor/libbacktrace#19](https://github.com/ianlancetaylor/libbacktrace/pull/19)
> > **edit: actually someone tried to use it with the newer libbacktrace code and was more trouble than it was worth**
> 
> Ian Lance Taylor is the upstream maintainer of libbacktrace. I suspect that if you gave that PR a nudge, possibly with suitable testing report, that it would move ahead.

I'm now thinking that https://github.com/bombela/backward-cpp might be the best library to embed. It's more modern, handles reading dwarf, has wide support across many platforms, and an MIT license.

PhilMiller:
Backward depends on either libbfd or libdw from elfutils. The former is GPL, the latter is dual GPL/LGPL. Either way, it's doing the pretty-printing, not the actual DWARF-reading.

lifflander:
> Backward depends on either libbfd or libdw from elfutils. The former is GPL, the latter is dual GPL/LGPL. Either way, it's doing the pretty-printing, not the actual DWARF-reading.

We don't need to include those in the source, right? It makes the call to those libraries to read the symbols (that's what I meant): `load_object_with_dwarf`.

I'm thinking we embed *backward*; we will only need those libs for the pretty backtrace.
> You will need to install some dependencies to get the ultimate stack trace. Two libraries are currently supported, the only difference is which one is the easiest for you to install, so pick your poison:

PhilMiller:
Re `-rdynamic`: https://stackoverflow.com/a/8624223/90002 See point number 4 at the bottom about static linking

PhilMiller:
How about libunwind:
https://www.nongnu.org/libunwind/man/libunwind(3).html
https://github.com/libunwind/libunwind
MIT licensed, seemingly standalone

PhilMiller:
> How about libunwind:
> https://www.nongnu.org/libunwind/man/libunwind(3).html
> https://github.com/libunwind/libunwind
> MIT licensed, seemingly standalone

Looks like it will similarly be limited in symbol naming by lack of `.dynsym` as the general `backtrace()`

lifflander:
@PhilMiller I have `-rdynamic` working now and the stack traces are printing correctly on gcc.

PhilMiller:
Even with fully static builds? That's great news.

cmake-modules/CheckLinkerFlag.cmake
```diff
@@ -0,0 +1,14 @@
...
+if(POLICY CMP0056)
+  # Honor link flags in try_compile() source-file signature.
+  cmake_policy(SET CMP0056 NEW)
+endif()
+
+# This is written because cmake does not support `check_linker_flag` until 3.18
+function(check_linker_flag flag out_var)
+  set(OLD_CMAKE_REQUIRED_FLAGS "${CMAKE_REQUIRED_FLAGS}")
+  set(CMAKE_REQUIRED_FLAGS "${CMAKE_REQUIRED_FLAGS} ${flag}")
+  check_cxx_compiler_flag("" ${out_var})
```
PhilMiller, line 12: Is there not a way to pass the `CMAKE_REQUIRED_FLAGS` as an argument to this, instead of saving and restoring it?

PhilMiller:
I'm  +1 on passing the `-rdynamic` flag as described, but can't at all knowledgeably comment on the CMake aspects of how you're doing this. Probably good to get someone who knows it better to look at that.

lifflander:
@nlslatt Can you look at the cmake here?

cmake-modules/CheckLinkerFlag.cmake
nlslatt: Can you help me understand why the code is written this way?  I think the usual way to call `check_cxx_compiler_flag(<flag> <var>)` would be `check_cxx_compiler_flag(${flag} ${out_var})` without modifying any other variables.  Do you need to do something different to get `${flag}` to be interpreted as a link flag instead?

lifflander: I actually took this from: https://reviews.llvm.org/D39431, but I might have gotten the policy wrong. It should be the OLD one I think.

lifflander: It doesn't see correct to set `set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} ${flag}")` because they it's permanently set when it might not work. Is that correct?

nlslatt: The summary says that the scope of the CMAKE_EXE_LINKER_FLAGS change is limited to the function and that it will automatically revert.  I don't know if that's true.

PR #919, opened 2020-07-13
908 - fix NVCC issues (mostly around serialization)

pnstickne:
Something broke with https://github.com/DARMA-tasking/vt/pull/911 and it stopped tracking new commits. This might have been related to the subsequent GitHub outage.. created a new PR in any case.

---

Fixes several NVCC issues to get compilation to work in 10.1 and 11 NVCC/Cuda compilers.

Build has outstanding issue of supplying both "-std=C++1y" and "std=C++14" options at the same time. This raises warnings in NVCC, although it should be operating in C++14 mode.

Fixes: #908

### DIFF
--- a/ci/build_cpp.sh
+++ b/ci/build_cpp.sh
@@ -5,6 +5,10 @@ set -ex
 source_dir=${1}
 build_dir=${2}
 
+# Dependency versions, when fetched via git.
+detector_rev=master
+checkpoint_rev=develop
+
 if test "${VT_DOXYGEN_ENABLED:-0}" -eq 1
 then
     token=${3}
@@ -32,7 +36,7 @@ if test -d "${source_dir}/lib/detector"
 then
     echo "Detector already in lib... not downloading, building, and installing"
 else
-    git clone -b master --depth 1 https://github.com/DARMA-tasking/detector.git
+    git clone -b "${detector_rev}" --depth 1 https://github.com/DARMA-tasking/detector.git
     export DETECTOR=$PWD/detector
     export DETECTOR_BUILD=${build_dir}/detector
     mkdir -p "$DETECTOR_BUILD"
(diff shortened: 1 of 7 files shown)


lifflander:
I've restarted a bunch of the tests as the failures are spurious (GitHub Actions were down).

ci/build_cpp.sh
```diff
@@ -5,6 +5,10 @@ set -ex
 source_dir=${1}
 build_dir=${2}
 
+# Dependency versions, when fetched via git.
+detector_rev=master
+checkpoint_rev=vt908-nvcc
```
pnstickne, line 10: Reset back to develop once changes in checkpoint (before this merge).

src/vt/messaging/message/message_serialize.h
```diff
@@ -77,7 +104,14 @@ struct has_own_serialize_member_t<U,
 template <typename T>
 static constexpr auto const has_own_serialize =
   ::checkpoint::SerializableTraits<T>::has_serialize_noninstrusive
-  or has_own_serialize_member_t<T>::value;
+  // NVCC (CUDA 10.1) does use lazy evaluation with 'and' substituted here.
+  // The check for has_any.. is also required for NVCC which will
+  // fail compilation at the decltype usage in has_own.., although
+  // correctly not-substitute in has_any..
+  or cxx14_conjunction<
```
pnstickne, line 111: TODO: try with `&&` instead of Conjunction. `and` fails, and either operator should have the same effect.

src/vt/rdmahandle/manager.impl.h
```diff
@@ -49,10 +49,14 @@
 #include "vt/rdmahandle/manager.h"
 #include "vt/rdmahandle/sub_handle.h"
 
+#include <type_traits>
+
 namespace vt { namespace rdma {
 namespace impl {
 
 struct HandleData {
+  using isByteCopyable = std::true_type;
```
pnstickne, line 58: Should be possible to remove this now..

lifflander: Does this only affect this message (maybe due to the union in `HandleKey`)? I'm confused to how far reaching this. Will the bug trigger on other `ReduceTMsg<MsgT>`

lifflander: I see another one where you explicitly added the prohibit macro. So maybe this more far-reaching. If this is only on nvcc though, it might be worth moving forward if you can't nail it down.

pnstickne: Forward moved!

PhilMiller:
I think this is good, but I'll hold off on formally marking approval in case there are other concerns

lifflander:
I just rebased this and am waiting to tests to pass

lifflander:
Done: 103 mins for 10.1 and 76 mins for 11.0

Issue #882, opened 2020-06-21
Implement diagnostics for VT


lifflander:
**What Needs to be Done?**

Now that #764 is complete and merged via #783, we should implement the component base classes with a universal interface for debugging and outputting general stats when the program ends for debugging and other diagnostic use cases. Also, `Bufferable` might be useful to make buffering on a condition a universal interface that safely saves messages with an epoch for later use.

The plan for `Diagnostic` is to have an interface whereby components can output information during exigent conditions for debugging. This would include what the termination detector does for printing epoch graphs when a hang occurs. Outputting outstanding messages that never were delivered would be useful.

Also, at the end of the program it would be nice to have some statistics on number of messages sent/received, memory usage high-water mark, epochs processed (partly done with the runtime now), average size of message send, time spent serializing, time outside vs. inside VT, VT overhead in non-user time, idle time, number of phases, persistence of phases, quality/frequency of LB, time spent in LBs, time spent in migrations, average/min/max grain size, etc.

PR #923, opened 2020-07-13
922 Implement static type switch on safe union

### LINKED ISSUE #922: Implement static switch based on type for safe union
**What Needs to be Done?**

Right now to switch on the types in a `vt::SafeUnion` one must write an if statement calling `u.is<T>()`, which is error prone if all cases must be properly covered. Implement a static functor dispatch across the union, which will simplify several use cases.

lifflander:
Fixes #922 

Another spin off of #882, that provides a lot more safety when switching.

### DIFF
--- a/tests/unit/utils/test_safe_union.cc
+++ b/tests/unit/utils/test_safe_union.cc
@@ -228,56 +228,6 @@ TEST_F(TestSafeUnion, test_safe_union_3) {
 
 struct MyTest4 { };
 
-template <typename T>
-struct TestFunctor;
-
-template <>
-struct TestFunctor<float> {
-  int operator()(vt::adt::SafeUnion<float, int, MyTest4>& in) {
-    EXPECT_FALSE(in.template is<int>());
-    EXPECT_FALSE(in.template is<MyTest4>());
-
-    EXPECT_TRUE(in.template is<float>());
-    EXPECT_GT(in.template get<float>(), 29.3);
-    EXPECT_LT(in.template get<float>(), 29.5);
-    return 1;
-  }
-};
-
-
-template <>
-struct TestFunctor<int> {
-  int operator()(vt::adt::SafeUnion<float, int, MyTest4>& in) {
-    EXPECT_FALSE(in.template is<float>());
-    EXPECT_FALSE(in.template is<MyTest4>());
-
-    EXPECT_TRUE(in.template is<int>());
-    EXPECT_EQ(in.template get<int>(), 10);
-    return 2;
-  }
-};
-
-template <>
-struct TestFunctor<MyTest4> {
-  int operator()(vt::adt::SafeUnion<float, int, MyTest4>& in) {
-    // never should happen
-    EXPECT_FALSE(true);
-    return 3;
-  }
-};
-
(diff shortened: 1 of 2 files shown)


tests/unit/utils/test_safe_union.cc
```diff
@@ -226,5 +226,75 @@ TEST_F(TestSafeUnion, test_safe_union_3) {
...
+TEST_F(TestSafeUnion, test_safe_union_switch_4) {
+
+  vt::adt::SafeUnion<float, int, MyTest4> x;
+
+  EXPECT_EQ(x.template switchOn<TestFunctor>(x), 0);
+
+  x.init<int>();
+  x.get<int>() = 10;
+
+  EXPECT_EQ(x.template switchOn<TestFunctor>(x), 2);
```
PhilMiller, line 290: This is a really funny looking interface. If it's a member function of `x`, why should the caller have to pass `x` again? Shouldn't the functor be taking exactly an argument of the selected type?

PhilMiller:
I'm fine with approving this as is, if it's really what you want, but it seems like a very strange way to express it.

lifflander:
> I'm fine with approving this as is, if it's really what you want, but it seems like a very strange way to express it.

@PhilMiller I've designed it so you can pass whatever arguments you want to the type-switched functor. In the actual use case, I have several args to pass. Unless I always pass as the first argument, I don't know how to easily "automatlly" pass the union, if that's what is needed, while keeping the arguments aligned in caller/callee.

If you have a better idea, I'm completely open to it.

PhilMiller:
I think my instinct is to let the functor capture any other arguments it may need, and to pass the active value from the union as the only argument. That eliminates the variadic templating and the argument forwarding entirely.

lifflander:
> I think my instinct is to let the functor capture any other arguments it may need, and to pass the active value from the union as the only argument. That eliminates the variadic templating and the argument forwarding entirely.

The functor can't be a lambda in C++14 since templated lambdas don't exist. Thus, there's no way to "capture". Am I misunderstanding what you are suggesting?

PhilMiller:
[Generic lambdas](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3649.html) is in C++14

PhilMiller:
Or, the caller can actually define a functor, just like you did in the test, and construct it with suitable arguments

lifflander:
I'm still missing how this would work. The functor is templated:
```C++
template <typename T>
struct MyFunctor {
  void operator()() const { }
};
```
The switch instantiates the functor with the right type. The user can't create one, because they don't know what type the union is on, and thus wouldn't be able to construct it. Are you suggesting that we move the template inside to the `operator()` and template it over `T`?

PhilMiller:
Along those lines, the functor's `operator()` can either be a template, or an overload set. The union's switch would just need to call it with the appropriate argument type, and the compiler will do the right thing.

lifflander:


```diff

Clones added
============
- tests/unit/utils/test_safe_union.cc  6
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=5845361&bid=19137955)

PR #933, opened 2020-07-16
876 Add vt_ to backend macros

### LINKED ISSUE #876: Rename macro "backend_check_enabled" left from a long time ago
**What Needs to be Done?**

The compile-time feature check macro is `backend_check_enabled` from the initial implementation of VT, when it was a "backend". This macro bleeds into user code and thus shouldn't be exposed with that name. Rename it to `vt_check_enabled`

lifflander:
Fixes #876

### DIFF
--- a/src/vt/configs/debug/debug_masterconfig.h
+++ b/src/vt/configs/debug/debug_masterconfig.h
@@ -106,6 +106,6 @@ using VTPrintConfig = Configuration<
 #define vt_default_threading openmp
 
 #define backend_no_threading                                             \
-  !backend_check_enabled(openmp) && !backend_check_enabled(stdthread)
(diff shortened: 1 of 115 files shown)


src/vt/configs/debug/debug_masterconfig.h
```diff
@@ -106,6 +106,6 @@ using VTPrintConfig = Configuration<
 #define vt_default_threading openmp
 
 #define backend_no_threading                                             \
-  !backend_check_enabled(openmp) && !backend_check_enabled(stdthread)
+  !vt_check_enabled(openmp) && !vt_check_enabled(stdthread)
```
PhilMiller, line 109: Perhaps parenthesize this, in case it's even embedded in a larger expression?

PR #938, opened 2020-07-17
937 Automatically generate workflows

### LINKED ISSUE #937: Generate CI workflows
**What Needs to be Done?**

Reduce the copy-paste and write a script that generates them

lifflander:
Fixes #937

- Reduce duplicated code by generating all the workflows using a template and an ini file that configures them consistently.
- Now, we just run the script when the configuration changes or we want to add a test


lifflander:
The workflow files should be exactly the same. Some of the default values are now in them, but they have the same default value just set consistently.

In theory, I could write a hook that generates the rest of the tests automatically instead of the user running the script:

```bash
./scripts/generate-workflow.pl ./scripts/workflow-template.yml ./scripts/workflows.ini
```

lifflander:
@PhilMiller @pnstickne What do you guys think about this?

.github/workflows/dockerimage-clang-3.9-ubuntu-mpich.yml
```diff
@@ -1,6 +1,9 @@
+###############################################################################
```
pnstickne, line 1: There we go!

scripts/workflows.ini
```diff
@@ -0,0 +1,101 @@
...
+# compiler = icc-18
+# cache_name = ubuntu-intel-18-cache
+# vt_trace = 1
+# vt_pool = 0
+# output_name = .github/workflows/dockerimage-intel-18-ubuntu-mpich-extended.yml
+# condition = "    if: github.event.label.name == 'runextendedtests'"
+# triggers = "on:\\n  pull_request:\\n    types:\\n      - labeled"
+
+[PR-tests-gcc-5]
+test_configuration = "gcc-5, ubuntu, mpich"
```
pnstickne, line 51: Generating the test configuration name _from_ the other properties would avoid accidental mis-alignment of name and actual settings,

pnstickne: Maybe an approach: before the reading the ini for the templates, re-generate these values in the .ini source file, such that they will also be picked up in a git diff/commit if (or as) they change.

scripts/workflows.ini
```diff
@@ -0,0 +1,101 @@
...
+compiler_type = gnu
+compiler = gcc-5
+output_name = .github/workflows/dockerimage-gcc-5-ubuntu-mpich.yml
+
+[PR-tests-gcc-6]
+test_configuration = "gcc-6, ubuntu, mpich"
```
pnstickne, line 57: Aside: any non-mpich targets?

lifflander: Not currently, we need to add them

PR #945, opened 2020-07-23
912 Remove Travis CI, expand tests and platforms

lifflander:
Fixes #912

- Eradicate Travis CI
- Get code coverage working with GitHub actions
  - New code coverage will now use examples to calculate coverage also (percent seems to have gone up)
  - Turns out that given how code coverage options were being applied, the flag was not getting appended to every cc file.
- Add gcc-9 with ZoltanLB
- Add gcc-10 with no LB and `openmpi`
- Fix compilation bug in ZoltanLB with load modeling missed before due to lack of testing (`CommVolume` change)

PR #947, opened 2020-07-24
946 Fix LB bug in load model and add new LB tests

### LINKED ISSUE #946: Fix bug when running LB with new load models
**Describe the bug**
To reproduce, run `lb_iter` with any load balancer. The phase offset does an invalid lookup in `BaseLB::importProcessorData`.

lifflander:
Fixes #946

### DIFF
--- a/tests/unit/collection/test_lb.extended.cc
+++ b/tests/unit/collection/test_lb.extended.cc
@@ -85,6 +85,9 @@ TEST_P(TestLoadBalancer, test_load_balancer_1) {
 
   vt::arguments::ArgConfig::vt_lb = true;
   vt::arguments::ArgConfig::vt_lb_name = lb_name;
+  if (vt::theContext()->getNode() == 0) {
+    fmt::print("Testing lb {}\n", lb_name);
+  }
 
   vt::theCollective()->barrier();
 
--- a/tests/unit/collection/test_lb.extended.cc
+++ b/tests/unit/collection/test_lb.extended.cc
@@ -123,6 +123,9 @@ INSTANTIATE_TEST_SUITE_P(
     "HierarchicalLB",
     "GossipLB",
     "GreedyLB"
+#   if vt_check_enabled(zoltan)
+    , "ZoltanLB"
+#   endif
   )
 );
 
--- a/tests/unit/collection/test_lb.extended.cc
+++ b/tests/unit/collection/test_lb.extended.cc
@@ -54,7 +54,7 @@
 
 namespace vt { namespace tests { namespace unit {
 
-static constexpr int const num_elms = 1024;
+static constexpr int const num_elms = 64;
 static constexpr int const num_phases = 10;
 
 struct MyCol : vt::Collection<MyCol,vt::Index1D> {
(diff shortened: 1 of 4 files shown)


lifflander:
@PhilMiller This needs to be sequenced after #945 and ZoltanLB should be added since that PR fixes some bugs in that.

Also, the test fails because there is a bug in GossipLB. Not sure if its related to load models or not. A bad access in a map it looks like.

tests/unit/collection/test_lb.extended.cc
```diff
@@ -0,0 +1,131 @@
...
+}
+
+INSTANTIATE_TEST_SUITE_P(
+  LoadBalancerExplode, TestLoadBalancer,
+  ::testing::Values(
+    "RandomLB",
+    "RotateLB",
+    "HierarchicalLB",
+    "GossipLB",
+    "GreedyLB"
```
lifflander, line 125: TODO: add Zoltan once the #945 is merged which fixes a bug with load modeling and Zoltan.

lifflander:
I'm trying to get to the bottom of the GossipLB bug

PhilMiller:
Would it make sense to open a new issue for the Gossip LB issue, and maybe comment it out of this test, if it's the only one that's failing? It kinda looked like Hierarchical may have been failing too, but the output was kinda jumbled-seeming, possibly due to how the test harness operates.

PhilMiller:
Ok, yeah, the output is a little confusing because of the argument modification and re-initialization. When the vt banner prints out, it's describing the choice of balancer for the test that *just ran*

lifflander:
> Ok, yeah, the output is a little confusing because of the argument modification and re-initialization. When the vt banner prints out, it's describing the choice of balancer for the test that _just ran_

Yes, it's a bit screwy the way I'm setting the LB. We need an interface to configure ArgConfig before startup.

PhilMiller:
> Would it make sense to open a new issue for the Gossip LB issue, and maybe comment it out of this test, if it's the only one that's failing?

I checked that GossipLB is the only failing LB, and opened #948 regarding it. I'd like to move the rest of this in, since it does provide a useful advance for everything that isn't broken anyway.

lifflander:
I agree, let's get this merged.

PhilMiller:
Which of the CI configurations actually runs the extended tests?

PhilMiller:
They weren't an issue for compiling or running with clang or gcc, so maybe they should be on for those?

lifflander:
> Which of the CI configurations actually runs the extended tests?

All the configurations except Intel run extended tests currently. nvcc will be added to that list once we get that added to CI.

PhilMiller:
Looks like they get built, but not run.

PhilMiller:
Or I may be confounding that with whether LB is disabled, which would mask out the relevant test.

I'm going to bed - it's late here.

lifflander:
> Or I may be confounding that with whether LB is disabled, which would mask out the relevant test.
> 
> I'm going to bed - it's late here.

Looks like there's a failure with `RandomLB` on gcc-7. Which ones are skipping it? The only one with LB disabled should be gcc-10.

Issue #956, opened 2020-07-29
Unused variable warnings in load models


PhilMiller:
```
In file included from /Users/jliffla/codes/vt/virtual-transport/src/vt/vrt/collection/balance/model/linear_model.cc:45:
/Users/jliffla/codes/vt/virtual-transport/src/vt/vrt/collection/balance/model/linear_model.h:76:7: warning: private field 'past_len_' is not used [-Wunused-private-field]
  int past_len_ = 0;
      ^
1 warning generated.
```
This is a bug - `past_len_` should limit the number of phases considered

```
In file included from /Users/jliffla/codes/vt/virtual-transport/src/vt/vrt/collection/balance/model/norm.cc:46:
/Users/jliffla/codes/vt/virtual-transport/src/vt/vrt/collection/balance/model/norm.h:71:7: warning: private field 'num_subphases_' is not used [-Wunused-private-field]
  int num_subphases_;
      ^
1 warning generated.
```
This variable is no longer used, and can be deleted.

PR #957, opened 2020-07-29
956 Load models: Correct unused variables

PhilMiller:
One bug, one disused variable.

Fixes: #956


lifflander:
Something is wrong with the Mac build. It shouldn't take more than ~30 mins

lifflander:
> * the second commit fixes something that isn't really described in #956
> * is there a unit test necessary for the second commit (since it changes the behavior)?

Good point on the unit test. I will create an issue to write more tests for this...but I'm going to merge it.

PR #961, opened 2020-07-31
960 Time-based scheduler triggers

### LINKED ISSUE #960: Add time-based triggers to scheduler
**What Needs to be Done?**

For diagnostics, it would be useful to trigger an event with some associated time period for recording diagnostic values over time.

lifflander:
Fixes #960 

- Implement new time-based triggers that can be registered with the scheduler.
- Write unit tests to cover them and ensure they are triggered when expected.

### DIFF
--- a/src/vt/scheduler/scheduler.cc
+++ b/src/vt/scheduler/scheduler.cc
@@ -187,11 +187,6 @@ void Scheduler::printMemoryUsage() {
   }
 }
 
-void Scheduler::progressTimeTriggers() {
-  auto const cur_time = timing::Timing::getCurrentTime();
-  time_triggers_.triggerReady(cur_time);
-}
-
 void Scheduler::runProgress(bool msg_only) {
   /*
    * Run through the progress functions `num_iter` times, making forward
@@ -211,8 +206,6 @@ void Scheduler::runProgress(bool msg_only) {
   if (arguments::ArgConfig::vt_print_memory_at_threshold) {
     printMemoryUsage();
   }
-
-  progressTimeTriggers();
 
   // Reset count of processed handlers since the last time progress was invoked
   processed_after_last_progress_ = 0;
@@ -336,16 +329,6 @@ void Scheduler::registerTriggerOnce(
     event_triggers.size() >= event, "Must be large enough to hold this event"
   );
   event_triggers_once[event].push_back(trigger);
-}
-
-int Scheduler::registerTimeTrigger(
-  std::chrono::milliseconds period, TriggerType trigger
-) {
-  return time_triggers_.addTrigger(period, trigger);
-}
-
-void Scheduler::unregisterTimeTrigger(int handle) {
-  time_triggers_.removeTrigger(handle);
 }
 
 }} //end namespace vt::sched
--- a/src/vt/scheduler/scheduler.cc
+++ b/src/vt/scheduler/scheduler.cc
@@ -187,6 +187,30 @@ void Scheduler::printMemoryUsage() {
   }
 }
 
+void Scheduler::progressTimeTriggers() {
+  auto const cur_time = timing::Timing::getCurrentTime();
+
+  // Check if the soonest next trigger is ready
+  if (cur_time < next_trigger_time_ and next_trigger_time_ != 0) {
+    return;
+  }
+
+  next_trigger_time_ = std::numeric_limits<TimeType>::max();
+
+  auto iter = time_triggers_.begin();
+  while (iter != time_triggers_.end()) {
+    bool should_trigger = iter->second.shouldTrigger(cur_time);
+    if (should_trigger) {
+      iter->second.trigger(cur_time);
+    }
+    next_trigger_time_ = std::min(
+      next_trigger_time_,
+      iter->second.getLastTriggerTime() + static_cast<TimeType>(iter->first) / 1000
+    );
+    iter++;
+  }
+}
+
 void Scheduler::runProgress(bool msg_only) {
   /*
    * Run through the progress functions `num_iter` times, making forward
@@ -205,6 +229,10 @@ void Scheduler::runProgress(bool msg_only) {
 
   if (arguments::ArgConfig::vt_print_memory_at_threshold) {
     printMemoryUsage();
+  }
+
+  if (time_triggers_.size() > 0) {
+    progressTimeTriggers();
   }
 
   // Reset count of processed handlers since the last time progress was invoked
@@ -329,6 +357,29 @@ void Scheduler::registerTriggerOnce(
     event_triggers.size() >= event, "Must be large enough to hold this event"
   );
   event_triggers_once[event].push_back(trigger);
+}
+
+int Scheduler::registerTimeTrigger(Milliseconds period, TriggerType trigger) {
+  auto iter = time_triggers_.find(period);
+  if (iter == time_triggers_.end()) {
+    time_triggers_.emplace(
+      std::piecewise_construct,
+      std::forward_as_tuple(period),
+      std::forward_as_tuple(TimeTrigger{period})
+    );
+    iter = time_triggers_.find(period);
+  }
+  return iter->second.addTrigger(trigger);
+}
+
+void Scheduler::unregisterTimeTrigger(Milliseconds period, int handle) {
+  auto iter = time_triggers_.lower_bound(period);
+  if (iter != time_triggers_.end() and iter->first == period) {
+    iter->second.removeTrigger(handle);
+    if (iter->second.numTriggers() == 0) {
+      time_triggers_.erase(iter);
+    }
+  }
 }
 
 }} //end namespace vt::sched
--- a/src/vt/scheduler/scheduler.cc
+++ b/src/vt/scheduler/scheduler.cc
@@ -189,26 +189,7 @@ void Scheduler::printMemoryUsage() {
 
 void Scheduler::progressTimeTriggers() {
   auto const cur_time = timing::Timing::getCurrentTime();
-
-  // Check if the soonest next trigger is ready
-  if (cur_time < next_trigger_time_ and next_trigger_time_ != 0) {
-    return;
-  }
-
-  next_trigger_time_ = std::numeric_limits<TimeType>::max();
-
-  auto iter = time_triggers_.begin();
-  while (iter != time_triggers_.end()) {
-    bool should_trigger = iter->second.shouldTrigger(cur_time);
-    if (should_trigger) {
-      iter->second.trigger(cur_time);
-    }
-    next_trigger_time_ = std::min(
-      next_trigger_time_,
-      iter->second.getLastTriggerTime() + static_cast<TimeType>(iter->first) / 1000
-    );
-    iter++;
-  }
+  time_triggers_.triggerReady(cur_time);
 }
 
 void Scheduler::runProgress(bool msg_only) {
@@ -231,9 +212,7 @@ void Scheduler::runProgress(bool msg_only) {
     printMemoryUsage();
   }
 
-  if (time_triggers_.size() > 0) {
-    progressTimeTriggers();
-  }
+  progressTimeTriggers();
 
   // Reset count of processed handlers since the last time progress was invoked
   processed_after_last_progress_ = 0;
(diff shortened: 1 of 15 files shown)


src/vt/scheduler/scheduler.h
```diff
@@ -97,6 +99,8 @@ struct Scheduler : runtime::component::Component<Scheduler> {
   using TriggerType          = std::function<void()>;
   using TriggerContainerType = std::list<TriggerType>;
   using EventTriggerContType = std::vector<TriggerContainerType>;
+  using Milliseconds         = typename TimeTrigger::Milliseconds;
```
PhilMiller, line 102: Could/should we use `std::chrono::milliseconds` and associated types, rather than defining our own?

PhilMiller: https://en.cppreference.com/w/cpp/chrono/duration Since C++11, so no compiler version issues expected

src/vt/scheduler/scheduler.h
```diff
@@ -172,6 +176,29 @@ struct Scheduler : runtime::component::Component<Scheduler> {
...
+
+  /**
+   * \brief Register a time-based trigger with a specific period
+   *
+   * \param[in] period the period in milliseconds
+   * \param[in] trigger the trigger to execute
+   *
+   * \return the trigger ID (can be used for removal)
+   */
+  int registerTimeTrigger(Milliseconds period, TriggerType trigger);
```
PhilMiller, line 192: I take it a repeating trigger will be handled internally, by the trigger resetting itself?

src/vt/scheduler/scheduler.cc
```diff
@@ -187,6 +187,30 @@ void Scheduler::printMemoryUsage() {
...
+  auto const cur_time = timing::Timing::getCurrentTime();
+
+  // Check if the soonest next trigger is ready
+  if (cur_time < next_trigger_time_ and next_trigger_time_ != 0) {
+    return;
+  }
+
+  next_trigger_time_ = std::numeric_limits<TimeType>::max();
+
+  auto iter = time_triggers_.begin();
```
PhilMiller, line 200: I don't see anything about the control flow here that would need this to not be a `for` loop

PhilMiller:
I feel like the responsibility for determining which triggers to run at what times is very 'mushy' between the the scheduler and the time trigger class. Is there some higher level design rationale here that I'm missing?

src/vt/scheduler/time_trigger.h
```diff
@@ -0,0 +1,147 @@
...
+#include <map>
+
+namespace vt { namespace sched {
+
+/**
+ * \struct TimeTrigger
+ *
+ * \internal \brief A time-based set of triggers with a given period
+ */
+struct TimeTrigger {
```
pnstickne, line 60: Might be worth considering each trigger (one trigger/action) is separate in a priority queue. Repeating / interval triggers then just-re-add on their expiration while singles don't.

src/vt/scheduler/scheduler.h
lifflander: I considered that. But, I went with an `int` to make it more simple. I can change that.

src/vt/scheduler/scheduler.cc
lifflander: Good point, it used to be more complex, but I simplified it

src/vt/scheduler/scheduler.cc
```diff
@@ -187,6 +187,30 @@ void Scheduler::printMemoryUsage() {
...
+  while (iter != time_triggers_.end()) {
+    bool should_trigger = iter->second.shouldTrigger(cur_time);
```
pnstickne, line 202: If can keep ordered this wouldn't have to iterate all trigger sets each time. Depends on expected / possible sizes I suppose.

lifflander:
> I feel like the responsibility for determining which triggers to run at what times is very 'mushy' between the the scheduler and the time trigger class. Is there some higher level design rationale here that I'm missing?

The `shouldTrigger` and `trigger` calls take a current time to avoid calling the timer a bunch of times. The logic for the `next_trigger_time_` is geared toward avoiding checking all timers every progress function call. The `TimeTrigger` just holds triggers for a certain interval. I could create a class that holds all the time intervals to encapsulate it more.

src/vt/scheduler/scheduler.cc
lifflander: Yes, I was thinking the same but wasn't sure if it was worth the cost of keeping it sorted.

src/vt/scheduler/scheduler.h
lifflander: Done

src/vt/scheduler/scheduler.cc
lifflander: Simplified

src/vt/scheduler/time_trigger.h
lifflander: Switched to priority queue

lifflander:
I had to increase the tolerance on the test quite a bit. On some of the CI platforms, for the 10ms test periods, we would get ~14ms trigger period. The 100ms ones were pretty accurate. Might be related to how must dedicated time the containers get.

src/vt/scheduler/time_trigger.h
```diff
@@ -0,0 +1,189 @@
...
+  }
+
+  /**
+   * \brief Trigger any read time-based triggers
+   *
+   * \param[in] cur_time the current time
+   */
+  void triggerReady(TimeType cur_time) {
+    while (not queue_.empty()) {
+      auto iter = removed_.find(queue_.top().getID());
```
lifflander, line 162: I'm not terribly pleased with how deletions are implemented, but this was the best I could do. I can't generally traverse the priority queue easily, so I just delay the removal until the trigger comes up again to the top of the heap.

src/vt/scheduler/time_trigger.h
```diff
@@ -0,0 +1,189 @@
...
+   *
+   * \param[in] period time period to trigger action
+   * \param[in] action action to trigger
+   *
+   * \return id to trigger for removal
+   */
+  int addTrigger(std::chrono::milliseconds period, ActionType action) {
+    auto const cur_id = next_trigger_id_++;
+    TimeTrigger trigger{period, action, cur_id};
+    trigger.runAction();
```
lifflander, line 141: This semantic might be questionable. I decided that if the user submits a trigger on a period I just trigger it right away. I could switch this to waiting the period duration before the first trigger... Opinions?

lifflander:
The problem this doesn't solve is a "synchronized" time trigger, which we need approximately want for diagnostics. The user can sync up before they register the trigger, which probably reasonable enough for that use case.

src/vt/scheduler/time_trigger.h
PhilMiller: Seems pretty reasonable to me.

PhilMiller:
I recognize that if these were a separate component, they should probably still get special treatment that would see them called even in the Scheduler's 'message only' progress function.

lifflander:
@PhilMiller I could easily make it a component. BTW, the message-only progress special case is a hack that exists to inhibit "early termination" during setup when work has yet to be produced. We could probably refactor our way out of that fairly easily now. I'm going to create an issue for that.

src/vt/scheduler/scheduler.cc
```diff
@@ -331,6 +338,16 @@ void Scheduler::registerTriggerOnce(
   event_triggers_once[event].push_back(trigger);
 }
 
+int Scheduler::registerTimeTrigger(
```
pnstickne, line 341: Could use explicit type.

src/vt/scheduler/scheduler.h
```diff
@@ -172,6 +174,28 @@ struct Scheduler : runtime::component::Component<Scheduler> {
     SchedulerEventType const& event, TriggerType trigger
   );
 
+  /**
+   * \brief Make progress on time-based triggers
+   */
+  void progressTimeTriggers();
+
+  /**
+   * \brief Register a time-based trigger with a specific period
```
pnstickne, line 183: Should define/document if this is "period/interval between the start of callbacks" or "period/interval since the end of last callback".

src/vt/scheduler/time_trigger.h
```diff
@@ -0,0 +1,189 @@
...
+   */
+  TimeType getLastTriggerTime() const { return last_trigger_time_; }
+
+  /**
+   * \brief Get the next trigger time
+   *
+   * \return the next time this should be triggered
+   */
+  TimeType nextTriggerTime() const {
+    return (last_trigger_time_*1000. + period_.count())/1000.;
```
pnstickne, line 90: Add the trailing 0.. :|

src/vt/scheduler/time_trigger.h
```diff
@@ -0,0 +1,189 @@
...
+  }
+
+  /**
+   * \brief Get the ID for the trigger
+   *
+   * \return the ID
+   */
+  int getID() const { return id_; }
+
+  friend bool operator<(TimeTrigger const& lhs, TimeTrigger const& rhs) {
```
pnstickne, line 108: Feels swapped.

src/vt/scheduler/time_trigger.h
pnstickne: Usually it's to wait one interval (ubiquitous case is setInterval).

I would probably have the default to wait, and perhaps accept a flag to 'fire immediately as well', if such proves useful.

src/vt/scheduler/time_trigger.h
```diff
@@ -0,0 +1,189 @@
...
+  /**
+   * \brief Add a new trigger
+   *
+   * \param[in] period time period to trigger action
+   * \param[in] action action to trigger
+   *
+   * \return id to trigger for removal
+   */
+  int addTrigger(std::chrono::milliseconds period, ActionType action) {
+    auto const cur_id = next_trigger_id_++;
```
pnstickne, line 139: `int` is actually less typing >_<

src/vt/scheduler/time_trigger.h
pnstickne: The queued item itself could contain the "should remove" state. For a low amount of churn it can just stay in the queue until it hits the trigger time.

src/vt/scheduler/time_trigger.h
```diff
@@ -0,0 +1,189 @@
...
+    while (not queue_.empty()) {
+      auto iter = removed_.find(queue_.top().getID());
+      if (iter != removed_.end()) {
+        queue_.pop();
+        removed_.erase(iter);
+        continue;
+      }
+
+      if (queue_.top().nextTriggerTime() < cur_time) {
+        auto t = queue_.top();
```
pnstickne, line 170: `auto t = queue_.top();` can be moved up, replacing all other uses of top().

src/vt/scheduler/time_trigger.h
pnstickne, line 90: This could also be computed at the end of the last trigger - micro, and it would save a wee bit of math on each loop check vs `queue_.top().nextTriggerTime()`.

src/vt/scheduler/time_trigger.h
pnstickne: Uhg. Nevermind. C++ queue access limits.

src/vt/scheduler/time_trigger.h
```diff
@@ -0,0 +1,189 @@
...
+  void triggerReady(TimeType cur_time) {
+    while (not queue_.empty()) {
+      auto iter = removed_.find(queue_.top().getID());
+      if (iter != removed_.end()) {
+        queue_.pop();
+        removed_.erase(iter);
+        continue;
+      }
+
+      if (queue_.top().nextTriggerTime() < cur_time) {
```
pnstickne, line 169: Could still move the check to only after ensuring the time elapsed, which should result in _most loops_ checking the top element and stopping immediately w/o any other work.

    auto& t = queue_.top();
    if (t.nextTriggerTime() < cur_time) {
        if (removed_.find(t.getID()) // only ever accessed once time comes up

src/vt/scheduler/time_trigger.h
lifflander: Added

src/vt/scheduler/scheduler.h
lifflander: Done

src/vt/scheduler/time_trigger.h
lifflander: Yes, I think it might be undefined if you change how the operator will order the heap while it exists in the heap. That's why I avoided that.

src/vt/scheduler/time_trigger.h
lifflander: I'm going to leave it as is because I think it's easier to read compared to saving the state.

src/vt/scheduler/time_trigger.h
lifflander: Ok, changed

PR #972, opened 2020-08-05
971 Fix component construction

### LINKED ISSUE #971: Fix perfect forwarding for move-only lambdas in component construction
**What Needs to be Done?**

For instance, current components constructor code does not handle passing a unique_ptr to a component constructor.

lifflander:
Fixes #971

### DIFF
--- a/src/vt/runtime/component/component_pack.cc
+++ b/src/vt/runtime/component/component_pack.cc
@@ -61,7 +61,7 @@ void ComponentPack::construct() {
 
     auto iter = construct_components_.find(next);
     if (iter != construct_components_.end()) {
-      iter->second();
+      iter->second->invoke();
     } else {
       vtAbort("Could not find component to construct");
     }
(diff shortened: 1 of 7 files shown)


src/vt/runtime/component/component_pack.cc
```diff
@@ -105,9 +105,15 @@ int ComponentPack::progress() {
 
 std::list<int> ComponentPack::topoSort() {
   std::list<int> order;
+
+  registry::AutoHandlerType max_idx = 0;
+  for (auto&& r : registered_components_) {
+    max_idx = std::max(r, max_idx);
+  }
+
+  auto visited = std::make_unique<bool[]>(max_idx);
```
PhilMiller, line 114: Are the values gotten above coming from a counter somewhere, or are they more arbitrary bit-field or address like things? If it's not from a counter, this array could be unpleasantly large

lifflander: They are generated from a type registry is linearized and compact. However, when I started writing tests that just register and use a single test type it becomes sparse. So dense for all types ever registered (for the common case in VT that will be all components), but sparse when writing tests or just constructing a subset of the system.

lifflander:
@cz4rs This should solve the problems that you were having with unique_ptr.

PR #982, opened 2020-08-12
981 Relax time trigger test again using new percent tolerance

### LINKED ISSUE #981: Fix time trigger again...
**What Needs to be Done?**

Seems to still break. Tolerance of +/-5 is still too low.

lifflander:
Fixes #981


lifflander:
> any reason to use percentage based tolerance only in second test?
> 
> overall, looks good to me

The first test uses a fixed period so a percent isn't needed across them.

Issue #986, opened 2020-08-17
MPI collective operations may interact badly with TD


PhilMiller:
https://github.com/DARMA-tasking/vt/pull/975#discussion_r470863870

> I think this test will only pass as modified by accident. I believe it actually illustrates a latent mistake we made in the async MPI collectives - they should produce on the active epoch when queued up, and consume on that same epoch when actually executed. The reduction to CollectiveAlg::runCollective will take care of things in the simplest cases, but I think it can go wrong when there are multiple concurrent collectives potentially in flight, with several of them queuing and returning.

**Expected behavior**

Invoking a collective op in an epoch and then waiting on that epoch should maybe guarantee that the op has completed. It's not clear whether or not it does, or whether or not that expectation is reasonable.

At a minimum, we need to document the intended usage and behavior and interaction.

PhilMiller:
I don't think there's any concern in the synchronous case, since that will spin the scheduler until the instant operation requested has completed, independent of active epoch.

PhilMiller:
We concluded this can actually fail in practice. Will fix with produce/consume

PR #994, opened 2020-08-18
#986: Run MPI collectives as part of active epoch at initial call

PhilMiller:
Fixes: #986

### DIFF
--- a/src/vt/collective/collective_scope.h
+++ b/src/vt/collective/collective_scope.h
@@ -66,12 +66,13 @@ private:
 
 private:
   struct CollectiveInfo {
-    CollectiveInfo(TagType in_seq, ActionType in_action)
-      : seq_(in_seq), action_(in_action)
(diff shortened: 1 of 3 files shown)


src/vt/collective/collective_scope.h
```diff
@@ -66,12 +66,13 @@ struct ScopeImpl {
 
 private:
   struct CollectiveInfo {
-    CollectiveInfo(TagType in_seq, ActionType in_action)
-      : seq_(in_seq), action_(in_action)
+    CollectiveInfo(TagType in_seq, ActionType in_action, EpochType in_epoch)
```
lifflander, line 69: Do we actually need to store the epoch here? Won't it be on the message that is reduced/broadcast out?

lifflander: Resovled

lifflander:
Wait for CUDA or merge...?

PR #996, opened 2020-08-19
987 Unit test time-based triggers independent of clock

### LINKED ISSUE #987: Unit test time-based triggers independent of clock
**Describe the bug**

The tests for the time-based trigger code are very sensitive to the execution environment, because they reference actual clock time. The code almost entirely can be tested without reference to an actual clock. The one sore point in the design is that `Trigger::runAction()` queries the current time, rather than taking it as an argument.

**To Reproduce**

See #976 for instance

**Expected behavior**

No clock calls should appear inside any of the time trigger code or its tests except `TimeTriggerManager::progress()`, which abstracts its behavior from the scheduler. As a corollary, the unit test should have no need to call the scheduler -- the test can prescribe arbitrary times to get whatever behaviors are desired.

cz4rs:
fixes #987

### DIFF
--- a/src/vt/timetrigger/time_trigger_manager.cc
+++ b/src/vt/timetrigger/time_trigger_manager.cc
@@ -79,7 +79,7 @@ void TimeTriggerManager::triggerReady(TimeType cur_time) {
       }
 
       queue_.pop();
-      t.runAction(timing::Timing::getCurrentTime());
+      t.runAction(cur_time);
       queue_.push(t);
     } else {
       // all other triggers will not be ready if this one isn't
--- a/src/vt/timetrigger/time_trigger_manager.cc
+++ b/src/vt/timetrigger/time_trigger_manager.cc
@@ -48,14 +48,15 @@
 namespace vt { namespace timetrigger {
 
 int TimeTriggerManager::addTrigger(
-  std::chrono::milliseconds period, ActionType action, bool fire_immediately
+  TimeType current_time, std::chrono::milliseconds period,
+  ActionType action, bool fire_immediately
 ) {
   int const cur_id = next_trigger_id_++;
   Trigger trigger{period, action, cur_id};
   if (fire_immediately) {
-    trigger.runAction(timing::Timing::getCurrentTime());
+    trigger.runAction(current_time);
   } else {
-    trigger.setLastTriggerTime(timing::Timing::getCurrentTime());
+    trigger.setLastTriggerTime(current_time);
   }
   queue_.push(trigger);
   return cur_id;
@@ -67,7 +68,7 @@ void TimeTriggerManager::removeTrigger(int id) {
 
 void TimeTriggerManager::triggerReady(TimeType cur_time) {
   while (not queue_.empty()) {
-    if (queue_.top().nextTriggerTime() < cur_time) {
+    if (queue_.top().ready(cur_time)) {
       auto t = queue_.top();
 
       auto iter = removed_.find(t.getID());
--- a/src/vt/timetrigger/time_trigger_manager.cc
+++ b/src/vt/timetrigger/time_trigger_manager.cc
@@ -53,7 +53,7 @@ int TimeTriggerManager::addTrigger(
   int const cur_id = next_trigger_id_++;
   Trigger trigger{period, action, cur_id};
   if (fire_immediately) {
-    trigger.runAction();
+    trigger.runAction(timing::Timing::getCurrentTime());
   } else {
     trigger.setLastTriggerTime(timing::Timing::getCurrentTime());
   }
@@ -78,7 +78,7 @@ void TimeTriggerManager::triggerReady(TimeType cur_time) {
       }
 
       queue_.pop();
-      t.runAction();
+      t.runAction(cur_time);
       queue_.push(t);
     } else {
       // all other triggers will not be ready if this one isn't
--- a/src/vt/timetrigger/time_trigger_manager.cc
+++ b/src/vt/timetrigger/time_trigger_manager.cc
@@ -78,7 +78,7 @@ void TimeTriggerManager::triggerReady(TimeType cur_time) {
       }
 
       queue_.pop();
-      t.runAction(cur_time);
+      t.runAction(timing::Timing::getCurrentTime());
       queue_.push(t);
     } else {
       // all other triggers will not be ready if this one isn't
--- a/tests/unit/timetrigger/test_time_trigger.extended.cc
+++ b/tests/unit/timetrigger/test_time_trigger.extended.cc
@@ -57,7 +57,30 @@ namespace vt { namespace tests { namespace unit {
 
 using TestTimeTrigger = TestParallelHarness;
 
-TEST_F(TestTimeTrigger, test_time_trigger_1) {
+TEST_F(TestTimeTrigger, test_time_trigger) {
+  using namespace std::chrono_literals;
+
+  std::chrono::milliseconds trigger_period = 100ms;
+  int trigger_id = 42;
+  int triggered = 0;
+
+  vt::timetrigger::Trigger trigger{trigger_period, [&](){
+    triggered++;
+  }, trigger_id};
+
+  EXPECT_EQ(trigger.getID(), trigger_id);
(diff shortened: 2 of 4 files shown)


tests/unit/timetrigger/test_time_trigger.extended.cc
```diff
@@ -57,7 +57,30 @@ namespace vt { namespace tests { namespace unit {
...
+  std::chrono::milliseconds trigger_period = 100ms;
+  int trigger_id = 42;
+  int triggered = 0;
+
+  vt::timetrigger::Trigger trigger{trigger_period, [&](){
+    triggered++;
+  }, trigger_id};
+
+  EXPECT_EQ(trigger.getID(), trigger_id);
+  EXPECT_EQ(trigger.getLastTriggerTime(), 0.);
```
pnstickne, line 72: `0.0`

pnstickne, line 72: Recommend: `0.0`.

src/vt/timetrigger/time_trigger_manager.cc
```diff
@@ -78,7 +78,7 @@ void TimeTriggerManager::triggerReady(TimeType cur_time) {
       }
 
       queue_.pop();
-      t.runAction();
+      t.runAction(cur_time);
```
pnstickne, line 81: All the actions will run with the same 'current time', even if some actions are delayed. This feels like it can compress events together over time because next is based on current +.

tests/unit/timetrigger/test_time_trigger.extended.cc
```diff
@@ -57,7 +57,30 @@ namespace vt { namespace tests { namespace unit {
...
+  EXPECT_EQ(trigger.nextTriggerTime(), 0.1);
+  EXPECT_EQ(triggered, 0);
+
+  trigger.runAction(4.);
+
+  EXPECT_EQ(trigger.getLastTriggerTime(), 4.);
+  EXPECT_EQ(trigger.nextTriggerTime(), 4.1);
```
pnstickne, line 79: Feels like there is a potential for rounding errors.. just dodged based on selected values?

tests/unit/timetrigger/test_time_trigger.extended.cc
```diff
@@ -57,7 +57,30 @@ namespace vt { namespace tests { namespace unit {
...
+  vt::timetrigger::Trigger trigger{trigger_period, [&](){
+    triggered++;
+  }, trigger_id};
+
+  EXPECT_EQ(trigger.getID(), trigger_id);
+  EXPECT_EQ(trigger.getLastTriggerTime(), 0.);
+  EXPECT_EQ(trigger.nextTriggerTime(), 0.1);
+  EXPECT_EQ(triggered, 0);
+
+  trigger.runAction(4.);
```
pnstickne, line 76: Using a variable here can lead to more clear meaning:

     auto start_time = 4.0;
     auto duration = 0.1; // or use with chronos..

     EXPECT_EQ(trigger.getLastTriggerTime(), start_time);
     EXPECT_EQ(trigger.nextTriggerTime(), start_time + duration);

tests/unit/timetrigger/test_time_trigger.extended.cc
```diff
@@ -103,7 +126,7 @@ TEST_F(TestTimeTrigger, test_time_trigger_1) {
   EXPECT_EQ(prev_triggered, triggered);
 }
 
-TEST_F(TestTimeTrigger, test_time_trigger_2) {
+TEST_F(TestTimeTrigger, test_time_trigger_manager_2) {
   using namespace std::chrono_literals;
 
   std::chrono::milliseconds trigger_period[3] = {100ms, 10ms, 1000ms};
```
pnstickne, line 132: The code below here is still dependent upon the current time.

tests/unit/timetrigger/test_time_trigger.extended.cc
cz4rs: switched to googletest recommended `EXPECT_DOUBLE_EQ`, according to their docs the tolerance is
> By "almost equal" we mean the values are within 4 ULP's from each other.

tests/unit/timetrigger/test_time_trigger.extended.cc
cz4rs: corrected

cz4rs: corrected

tests/unit/timetrigger/test_time_trigger.extended.cc
cz4rs: corrected

src/vt/timetrigger/time_trigger_manager.cc
cz4rs: I agree, this seems too imprecise - for now I've stuck with just using `timing::Timing::getCurrentTime()`

this means that there is no change in behavior in `TimeTriggerManager`, the invocations of `timing::Timing::getCurrentTime()` were just moved out of `RunAction` (well, the trigger is now run after calling the clock instead of before, but I don't think that's a problem)

tests/unit/timetrigger/test_time_trigger.extended.cc
cz4rs: @PhilMiller can you have a look as well?
on the good side, there are no calls to `getCurrentTime` left in `Trigger` code nor the testing code for it

> The code below here is still dependent upon the current time.

there are calls to `getCurrentTime` left both in `TimeTriggerManager` code and the tests for it 

- `TimeTriggerManager::progress` - considered ok
- `TimeTriggerManager::triggerReady` - discussed above - https://github.com/DARMA-tasking/vt/pull/996#discussion_r473648754 - passing the time from outside seems dangerous here
- I have changed `addTrigger` so it takes current time as an argument as well, but I have mixed feelings about it (using this interface is more complicated now)

calls in test code:
I have tested part of `TimeTriggerManager` without using actual time, but it seems to be difficult to get rid of `getCurrentTime` completely in the test code - unless we want to mock the clock (using static dependency injection?)

and the (problematic) tests I have left unchanged still feel valuable even if coverage-wise removing them would not make a difference now

any ideas how to progress from here?

src/vt/timetrigger/time_trigger_manager.cc
```diff
@@ -78,7 +79,7 @@ void TimeTriggerManager::triggerReady(TimeType cur_time) {
       }
 
       queue_.pop();
-      t.runAction();
+      t.runAction(timing::Timing::getCurrentTime());
```
PhilMiller, line 82: I'm concerned that this is going to continue to make testing flaky, and breaks isolation. I think the risk of 'compression' of events towards each other is fine in comparison

tests/unit/timetrigger/test_time_trigger.extended.cc
```diff
@@ -57,97 +57,89 @@ namespace vt { namespace tests { namespace unit {
...
 
+  // test unregisteration of triggers
   auto prev_triggered = triggered;
-  testTime->removeTrigger(id);
+  trigger_manager->removeTrigger(id);
 
-  sleep_for(110ms);
-  testTime->progress();
-  testTime->progress();
+  sleep_for(trigger_period_ms);
```
PhilMiller, line 137: I opened the initial issue with the intention of getting rid of flaky sleep calls in this test suite. Really, truly, the calls to `getCurrentTime` inside the trigger code need to be removed.

src/vt/timetrigger/time_trigger_manager.cc
cz4rs: switched back to `cur_time` as per comments from Phil

tests/unit/timetrigger/test_time_trigger.extended.cc
cz4rs: ok, 87f52ff removes the last `getCurrentTime` outside of `progress` and allows to clean the testing code as well

'progress()` is not tested explicitly anymore, but since all it does is getting the time and passing it to `triggerReady()` testing the latter should suffice

cz4rs:
@PhilMiller  I think it's good to go - the only test failing is the problematic LB reader test (not connected with work here)

PR #1000, opened 2020-08-24
970 remove root from reduce overload

### LINKED ISSUE #970: Ambiguous callback/root specification in reduce interface
Consider these two programs. The first one creates an active message callback (registers a handler) that explicitly runs on node 0. All nodes will create this callback and the ID for it will encode where it should run (ever node has the same "pipe ID"). Additionally, the user has to specify in the reduce interface where the root of the reduction occurs. This is for historical reasons before callbacks existed and the root was how a user controlled delivery of the reduce data. Now, the root doesn't matter much as along as the callback can be invoked as desired by the user.

```C++
  auto msg = makeMessage<MyReduceMsg>();
  auto cb = vt::theCB()->makeSend<MyReduceMsg, test_handler_reduce>(0); // run callback on node 0
  theCollective()->global()->reduce<vt::collective::None>(0 /* run root on node 0 */, msg.get(), cb);
```

This second program is very similar, but dangerously (and reproducibly) ambiguous. In the case, we create a callback to a lambda, which seems innocent enough. However, every node does this so we have N callbacks with very different IDs that point to a function pointer on different nodes. The reduce combine function ignores the callback so the order the reduction gets combined will determine which callback actually gets invoked. The root node here is relevant, but not useful for controlling with lambda gets called. A lambda/function pointer callback can be invoked anywhere in the system.

```C++
  auto msg = makeMessage<MyReduceMsg>();
  auto cb = vt::theCB()->makeSend<MyReduceMsg>(&test_handler_reduce);
  theCollective()->global()->reduce<vt::collective::None>(0, msg.get(), cb);
```

@PhilMiller What do you think a good solution for this would be?

cz4rs:
fixes #970

### DIFF
--- a/src/vt/collective/reduce/reduce.h
+++ b/src/vt/collective/reduce/reduce.h
@@ -168,7 +168,6 @@ struct Reduce : virtual collective::tree::Tree {
   /**
    * \brief Reduce a message up the tree
    *
-   * \param[in] root the root node where the final handler provides the result
    * \param[in] msg the message to reduce on this node
    * \param[in] cb the callback to trigger on the root node
    * \param[in] id the reduction stamp (optional), provided if out-of-order
@@ -182,7 +181,7 @@ struct Reduce : virtual collective::tree::Tree {
     ActiveTypedFnType<MsgT> *f
   >
   PendingSendType reduce(
-    NodeType const& root, MsgT* msg, Callback<MsgT> cb,
+    MsgT* msg, Callback<MsgT> cb,
     detail::ReduceStamp id = detail::ReduceStamp{},
     ReduceNumType const& num_contrib = 1
   );
--- a/src/vt/collective/reduce/reduce.impl.h
+++ b/src/vt/collective/reduce/reduce.impl.h
@@ -88,10 +88,11 @@ void Reduce::reduceRootRecv(MsgT* msg) {
 
 template <typename OpT, typename MsgT, ActiveTypedFnType<MsgT> *f>
 Reduce::PendingSendType Reduce::reduce(
-  NodeType const& root, MsgT* msg, Callback<MsgT> cb, detail::ReduceStamp id,
+  MsgT* msg, Callback<MsgT> cb, detail::ReduceStamp id,
   ReduceNumType const& num_contrib
 ) {
   msg->setCallback(cb);
+  NodeType root = 0;
   return reduce<MsgT,f>(root,msg,id,num_contrib);
 }
 
--- a/src/vt/pipe/pipe_manager.impl.h
+++ b/src/vt/pipe/pipe_manager.impl.h
@@ -105,8 +105,9 @@ template <typename MsgT>
 Callback<MsgT> PipeManager::makeSend(
   FuncMsgType<MsgT> fn, NodeType const& node
 ) {
-  return (node == theContext()->getNode()) ?
-    makeCallbackSingleAnon<MsgT,Callback<MsgT>>(fn) : Callback<MsgT>();
+  // return (node == theContext()->getNode()) ?
+  //   makeCallbackSingleAnon<MsgT,Callback<MsgT>>(fn) : Callback<MsgT>();
+  return Callback<MsgT>();
 }
 
 template <typename FunctorT, typename MsgT>
--- a/src/vt/pipe/pipe_manager.impl.h
+++ b/src/vt/pipe/pipe_manager.impl.h
@@ -101,6 +101,14 @@ Callback<MsgT> PipeManager::makeSend(NodeType const& node) {
   return makeCallbackSingleSend<MsgT,f>(node);
 }
 
+template <typename MsgT>
+Callback<MsgT> PipeManager::makeSend(
+  FuncMsgType<MsgT> fn, NodeType const& node
+) {
+  return (node == theContext()->getNode()) ?
+    makeCallbackSingleAnon<MsgT,Callback<MsgT>>(fn) : Callback<MsgT>();
+}
+
 template <typename FunctorT, typename MsgT>
 Callback<MsgT> PipeManager::makeSend(NodeType const& node) {
   return makeCallbackFunctorSend<FunctorT,MsgT>(node);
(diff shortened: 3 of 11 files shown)


src/vt/collective/reduce/reduce.impl.h
```diff
@@ -81,10 +81,11 @@ void Reduce::reduceRootRecv(MsgT* msg) {
 
 template <typename OpT, typename MsgT, ActiveTypedFnType<MsgT> *f>
 detail::ReduceStamp Reduce::reduce(
-  NodeType const& root, MsgT* msg, Callback<MsgT> cb, detail::ReduceStamp id,
+  MsgT* msg, Callback<MsgT> cb, detail::ReduceStamp id,
   ReduceNumType const& num_contrib
 ) {
   msg->setCallback(cb);
+  NodeType root = 0;
```
cz4rs, line 88: not sure about just passing `0` here

note:
this overload was used in 3 locations and in all these cases the provided root value was 0:
- `CollectiveScope::mpiCollectiveAsync(ActionType action)` in `collective_scope.cc`
- `TerminationDetector::startEpochGraphBuild()` in `termination.cc`
- `CollectionManager::reduceConstruction(VirtualProxyType const& proxy)` in `manager.impl.h`

src/vt/pipe/pipe_manager.impl.h
```diff
@@ -99,7 +99,9 @@ Callback<MsgT> PipeManager::makeSend(NodeType const& node) {
 
 template <typename FunctorT, typename MsgT>
 Callback<MsgT> PipeManager::makeSend(NodeType const& node) {
-  return makeCallbackFunctorSend<FunctorT,MsgT>(node);
+  auto const current_node = theContext()->getNode();
```
pnstickne, line 102: Consider not all `auto` is beneficial over a simple `NodeType` (eg.)

src/vt/collective/reduce/reduce.h
```diff
@@ -126,7 +126,6 @@ struct Reduce : virtual collective::tree::Tree {
   /**
    * \brief Reduce a message up the tree
    *
-   * \param[in] root the root node where the final handler provides the result
```
pnstickne, line 129: Would be good to expand method doc. What node(s) does the callback run on?

src/vt/collective/reduce/reduce.impl.h
lifflander: @cz4rs I just confirmed that in all three of cases we don't need the root specified because they have a callback that is either root independent or specify the location for the callback to be fired. The ones we need to resolve are the lambdas that don't pass a node (which probably don't exist due to the bug).

cz4rs: Ack. I've had my schedule interrupted today so I will continue the work on this later in the evening and tomorrow.

cz4rs:
after passing `uninitialized_destination`, function callback behavior:
```
$ mpiexec -n 4 ./tests/test_collectives_reduce_extended --gtest_filter=TestReduce.test_reduce_none_op_cb --vt_debug_reduce
(...)
vt: [0] reduce: callbackFn: cur=0x562a6b0bf2b0: is_root=true, count=1
vt: [1] reduce: callbackFn: cur=0x565162af34d0: is_root=true, count=1
vt: [2] reduce: callbackFn: cur=0x5614a7be8450: is_root=true, count=1
vt: [3] reduce: callbackFn: cur=0x5648ba2fc850: is_root=true, count=1
vt: [3] reduce: callbackFn: cur=0x5648ba2fc450: is_root=true, count=1
vt: [0] reduce: callbackFn: cur=0x562a6b0bf1f0: is_root=true, count=1
vt: [1] reduce: callbackFn: cur=0x565162af37d0: is_root=true, count=1
vt: [2] reduce: callbackFn: cur=0x5614a7be8850: is_root=true, count=1
vt: [3] reduce: callbackFn: cur=0x5648ba2fc850: is_root=true, count=1
vt: [0] reduce: callbackFn: cur=0x562a6b0bf2b0: is_root=true, count=1
vt: [2] reduce: callbackFn: cur=0x5614a7be8450: is_root=true, count=1
vt: [1] reduce: callbackFn: cur=0x565162af37d0: is_root=true, count=1
vt: [3] reduce: callbackFn: cur=0x5648ba2fc450: is_root=true, count=1
../tests/unit/collectives/test_collectives_reduce.extended.cc:126: Failure
Expected equality of these values:
  called
    Which is: 3
  theContext()->getNumNodes()
    Which is: 4
../tests/unit/collectives/test_collectives_reduce.extended.cc:126: Failure
Expected equality of these values:
  called
    Which is: 3
  theContext()->getNumNodes()
    Which is: 4
../tests/unit/collectives/test_collectives_reduce.extended.cc:126: Failure
Expected equality of these values:
  called
    Which is: 3
  theContext()->getNumNodes()
    Which is: 4
```

after adding `makeSend` for lambda callback:
```
$ mpiexec -n 4 ./tests/test_collectives_reduce_extended --gtest_filter=TestReduce.test_reduce_none_op_lambda_cb --vt_debug_reduce
(...)
vt: [0] reduce: lambda callback: cur=0x561001828560: is_root=true, count=1
vt: [0] reduce: lambda callback: cur=0x561001828430: is_root=true, count=1
vt: [0] reduce: lambda callback: cur=0x5610018280f0: is_root=true, count=1
../tests/unit/collectives/test_collectives_reduce.extended.cc:148: Failure
Expected equality of these values:
  called
    Which is: 3
  theContext()->getNumNodes()
    Which is: 4
../tests/unit/collectives/test_collectives_reduce.extended.cc:148: Failure
Expected equality of these values:
  called
    Which is: 0
  theContext()->getNumNodes()
    Which is: 4
../tests/unit/collectives/test_collectives_reduce.extended.cc:148: Failure
Expected equality of these values:
  called
    Which is: 0
  theContext()->getNumNodes()
    Which is: 4
../tests/unit/collectives/test_collectives_reduce.extended.cc:148: Failure
Expected equality of these values:
  called
    Which is: 0
  theContext()->getNumNodes()
    Which is: 4
```

src/vt/collective/reduce/reduce.h
lifflander: Agreed, we should specify what happens now that the interface changed

lifflander:
this doesn't contain a fix so we are closing it.

PR #1008, opened 2020-08-27
1007 Add missing cleanup for restoreFromFile

### LINKED ISSUE #1007: Collection restoreFromFile leaks a collection pointer?
**Describe the bug**
It calls the copy constructor and then proceeds to leak a copy of the data that was serialized.

lifflander:
Fixes #1007

- Write a test that triggers the destruction bug
- Add missing cleanup functions (destroy not working) for token-based collection insertion
- Refactor to avoid calling both copy constructor and move constructor when restoring from file
- Create a helper for doing the insertion once the unique pointer is available

### DIFF
--- a/src/vt/vrt/collection/manager.impl.h
+++ b/src/vt/vrt/collection/manager.impl.h
@@ -1679,20 +1679,46 @@ VirtualProxyType CollectionManager::makeDistProxy(TagType const& tag) {
 /* end SPMD distributed collection support */
 
 
+template <typename ColT>
+void CollectionManager::staticInsertColPtr(
+  VirtualProxyType proxy, typename ColT::IndexType idx,
+  std::unique_ptr<ColT> ptr
+) {
+  using IndexT = typename ColT::IndexType;
+  using BaseIdxType = vt::index::BaseIndex;
+
+  auto map_han = UniversalIndexHolder<>::getMap(proxy);
+  auto holder = findColHolder<ColT, IndexT>(proxy);
+  auto range = holder->max_idx;
+  auto const num_elms = range.getSize();
+  auto fn = auto_registry::getHandlerMap(map_han);
+  auto const num_nodes = theContext()->getNumNodes();
+  auto const cur = static_cast<BaseIdxType*>(&idx);
+  auto const max = static_cast<BaseIdxType*>(&range);
+  auto const home_node = fn(cur, max, num_nodes);
+
+  // Through the attorney, setup all the properties on the newly constructed
+  // collection element: index, proxy, number of elements. Note: because of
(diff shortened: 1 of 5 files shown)


PhilMiller:
Could we delete the copy constructor for the base collection element class, so that this sort of issue is prevented at compile time?

lifflander:
> Could we delete the copy constructor for the base collection element class, so that this sort of issue is prevented at compile time?

I not sure its ideal  to force collection elements to never be copyable. I don't want to prescribe that on users.

PhilMiller:
I guess that kinda comes down to a broader philosophical point of whether an instance of a collection element class is always actually part of a collection. Many members are going to be poorly defined if the answer is no, so I think the answer has to be yes. At which point, what does it mean to have a copy of such an element?

lifflander:
> I guess that kinda comes down to a broader philosophical point of whether an instance of a collection element class is always actually part of a collection. Many members are going to be poorly defined if the answer is no, so I think the answer has to be yes. At which point, what does it mean to have a copy of such an element?

I'm thinking of dual-use cases, where one might not always use classes that inherit from collection elements always as a member of the collection. Also, I think the less we enforce on requirements for a user's class the better. Eventually, I could imagine an interface where we don't even require inheritance from the user for a collection element.

lifflander:
I do prefer the design I went with for ObjGroups which make them very non-intrusive.

src/vt/vrt/collection/manager.impl.h
```diff
@@ -1679,20 +1679,46 @@ VirtualProxyType CollectionManager::makeDistProxy(TagType const& tag) {
...
+  auto const home_node = fn(cur, max, num_nodes);
+
+  // Through the attorney, setup all the properties on the newly constructed
+  // collection element: index, proxy, number of elements. Note: because of
+  // how the constructor works, the index is not currently available through
+  // "getIndex"
+  CollectionTypeAttorney::setup(ptr.get(), num_elms, idx, proxy);
+
+  VirtualPtrType<ColT, IndexT> col_ptr(
+    static_cast<CollectionBase<ColT, IndexT>*>(ptr.release())
```
PhilMiller, line 1707: Is this casting actually necessary? Can't `unique_ptr` do that conversion implicitly?

PhilMiller:
> I do prefer the design I went with for ObjGroups which make them very non-intrusive.

If we can go that direction for collection elements, I'd support it.

src/vt/vrt/collection/manager.impl.h
lifflander: No, it can't in this case.

Issue #1003, opened 2020-08-25
"Nested schedulers never expected from idle context" can be asserted


pnstickne:
@nlslatt was able to cause this (not expected) assertion. Resolve.

PhilMiller:
Stack trace

[backtrace.txt](https://github.com/DARMA-tasking/vt/files/5125542/backtrace.txt)

PR #1010, opened 2020-09-01
#1003 schedule- remove idle-assert

pnstickne:
- Not neccessary and over-restriction for edge case.


PhilMiller:
It's not clear to me what exactly the edge case is, where a nested scheduler can be called in a situation where `is_idle == true`. I'm concerned that the assertion is catching something wrong elsewhere (i.e. a missing/excess/misplaced assignment to `is_idle`), that we should be fixing instead.

PR #1013, opened 2020-09-01
#1009 add memory footprinting support for VT components

### LINKED ISSUE #1009: Monitor memory usage of VT components
**What Needs to be Done?**

We've been trying to solve some memory growth in EMPIRE. However, it's difficult to tell where memory increase might occur. Thus, the idea is to add a virtual component method for reporting memory usage across the all the VT components. To do this, we should add serializers (that count the size at the least---doesn't have to serialize properly) to the VT components and then invoke `checkpoint::getSize` on the component.

cz4rs:
fixes #1009

### DIFF
--- a/examples/rdma/rdma_simple_get.cc
+++ b/examples/rdma/rdma_simple_get.cc
@@ -93,7 +93,6 @@ static vt::RDMA_GetType test_get_fn(
 
 int main(int argc, char** argv) {
   vt::initialize(argc, argv);
-  vt::rt->printMemoryFootprint();
 
   vt::NodeType this_node = vt::theContext()->getNode();
 
@@ -117,7 +116,6 @@ int main(int argc, char** argv) {
     vt::theMsg()->broadcastMsg<HandleMsg, tell_handle>(msg);
   }
 
-  vt::rt->printMemoryFootprint();
   vt::finalize();
 
   return 0;
--- a/examples/rdma/rdma_simple_get.cc
+++ b/examples/rdma/rdma_simple_get.cc
@@ -93,6 +93,7 @@ static vt::RDMA_GetType test_get_fn(
 
 int main(int argc, char** argv) {
   vt::initialize(argc, argv);
+  vt::rt->printMemoryFootprint();
 
   vt::NodeType this_node = vt::theContext()->getNode();
 
@@ -116,6 +117,7 @@ int main(int argc, char** argv) {
     vt::theMsg()->broadcastMsg<HandleMsg, tell_handle>(msg);
   }
 
+  vt::rt->printMemoryFootprint();
   vt::finalize();
 
   return 0;
--- a/src/vt/configs/arguments/app_config.h
+++ b/src/vt/configs/arguments/app_config.h
@@ -150,7 +150,7 @@ struct AppConfig {
   bool vt_term_rooted_use_wave = false;
   int64_t vt_hang_freq         = 1024;
 
-#if vt_check_enabled(diagnostics_runtime)
+#if (vt_diagnostics_runtime != 0)
   bool vt_diag_enable = true;
 #else
   bool vt_diag_enable = false;
--- a/src/vt/configs/arguments/app_config.h
+++ b/src/vt/configs/arguments/app_config.h
@@ -102,7 +102,7 @@ struct AppConfig {
   bool vt_print_memory_at_threshold   = false;
   std::string vt_print_memory_threshold = "1 GiB";
   int32_t vt_print_memory_sched_poll    = 100;
-  bool vt_print_memory_footprint  = false;
+  bool vt_print_memory_footprint  = true;
 
   bool vt_no_warn_stack     = false;
   bool vt_no_assert_stack   = false;
--- a/src/vt/configs/arguments/app_config.h
+++ b/src/vt/configs/arguments/app_config.h
@@ -102,6 +102,7 @@ struct AppConfig {
   bool vt_print_memory_at_threshold   = false;
   std::string vt_print_memory_threshold = "1 GiB";
   int32_t vt_print_memory_sched_poll    = 100;
+  bool vt_print_memory_footprint  = false;
 
   bool vt_no_warn_stack     = false;
   bool vt_no_assert_stack   = false;
@@ -249,6 +250,7 @@ struct AppConfig {
       | vt_print_memory_at_threshold
       | vt_print_memory_threshold
       | vt_print_memory_sched_poll
+      | vt_print_memory_footprint
 
       | vt_no_warn_stack
       | vt_no_assert_stack
--- a/src/vt/configs/arguments/app_config.h
+++ b/src/vt/configs/arguments/app_config.h
@@ -102,7 +102,7 @@ struct AppConfig {
   bool vt_print_memory_at_threshold   = false;
   std::string vt_print_memory_threshold = "1 GiB";
   int32_t vt_print_memory_sched_poll    = 100;
-  bool vt_print_memory_footprint  = true;
+  bool vt_print_memory_footprint  = false;
 
   bool vt_no_warn_stack     = false;
   bool vt_no_assert_stack   = false;
--- a/src/vt/configs/arguments/app_config.h
+++ b/src/vt/configs/arguments/app_config.h
@@ -353,7 +353,6 @@ struct AppConfig {
 
       | argv_prog_name
 
-      | mpi_init_args
       | passthru_args;
   }
 };
--- a/src/vt/context/context.h
+++ b/src/vt/context/context.h
@@ -69,6 +69,8 @@ namespace vt {  namespace ctx {
  * functionality analogous to \c MPI_Comm_size and \c MPI_Comm_rank.
  */
 struct Context : runtime::component::Component<Context> {
+  checkpoint_virtual_serialize_derived_from(Component)
+
   /**
    * \internal
    * \brief Construct the context.
@@ -143,8 +145,13 @@ struct Context : runtime::component::Component<Context> {
 
   std::string name() override { return "Context"; }
 
-  template <typename Serializer>
-  void serialize(Serializer& s) {
+  template <
+    typename SerializerT,
+    typename = std::enable_if_t<
+      std::is_same<SerializerT, checkpoint::Footprinter>::value
+    >
+  >
+  void serialize(SerializerT& s) {
     s | thisNode_
       | numNodes_
       | numWorkers_;
--- a/src/vt/context/context.h
+++ b/src/vt/context/context.h
@@ -147,8 +147,7 @@ struct Context : runtime::component::Component<Context> {
   void serialize(Serializer& s) {
     s | thisNode_
       | numNodes_
-      | numWorkers_
-      | is_comm_world_;
+      | numWorkers_;
 
     s.countBytes(communicator_);
   }
--- a/src/vt/context/context.h
+++ b/src/vt/context/context.h
@@ -48,6 +48,8 @@
 #include <memory>
 #include <mpi.h>
 
+#include <checkpoint/checkpoint.h>
+
 #include "vt/config.h"
 #include "vt/runtime/component/component_pack.h"
 #include "vt/context/context_attorney_fwd.h"
@@ -69,8 +71,6 @@ namespace vt {  namespace ctx {
  * functionality analogous to \c MPI_Comm_size and \c MPI_Comm_rank.
  */
 struct Context : runtime::component::Component<Context> {
-  checkpoint_virtual_serialize_derived_from(Component)
-
   /**
    * \internal
    * \brief Construct the context.
--- a/src/vt/context/context.h
+++ b/src/vt/context/context.h
@@ -48,8 +48,6 @@
 #include <memory>
 #include <mpi.h>
 
-#include <checkpoint/checkpoint.h>
-
 #include "vt/config.h"
 #include "vt/runtime/component/component_pack.h"
 #include "vt/context/context_attorney_fwd.h"
@@ -145,12 +143,7 @@ struct Context : runtime::component::Component<Context> {
 
   std::string name() override { return "Context"; }
 
-  template <
-    typename SerializerT,
-    typename = std::enable_if_t<
-      std::is_same<SerializerT, checkpoint::Footprinter>::value
-    >
-  >
+  template <typename SerializerT>
   void serialize(SerializerT& s) {
     s | thisNode_
       | numNodes_
--- a/src/vt/context/context.h
+++ b/src/vt/context/context.h
@@ -149,7 +149,8 @@ struct Context : runtime::component::Component<Context> {
       | numNodes_
       | numWorkers_
       | is_comm_world_;
-      // | communicator_; // ompi_communicator_t
+
+    s.countBytes(communicator_);
   }
 
 protected:
--- a/src/vt/context/context.h
+++ b/src/vt/context/context.h
@@ -147,9 +147,8 @@ struct Context : runtime::component::Component<Context> {
   void serialize(SerializerT& s) {
     s | thisNode_
       | numNodes_
-      | numWorkers_;
-
-    s.countBytes(communicator_);
+      | numWorkers_
+      | communicator_;
   }
 
 protected:
--- a/src/vt/event/event.h
+++ b/src/vt/event/event.h
@@ -194,9 +194,8 @@ struct AsyncEvent : runtime::component::PollableComponent<AsyncEvent> {
       | polling_event_container_
       | eventPollCount
       | eventSizeGauge
-      | mpiEventWaitTime;
-
-    s.countBytes(lookup_container_);
+      | mpiEventWaitTime
+      | lookup_container_;
 
   # if vt_check_enabled(trace_enabled)
     s | trace_event_polling;
--- a/src/vt/event/event.h
+++ b/src/vt/event/event.h
@@ -53,6 +53,8 @@
 #include "vt/event/event_holder.h"
 #include "vt/event/event_msgs.h"
 
+#include <checkpoint/checkpoint.h>
+
 #include <memory>
 #include <vector>
 #include <list>
@@ -84,8 +86,6 @@ enum class EventState : int8_t {
  * Component to track events in the system to trigger actions or other events
  */
 struct AsyncEvent : runtime::component::PollableComponent<AsyncEvent> {
-  checkpoint_virtual_serialize_derived_from(PollableComponent)
-
   using EventRecordTypeType = eEventRecord;
   using EventManagerType = EventIDManager;
   using EventStateType = EventState;
--- a/src/vt/event/event.h
+++ b/src/vt/event/event.h
@@ -84,6 +84,8 @@ enum class EventState : int8_t {
  * Component to track events in the system to trigger actions or other events
  */
 struct AsyncEvent : runtime::component::PollableComponent<AsyncEvent> {
+  checkpoint_virtual_serialize_derived_from(PollableComponent)
+
   using EventRecordTypeType = eEventRecord;
   using EventManagerType = EventIDManager;
   using EventStateType = EventState;
@@ -187,8 +189,13 @@ struct AsyncEvent : runtime::component::PollableComponent<AsyncEvent> {
 
   std::string name() override { return "AsyncEvent"; }
 
-  template <typename Serializer>
-  void serialize(Serializer& s) {
+  template <
+    typename SerializerT,
+    typename = std::enable_if_t<
+      std::is_same<SerializerT, checkpoint::Footprinter>::value
+    >
+  >
+  void serialize(SerializerT& s) {
     s | cur_event_
       | event_container_
       | polling_event_container_;
--- a/src/vt/event/event.h
+++ b/src/vt/event/event.h
@@ -53,8 +53,6 @@
 #include "vt/event/event_holder.h"
 #include "vt/event/event_msgs.h"
 
-#include <checkpoint/checkpoint.h>
-
 #include <memory>
 #include <vector>
 #include <list>
@@ -189,12 +187,7 @@ struct AsyncEvent : runtime::component::PollableComponent<AsyncEvent> {
 
   std::string name() override { return "AsyncEvent"; }
 
-  template <
-    typename SerializerT,
-    typename = std::enable_if_t<
-      std::is_same<SerializerT, checkpoint::Footprinter>::value
-    >
-  >
+  template <typename SerializerT>
   void serialize(SerializerT& s) {
     s | cur_event_
       | event_container_
--- a/src/vt/event/event.h
+++ b/src/vt/event/event.h
@@ -191,7 +191,10 @@ struct AsyncEvent : runtime::component::PollableComponent<AsyncEvent> {
   void serialize(SerializerT& s) {
     s | cur_event_
       | event_container_
-      | polling_event_container_;
+      | polling_event_container_
+      | eventPollCount
+      | eventSizeGauge
+      | mpiEventWaitTime;
 
     s.countBytes(lookup_container_);
 
--- a/src/vt/event/event_record.h
+++ b/src/vt/event/event_record.h
@@ -73,9 +73,7 @@ union uEventPayload {
 
   template <typename Serializer>
   void serialize(Serializer& s) {
-    s | event_list;
-
-    s.countBytes(mpi_req);
+    s.countBytes(*this);
   }
 };
 
--- a/src/vt/event/event_record.h
+++ b/src/vt/event/event_record.h
@@ -117,6 +117,10 @@ struct EventRecord {
       | event_union_
       | event_id_
       | type_;
+
+  # if vt_check_enabled(diagnostics)
+    s | creation_time_stamp_;
+  # endif
   }
 
 private:
--- a/src/vt/group/group_manager.h
+++ b/src/vt/group/group_manager.h
@@ -382,9 +382,8 @@ public:
       | cur_collective_id_
       | continuation_actions_
       | cleanup_actions_
-      | collective_scope_;
-
-    s.countBytes(default_comm_);
+      | collective_scope_
+      | default_comm_;
   }
 
 private:
--- a/src/vt/group/group_manager.h
+++ b/src/vt/group/group_manager.h
@@ -375,9 +375,9 @@ public:
   void serialize(Serializer& s) {
     s | next_group_id_
       | next_collective_group_id_
-      // | local_collective_group_info_
-      // | local_group_info_
-      // | remote_group_info_
+      | local_collective_group_info_
+      | local_group_info_
+      | remote_group_info_
       | cur_id_
       | cur_collective_id_
       | continuation_actions_
--- a/src/vt/group/group_manager.h
+++ b/src/vt/group/group_manager.h
@@ -64,6 +64,8 @@
 #include "vt/collective/collective_scope.h"
 #include "vt/runtime/component/component_pack.h"
 
+#include <checkpoint/checkpoint.h>
+
 #include <memory>
 #include <unordered_map>
 #include <cstdlib>
@@ -84,8 +86,6 @@ namespace vt { namespace group {
  * can dynamically create groups in a fully distributed manner.
  */
 struct GroupManager : runtime::component::Component<GroupManager> {
-  checkpoint_virtual_serialize_derived_from(Component)
-
   using RegionType = region::Region;
   using RegionPtrType = std::unique_ptr<RegionType>;
   using GroupInfoType = Info;
--- a/src/vt/group/group_manager.h
+++ b/src/vt/group/group_manager.h
@@ -64,8 +64,6 @@
 #include "vt/collective/collective_scope.h"
 #include "vt/runtime/component/component_pack.h"
 
-#include <checkpoint/checkpoint.h>
-
 #include <memory>
 #include <unordered_map>
 #include <cstdlib>
@@ -373,12 +371,7 @@ public:
    */
   RemoteOperationIDType getNextOpID();
 
-  template <
-    typename SerializerT,
-    typename = std::enable_if_t<
-      std::is_same<SerializerT, checkpoint::Footprinter>::value
-    >
-  >
+  template <typename SerializerT>
   void serialize(SerializerT& s) {
     s | next_group_id_
       | next_collective_group_id_
--- a/src/vt/group/group_manager.h
+++ b/src/vt/group/group_manager.h
@@ -383,6 +383,8 @@ public:
       | continuation_actions_
       | cleanup_actions_
       | collective_scope_;
+
+    s.countBytes(default_comm_);
   }
 
 private:
--- a/src/vt/group/group_manager.h
+++ b/src/vt/group/group_manager.h
@@ -84,6 +84,8 @@ namespace vt { namespace group {
  * can dynamically create groups in a fully distributed manner.
  */
 struct GroupManager : runtime::component::Component<GroupManager> {
+  checkpoint_virtual_serialize_derived_from(Component)
+
   using RegionType = region::Region;
   using RegionPtrType = std::unique_ptr<RegionType>;
   using GroupInfoType = Info;
@@ -371,8 +373,13 @@ public:
    */
   RemoteOperationIDType getNextOpID();
 
-  template <typename Serializer>
-  void serialize(Serializer& s) {
+  template <
+    typename SerializerT,
+    typename = std::enable_if_t<
+      std::is_same<SerializerT, checkpoint::Footprinter>::value
+    >
+  >
+  void serialize(SerializerT& s) {
     s | next_group_id_
       | next_collective_group_id_
       | local_collective_group_info_
--- a/src/vt/messaging/active.h
+++ b/src/vt/messaging/active.h
@@ -155,8 +155,9 @@ struct InProgressBase {
     s | buf
       | probe_bytes
       | sender
-      | req
       | valid;
+
+    s.countBytes(req);
   }
 
   char* buf = nullptr;
--- a/src/vt/messaging/active.h
+++ b/src/vt/messaging/active.h
@@ -117,6 +117,8 @@ struct PendingRecv {
   PriorityType priority = no_priority;
   bool is_user_buf = false;
 
+  PendingRecv() = default;
+
   PendingRecv(
     int in_nchunks, void* in_user_buf, ContinuationDeleterType in_cont,
     ActionType in_dealloc_user_buf, NodeType node,
@@ -125,6 +127,15 @@ struct PendingRecv {
       dealloc_user_buf(in_dealloc_user_buf), sender(node),
       priority(in_priority), is_user_buf(in_is_user_buf)
   { }
+
+  template <typename Serializer>
+  void serialize(Serializer& s) {
+    s | user_buf
+      | cont
+      | dealloc_user_buf
+      | recv_node
+      | priority;
+  }
 };
 
 /**
@@ -249,11 +260,20 @@ struct BufferedActiveMsg {
   NodeType from_node;
   ActionType cont;
 
+  BufferedActiveMsg() : buffered_msg(nullptr) {}
+
   BufferedActiveMsg(
     MessageType const& in_buffered_msg, NodeType const& in_from_node,
     ActionType in_cont
   ) : buffered_msg(in_buffered_msg), from_node(in_from_node), cont(in_cont)
   { }
+
+  template <typename Serializer>
+  void serialize(Serializer& s) {
+    s | buffered_msg
+      | from_node
+      | cont;
+  }
 };
 
 // forward-declare for header
@@ -1667,13 +1687,13 @@ struct ActiveMessenger : runtime::component::PollableComponent<ActiveMessenger> 
       | current_priority_context_
       | current_priority_level_context_
       | maybe_ready_tag_han_
-      // | pending_handler_msgs_
-      // | pending_recvs_
+      | pending_handler_msgs_
+      | pending_recvs_
       | cur_direct_buffer_tag_
-      // | epoch_stack_ // std::stack
-      // | send_listen_
-      // | in_progress_active_msg_irecv
-      // | in_progress_data_irecv
+      | epoch_stack_
+      | send_listen_
+      | in_progress_active_msg_irecv
+      | in_progress_data_irecv
       | this_node_;
 
   # if vt_check_enabled(trace_enabled)
--- a/src/vt/messaging/active.h
+++ b/src/vt/messaging/active.h
@@ -117,8 +117,6 @@ struct PendingRecv {
   PriorityType priority = no_priority;
   bool is_user_buf = false;
 
-  PendingRecv() = default;
-
   PendingRecv(
     int in_nchunks, void* in_user_buf, ContinuationDeleterType in_cont,
     ActionType in_dealloc_user_buf, NodeType node,
@@ -260,8 +258,6 @@ struct BufferedActiveMsg {
   MessageType buffered_msg;
   NodeType from_node;
   ActionType cont;
-
-  BufferedActiveMsg() : buffered_msg(nullptr) {}
 
   BufferedActiveMsg(
     MessageType const& in_buffered_msg, NodeType const& in_from_node,
--- a/src/vt/messaging/active.h
+++ b/src/vt/messaging/active.h
@@ -49,6 +49,8 @@
 #include <memory>
 #include <mpi.h>
 
+#include <checkpoint/checkpoint.h>
+
 #include "vt/config.h"
 #include "vt/activefn/activefn.h"
 #include "vt/messaging/active.fwd.h"
@@ -312,8 +314,6 @@ struct MultiMsg;
  *  - \ref sendpayload
  */
 struct ActiveMessenger : runtime::component::PollableComponent<ActiveMessenger> {
-  checkpoint_virtual_serialize_derived_from(PollableComponent)
-
   using BufferedMsgType      = BufferedActiveMsg;
   using MessageType          = ShortMessage;
   using CountType            = int32_t;
--- a/src/vt/messaging/active.h
+++ b/src/vt/messaging/active.h
@@ -312,6 +312,8 @@ struct MultiMsg;
  *  - \ref sendpayload
  */
 struct ActiveMessenger : runtime::component::PollableComponent<ActiveMessenger> {
+  checkpoint_virtual_serialize_derived_from(PollableComponent)
+
   using BufferedMsgType      = BufferedActiveMsg;
   using MessageType          = ShortMessage;
   using CountType            = int32_t;
@@ -1676,8 +1678,13 @@ struct ActiveMessenger : runtime::component::PollableComponent<ActiveMessenger> 
     send_listen_.clear();
   }
 
-  template <typename Serializer>
-  void serialize(Serializer& s) {
+  template <
+    typename SerializerT,
+    typename = std::enable_if_t<
+      std::is_same<SerializerT, checkpoint::Footprinter>::value
+    >
+  >
+  void serialize(SerializerT& s) {
     s | current_handler_context_
       | current_node_context_
       | current_epoch_context_
--- a/src/vt/messaging/active.h
+++ b/src/vt/messaging/active.h
@@ -49,8 +49,6 @@
 #include <memory>
 #include <mpi.h>
 
-#include <checkpoint/checkpoint.h>
-
 #include "vt/config.h"
 #include "vt/activefn/activefn.h"
 #include "vt/messaging/active.fwd.h"
@@ -1678,12 +1676,7 @@ struct ActiveMessenger : runtime::component::PollableComponent<ActiveMessenger> 
     send_listen_.clear();
   }
 
-  template <
-    typename SerializerT,
-    typename = std::enable_if_t<
-      std::is_same<SerializerT, checkpoint::Footprinter>::value
-    >
-  >
+  template <typename SerializerT>
   void serialize(SerializerT& s) {
     s | current_handler_context_
       | current_node_context_
--- a/src/vt/messaging/active.h
+++ b/src/vt/messaging/active.h
@@ -128,11 +128,13 @@ struct PendingRecv {
 
   template <typename Serializer>
   void serialize(Serializer& s) {
-    s | user_buf
+    s | nchunks
+      | user_buf
       | cont
       | dealloc_user_buf
-      | recv_node
-      | priority;
+      | sender
+      | priority
+      | is_user_buf;
   }
 };
 
@@ -154,8 +156,6 @@ struct InProgressBase {
       | probe_bytes
       | sender
       | valid;
-
-    s.countBytes(req);
   }
 
   char* buf = nullptr;
--- a/src/vt/messaging/active.h
+++ b/src/vt/messaging/active.h
@@ -232,9 +232,8 @@ struct InProgressDataIRecv : InProgressBase {
       | dealloc_user_buf
       | next
       | priority
-      | cur;
-
-    s.countBytes(reqs);
+      | cur
+      | reqs;
   }
 
   void* user_buf = nullptr;
--- a/src/vt/messaging/message/smart_ptr.h
+++ b/src/vt/messaging/message/smart_ptr.h
@@ -48,6 +48,7 @@
 #include "vt/config.h"
 #include "vt/messaging/message/message.h"
 #include "vt/messaging/message/refs.h"
+#include "vt/serialization/sizer.h"
 
 #include <iosfwd>
 #include <cassert>
@@ -215,9 +216,13 @@ struct MsgSharedPtr final {
               << ")";
   }
 
-  template <typename Serializer>
-  void serialize(Serializer& s) {
-    s.countBytes(*this);
+  template <typename SerializerT>
+  void serialize(SerializerT& s) {
+    if (ownsMessage()) {
+      auto ptr = get();
+      auto const msg_size = vt::serialization::MsgSizer<MsgType>::get(ptr);
+      s.addBytes(msg_size);
+    }
   }
 
 private:
--- a/src/vt/messaging/message/smart_ptr.h
+++ b/src/vt/messaging/message/smart_ptr.h
@@ -216,7 +216,12 @@ struct MsgSharedPtr final {
               << ")";
   }
 
-  template <typename SerializerT>
+  template <
+    typename SerializerT,
+    typename = std::enable_if_t<
+      std::is_same<SerializerT, checkpoint::Footprinter>::value
+    >
+  >
   void serialize(SerializerT& s) {
     if (ownsMessage()) {
       auto ptr = get();
--- a/src/vt/messaging/message/smart_ptr.h
+++ b/src/vt/messaging/message/smart_ptr.h
@@ -217,8 +217,7 @@ struct MsgSharedPtr final {
 
   template <typename Serializer>
   void serialize(Serializer& s) {
-    // s | impl_
-    //   | ptr_;
+    s.countBytes(*this);
   }
 
 private:
--- a/src/vt/objgroup/manager.h
+++ b/src/vt/objgroup/manager.h
@@ -82,6 +82,8 @@ namespace vt { namespace objgroup {
  * implemented as object groups, such as the load balancers.
  */
 struct ObjGroupManager : runtime::component::Component<ObjGroupManager> {
+  checkpoint_virtual_serialize_derived_from(Component)
+
   template <typename ObjT>
   using ProxyType           = proxy::Proxy<ObjT>;
   template <typename ObjT>
@@ -362,8 +364,13 @@ struct ObjGroupManager : runtime::component::Component<ObjGroupManager> {
     MsgSharedPtr<ShortMessage> msg, HandlerType han, EpochType ep
   );
 
-  template <typename Serializer>
-  void serialize(Serializer& s) {
+  template <
+    typename SerializerT,
+    typename = std::enable_if_t<
+      std::is_same<SerializerT, checkpoint::Footprinter>::value
+    >
+  >
+  void serialize(SerializerT& s) {
     s | cur_obj_id_
       | dispatch_
       | objs_
--- a/src/vt/objgroup/manager.h
+++ b/src/vt/objgroup/manager.h
@@ -58,6 +58,8 @@
 #include "vt/messaging/message/smart_ptr.h"
 #include "vt/messaging/pending_send.h"
 
+#include <checkpoint/checkpoint.h>
+
 #include <memory>
 #include <functional>
 #include <unordered_map>
@@ -82,7 +84,6 @@ namespace vt { namespace objgroup {
  * implemented as object groups, such as the load balancers.
  */
 struct ObjGroupManager : runtime::component::Component<ObjGroupManager> {
-  checkpoint_virtual_serialize_derived_from(Component)
 
   template <typename ObjT>
   using ProxyType           = proxy::Proxy<ObjT>;
--- a/src/vt/objgroup/manager.h
+++ b/src/vt/objgroup/manager.h
@@ -365,7 +365,7 @@ struct ObjGroupManager : runtime::component::Component<ObjGroupManager> {
   template <typename Serializer>
   void serialize(Serializer& s) {
     s | cur_obj_id_
-      // | dispatch_
+      | dispatch_
       // | objs_
       | obj_to_proxy_
       // | pending_
--- a/src/vt/objgroup/manager.h
+++ b/src/vt/objgroup/manager.h
@@ -366,9 +366,9 @@ struct ObjGroupManager : runtime::component::Component<ObjGroupManager> {
   void serialize(Serializer& s) {
     s | cur_obj_id_
       | dispatch_
-      // | objs_
+      | objs_
       | obj_to_proxy_
-      // | pending_
+      | pending_
       | derived_to_bases_;
   }
 
--- a/src/vt/objgroup/manager.h
+++ b/src/vt/objgroup/manager.h
@@ -58,8 +58,6 @@
 #include "vt/messaging/message/smart_ptr.h"
 #include "vt/messaging/pending_send.h"
 
-#include <checkpoint/checkpoint.h>
-
 #include <memory>
 #include <functional>
 #include <unordered_map>
@@ -366,11 +364,7 @@ struct ObjGroupManager : runtime::component::Component<ObjGroupManager> {
   );
 
   template <
-    typename SerializerT,
-    typename = std::enable_if_t<
-      std::is_same<SerializerT, checkpoint::Footprinter>::value
-    >
-  >
+    typename SerializerT>
   void serialize(SerializerT& s) {
     s | cur_obj_id_
       | dispatch_
--- a/src/vt/objgroup/manager.h
+++ b/src/vt/objgroup/manager.h
@@ -363,8 +363,7 @@ struct ObjGroupManager : runtime::component::Component<ObjGroupManager> {
     MsgSharedPtr<ShortMessage> msg, HandlerType han, EpochType ep
   );
 
-  template <
-    typename SerializerT>
+  template <typename SerializerT>
   void serialize(SerializerT& s) {
     s | cur_obj_id_
       | dispatch_
--- a/src/vt/rdmahandle/manager.h
+++ b/src/vt/rdmahandle/manager.h
@@ -229,7 +229,12 @@ public:
 
   template <typename Serializer>
   void serialize(Serializer& s) {
-    //
+    s | cur_handle_obj_group_
+      | cur_handle_collection_
+      | collection_to_manager_
+      | proxy_
+      // holder_
+      | collective_scope_;
   }
 
 private:
--- a/src/vt/rdmahandle/manager.h
+++ b/src/vt/rdmahandle/manager.h
@@ -227,6 +227,11 @@ private:
 public:
   static std::unique_ptr<Manager> construct();
 
+  template <typename Serializer>
+  void serialize(Serializer& s) {
+    //
+  }
+
 private:
   /// Current collective handle for a given objgroup proxy
   std::unordered_map<ObjGroupProxyType, RDMA_HandleType> cur_handle_obj_group_;
--- a/src/vt/rdmahandle/manager.h
+++ b/src/vt/rdmahandle/manager.h
@@ -59,6 +59,8 @@
 #include "vt/runtime/component/component_pack.h"
 #include "vt/collective/collective_scope.h"
 
+#include <checkpoint/checkpoint.h>
+
 namespace vt { namespace rdma {
 
 /** \file */
@@ -81,8 +83,6 @@ struct InformRDMAMsg;
  * \brief RDMA Handle Manager for creation of node- or index-level handles
  */
 struct Manager : runtime::component::Component<Manager> {
-  checkpoint_virtual_serialize_derived_from(Component)
-
   using ProxyType       = vt::objgroup::proxy::Proxy<Manager>;
   using ElemToHandle    = std::unordered_map<int64_t, RDMA_HandleType>;
   using HandleToManager = std::unordered_map<RDMA_HandleType, ObjGroupProxyType>;
--- a/src/vt/rdmahandle/manager.h
+++ b/src/vt/rdmahandle/manager.h
@@ -59,8 +59,6 @@
 #include "vt/runtime/component/component_pack.h"
 #include "vt/collective/collective_scope.h"
 
-#include <checkpoint/checkpoint.h>
-
 namespace vt { namespace rdma {
 
 /** \file */
@@ -231,12 +229,7 @@ private:
 public:
   static std::unique_ptr<Manager> construct();
 
-  template <
-    typename SerializerT,
-    typename = std::enable_if_t<
-      std::is_same<SerializerT, checkpoint::Footprinter>::value
-    >
-  >
+  template <typename SerializerT>
   void serialize(SerializerT& s) {
     s | cur_handle_obj_group_
       | cur_handle_collection_
--- a/src/vt/rdmahandle/manager.h
+++ b/src/vt/rdmahandle/manager.h
@@ -81,6 +81,8 @@ struct InformRDMAMsg;
  * \brief RDMA Handle Manager for creation of node- or index-level handles
  */
 struct Manager : runtime::component::Component<Manager> {
+  checkpoint_virtual_serialize_derived_from(Component)
+
   using ProxyType       = vt::objgroup::proxy::Proxy<Manager>;
   using ElemToHandle    = std::unordered_map<int64_t, RDMA_HandleType>;
   using HandleToManager = std::unordered_map<RDMA_HandleType, ObjGroupProxyType>;
@@ -227,8 +229,13 @@ private:
 public:
   static std::unique_ptr<Manager> construct();
 
-  template <typename Serializer>
-  void serialize(Serializer& s) {
+  template <
+    typename SerializerT,
+    typename = std::enable_if_t<
+      std::is_same<SerializerT, checkpoint::Footprinter>::value
+    >
+  >
+  void serialize(SerializerT& s) {
     s | cur_handle_obj_group_
       | cur_handle_collection_
       | collection_to_manager_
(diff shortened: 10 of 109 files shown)


cz4rs:
pushed 6a34657 as a basis for discussion

- based on [sizer test](https://github.com/DARMA-tasking/checkpoint/blob/develop/tests/unit/test_sizer.cc), `serialize` method needs to be implemented in all components (and the types they are composed of)
- is there any need for `using isByteCopyable = std::false_type;`?

src/vt/configs/arguments/app_config.h
```diff
@@ -57,6 +59,11 @@ namespace vt { namespace arguments {
  * \brief Configuration parameters for a VT instance
  */
 struct AppConfig {
+  template <typename Serializer>
```
pnstickne, line 62: The serializers usually appear lower in types. I wonder if there is any advantage (in checkpoint or otherwise) to use out-of-line serializers.

src/vt/configs/arguments/app_config.h
```diff
@@ -57,6 +59,11 @@ namespace vt { namespace arguments {
...
+  void serialize(Serializer& s) {
+    s | vt_color | vt_no_color | vt_sched_num_progress | vt_memory_reporters;
```
pnstickne, line 64: This surely isn't all the state..

If partial serialization is all that is required/useful, a comment to the effect/scope might be warranted.

cz4rs: @pnstickne correct, the idea is to measure memory footprint of a whole component, this (partial implementation) was just a basis for discussion during last Tuesday's meeting. The decision is to make changes in checkpoint library first - I will be preparing the merge request there today, and get back to this issue then.

src/vt/configs/arguments/app_config.h
cz4rs: not sure about that, but that can be changed easily if there's any preference (I guess for this issue, where the goal is some diagnostics, it might be better to use out-of-line variant to keep the class interface unchanged?)

cz4rs:
This depends on following PR in `checkpoint` library:
https://github.com/DARMA-tasking/checkpoint/pull/124

lifflander:
The corresponding PR in checkpoint: https://github.com/DARMA-tasking/checkpoint/pull/124 has now been merged!

lifflander:


```diff

Clones added
============
- src/vt/scheduler/priority_queue.h  1
- src/vt/scheduler/queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/priority_queue.h  1
- src/vt/scheduler/queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/priority_queue.h  1
- src/vt/scheduler/queue.h  1
         

Clones removed
==============
+ src/vt/worker/worker_openmp.h  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/queue.h  1
- src/vt/scheduler/priority_queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/queue.h  1
- src/vt/scheduler/priority_queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/priority_queue.h  1
- src/vt/scheduler/queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/queue.h  1
- src/vt/scheduler/priority_queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/queue.h  1
- src/vt/scheduler/priority_queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/queue.h  1
- src/vt/scheduler/priority_queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

cz4rs:
| component | sizeof | getMemoryFootprint |
| --- | --- | --- |
| test_lb_manager | 160 | 160 |
| test_stats_restart_reader | 256 | 256 |
| test_node_stats | 448 | 448 |
| test_memory_usage | 56 | 56 |
| test_rdma_manager_handle | 208 | 208 |
| test_collective_alg | 440 | 440 |
| test_obj_group_manager | 360 | 360 |
| test_pipe_manager | 176 | 176 |
| test_group_manager | 304 | 304 |
| test_collection_manager | 624 | 624 |
| test_location_manager | 96 | 96 |
| test_termination_detector | 880 | 880 |
| test_sequencer_virtual | 208 | 208 |
| test_sequencer | 152 | 152 |
| test_scheduler | 160 | 772 |
| test_registry | 152 | 152 |
| test_rdma_manager | 232 | 232 |
| test_param | 24 | 24 |
| test_active_messenger | 408 | 408 |
| test_async_event | 144 | 144 |
| test_worker_group_type | - | - |
| test_virtual_context_manager | 280 | 280 |
| test_pool | 88 | 88 |
| test_context | 32 | 32 |
| test_trace | 464 | 33347 |
| test_pmpi_component | 24 | 24 |

lifflander:


```diff

Clones added
============
- src/vt/scheduler/queue.h  1
- src/vt/scheduler/priority_queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

PhilMiller:
In the scheduler or active messenger, newly-enqueued messages should have their size added to some gauge variable to track their total footprint, and be subtracted when they're processed. That variable's value can then be added to the scheduler component's footprint when that's being calculated. If there's no method on the footprinter to add an arbitrary value to the computed footprint, that should get implemented.

lifflander:


```diff

Clones added
============
- src/vt/scheduler/priority_queue.h  1
- src/vt/scheduler/queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/queue.h  1
- src/vt/scheduler/priority_queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/queue.h  1
- src/vt/scheduler/priority_queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/priority_queue.h  1
- src/vt/scheduler/queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/priority_queue.h  1
- src/vt/scheduler/queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/queue.h  1
- src/vt/scheduler/priority_queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/priority_queue.h  1
- src/vt/scheduler/queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/priority_queue.h  1
- src/vt/scheduler/queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/priority_queue.h  1
- src/vt/scheduler/queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/priority_queue.h  1
- src/vt/scheduler/queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/priority_queue.h  1
- src/vt/scheduler/queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/priority_queue.h  1
- src/vt/scheduler/queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/priority_queue.h  1
- src/vt/scheduler/queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/priority_queue.h  1
- src/vt/scheduler/queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/queue.h  1
- src/vt/scheduler/priority_queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

cz4rs:
FIXME:
```
/vt/src/vt/topos/location/location.impl.h:645:6: warning: declaration of 'serialize' shadows a member of 'vt::location::EntityLocationCoord<vt::vrt::collection::VrtElmProxy<vt::tests::unit::TestCol<int>, vt::index::DenseIndexArray<int, 2> > >' [-Wshadow]
 void EntityLocationCoord<EntityID>::routeMsg(
      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /vt/src/vt/topos/location/location_headers.h:49:0,
                 from /vt/src/vt/vrt/context/context_vrtmanager.impl.h:53,
                 from /vt/src/vt/vrt/context/context_vrtmanager.h:216,
                 from /vt/src/vt/vrt/context/context_vrtheaders.h:51,
                 from /vt/src/vt/sequence/seq_matcher_virtual.h:55,
                 from /vt/src/vt/sequence/sequencer_virtual.h:51,
                 from /vt/src/vt/sequence/sequencer_headers.h:49,
                 from /vt/src/vt/transport.h:66,
                 from /vt/tests/unit/rdma/test_rdma_collection_handle.extended.cc:47:
/vt/src/vt/topos/location/location.h:425:8: note: shadowed declaration is here
   void serialize(Serializer& s) {
        ^~~~~~~~~
```

lifflander:


```diff

Clones added
============
- src/vt/scheduler/queue.h  1
- src/vt/scheduler/priority_queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:
@cz4rs Looks like there are more `SerializableTraits` that we will need to change in VT.

lifflander:
For backwards compatibility, maybe we should include a default argument on the traits? Or maybe that will cause more errors?

cz4rs:
> For backwards compatibility, maybe we should include a default argument on the traits? Or maybe that will cause more errors?

you mean defaulting to `checkpoint::Serializer`? that should work

lifflander:
> > For backwards compatibility, maybe we should include a default argument on the traits? Or maybe that will cause more errors?
> 
> you mean defaulting to `checkpoint::Serializer`? that should work

Yes, I'm thinking once we merge the checkpoint branch into checkpoint develop the release branch will cease to work with the latest checkpoint. So if we provide the default template argument for `checkpoint::Serializer` we will avoid that problem.

lifflander:


```diff

Clones added
============
- src/vt/scheduler/queue.h  1
- src/vt/scheduler/priority_queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

cz4rs:
> Yes, I'm thinking once we merge the checkpoint branch into checkpoint develop the release branch will cease to work with the latest checkpoint. So if we provide the default template argument for `checkpoint::Serializer` we will avoid that problem.

I will implement and test it in a minute, in the meantime - here's the list of classes that had
```
#include <checkpoint/checkpoint.h>
```
added for virtual serialization, some of these are pretty low-level so the propagation would have significant effect
- `vt::messaging::BaseMsg` (may be unnecessary, but I didn't go into detail with `ActiveMsg` / `vt_msg_serialize_mode_`)
- `vt::messaging::MsgSharedPtr<T>`
- `vt::runtime::component::BaseComponent`
- `vt::messaging::Listener`
- `vt::objgroup::dispatch::DispatchBase`
- `vt::objgroup::holder::HolderBase`
- `vt::vrt::collection::balance::LoadModel`

lifflander:


```diff

Clones added
============
- src/vt/scheduler/queue.h  1
- src/vt/scheduler/priority_queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/priority_queue.h  1
- src/vt/scheduler/queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/priority_queue.h  1
- src/vt/scheduler/queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/queue.h  1
- src/vt/scheduler/priority_queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/queue.h  1
- src/vt/scheduler/priority_queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/queue.h  1
- src/vt/scheduler/priority_queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/queue.h  1
- src/vt/scheduler/priority_queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/priority_queue.h  1
- src/vt/scheduler/queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/queue.h  1
- src/vt/scheduler/priority_queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/priority_queue.h  1
- src/vt/scheduler/queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/priority_queue.h  1
- src/vt/scheduler/queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/priority_queue.h  1
- src/vt/scheduler/queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/queue.h  1
- src/vt/scheduler/priority_queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

src/vt/vrt/collection/balance/model/raw_data.h
```diff
@@ -74,6 +77,22 @@ struct RawData : public LoadModel {
...
+      std::is_same<
+        SerializerT,
+        checkpoint::Footprinter
+      >::value
+    >
+  >
+  void serialize(SerializerT& s) {
+    s | proc_load_
+      | proc_subphase_load_
+      | proc_comm_
```
PhilMiller, line 92: Given that these pointers are generally intended to just be observers of the data actually owned by `NodeStats`, I think they should just be counted as the size of the pointers here. It's not 100% clear to me whether that's actually the case.

lifflander:


```diff

Clones added
============
- src/vt/scheduler/queue.h  1
- src/vt/scheduler/priority_queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/priority_queue.h  1
- src/vt/scheduler/queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

src/vt/vrt/collection/balance/model/raw_data.h
cz4rs: daf9bfb fixes this (`countBytes` method simply applies `sizeof` and adds the result to total footprint)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/queue.h  1
- src/vt/scheduler/priority_queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/queue.h  1
- src/vt/scheduler/priority_queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/priority_queue.h  1
- src/vt/scheduler/queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

PhilMiller:
How is this code setting up to be able to call the `serialize` methods added to the various components? Is it going to be using Checkpoint's virtual serialization, or something in the component registry that will known the dynamic type directly?

tests/unit/runtime/test_memory_footprint.cc
```diff
@@ -0,0 +1,208 @@
...
+
+using TestMemoryFootprinting = TestParallelHarness;
+
+template<typename T>
+void printMemoryFootprint(T* obj) {
+  fmt::print("sizeof(obj):\t{}\n", sizeof(*obj));
+
+  auto size = checkpoint::getMemoryFootprint<T>(*obj);
+  fmt::print("footprint(obj):\t{}\n", size);
+}
```
PhilMiller, line 85: This acts as a smoke test right now. Do we have a plan for validating the results in some fashion?

lifflander:
> How is this code setting up to be able to call the `serialize` methods added to the various components? Is it going to be using Checkpoint's virtual serialization, or something in the component registry that will known the dynamic type directly?

I think the ComponentPack should traverse the active components and ask for their footprint.

PhilMiller:
@cz4rs Just want to call your attention to a change in checkpoint, https://github.com/DARMA-tasking/checkpoint/pull/133
It simplifies the interface for the virtual class hierarchy serialization a little bit, by dropping the "what type am I?" arguments from the macros, leaving just the parent class argument in the macro for derived classes.

src/vt/configs/arguments/app_config.h
PhilMiller: I'd prefer leaving them intrusive to the classes, so that we don't end up having to make data members public unnecessarily.

lifflander:


```diff

Clones added
============
- src/vt/scheduler/priority_queue.h  1
- src/vt/scheduler/queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/priority_queue.h  1
- src/vt/scheduler/queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/queue.h  1
- src/vt/scheduler/priority_queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/queue.h  1
- src/vt/scheduler/priority_queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

cz4rs:
> @cz4rs Just want to call your attention to a change in checkpoint, [DARMA-tasking/checkpoint#133](https://github.com/DARMA-tasking/checkpoint/pull/133)
> It simplifies the interface for the virtual class hierarchy serialization a little bit, by dropping the "what type am I?" arguments from the macros, leaving just the parent class argument in the macro for derived classes.

thanks for the heads-up, a5390bc  addresses this change

btw does this mean that https://github.com/DARMA-tasking/checkpoint/issues/132 can be closed?

lifflander:


```diff

Clones added
============
- src/vt/scheduler/priority_queue.h  1
- src/vt/scheduler/queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/priority_queue.h  1
- src/vt/scheduler/queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:
> > @cz4rs Just want to call your attention to a change in checkpoint, [DARMA-tasking/checkpoint#133](https://github.com/DARMA-tasking/checkpoint/pull/133)
> > It simplifies the interface for the virtual class hierarchy serialization a little bit, by dropping the "what type am I?" arguments from the macros, leaving just the parent class argument in the macro for derived classes.
> 
> thanks for the heads-up, [a5390bc](https://github.com/DARMA-tasking/vt/commit/a5390bcfbbe9e2f9ee1e901dfc21e66a6d8087df) addresses this change
> 
> btw does this mean that [DARMA-tasking/checkpoint#132](https://github.com/DARMA-tasking/checkpoint/issues/132) can be closed?

There's a second part we are still working on to completely eliminate the need to name the class. The last PR removed one of the parameters, but not the other.

src/vt/runtime/component/base.h
```diff
@@ -45,6 +45,8 @@
 #if !defined INCLUDED_VT_RUNTIME_COMPONENT_BASE_H
 #define INCLUDED_VT_RUNTIME_COMPONENT_BASE_H
 
+#include <checkpoint/checkpoint.h>
```
cz4rs, line 48: TODO: verify if a "smaller" header can be used instead

lifflander:


```diff

Clones added
============
- src/vt/scheduler/queue.h  1
- src/vt/scheduler/priority_queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

src/vt/runtime/component/base.h
```diff
@@ -68,6 +70,8 @@ struct ComponentPack;
  * \brief The abstract \c BaseComponent for VT runtime component pack
  */
 struct BaseComponent : Diagnostic, Bufferable, Progressable {
+  checkpoint_virtual_serialize_root()
```
cz4rs, line 73: TODO: add `checkpoint_virtual_serialize_derived_from` macro for all derived classes

tests/unit/runtime/test_memory_footprint.cc
cz4rs: the simplest option I can see right now is hard-coding the values that we get at the moment, e.g. `EXPECT_EQ(trace_footprint, 434);`

cons:
- probably a pain to maintain
- not sure if the footprint will be exactly the same on macosx+ clang build

cz4rs:
> > How is this code setting up to be able to call the `serialize` methods added to the various components? Is it going to be using Checkpoint's virtual serialization, or something in the component registry that will known the dynamic type directly?
> 
> I think the ComponentPack should traverse the active components and ask for their footprint.

168e53c8410ebd15bff9f2742d28730dace2e7e0 shows how this could be done (with more / better formatted information included + virtual serialization support completed for components). This could be plugged in somewhere between the phases eventually?

lifflander:


```diff

Clones added
============
- src/vt/scheduler/priority_queue.h  1
- src/vt/scheduler/queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/priority_queue.h  1
- src/vt/scheduler/queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

cz4rs:
TODO:

- [ ] fix virtual serialization issue
- [ ] find any missing classes / members not being serialized
- [ ] improve the unit tests (only smoke tests currently)
- [ ] improve support for collections / `vt::vrt::collection::BaseHolder`

lifflander:


```diff

Clones added
============
- src/vt/scheduler/queue.h  1
- src/vt/scheduler/priority_queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/priority_queue.h  1
- src/vt/scheduler/queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/queue.h  1
- src/vt/scheduler/priority_queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/queue.h  1
- src/vt/scheduler/priority_queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/utils/memory/memory_usage.h  2
- src/vt/scheduler/queue.h  1
- src/vt/scheduler/priority_queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

src/vt/configs/arguments/app_config.h
```diff
@@ -92,7 +92,7 @@ struct AppConfig {
   bool vt_no_sigsegv   = false;
   bool vt_no_terminate = false;
   std::string vt_memory_reporters =
-# if vt_check_enabled(mimalloc)
+# if (vt_feature_mimalloc != 0)
```
cz4rs, line 95: macro expanded to avoid pulling dependencies into this header

lifflander:


```diff

Clones added
============
- src/vt/scheduler/queue.h  1
- src/vt/scheduler/priority_queue.h  1
- src/vt/utils/memory/memory_usage.h  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/priority_queue.h  1
- src/vt/scheduler/queue.h  1
- src/vt/utils/memory/memory_usage.h  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

cz4rs:
@lifflander 
`error: ld returned 1 exit status` (`gcc-7` build) strikes again:
https://github.com/DARMA-tasking/vt/pull/1013/checks?check_run_id=1275153716

also, `gcc-8` build fails without any details available (I cannot expand "Build the Docker image section"):
Screenshot from 2020-10-19 15-55-52

lifflander:


```diff

Clones added
============
- src/vt/scheduler/priority_queue.h  1
- src/vt/utils/memory/memory_usage.h  2
- src/vt/scheduler/queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/utils/memory/memory_usage.h  2
- src/vt/scheduler/priority_queue.h  1
- src/vt/scheduler/queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/utils/memory/memory_usage.h  2
- src/vt/scheduler/queue.h  1
- src/vt/scheduler/priority_queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/priority_queue.h  1
- src/vt/scheduler/queue.h  1
- src/vt/utils/memory/memory_usage.h  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

src/vt/configs/arguments/app_config.h
lifflander: I agree

lifflander:


```diff

Clones added
============
- src/vt/scheduler/priority_queue.h  1
- src/vt/utils/memory/memory_usage.h  2
- src/vt/scheduler/queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/queue.h  1
- src/vt/utils/memory/memory_usage.h  2
- src/vt/scheduler/priority_queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/utils/memory/memory_usage.h  2
- src/vt/scheduler/queue.h  1
- src/vt/scheduler/priority_queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/queue.h  1
- src/vt/utils/memory/memory_usage.h  2
- src/vt/scheduler/priority_queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/priority_queue.h  1
- src/vt/scheduler/queue.h  1
- src/vt/utils/memory/memory_usage.h  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/priority_queue.h  1
- src/vt/scheduler/queue.h  1
- src/vt/utils/memory/memory_usage.h  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/utils/memory/memory_usage.h  2
- src/vt/scheduler/queue.h  1
- src/vt/scheduler/priority_queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/priority_queue.h  1
- src/vt/scheduler/queue.h  1
- src/vt/utils/memory/memory_usage.h  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/priority_queue.h  1
- src/vt/scheduler/queue.h  1
- src/vt/utils/memory/memory_usage.h  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

src/vt/rdmahandle/manager.h
```diff
@@ -227,6 +229,21 @@ struct Manager : runtime::component::Component<Manager> {
...
+    typename = std::enable_if_t<
+      std::is_same<SerializerT, checkpoint::Footprinter>::value
+    >
+  >
+  void serialize(SerializerT& s) {
+    s | cur_handle_obj_group_
+      | cur_handle_collection_
+      | collection_to_manager_
+      | proxy_
+      // holder_
```
cz4rs, line 243: `holder_` is declared as:
```cpp
  template <typename T, HandleEnum E>
  static std::unordered_map<HandleKey, Holder<T,E>> holder_;
```

so we'd have to template `serialize` over `typename T, HandleEnum E` to call
```
s | holder_<T,E>;
```
which breaks `serialize` detection (and doesn't really solve template deduction either?)

lifflander:


```diff

Clones added
============
- src/vt/utils/memory/memory_usage.h  2
- src/vt/scheduler/priority_queue.h  1
- src/vt/scheduler/queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/scheduler/priority_queue.h  1
- src/vt/scheduler/queue.h  1
- src/vt/utils/memory/memory_usage.h  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

lifflander:


```diff

Clones added
============
- src/vt/utils/memory/memory_usage.h  2
- src/vt/scheduler/queue.h  1
- src/vt/scheduler/priority_queue.h  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6106220&bid=19987069)

src/vt/rdmahandle/manager.h
lifflander: Yes, so this is the tricky case. I think this comes up in several places in VT.

I think the solution (for places where we don't just want to skip them), is to register lambdas that type-erase (e.g., `Holder<T,E>`) and just return out the size. In this case, when something is added to `holder_`, we can have a second unordered_map from `HandleKey` to a new `std::function<std::size_t()>` that captures the `Holder<T,E>` and calls footprint and returns the size. I realize this won't all the code to perfectly recurse downward, but then we can just call that explicitly in the RDMAManager's footprinting serializer.

@cz4rs Does that make sense?
@PhilMiller What do you think of this solution?

src/vt/rdmahandle/manager.impl.h
```diff
@@ -133,6 +135,9 @@ Handle<T, E> Manager::makeHandleCollectiveObjGroup(
   auto key = HandleKey{typename HandleKey::ObjGroupTag{}, proxy_bits, next_handle};
   auto han = Handle<T, E>{typename Handle<T, E>::NodeTagType{}, key, size};
   holder_<T,E>[key].template addHandle<ObjGroupProxyType>(key, sub, han, size, uniform_size);
+  holder_footprint_[key] = [&holder = holder_<T,E>[key]]() {
```
lifflander, line 138: This looks great to me! I think this is the best way to address these types of containers.

cz4rs:
@lifflander 
- removing all virtual serialization allows the builds to pass (with tests enabled) - ab36a3f
- re-enabling it for some simple cases is enough to trigger 'collect2: error: ld returned 1 exit status' in gcc-7 build - 62ec69a

src/vt/runtime/runtime.cc
```diff
@@ -1001,8 +1003,35 @@ void Runtime::initializeWorkers(WorkerCountType const num_workers) {
...
+      component->name(),
+      checkpoint::getMemoryFootprint(*component)
+    );
+  }
+}
+
 void Runtime::printMemoryFootprint() const {
-  p_->printMemoryFootprint();
+  p_->foreach([&](component::BaseComponent* base) {
+    if (base->name() == "ArgConfig")
```
cz4rs, line 1019: This should be enough to deduce the actual types of live components without virtual serialization - not pretty and somewhat verbose, but it does the job. If it looks good, I will cover the rest of components in the same fashion.

- [ ] TODO: add the rest of components

cz4rs: also, for the very basic stats, `printMemoryFootprint` can be plugged into startup / shutdown banner - @lifflander did you have any specific place in mind? (+ should this be covered with a command line argument? something like `--vt_print_memory_footprint`?)

cz4rs:
bb08a84 removes all `checkpoint_virtual_serialize*` macros which should help with getting the builds green, but does not resolve the root issues with `gcc-7` (known problem), `gcc-6` and `gcc-8` (still not diagnosed) builds.

- this makes the footprinting less precise, which is not a problem for components themselves (majority of cases where we skip `BaseComponent : Diagnostic, Bufferable, Progressable` hierarchy), but it is necessary for cases like 62ec69a
- code coverage correctly shows a drop, because without virtual serialization some of the `serialize` methods are never visited now

cz4rs:
comparison of footprint with (febde03e6) and without virtual serialization (6b55458fe)

test used:
```shell
cmake --build ./build --config Debug --target runtime_basic -- -j 10
./tests/runtime_basic --gtest_filter=TestMemoryFootprinting.test_live_components
```

| component  | no virtual serialization | virtual serialization |
| ------------- | ------------- | ------------- |
| ArgConfig | 1200 |       1199 |
| Context |   96 | 80 |
| Registry |  208 |        128 |
| MemoryPool |        8880 |      8880 |
| Trace |     33359 |   33359 |
| AsyncEvent |        280 |      120 |
| ActiveMessenger |   856 | 375 |
| ObjGroupManager |   808 | 2525 |
| PhaseManager |      861 |    861 |
| MemoryUsage |       112 |     135 |
| PMPI |      80 |    80 |
| Scheduler | 964 |       964 |
| TerminationDetector |       936 |     491 |
| Collective |        504 |      407 |
| GroupManager |      368 |    277 |
| NodeStats | 560 |       417 |
| LBManager | 248 |       194 |
| LocationManager |   216 | 1992 |
| CollectionManager | 656 |       576 |
| PipeManager |       232 |     80 |
| RDMAManager |       288 |     201 |
| Param |     80 |   80 |
| Sequencer | 208 |       80 |
| VirtualContextManager |     336 |   246 |
| VirtualSequencer |  264 |   80 |
| HandleRDMA |        320 |      181 |
| TimeTriggerManager |        176 |      92 |

lifflander:
> comparison of footprint with ([febde03](https://github.com/DARMA-tasking/vt/commit/febde03e6fa2a8eed620d3708c57ab4a3ca74611)) and without virtual serialization ([6b55458](https://github.com/DARMA-tasking/vt/commit/6b55458fe4eb9aa4398f269fecef581e3edf1ac4))

Is this mostly due to diagnostics?

cz4rs:
quotation for the context, since rebase makes it hard to follow conversation:
> comparison of footprint with ([febde03](https://github.com/DARMA-tasking/vt/commit/febde03e6fa2a8eed620d3708c57ab4a3ca74611)) and without virtual serialization ([6b55458](https://github.com/DARMA-tasking/vt/commit/6b55458fe4eb9aa4398f269fecef581e3edf1ac4))
> 
> test used:
> 
> ```shell
> cmake --build ./build --config Debug --target runtime_basic -- -j 10
> ./tests/runtime_basic --gtest_filter=TestMemoryFootprinting.test_live_components
> ```
> 
> component 	no virtual serialization 	virtual serialization
> ArgConfig 	1200 	1199
> Context 	96 	80
> Registry 	208 	128
> MemoryPool 	8880 	8880
> Trace 	33359 	33359
> AsyncEvent 	280 	120
> ActiveMessenger 	856 	375
> ObjGroupManager 	808 	2525
> PhaseManager 	861 	861
> MemoryUsage 	112 	135
> PMPI 	80 	80
> Scheduler 	964 	964
> TerminationDetector 	936 	491
> Collective 	504 	407
> GroupManager 	368 	277
> NodeStats 	560 	417
> LBManager 	248 	194
> LocationManager 	216 	1992
> CollectionManager 	656 	576
> PipeManager 	232 	80
> RDMAManager 	288 	201
> Param 	80 	80
> Sequencer 	208 	80
> VirtualContextManager 	336 	246
> VirtualSequencer 	264 	80
> HandleRDMA 	320 	181
> TimeTriggerManager 	176 	92
> 
> (for more extensive testing, one could set `bool vt_print_memory_footprint = true;` in `app_config.h` to print the footprint by default)

> Is this mostly due to diagnostics?

Partially yes. Please note that `checkpoint::getMemoryFootprint` returns `sizeof` instead of calculated footprint if the latter is smaller:
```cpp
  return size_offset + std::max(
    dispatch::Standard::footprint<T, Footprinter>(target),
    sizeof(target)
  );
```
The values in the summary are actually smaller for some cases with virtual serialization enabled (explanation: `sizeof(BaseComponent)` is 80, `sizeof` of the actual class will usually be bigger).

Notable exceptions are:
| component | no virtual serialization | virtual serialization enabled |
| --------------- | ---------------------------- | -------------------------- |
| LocationManager |	216 |	1992 |
| ObjGroupManager |	808 |	2525 |
where we can see that following inheritance hierarchies results in much bigger footprint with virtual serialization enabled.

lifflander:
I don't know why extended tests have been run. They shouldn't be running automatically here.

cz4rs:
> I don't know why extended tests have been run. They shouldn't be running automatically here.

no idea either (I did nothing to trigger them), but actually it is good to run them at least once for changes as big as this

I have solved a bug in RDMA State footprinting and improved the coverage so `codecov` build is happy now, so if the extended tests are fine this should be ready to merge

cz4rs:
command line argument controlling printing during initialization / shutdown:
```--vt_print_memory_footprint```

sample output:
```
vt: Runtime Initializing: interop=true: mode: single-thread per rank
(...)
vt: Pass-through Arguments:
vt:     None. All arguments handled.
vt: Printing memory footprint for live components:
vt:     Memory footprint for ArgConfig: 1256
vt:     Memory footprint for Context:   152
vt:     Memory footprint for Registry:  264
vt:     Memory footprint for MemoryPool:        8936
vt:     Memory footprint for Trace:     33415
vt:     Memory footprint for AsyncEvent:        336
vt:     Memory footprint for ActiveMessenger:   912
vt:     Memory footprint for ObjGroupManager:   864
vt:     Memory footprint for PhaseManager:      917
vt:     Memory footprint for MemoryUsage:       168
vt:     Memory footprint for PMPI:      136
vt:     Memory footprint for Scheduler: 1020
vt:     Memory footprint for TerminationDetector:       992
vt:     Memory footprint for Collective:        560
vt:     Memory footprint for GroupManager:      424
vt:     Memory footprint for NodeStats: 616
vt:     Memory footprint for LBManager: 280
vt:     Memory footprint for LocationManager:   272
vt:     Memory footprint for CollectionManager: 712
vt:     Memory footprint for PipeManager:       288
vt:     Memory footprint for RDMAManager:       344
vt:     Memory footprint for Param:     136
vt:     Memory footprint for Sequencer: 264
vt:     Memory footprint for VirtualContextManager:     392
vt:     Memory footprint for VirtualSequencer:  320
vt:     Memory footprint for HandleRDMA:        376
vt:     Memory footprint for TimeTriggerManager:        232
vt: Total collective epochs processed: 0
vt: Total work units processed: 0
vt: Runtime Finalizing
```
note: this is disabled by default, you can change the default value in `app_config.h`; I have done it temporarily with 397ac50 to check all CI builds - no errors detected

File: src/vt/runtime/runtime.cc [hunk 11]
```diff
@@ -1001,6 +1007,159 @@ void Runtime::initializeWorkers(WorkerCountType const num_workers) {
...
+
+void Runtime::printMemoryFootprint() const {
+  fmt::print(
+    "{}{}Printing memory footprint for live components:{}\n",
+    debug::vtPre(),
+    debug::green(),
+    debug::reset()
+  );
+
+  p_->foreach([&](component::BaseComponent* base) {
```
cz4rs, line 1036: the casting here is somewhat verbose, but I couldn't think of anything clearly better while maintaining the simplicity :thinking: 
if this has potential to be useful elsewhere, I can extract it and refactor as needed

cz4rs:
this is ready for review (there are still some builds queued, but most of the tests are green already)

there are some random blank lines added throughout the review (leftovers after back-and-forth with virtual serialization), I will remove them along with addressing review comments

lifflander:
> command line argument controlling printing during initialization / shutdown:
> `--vt_print_memory_footprint`
> 
> sample output:
> 
> ```
> vt: Runtime Initializing: interop=true: mode: single-thread per rank
> (...)
> vt: Pass-through Arguments:
> vt:     None. All arguments handled.
> vt: Printing memory footprint for live components:
> vt:     Memory footprint for ArgConfig: 1256
> vt:     Memory footprint for Context:   152
> vt:     Memory footprint for Registry:  264
> vt:     Memory footprint for MemoryPool:        8936
> vt:     Memory footprint for Trace:     33415
> vt:     Memory footprint for AsyncEvent:        336
> vt:     Memory footprint for ActiveMessenger:   912
> vt:     Memory footprint for ObjGroupManager:   864
> vt:     Memory footprint for PhaseManager:      917
> vt:     Memory footprint for MemoryUsage:       168
> vt:     Memory footprint for PMPI:      136
> vt:     Memory footprint for Scheduler: 1020
> vt:     Memory footprint for TerminationDetector:       992
> vt:     Memory footprint for Collective:        560
> vt:     Memory footprint for GroupManager:      424
> vt:     Memory footprint for NodeStats: 616
> vt:     Memory footprint for LBManager: 280
> vt:     Memory footprint for LocationManager:   272
> vt:     Memory footprint for CollectionManager: 712
> vt:     Memory footprint for PipeManager:       288
> vt:     Memory footprint for RDMAManager:       344
> vt:     Memory footprint for Param:     136
> vt:     Memory footprint for Sequencer: 264
> vt:     Memory footprint for VirtualContextManager:     392
> vt:     Memory footprint for VirtualSequencer:  320
> vt:     Memory footprint for HandleRDMA:        376
> vt:     Memory footprint for TimeTriggerManager:        232
> vt: Total collective epochs processed: 0
> vt: Total work units processed: 0
> vt: Runtime Finalizing
> ```
> 
> note: this is disabled by default, you can change the default value in `app_config.h`; I have done it temporarily with [397ac50](https://github.com/DARMA-tasking/vt/commit/397ac50f816baa666769063f6295e06feb7fa216) to check all CI builds - no errors detected

Can you align the memory usage in the table with fmt? Also, can you call `std::tuple<std::string, double> getBestMemoryUnit(std::size_t bytes)` once #1113 is merged? That will get the best memory unit for the output.

cz4rs:
> Can you align the memory usage in the table with fmt? Also, can you call `std::tuple<std::string, double> getBestMemoryUnit(std::size_t bytes)` once #1113 is merged? That will get the best memory unit for the output.

:+1: for `getBestMemoryUnit`
regarding alignment - did you mean `libfort`?

lifflander:
> > Can you align the memory usage in the table with fmt? Also, can you call `std::tuple<std::string, double> getBestMemoryUnit(std::size_t bytes)` once #1113 is merged? That will get the best memory unit for the output.
> 
> 👍 for `getBestMemoryUnit`
> regarding alignment - did you mean `libfort`?

No, you can align with fmt. https://fmt.dev/latest/syntax.html#format-specification-mini-language

lifflander:
You can look at vt/src/vt/configs/error/pretty_print_message.cc for an example of alignment if you want.

examples/rdma/rdma_simple_get.cc
```diff
@@ -93,6 +93,7 @@ static vt::RDMA_GetType test_get_fn(
 
 int main(int argc, char** argv) {
   vt::initialize(argc, argv);
+  vt::rt->printMemoryFootprint();
```
lifflander, line 96: I think these probably weren't intended to be left here in the examples.

src/vt/configs/arguments/app_config.h
lifflander: Maybe in the future we can isolate that header into a single file. It's a pretty simple macro.

src/vt/context/context.h
```diff
@@ -143,6 +143,15 @@ struct Context : runtime::component::Component<Context> {
 
   std::string name() override { return "Context"; }
 
+  template <typename SerializerT>
+  void serialize(SerializerT& s) {
+    s | thisNode_
+      | numNodes_
+      | numWorkers_;
+
+    s.countBytes(communicator_);
```
lifflander, line 152: I'm not exactly sure what that MPI standard says specifically, but `MPI_Comm` might not be a cleanly byte-countable entity. I'm not worried about missing some memory counting...just that this overload works for everything that might be passed to this function depending on the MPI implementation.

src/vt/messaging/message/smart_ptr.h
```diff
@@ -215,6 +215,11 @@ struct MsgSharedPtr final {
               << ")";
   }
 
+  template <typename Serializer>
+  void serialize(Serializer& s) {
```
lifflander, line 219: Shouldn't this have an enable_if on whether the underlying message type is serializable and if so, call it's serializer for footprinting instead if just counting it as bytes?

src/vt/objgroup/manager.h
```diff
@@ -362,6 +363,17 @@ struct ObjGroupManager : runtime::component::Component<ObjGroupManager> {
     MsgSharedPtr<ShortMessage> msg, HandlerType han, EpochType ep
   );
 
+  template <
```
lifflander, line 366: This formatting is a little odd.

src/vt/runtime/runtime.cc
lifflander: I think this is fine for now

src/vt/runtime/runtime.cc
```diff
@@ -1001,6 +1007,159 @@ void Runtime::initializeWorkers(WorkerCountType const num_workers) {
...
+        static_cast<trace::Trace*>(base)
+      );
+    #endif
+    #if vt_check_enabled(mpi_access_guards)
+    } else if (name == "PMPI") {
+      printComponentFootprint(
+        static_cast<pmpi::PMPIComponent*>(base)
+      );
+    #endif
+    }
```
lifflander, line 1159: Maybe you should add an else statement and print a warning so if a component is added when it prints out the user will know that they missed something.

tests/unit/runtime/test_memory_footprint.cc
lifflander: Yes, this would be tricky to validate the results. Maybe, in the future, once we have the checkpoint analyzer running it can help.

lifflander (approved):
Overall, this looks great! I have a few minor points to be considered. The largest potential issue I see is not calling the message serializer to get the footprint of the message if the message is serializable.

src/vt/context/context.h
cz4rs: Yep, this was exemplified with errors in our openmpi build (when trying to serialize with `|` operator). 
This should be fine since all `countBytes` does is to apply `sizeof`:
https://github.com/DARMA-tasking/checkpoint/blob/8a3fea5cb56f92e815458c83b0b634e19e68c151/src/checkpoint/serializers/footprinter.h#L63-L66
(so in the end I have used it as a silver bullet for most of the problematic cases)

examples/rdma/rdma_simple_get.cc
cz4rs: removed

src/vt/objgroup/manager.h
cz4rs: fixed

src/vt/runtime/runtime.cc
cz4rs: warning added:
```
vt:     Warning: memory footprinting is not supported for SomeComponentName!
```

cz4rs:
> Can you align the memory usage in the table with fmt? Also, can you call `std::tuple<std::string, double> getBestMemoryUnit(std::size_t bytes)` once #1113 is merged? That will get the best memory unit for the output.

aligned and adapted to use `getBestMemoryUnit` (decimal precision limited to 1)
```
vt: Runtime Initializing: interop=true: mode: single-thread per rank
(...)
vt: Printing memory footprint for live components:
vt:     ArgConfig                  1.2 KiB
vt:     Context                  152.0 B
vt:     Registry                 264.0 B
vt:     MemoryPool                 8.7 KiB
vt:     Trace                     32.6 KiB
vt:     AsyncEvent               336.0 B
vt:     ActiveMessenger          912.0 B
vt:     ObjGroupManager          864.0 B
vt:     PhaseManager             917.0 B
vt:     MemoryUsage              168.0 B
vt:     PMPI                     136.0 B
vt:     Scheduler               1020.0 B
vt:     TerminationDetector      992.0 B
vt:     Collective               560.0 B
vt:     GroupManager             424.0 B
vt:     NodeStats                616.0 B
vt:     LBManager                280.0 B
vt:     LocationManager          272.0 B
vt:     CollectionManager        712.0 B
vt:     PipeManager              288.0 B
vt:     RDMAManager              344.0 B
vt:     Param                    136.0 B
vt:     Sequencer                264.0 B
vt:     VirtualContextManager    392.0 B
vt:     VirtualSequencer         320.0 B
vt:     HandleRDMA               376.0 B
vt:     TimeTriggerManager       232.0 B
vt: Total collective epochs processed: 0
vt: Total work units processed: 0
vt: Runtime Finalizing
```

src/vt/messaging/message/smart_ptr.h
cz4rs: 4f6a860 attempts to fix this using existing code (`vt::serialization::MsgSizer`), but I'm not sure if this is correct (especially if there are any consequences of using `get` - in regards to ownership)

cz4rs:
@lifflander 
all the minor issues (formatting, printing...) are solved, regarding serialization of the messages - I have made a separate comment under `smaprt_ptr.h`

current output sample:
```
vt: Runtime Initializing: interop=true: mode: single-thread per rank
(...)
vt: Pass-through Arguments:
vt:     None. All arguments handled.
vt: Printing memory footprint for live components:
vt:     ArgConfig                  1.2 KiB
vt:     Context                  152.0 B
vt:     Registry                 264.0 B
vt:     MemoryPool                 8.7 KiB
vt:     Trace                     32.6 KiB
vt:     AsyncEvent               336.0 B
vt:     ActiveMessenger          912.0 B
vt:     ObjGroupManager          864.0 B
vt:     PhaseManager             917.0 B
vt:     MemoryUsage              168.0 B
vt:     PMPI                     136.0 B
(...)
vt: Runtime Finalizing
```

src/vt/event/event.h
```diff
@@ -187,6 +187,19 @@ struct AsyncEvent : runtime::component::PollableComponent<AsyncEvent> {
...
+  void serialize(SerializerT& s) {
+    s | cur_event_
+      | event_container_
+      | polling_event_container_;
+
+    s.countBytes(lookup_container_);
+
+  # if vt_check_enabled(trace_enabled)
+    s | trace_event_polling;
+  # endif
```
PhilMiller, line 200: Should the diagnostics `eventPollCount` `eventSizeGauge`, and `mpiEventWaitTime` be counted?

src/vt/event/event.h
```diff
@@ -187,6 +187,19 @@ struct AsyncEvent : runtime::component::PollableComponent<AsyncEvent> {
 
   std::string name() override { return "AsyncEvent"; }
 
+  template <typename SerializerT>
+  void serialize(SerializerT& s) {
+    s | cur_event_
+      | event_container_
+      | polling_event_container_;
+
+    s.countBytes(lookup_container_);
```
PhilMiller, line 196: I think this would be footprintable?

src/vt/event/event_record.h
```diff
@@ -105,6 +110,15 @@ struct EventRecord {
...
 # endif
 
+  template <typename Serializer>
+  void serialize(Serializer& s) {
+    s | ready
+      | msg_
+      | event_union_
+      | event_id_
+      | type_;
+  }
```
PhilMiller, line 120: Should this conditionally cover `creation_time_stamp_`?

src/vt/group/group_manager.h
```diff
@@ -371,6 +371,20 @@ struct GroupManager : runtime::component::Component<GroupManager> {
    */
   RemoteOperationIDType getNextOpID();
 
+  template <typename SerializerT>
+  void serialize(SerializerT& s) {
+    s | next_group_id_
```
PhilMiller, line 376: `s.countBytes(default_comm_);` perhaps?

src/vt/messaging/active.h
```diff
@@ -207,7 +226,17 @@ struct InProgressDataIRecv : InProgressBase {
...
-public:
+  template <typename Serializer>
+  void serialize(Serializer& s) {
+    s | user_buf
+      | dealloc_user_buf
+      | next
+      | priority
+      | cur;
+
+    s.countBytes(reqs);
```
PhilMiller, line 237: Seeing this, would it make sense to add a footprint-specific serializer for `std::vector` that at least counts up `sizeof(T)*v.size()` if `T` itself isn't serializable?

src/vt/event/event.h
cz4rs: :+1: these were added "in the meantime" and I haven't noticed them in the review

src/vt/scheduler/scheduler.h
```diff
@@ -264,6 +264,24 @@ struct Scheduler : runtime::component::Component<Scheduler> {
...
+      | num_term_msgs_
+      | event_triggers
+      | event_triggers_once
+      | last_progress_time_
+      | progress_time_enabled_
+      | processed_after_last_progress_
+      | last_threshold_memory_usage_
+      | threshold_memory_usage_
+      | last_memory_usage_poll_;
+  }
```
PhilMiller, line 283: As noted in another component, the diagnostic members are omitted here.

src/vt/utils/mutex/omp_mutex.h
```diff
@@ -64,6 +64,11 @@ struct OMPMutex {
   void lock();
   void unlock();
   bool try_lock();
+
+  template <typename Serializer>
+  void serialize(Serializer& s) {
+    s | omp_lock;
```
PhilMiller, line 70: Since this isn't going to be a fully-working serializer, it should be footprinting-only

src/vt/vrt/collection/balance/lb_invoke/lb_manager.h
```diff
@@ -164,6 +164,16 @@ struct LBManager : runtime::component::Component<LBManager> {
...
 
+  template <typename SerializerT>
+  void serialize(SerializerT& s) {
+    s | cached_phase_
+      | cached_lb_
+      | destroy_lb_
+      | proxy_
+      | base_model_
+      | model_;
+  }
```
PhilMiller, line 175: Missing `lb_instances_`

src/vt/vrt/collection/balance/node_stats.h
```diff
@@ -243,6 +243,21 @@ struct NodeStats : runtime::component::Component<NodeStats> {
...
+  template <typename SerializerT>
+  void serialize(SerializerT& s) {
+    s | proxy_
+      | node_data_
+      | node_subphase_data_
+      | node_migrate_
+      | node_temp_to_perm_
+      | node_perm_to_temp_
+      | node_collection_lookup_
+      | node_comm_
```
PhilMiller, line 255: Need to add newly-added `node_subphase_comm_`

cz4rs:
note to self: `Counter`, `Gauge` and `Timer` should probably footprint their base classes (instead of just members)

lifflander (approved):
Looks ready to merge once you address Phil's comments.

src/vt/event/event.h
cz4rs: the values here are `std::list` iterators (precisely `std::list<EventHolderType>::iterator`), so this could be similar to the case in `active.h` (a container where T isn't serializable)

src/vt/runtime/component/meter/counter.h
```diff
@@ -99,6 +99,11 @@ struct Counter : DiagnosticMeter {
 #   endif
   }
 
+  template <typename SerializerT>
+  void serialize(SerializerT& s) {
+    s | impl_;
+  }
+
 private:
   detail::DiagnosticValue<T>* impl_ = nullptr; /**< The actual underlying value */
```
PhilMiller, line 108: @lifflander Should this perhaps be `unique_ptr` rather than a raw pointer? That would make the serializer more generally valid, and avoid leaking on destruction. I do see that `impl_` gets passed in as an argument, so maybe there's shared ownership in the picture, which I haven't looked back at yet.

src/vt/event/event.h
PhilMiller: So, suppose there are an excessive number of entries in `lookup_container_` (which, IIRC, was a thing that actually happened as a result of a bug). We definitely want to see the footprint of the container growing with its size, even if we can't capture everything it might reach recursively.

PhilMiller: Yeah, this is the same case as my comment in active.h

PhilMiller: I think the general thing to do would be to add specializations of the serializers for containers whose elements are not themselves serializable. It can do `s.countBytes(container); s.addBytes(container.size() * sizeof(T));`

tests/unit/runtime/test_memory_footprint.cc
PhilMiller: One thing we could do is pick something that we can intentionally enlarge, and make sure that the footprint grows with each enlargement. Though in cases where the growth is in a container whose capacity may be greater than its size, really, the footprint should just be non-decreasing, and at least `sizeof(T)*num_elements_inserted`

PhilMiller: That seems like something more for a unit test of the container footprinting serializers in checkpoint itself, though.

src/vt/event/event.h
PhilMiller: Since that'll presumably be a PR against checkpoint, please add me as a reviewer, and I'll try to get to it ASAP.

cz4rs: Ok, I will create an issue in checkpoint and try to implement this approach. I'm not sure yet how to detect such cases (`not is_bytecopyable and not has_serialize_function` comes to mind), but the test cases are pretty clear, so I'll see what works.

PhilMiller: If I had to come up with a distinct type trait for `is_footprintable`, it would be something like

```
template <typename T>
using is_footprintable = 
  is_same_v<size_t, 
            decltype(checkpoint::getMemoryFootprint(std::declval<T>()))
  >
```

cz4rs: note: using non-general approach (specializing for explicit types) will not work for some cases

for example `std::vector<MPI_Request>` is a vector if `int`s when using `mpich`, which would cause conflict at dispatch stage

cz4rs:
note to self: switch `checkpoint` branch back to develop before merging (revert 94d7528)

cz4rs:
this is passing the tests just fine, but it needs to have a292ff482 reverted (pointing `checkpoint_rev` back to develop) before merging

src/vt/topos/location/cache/cache.h
```diff
@@ -75,6 +75,14 @@ struct LocationCache {
   void insert(KeyT const& key, ValueT const& value);
   void printCache() const;
 
+  template <typename Serializer>
+  void serialize(Serializer& s) {
+    s | cache_
+      | max_size_;
+
+    s.countBytes(lookup_);
```
PhilMiller, line 83: This is one we want to change back to actually traversing, now that checkpoint knows how.

PhilMiller: Specifically, it's a structure that could grow arbitrarily large, and we want to see if that's happening.

src/vt/trace/trace_log.h
```diff
@@ -170,6 +189,11 @@ struct Log final {
       reset();
     }
 
+  template <typename Serializer>
+  void serialize(Serializer& s) {
+    s.countBytes(*this);
+  }
+
```
PhilMiller, line 196: I was looking at `countBytes` call sites in light of the updated checkpoint, and noticed that this seems to not be indented the same as its surroundings.

PhilMiller (approved):
Other than the location lookup cache, this all looks good to me.

src/vt/topos/location/cache/cache.h
cz4rs: done

src/vt/trace/trace_log.h
cz4rs: indent fixed, I took a look at the call sites as well and switched to operator `|` when it was possible

the remaining 7 occurences are unions and RDMA `State` internals

PhilMiller (approved):

Issue #868, opened 2020-06-17
Remove distinction between temporary and permanent object IDs in load balancing


PhilMiller:
**What Needs to be Done?**

**Is your feature request related to a problem? Please describe.**

**Describe potential solution outcome**

**Describe alternatives you've considered**

**Additional context**

PhilMiller:
Ideally, the IDs would be bijective with the pair {collection ID, index}

PhilMiller:
@jstrzebonski I'll get more details filled in on this for you later today.

PhilMiller:
I worked through part of the implementation #1017, and ran into the somewhat expected barrier over the use of the ability to extract current node assignments for each object in `objGetNode`.

The coupled bits of code that run into trouble with this are `BaseLB, GreedyLB, ZoltanLB, LBCommKey, and CollectionManager::recordStats`

PhilMiller:
I think some of those usages could be done away with, or have an alternative data source that would give the relevant current node for the object in question.

nlslatt:
My branch for removing temporary IDs is `868-remove-temporary-ids`.  I'll try to look at this a bit more on Monday before I'm out.

nlslatt:
@jliffla I'm ready to hand this off.  What I've pushed on `868-remove-temporary-ids` almost builds.  I'm not sure I understand `test_model_comm_overhead.nompi.cc` well enough not to break it. I also haven't attempted to update `ZoltanLB` because you said you were the best person to tackle that.

I left the unique ID part of the struct the same as the old permanent ID, which had bits encoding the rank.  There shouldn't be any logic remaining that depends on those bits, so any alternative means of making the IDs unique across ranks should work fine.

nlslatt:
@lifflander I made a comment to you above but messed up your name, so I'm not sure if you were notified.

PR #1017, opened 2020-09-01
#868: Use Permanent IDs everywhere in LB

PhilMiller:


src/vt/vrt/collection/balance/node_stats.h
```diff
@@ -111,7 +111,7 @@ struct NodeStats : runtime::component::Component<NodeStats> {
    *
    * \return the temporary ID for the object assigned for this phase
```
pnstickne, line 112: Comment outdated.

PhilMiller:
Nicole addressed this in a different PR

PR #1022, opened 2020-09-02
1021  MaybeSerialize wrapper

### LINKED ISSUE #1021: Implement safe maybe-serialize ADT
**What Needs to be Done?**

Implement a new data structure uses `sizeof(T)` to determine if it is byte-packed inline with no serialize (based on some size limit) or is serialized when the type exceeds a certain threshold.

This is the first step in #833 and also is needed for other refactoring for sending messages.

lifflander:
Fixes #1021

The new wrapper class takes a fixed size and depending on the size of
T decides whether to serialize it or store it inline.

I still need to implement a new base type, MaybeSerializeBase<k>,
which will elide the type so that classes like CollectionMessage do
not need ColT or IndexT. Once that is done, we can register the
collection/index type so we can pick it up on the receive side and
dispatch to the collection.

### DIFF
--- a/src/vt/utils/adt/maybe_serialize.h
+++ b/src/vt/utils/adt/maybe_serialize.h
@@ -1,0 +1,254 @@
+/*
+//@HEADER
+// *****************************************************************************
+//
+//                              maybe_serialize.h
+//                           DARMA Toolkit v. 1.0.0
+//                       DARMA/vt => Virtual Transport
+//
+// Copyright 2019 National Technology & Engineering Solutions of Sandia, LLC
+// (NTESS). Under the terms of Contract DE-NA0003525 with NTESS, the U.S.
+// Government retains certain rights in this software.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+// * Redistributions of source code must retain the above copyright notice,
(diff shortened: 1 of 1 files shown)


lifflander:


```diff

Clones added
============
- src/vt/utils/adt/maybe_serialize.h  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6113839&bid=20009134)

PhilMiller:
Why `shared_ptr` rather than `unique_ptr`?

pnstickne:
Isn't the need for serialization based on the types in question, regardless of size?

lifflander:
> Isn't the need for serialization based on the types in question, regardless of size?

This is for byte-copyable types that need to be embedded in fixed-size structures. Like collection message, which maybe templated over different indices, that conform to the index trait but don't have size restrictions in general. However, to de-template it, we need to be able to always store some indices based on some chosen size. This allows us to say if we are under some size we just copy in the buffer vs. serialize keeping the collection message type the same.

lifflander:
> Why `shared_ptr` rather than `unique_ptr`?

The structure is copyable so the underlying pointer needs to also be copyable. I could make it move-only but it might be more difficult to work with.

PR #1027, opened 2020-09-07
1002 Add option to disable signal handlers in vt::finalize

### LINKED ISSUE #1002: Disable signal handlers when finalized
**What Needs to be Done?**

Instead of keeping these active, disable the signal handlers when VT is finalized.

JacobDomagala:
Fixes #1002

### DIFF
--- a/src/vt/runtime/runtime.cc
+++ b/src/vt/runtime/runtime.cc
@@ -391,6 +391,14 @@ bool Runtime::initialize(bool const force_now) {
       }
     }
     setup();
+
+    // Runtime component is being re-initialized
+    // Setup signal handlers based on AppConfig
+    if (finalized_) {
+      setupSignalHandler();
+      setupSignalHandlerINT();
+    }
+
     sync();
     initialized_ = true;
     return true;
--- a/src/vt/runtime/runtime.cc
+++ b/src/vt/runtime/runtime.cc
@@ -394,9 +394,10 @@ bool Runtime::initialize(bool const force_now) {
 
     // Runtime component is being re-initialized
     // Setup signal handlers based on AppConfig
-    if (finalized_) {
+    if (finalized_ && sig_handlers_disabled_) {
       setupSignalHandler();
       setupSignalHandlerINT();
+      sig_handlers_disabled_ = false;
     }
 
     sync();
@@ -444,6 +445,7 @@ bool Runtime::finalize(bool const force_now, bool const disable_sig) {
       signal(SIGSEGV, SIG_DFL);
       signal(SIGUSR1, SIG_DFL);
       signal(SIGINT, SIG_DFL);
+      sig_handlers_disabled_ = true;
     }
 
     finalized_ = true;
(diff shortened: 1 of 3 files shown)


JacobDomagala:
Worth Noting: 
Signal handlers are setup in Runtime constructor, so calling vt::finalize(disable_sig = true) and then vt::initialize() will not re-enable signal handlers.

pnstickne:
1. What if the option has to be controlled via the normal Args? And perhaps defaults to true..; or
2. Handlers are always disabled on the finalize (and always re-added if specified) in initialize?

Is there a case that code might wish to only sometimes disable on finalize?

JacobDomagala:
Adding an option to re-enable handlers in initalize() is a good idea i think

JacobDomagala:
Ok so i added the flag which will indicate whether the signal handlers were disabled on last `vt::finalize()` call. This flag is now used inside `vt::initialize()` together with `finalized_` field which indicates if `Runtime` component is being re-initialized.

Based on the comment on top of the _runtime.cc_ file i assume that contents of `arg_config_` variable will be valid during `vt::initialize()` call (`setupSignalHandler()` and `setupSignalHandlerINT()` use  `arg_config_` to determine which handlers to setup)

JacobDomagala:
@PhilMiller @lifflander  Should we also disable termination handler?

src/vt/runtime/runtime.h
```diff
@@ -252,10 +253,11 @@ struct Runtime {
 
   /**
    * \internal \brief Try to finalize
+   * \param[in] disable_sig whether to disable signal handlers
    *
    * \return whether it succeeded
    */
-  bool tryFinalize();
+  bool tryFinalize(bool const disable_sig);
```
cz4rs, line 260: it should default to `true` if we want to keep it consistent with `finalize`

JacobDomagala: Updated.
I didn't use default arg here as `tryFinalize` is used only internally by `finalize` with its `disable_sig` param.

cz4rs (approved):
the code looks good (I don't know if there's anything missing from the functional point of view?)

src/vt/runtime/runtime.cc
```diff
@@ -391,6 +391,15 @@ bool Runtime::initialize(bool const force_now) {
       }
     }
     setup();
+
+    // Runtime component is being re-initialized
+    // Setup signal handlers based on AppConfig
```
pnstickne, line 396: No `AppConfig` check here..?

src/vt/runtime/runtime.cc
```diff
@@ -431,10 +440,18 @@ bool Runtime::finalize(bool const force_now) {
     }
 
     finalizeMPI();
+
+    if (disable_sig) {
+      signal(SIGSEGV, SIG_DFL);
+      signal(SIGUSR1, SIG_DFL);
+      signal(SIGINT, SIG_DFL);
+      sig_handlers_disabled_ = true;
```
pnstickne, line 448: Could also be in the positive case: `sig_handlers_attached_ = false;`.

lifflander:
@JacobDomagala Can you start your branch names with just the issue number, not a `#`? Thanks!

lifflander (approved):
Looks good to me

PR #1029, opened 2020-09-08
#1014: dump node stats before trimming

### LINKED ISSUE #1014: Fix NodeStats LB stat file output
**What Needs to be Done?**

Currently, after trimming NodeStats the LB mapping/stats file is no longer being output correctly.

jstrzebonski:
Fixes #1014

### DIFF
--- a/src/vt/vrt/collection/balance/lb_invoke/lb_manager.cc
+++ b/src/vt/vrt/collection/balance/lb_invoke/lb_manager.cc
@@ -296,6 +296,8 @@ void LBManager::releaseNow(PhaseType phase) {
     );
   }
 
+  theNodeStats()->outputStatsForPhase(phase);
+
   auto msg = makeMessage<CollectionPhaseMsg>();
 
   // Destruct the objgroup that was used for LB
--- a/src/vt/vrt/collection/balance/node_stats.cc
+++ b/src/vt/vrt/collection/balance/node_stats.cc
@@ -170,16 +170,24 @@ void NodeStats::releaseLB() {
   CollectionManager::releaseLBPhase(msg_hold.get());
 }
 
+void NodeStats::initialize() {
+#if vt_check_enabled(lblite)
+  if (theConfig()->vt_lb_stats) {
+    theNodeStats()->createStatsFile();
+  }
+#endif
+}
+
 void NodeStats::createStatsFile() {
   auto const node = theContext()->getNode();
-  auto const base_file = std::string(theConfig()->vt_lb_stats_file);
-  auto const dir = std::string(theConfig()->vt_lb_stats_dir);
+  auto const base_file = theConfig()->vt_lb_stats_file;
+  auto const dir = theConfig()->vt_lb_stats_dir;
   auto const file = fmt::format("{}.{}.out", base_file, node);
   auto const file_name = fmt::format("{}/{}", dir, file);
 
   vt_debug_print(
     lb, node,
-    "NodeStats: createStatsFile file={}\n", file_name
+    "NodeStats::createStatsFile: file={}\n", file_name
   );
 
   // Node 0 creates the directory
@@ -198,83 +206,91 @@ void NodeStats::createStatsFile() {
   }
 
   stats_file_ = fopen(file_name.c_str(), "w+");
+  vtAssertExpr(stats_file_ != nullptr);
+}
+
+void NodeStats::finalize() {
+  // If statistics are enabled, close output file and clear stats
+#if vt_check_enabled(lblite)
+  if (theConfig()->vt_lb_stats) {
+    closeStatsFile();
+    clearStats();
+  }
+#endif
 }
 
 void NodeStats::closeStatsFile() {
   if (stats_file_) {
     fclose(stats_file_);
-    stats_file_  = nullptr;
+    stats_file_ = nullptr;
   }
 }
 
-void NodeStats::outputStatsFile() {
-  if (stats_file_ == nullptr) {
-    createStatsFile();
+std::pair<ElementIDType, ElementIDType>
+getRecvSendDirection(CommKeyType const& comm) {
+  switch (comm.cat_) {
+  case CommCategory::SendRecv:
+  case CommCategory::Broadcast:
+    return std::make_pair(comm.toObj(), comm.fromObj());
+
+  case CommCategory::NodeToCollection:
+  case CommCategory::NodeToCollectionBcast:
+    return std::make_pair(comm.toObj(), comm.fromNode());
+
+  case CommCategory::CollectionToNode:
+  case CommCategory::CollectionToNodeBcast:
+    return std::make_pair(comm.toNode(), comm.fromObj());
+  }
+
+  vtAssert(false, "Invalid balance::CommCategory enum value");
+  return std::make_pair(ElementIDType{}, ElementIDType{});
+}
+
+void NodeStats::outputStatsForPhase(PhaseType phase) {
+  // Statistics output when LB is enabled and appropriate flag is enabled
+  if (!theConfig()->vt_lb_stats) {
+    return;
   }
 
   vtAssertExpr(stats_file_ != nullptr);
 
-  auto const num_iters = node_data_.size();
+  vt_print(lb, "NodeStats::outputStatsForPhase: phase={}\n", phase);
 
-  vt_print(lb, "NodeStats::outputStatsFile: file={}, iter={}\n", print_ptr(stats_file_), num_iters);
+  for (auto&& elm : node_data_.at(phase)) {
+    ElementIDType id = elm.first;
+    TimeType time = elm.second;
+    const auto& subphase_times = node_subphase_data_.at(phase)[id];
+    size_t subphases = subphase_times.size();
 
-  for (size_t i = 0; i < num_iters; i++) {
-    for (auto&& elm : node_data_.at(i)) {
-      ElementIDType id = elm.first;
-      TimeType time = elm.second;
-      const auto& subphase_times = node_subphase_data_.at(i)[id];
-      size_t subphases = subphase_times.size();
+    auto obj_str = fmt::format("{},{},{},{},[", phase, id, time, subphases);
 
-      auto obj_str = fmt::format("{},{},{},{},[", i, id, time, subphases);
-      for (size_t s = 0; s < subphases; s++) {
-        obj_str += std::to_string(subphase_times[s]);
-        if (s != subphases - 1)
-          obj_str += ",";
+    for (size_t s = 0; s < subphases; s++) {
+      if (s > 0) {
+        obj_str += ",";
       }
 
-      obj_str += "]\n";
-
-      fprintf(stats_file_, "%s", obj_str.c_str());
+      obj_str += std::to_string(subphase_times[s]);
     }
-    for (auto&& elm : node_comm_.at(i)) {
-      using E = typename std::underlying_type<CommCategory>::type;
 
-      auto const& key = elm.first;
-      auto const& val = elm.second;
-      auto const cat = static_cast<E>(key.cat_);
+    obj_str += "]\n";
 
-      if (
-        key.cat_ == CommCategory::SendRecv or
-        key.cat_ == CommCategory::Broadcast
-      ) {
-        auto const to   = key.toObj();
-        auto const from = key.fromObj();
-        auto obj_str = fmt::format("{},{},{},{},{}\n", i, to, from, val.bytes, cat);
-        fprintf(stats_file_, "%s", obj_str.c_str());
-      } else if (
-        key.cat_ == CommCategory::NodeToCollection or
-        key.cat_ == CommCategory::NodeToCollectionBcast
-      ) {
-        auto const to   = key.toObj();
-        auto const from = key.fromNode();
-        auto obj_str = fmt::format("{},{},{},{},{}\n", i, to, from, val.bytes, cat);
-        fprintf(stats_file_, "%s", obj_str.c_str());
-      } else if (
-        key.cat_ == CommCategory::CollectionToNode or
-        key.cat_ == CommCategory::CollectionToNodeBcast
-      ) {
-        auto const to   = key.toNode();
-        auto const from = key.fromObj();
-        auto obj_str = fmt::format("{},{},{},{},{}\n", i, to, from, val.bytes, cat);
-        fprintf(stats_file_, "%s", obj_str.c_str());
-      } else {
-        vtAssert(false, "Invalid balance::CommCategory enum value");
-      }
-    }
+    fprintf(stats_file_, "%s", obj_str.c_str());
   }
-  fflush(stats_file_);
 
-  closeStatsFile();
+  for (auto&& elm : node_comm_.at(phase)) {
+    using E = typename std::underlying_type<CommCategory>::type;
+
+    auto const& comm = elm.first;
+    auto const recvSend = getRecvSendDirection(comm);
+    auto const cat = static_cast<E>(comm.cat_);
+    auto obj_str = fmt::format(
+      "{},{},{},{},{}\n", phase, recvSend.first, recvSend.second,
+      elm.second.bytes, cat
+    );
+    fprintf(stats_file_, "%s", obj_str.c_str());
+  }
+
+  fflush(stats_file_);
 }
 
 ElementIDType NodeStats::addNodeStats(
--- a/src/vt/vrt/collection/balance/node_stats.h
+++ b/src/vt/vrt/collection/balance/node_stats.h
@@ -133,7 +133,7 @@ public:
   void releaseLB();
 
   /**
-   * \internal \brief Output stats file based on instrumented data
+   * \internal \brief Output stats file for given phase based on instrumented data
    *
    * The contents of the file consist of a series of records separated
    * by newlines. Each record consists of comma separated fields. The
@@ -161,7 +161,7 @@ public:
    * recipient and distinguishing point-to-point messages from
    * broadcasts, as a decimal integer.
    */
-  void outputStatsFile();
+  void outputStatsForPhase(PhaseType phase);
 
   /**
    * \internal \brief Generate the next object element ID for LB
@@ -236,6 +236,9 @@ public:
    * otherwise \c no_vrt_proxy
    */
   VirtualProxyType getCollectionProxyForElement(ElementIDType temp_id) const;
+
+  void initialize() override;
+  void finalize() override;
 
 private:
   /**
--- a/tests/unit/test_parallel_harness.h
+++ b/tests/unit/test_parallel_harness.h
@@ -70,6 +70,7 @@ struct MPISingletonMultiTest {
     comm_ = MPI_COMM_WORLD;
     MPI_Barrier(comm_);
   }
+
   virtual ~MPISingletonMultiTest() {
     MPI_Barrier(comm_);
     MPI_Finalize();
@@ -88,23 +89,26 @@ struct TestParallelHarnessAny : TestHarnessAny<TestBase> {
 
     TestHarnessAny<TestBase>::SetUp();
 
+    // TODO (STRZ) - do we want to use additional args here as well?
     if (mpi_singleton == nullptr) {
-      mpi_singleton = std::make_unique<MPISingletonMultiTest>(
-        test_argc, test_argv
-      );
+      mpi_singleton =
+        std::make_unique<MPISingletonMultiTest>(test_argc, test_argv);
     }
 
-    CollectiveOps::initialize(test_argc, test_argv, no_workers, true);
+    auto const new_args = injectAdditionalArgs(test_argc, test_argv);
+    auto custom_argc = new_args.first;
+    auto custom_argv = new_args.second;
+    CollectiveOps::initialize(custom_argc, custom_argv, no_workers, true);
 
-#   if vt_feature_cmake_test_trace_on
+#if vt_feature_cmake_test_trace_on
     vt::theConfig()->vt_trace = true;
-#   endif
+#endif
 
-    #if DEBUG_TEST_HARNESS_PRINT
-      auto const& my_node = theContext()->getNode();
-      auto const& num_nodes = theContext()->getNumNodes();
-      fmt::print("my_node={}, num_nodes={}\n", my_node, num_nodes);
-    #endif
+#if DEBUG_TEST_HARNESS_PRINT
+    auto const& my_node = theContext()->getNode();
+    auto const& num_nodes = theContext()->getNumNodes();
+    fmt::print("my_node={}, num_nodes={}\n", my_node, num_nodes);
+#endif
   }
 
   virtual void TearDown() {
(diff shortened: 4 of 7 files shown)


lifflander:


```diff

Clones added
============
- src/vt/vrt/collection/balance/node_stats.cc  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6142824&bid=20080880)

jstrzebonski:
@lifflander I think that the easiest way to do it is to call `outputStatsFile()` with a given phase to dump the stats for (something like this: `void NodeStats::outputStatsFileForPhase(PhaseType phase)`). Now the questions are:

1. When to do it? I'm not yet that familiar with the code, but my blind guess would be to do it just before `theNodeStats()->startIterCleanup(phase, model_->getNumPastPhasesNeeded());`. Is that OK?

2. (_this might be a very stupid question, but I want to be sure_)
If 1. is OK then would doing so dump stats for all phases, or could there be some (the last one) that still needs to be saved to file before the program ends? After running lb_iter example I have an impression that stats for all phases are written down.

3. (_next not very smart question..._)
I'm not sure of the size of calculations that vt might be used with, but I assume they could be gigantic. But anyway, does multiple opening and closing a file might be a performance issue for any of vt usage scenarios? I didn't do any benchmark, but I found this -> https://stackoverflow.com/questions/26250744/efficiency-of-fopen-fclose/26251167 and it shows that on some machine (2.2GHz Core i7, 8GB of RAM running OS X) opening and closing a file 100000 times takes about 0.041s. I can imagine that generally vt would be used on enormous clusters, so opening/closing a file is negligible regardless of how many times it's done, but I just want to make sure.

4. Is it possible to change how frequent LB is being run? I mean - every phase, every two phases, and so on.

src/vt/vrt/collection/balance/lb_invoke/lb_manager.cc
```diff
@@ -178,6 +178,14 @@ LBManager::makeLB(MsgSharedPtr<StartLBMsg> msg) {
       lb, node,
       "LBManager: finished migrations\n"
     );
+
+  // Statistics output when LB is enabled and appropriate flag is enabled
+#if vt_check_enabled(lblite)
+  if (theConfig()->vt_lb_stats) {
+    theNodeStats()->outputStatsForPhase(phase);
+  }
+#endif
```
jstrzebonski, line 187: Or maybe it should be done directly in `startIterCleanup()`, just before stats trimming?

src/vt/vrt/collection/balance/node_stats.cc
```diff
@@ -198,78 +198,99 @@ void NodeStats::createStatsFile() {
...
 }
 
-void NodeStats::outputStatsFile() {
+void NodeStats::outputStatsForPhase(PhaseType phase) {
+  // TODO (STRZ) - obviously directory cannot be created here. So where?
+  // if (stats_file_ == nullptr) {
+  //   phase == 0
+  //     ? createStatsFile()
+  //     : openStatsFile();
+  // }
```
jstrzebonski, line 229: That's another question - I think that if stats are enabled creating directory should be done at the beginning, but I done know, where that "beginning" is.

PhilMiller:
I've not gotten into the details of the design or code at all, but just responding to the question about repeat open/close of a file - on supercomputer platforms, this can be **much more expensive** than on a standard workstation with a local file system. Files that are going to be written successively should be opened once, at initialization or first need (my preference is initialization, for early error reporting), and kept open for the duration of the run.

src/vt/vrt/collection/balance/node_stats.cc
```diff
@@ -198,78 +198,99 @@ void NodeStats::createStatsFile() {
...
 }
 
-void NodeStats::outputStatsFile() {
+void NodeStats::outputStatsForPhase(PhaseType phase) {
+  // TODO (STRZ) - directory cannot be created here. So where?
+  // if (stats_file_ == nullptr) {
+  //   phase == 0
+  //     ? createStatsFile()
+  //     : openStatsFile();
+  // }
```
jstrzebonski, line 229: That's another question - I think that if stats are enabled creating directory should be done at the beginning, but I done know, where that "beginning" is.

lifflander:


```diff

Clones added
============
- src/vt/vrt/collection/balance/node_stats.cc  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6142824&bid=20080880)

src/vt/vrt/collection/balance/node_stats.cc
jstrzebonski: Just do it in `initialize()`

src/vt/vrt/collection/balance/node_stats.cc
```diff
@@ -180,21 +180,6 @@ void NodeStats::createStatsFile() {
...
-  }
-
-  // Barrier: wait for node 0 to create directory before trying to put a file in
-  // the stats destination directory
-  if (curRT) {
-    curRT->systemSync();
-  } else {
-    // Something is wrong
-    vtAssert(false, "Trying to dump stats when VT runtime is deallocated?");
-  }
```
jstrzebonski, line 196: Now it's done in `CollectionManager::initialize()`.

src/vt/vrt/collection/balance/node_stats.cc
```diff
@@ -205,74 +190,67 @@ void NodeStats::closeStatsFile() {
   }
 }
 
-void NodeStats::outputStatsFile() {
-  if (stats_file_ == nullptr) {
-    createStatsFile();
-  }
```
jstrzebonski, line 211: File creation is now done in `CollectionManager::initialize()`.

src/vt/vrt/collection/balance/node_stats.cc
```diff
@@ -205,74 +190,67 @@ void NodeStats::closeStatsFile() {
...
+      auto const from = key.fromObj();
+      auto obj_str = fmt::format("{},{},{},{},{}\n", i, to, from, val.bytes, cat);
+      fprintf(stats_file_, "%s", obj_str.c_str());
+    } else {
+      vtAssert(false, "Invalid balance::CommCategory enum value");
     }
   }
-  fflush(stats_file_);
 
-  closeStatsFile();
```
jstrzebonski, line 275: Closing a file is now done in `CollectionManager::finalize()`.

jstrzebonski:
@PhilMiller @lifflander 
First implementation is ready.

src/vt/vrt/collection/balance/lb_invoke/lb_manager.cc
PhilMiller: I think I'd prefer that

src/vt/vrt/collection/balance/node_stats.cc
PhilMiller: Why did this code have to move?

src/vt/vrt/collection/manager.h
```diff
@@ -157,6 +157,7 @@ struct CollectionManager
 
   virtual ~CollectionManager();
 
+  void initialize() override;
```
PhilMiller, line 160: Why can't `NodeStats` itself have an `initialize()` method added? Isn't it a `vt::Component` also?

jstrzebonski: Of course, you're correct.

Does it also mean that
```
#if vt_check_enabled(lblite)
  if (theConfig()->vt_lb_stats) {
    theNodeStats()->closeStatsFile();
    theNodeStats()->clearStats();
  }
#endif
```
could be moved from `CollectionManager::finalize()` to `NodeStats::finalize()` ?

src/vt/vrt/collection/balance/node_stats.cc
jstrzebonski: Yeah, that was stupid.

I'm having an error when calling `systemSync()`:

```
An error occurred in MPI_Barrier
reported by process [1806237697,0]
on communicator MPI_COMM_WORLD
MPI_ERR_COMM: invalid communicator
MPI_ERRORS_ARE_FATAL (processes in this communicator will now abort, and potentially your MPI job)
```

so I was thinking, let's just move it out, and omit SystemSync(), but that doesn't make much sense, other nodes still needs to wait for a 0th to create a directory.

So I'm moving it back, but still I need a help with that error..

jstrzebonski: @lifflander @PhilMiller I will need your help with this error.

src/vt/vrt/collection/balance/node_stats.cc
```diff
@@ -142,6 +142,13 @@ void NodeStats::startIterCleanup(PhaseType phase, int look_back) {
   }
   node_data_[phase] = std::move(new_data);
 
+  // Statistics output when LB is enabled and appropriate flag is enabled
+#if vt_check_enabled(lblite)
```
pnstickne, line 146: If none of the code here is conditionally dependent upon the define (eg. is only depending on `vt_lb_stats` having a value), then it might be reasonable to remove the conditional guards here.

I would check other location for consistency, although the the `theConfig` entries themselves appears unguarded.

src/vt/vrt/collection/balance/node_stats.cc
```diff
@@ -207,74 +233,67 @@ void NodeStats::closeStatsFile() {
...
+  auto i = phase;
+  for (auto&& elm : node_data_.at(i)) {
+    ElementIDType id = elm.first;
+    TimeType time = elm.second;
+    const auto& subphase_times = node_subphase_data_.at(i)[id];
+    size_t subphases = subphase_times.size();
 
-  vt_print(lb, "NodeStats::outputStatsFile: file={}, iter={}\n", print_ptr(stats_file_), num_iters);
+    auto obj_str = fmt::format("{},{},{},{},[", i, id, time, subphases);
+    for (size_t s = 0; s < subphases; s++) {
```
pnstickne, line 249: I would consider `stringstream` as a general construct for appending an arbitrary sequence.

Probably doesn't matter here; just going for uniform consistency.

src/vt/vrt/collection/balance/node_stats.cc
```diff
@@ -207,74 +233,67 @@ void NodeStats::closeStatsFile() {
...
+      obj_str += std::to_string(subphase_times[s]);
+      if (s != subphases - 1)
```
pnstickne, line 251: I'm a fan of moving this to the loop start:

   if (s > 0) obj_str += ",";

src/vt/vrt/collection/manager.h
lifflander: Yes, I think that finalize can do it!

src/vt/vrt/collection/balance/node_stats.cc
lifflander: I actually strongly prefer using fmt unless performance is really an issue.

src/vt/vrt/collection/balance/node_stats.cc
```diff
@@ -207,74 +233,67 @@ void NodeStats::closeStatsFile() {
...
-      obj_str += "]\n";
+    auto const& key = elm.first;
+    auto const& val = elm.second;
+    auto const cat = static_cast<E>(key.cat_);
 
+    if (
+      key.cat_ == CommCategory::SendRecv or
+      key.cat_ == CommCategory::Broadcast
+    ) {
+      auto const to   = key.toObj();
```
pnstickne, line 271: `to/from` variables can be lifted.

src/vt/vrt/collection/balance/node_stats.cc
lifflander: Agreed, you can probably drop the check here because `vt_lb_stats` will exist either way. But, we might want to check, if we don't already, that the uses isn't trying to dump stats when LB is compile-time disabled. Or maybe that's fine and the stats will just be empty?

src/vt/vrt/collection/balance/node_stats.cc
```diff
@@ -207,74 +233,67 @@ void NodeStats::closeStatsFile() {
...
-        obj_str += std::to_string(subphase_times[s]);
-        if (s != subphases - 1)
-          obj_str += ",";
-      }
+  for (auto&& elm : node_comm_.at(i)) {
+    using E = typename std::underlying_type<CommCategory>::type;
 
-      obj_str += "]\n";
+    auto const& key = elm.first;
+    auto const& val = elm.second;
```
pnstickne, line 264: Only using `val.bytes` below.

lifflander:
This is looking good to me!

src/vt/vrt/collection/balance/node_stats.cc
jstrzebonski: It's guarded in `CollectionManager::startPhaseCollective()` so can be remove here.

src/vt/vrt/collection/manager.h
jstrzebonski: Done.

src/vt/vrt/collection/balance/node_stats.cc
jstrzebonski: Done.

src/vt/vrt/collection/balance/node_stats.cc
jstrzebonski: Left as it is.

src/vt/vrt/collection/balance/node_stats.cc
jstrzebonski: Done.

src/vt/vrt/collection/balance/node_stats.cc
jstrzebonski: Done.

src/vt/vrt/collection/balance/node_stats.cc
jstrzebonski: Done.

lifflander:


```diff

Clones removed
==============
+ src/vt/vrt/collection/balance/node_stats.cc  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6142824&bid=20080880)

jstrzebonski:
So I still have a problem with directory creation.

Currently node 0 creates directory and other nodes should wait for it to finish, because right after that all of them would like to create files for their stats. But calling `systemSync()` to make them wait triggers an error:

```
An error occurred in MPI_Barrier
reported by process [1806237697,0]
on communicator MPI_COMM_WORLD
MPI_ERR_COMM: invalid communicator
MPI_ERRORS_ARE_FATAL (processes in this communicator will now abort, and potentially your MPI job)
```

@lifflander AFAIR you said that you have dealt with directory creation in trace? But all I can see there is just calling `mkdir()`. In my case it's not that simple (I guess), because just after directory is created, all nodes need their files to be opened.

I came up with a naive workaround if needed - every node could create a directory with it's rank in name, and that way no node needs to wait for another. It's far from perfect, but it's something.

Anyway, I someone could give me a better solution, that would be great.

lifflander:
> So I still have a problem with directory creation.
> Anyway, I will really appreciate if someone could give me a hand with this.

@jstrzebonski: I've fixed the problem on the branch and rebased it. Basically, while the components are still initializing the communicator was not active in the runtime. I fixed this with some addition logic in `sync()`.

While running `lb_iter`, it seems the stat file output is working when a LB is enabled (for example, `--vt_lb --vt_lb_name=GossipLB`). However, if `--vt_lb_stats` is on the command line _without_ a LB selected, the stats do not output correctly. So I think you need to change that in your implementation.

lifflander:


```diff

Clones removed
==============
+ src/vt/vrt/collection/balance/node_stats.cc  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6142824&bid=20080880)

src/vt/vrt/collection/balance/node_stats.cc
```diff
@@ -142,6 +142,11 @@ void NodeStats::startIterCleanup(PhaseType phase, int look_back) {
   }
   node_data_[phase] = std::move(new_data);
 
+  // Statistics output when LB is enabled and appropriate flag is enabled
```
lifflander, line 145: This only runs when the LB actually runs. So we also need to add a hook when the LB doesn't run (either never or on some non-1 interval). You might want to look at `sysReleaseLB`

lifflander:
@jstrzebonski Can you add some tests for the LB stats file if there aren't any? You might be able to add them to `test_lb`

lifflander:


```diff

Clones removed
==============
+ src/vt/vrt/collection/balance/node_stats.cc  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6142824&bid=20080880)

src/vt/vrt/collection/balance/node_stats.cc
jstrzebonski: Done.

lifflander:


```diff

Clones removed
==============
+ src/vt/vrt/collection/balance/node_stats.cc  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6142824&bid=20080880)

lifflander:


```diff

Clones removed
==============
+ src/vt/vrt/collection/balance/node_stats.cc  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6142824&bid=20080880)

src/vt/vrt/collection/balance/lb_invoke/lb_manager.h
```diff
@@ -263,6 +264,16 @@ struct LBManager : runtime::component::Component<LBManager> {
   template <typename LB>
   void makeLB(MsgSharedPtr<StartLBMsg> msg);
 
+private:
+  /**
+   * \internal \brief Communicate to the NodeStats component that a new phase
```
pnstickne, line 269: Comment doesn't fit name very well, probably as it talks a bout an implementation detail.

src/vt/vrt/collection/balance/node_stats.cc
```diff
@@ -207,74 +228,75 @@ void NodeStats::closeStatsFile() {
...
+    auto const recvSend = getRecvSend(comm);
+    auto const cat = static_cast<E>(comm.cat_);
+    auto obj_str = fmt::format("{},{},{},{},{}\n", i, recvSend.first, recvSend.second, elm.second.bytes, cat);
+    fprintf(stats_file_, "%s", obj_str.c_str());
   }
+
   fflush(stats_file_);
+}
+
+std::pair<ElementIDType, ElementIDType> NodeStats::getRecvSend(CommKeyType const& comm) {
```
pnstickne, line 276: In a case like this, where all usages a local, another option to just `void fn(.., &, &)` instead of returning a pair. Just a consideration.

pnstickne, line 276: Make a TU (.cc) - local function. It is a helper that does not need to be a helper.

Could also use a `switch` instead here likely, which would lead to shorter code.

I might also be tempted to return the values by ref (an actual _advantage here_, is it actually keeps names vs. the pairs first/second).

src/vt/vrt/collection/balance/node_stats.h
```diff
@@ -248,6 +251,8 @@ struct NodeStats : runtime::component::Component<NodeStats> {
    */
   void closeStatsFile();
 
+  static std::pair<ElementIDType, ElementIDType> getRecvSend(CommKeyType const& comm);
```
pnstickne, line 254: Recommend keeping TU local, not a method.

Also might consider a different name. Can't quite put my finger on it. Perhaps `getCommunicationDirection` (although not that long ^_^) or even `getRecvSendDirection`, perhaps.

src/vt/vrt/collection/balance/node_stats.cc
```diff
@@ -207,74 +228,75 @@ void NodeStats::closeStatsFile() {
...
-  auto const num_iters = node_data_.size();
+  vt_print(lb, "NodeStats::outputStatsForPhase: file={}, phase={}\n", print_ptr(stats_file_), phase);
 
-  vt_print(lb, "NodeStats::outputStatsFile: file={}, iter={}\n", print_ptr(stats_file_), num_iters);
+  auto i = phase;
+  for (auto&& elm : node_data_.at(i)) {
+    ElementIDType id = elm.first;
+    TimeType time = elm.second;
+    const auto& subphase_times = node_subphase_data_.at(i)[id];
+    size_t subphases = subphase_times.size();
```
pnstickne, line 246: If `elm * subphases` counts can be "not trivial" in size, consider a stringstream over many-concats.

src/vt/vrt/collection/balance/node_stats.cc
pnstickne: As long as the problem bounds are _known_.

lifflander:
> @lifflander @PhilMiller

Sorry I've been slow to getting to answering these questions.

> 1. When to do it? I'm not yet that familiar with the code, but my blind guess would be to do it just before `theNodeStats()->startIterCleanup(phase, model_->getNumPastPhasesNeeded());`. Is that OK?

This sounds approximately right to me, except for we want tor run it every phase regardless of if the LB runs. But, it should always happen _after_ LB runs. I think that putting it in `LBManager::releaseNow` would be effective.

> 2. (_this might be a very stupid question, but I want to be sure_)
>    If 1. is OK then would doing so dump stats for all phases, or could there be some (the last one) that still needs to be saved to file before the program ends? After running lb_iter example I have an impression that stats for all phases are written down.

The stats for each phase get created regardless of whether LB actually runs. And we want to be able to dump stats even if LB never actually runs. We use the stats file with the python LBAF (Load Balancing Analysis Framework) to analyze the loads _a posteriori_ (https://github.com/DARMA-tasking/LB-analysis-framework).

> 1. Is it possible to change how frequent LB is being run? I mean - every phase, every two phases, and so on.

Yes, you can change the interval of how frequent LB actually runs. But, the system always calls into `sysLB` or `sysReleaseLB` every phase. That's why I'm suggesting running it every phase regardless. Eventually, we could provide some flexibility on how often it gets dumped to file.

src/vt/vrt/collection/balance/lb_invoke/lb_manager.h
jstrzebonski: Removed.

src/vt/vrt/collection/balance/node_stats.cc
jstrzebonski: Done. Except returning by ref param - I like when the function just returns by `return`. The best would be to have C++17's and structured bindings :- )

src/vt/vrt/collection/balance/node_stats.h
jstrzebonski: `getCommDirection`?

tests/unit/test_parallel_harness.h
```diff
@@ -114,15 +123,32 @@ struct TestParallelHarnessAny : TestHarnessAny<TestBase> {
...
+#endif
 
     CollectiveOps::finalize();
 
     TestHarnessAny<TestBase>::TearDown();
   }
+
+protected:
+  template <typename Arg>
+  void AddArgs(Arg& arg) {
```
jstrzebonski, line 138: Should be `addArgs`

tests/unit/test_parallel_harness.h
```diff
@@ -114,15 +123,32 @@ struct TestParallelHarnessAny : TestHarnessAny<TestBase> {
...
+    this->additional_args_.emplace_back(&arg[0]);
+  }
+
+  template <typename Arg, typename... Args>
+  void AddArgs(Arg& arg, Args&... args) {
```
jstrzebonski, line 143: Should be `addArgs`

tests/unit/test_parallel_harness.h
```diff
@@ -88,23 +89,31 @@ struct TestParallelHarnessAny : TestHarnessAny<TestBase> {
...
+      mpi_singleton =
+        std::make_unique<MPISingletonMultiTest>(test_argc, test_argv);
     }
 
-    CollectiveOps::initialize(test_argc, test_argv, no_workers, true);
+    std::copy(
+      test_argv, test_argv + test_argc, std::back_inserter(additional_args_)
+    );
+
+    AddAdditionalArgs();
```
jstrzebonski, line 102: Should be `addAdditionalArgs`

tests/unit/test_parallel_harness.h
```diff
@@ -88,23 +89,31 @@ struct TestParallelHarnessAny : TestHarnessAny<TestBase> {
...
+
+    int custom_argc = additional_args_.size();
+    char** custom_argv = additional_args_.data();
```
jstrzebonski, line 105: Extract to method.

tests/unit/test_parallel_harness.h
jstrzebonski: Done.

tests/unit/test_parallel_harness.h
jstrzebonski: Done.

tests/unit/test_parallel_harness.h
jstrzebonski: Done.

tests/unit/test_parallel_harness.h
jstrzebonski: Done.

lifflander:


```diff

Clones added
============
- tests/unit/collection/test_lb.extended.cc  4
         

Clones removed
==============
+ src/vt/vrt/collection/balance/node_stats.cc  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6142824&bid=20080880)

lifflander (approved):
This is looking great. I think it's ready to merge

lifflander:


```diff

Clones added
============
- tests/unit/collection/test_lb.extended.cc  4
         

Clones removed
==============
+ src/vt/vrt/collection/balance/node_stats.cc  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/lifflander/vt/pullRequest?prid=6142824&bid=20080880)

PR #1043, opened 2020-09-11
1042 Cleanup memory growing unbounded in reduce manager

### LINKED ISSUE #1042: Clear reduce state holder outer state after reduction completes
**Describe the bug**
Leaving behind data structures that need to be deleted

lifflander:
Fixes #1042

### DIFF
--- a/src/vt/collective/reduce/reduce.impl.h
+++ b/src/vt/collective/reduce/reduce.impl.h
@@ -291,6 +291,8 @@ void Reduce::startReduce(detail::ReduceStamp id, bool use_num_contrib) {
       theMsg()->sendMsg<MsgT,ReduceManager::reduceUp<MsgT>>(parent, typed_msg);
     }
   }
+
+  state_.erase(lookup);
 }
 
 template <typename MsgT>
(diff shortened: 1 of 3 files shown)


src/vt/collective/reduce/reduce.impl.h
```diff
@@ -291,6 +291,8 @@ void Reduce::startReduce(detail::ReduceStamp id, bool use_num_contrib) {
       theMsg()->sendMsg<MsgT,ReduceManager::reduceUp<MsgT>>(parent, typed_msg);
     }
   }
+
+  state_.erase(lookup);
```
PhilMiller, line 295: Should this be here, or inside the `if (ready)` block?

lifflander: Oops, good catch. I got this right on release branch, but screwed it up here.

PhilMiller:
I've been testing locally, and this seems to fix the CI failures

PhilMiller:
Make sure to make the same fix on the stable branch PR.

PhilMiller (approved):

lifflander:
There is a reentrancy bug here due to the location of the erase.

PhilMiller (approved):

lifflander:
The gcc-7 workflow keeps failing on copying a file during the install process. Is this an issue with CI? Or maybe with trace enabled we are reaching the disk usage limit on the container?

lifflander:
Also gcc-5 has hung for nearly 4 hours. Never seen that before and can't see to access the logs right now.

Issue #954, opened 2020-07-28
Add documentation about epoch tracking during reductions


lifflander:
**What Needs to be Done?**

The reason why epochs are tracked correctly during reductions is a little tricky and not obvious from reading the code. Add some details about how this works to the documentation. Maybe include some info about other components also.

PhilMiller:
Would you like me to pick this up? I understand the reasoning well enough to write it up.

lifflander:
Yes, good idea.. I'll assign it to you.

PR #1046, opened 2020-09-11
#954: Document TD reasoning of reductions

PhilMiller:
Fixes: #954


lifflander (approved):
Text looks good to me!

Issue #1056, opened 2020-09-15
Offset default LB invocation from phase 0 to avoid initialization and warm-up transients


PhilMiller:
**What Needs to be Done?**

When using simple arguments for LB interval, we want it happen first on phase 1 rather than phase 0. If the interval is 1, then that special case needs to be taken into account.

**Is your feature request related to a problem? Please describe.**

Applications have all sorts of initialization before their normal work starts, and phase zero timings will include all of that, as well as potentially being skewed by warm-up effects.

**Describe potential solution outcome**

Change the logic to look for `phase % interval == 1 || interval == 1`

PR #1057, opened 2020-09-15
#1056: Run LB starting after phase 1 by default

PhilMiller:
Fixes: #1056


lifflander (approved):
Looks good to me!

lifflander:
Rebase and the CI failure will go away. I've fixed them on develop.

PhilMiller:
I just had to rebase and await CI checks again because my rebase on an SNL server wasn't signed. We should probably add that as a commit check, so it's obvious that merging will be impossible before waiting on all the other tests to pass.

PhilMiller:
Opened #1061 accordingly

PhilMiller:
At least ccache is working reasonably for many of the checks

PR #1089, opened 2020-09-25
#1088: Always use -std=c++14 for GCC and not earlier experimental -std=c++1y

### LINKED ISSUE #1088: nvcc complains about use of -std=c++1y vs -std=c++14
**Describe the bug**

Compilation with Nvidia's compiler produces these messages for every file compiled, for versions 10.1 and 11.0:
```
nvcc_wrapper does not accept standard flags -std=c++1y since partial standard flags and standards after C++14 are not supported. nvcc_wrapper will use -std=c++14 instead. It is undefined behavior to use this flag. This should only be occurring during CMake configuration.
nvcc_wrapper - *warning* you have set multiple standard flags (-std=c++1* or --std=c++1*), only the last is used because nvcc can only accept a single std setting
```

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

PhilMiller:
Fixes #1088

### DIFF
--- a/cmake-modules/SetCXXCompilerFlags.cmake
+++ b/cmake-modules/SetCXXCompilerFlags.cmake
@@ -6,7 +6,7 @@ macro(set_darma_compiler_flags)
 set(CXX_STANDARD_FLAGS)
 if ("${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU")
   # 4.9.3 complains about std::min not being constexpr
-  set(CXX_STANDARD_FLAGS -std=c++1y)
+  set(CXX_STANDARD_FLAGS -std=c++14)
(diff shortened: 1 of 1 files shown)


lifflander (approved):

PhilMiller:
It cuts down half the errors. Now just getting
```
nvcc_wrapper - *warning* you have set multiple standard flags (-std=c++1* or --std=c++1*), only the last is used because nvcc can only accept a single std setting
```

PhilMiller:
Detector and Checkpoint both produce this complaint, which may be the culprit if the flags are inherited and incorporated:
```
nvcc_wrapper has been given GNU extension standard flag -std=gnu++14 - reverting flag to -std=c++14
```

PhilMiller:
I'm inclined to get this bit merged, and leave the other warning for a new issue, for someone else to dig into.

PhilMiller:
Opened #1090 for the latter

PR #1091, opened 2020-09-26
#1069 runtime - don't re-initialize MPI in interop mode

### LINKED ISSUE #1069: Argument parsing error unconditionally triggers MPI_Init
**Describe the bug**
When an invalid command-line argument is passed to vt, it calls `MPI_Init` (`src/vt/runtime/runtime.cc:153`) in the process of handling the error.  This is problematic when vt is used within EMPIRE because MPI was already initialized.

**To Reproduce**
1. Call MPI_Init in your driver before initializing vt
2. Pass an invalid command-line argument to your executable

**Expected behavior**
If MPI was already initialized, it should use the existing communicator.

pnstickne:
Don't crash processes by attempting to improperly call MPI_Init again.

Will still terminate a process, with diagnostic message, when there are invalid VT args. Invalid args are limited to those in the form `--vt_bad`.

Also moved any MPI_Init (when not interop) into the Runtime class itself, for parity with MPI_Finalize, full reliance on MPI_Init eating its own arguments, and some future update work.

Minor breaking: `--vt_mpi_args` is redundant and no longer supported.

fixes #1069

### DIFF
--- a/src/vt/collective/collective_ops.cc
+++ b/src/vt/collective/collective_ops.cc
@@ -63,9 +63,11 @@ RuntimePtrType CollectiveAnyOps<instance>::initialize(
   using vt::runtime::Runtime;
   using vt::runtime::eRuntimeInstance;
 
+  MPI_Comm resolved_comm = comm not_eq nullptr ? *comm : MPI_COMM_WORLD;
+
 #pragma sst global rt
   RuntimeInst<instance>::rt = std::make_unique<Runtime>(
-    argc, argv, num_workers, is_interop, comm
+    argc, argv, num_workers, is_interop, resolved_comm
   );
 
 #pragma sst global rt
--- a/src/vt/context/context.cc
+++ b/src/vt/context/context.cc
@@ -56,46 +56,31 @@
 
 namespace vt { namespace ctx {
 
-Context::Context(int argc, char** argv, bool const is_interop, MPI_Comm* comm) {
+Context::Context(bool const is_interop, MPI_Comm comm) {
   #if DEBUG_VT_CONTEXT
     fmt::print(
       "Context::Context is_interop={}, comm={}\n", print_bool(is_interop), comm
     );
   #endif
 
-  if (not is_interop) {
-    MPI_Init(&argc, &argv);
-  }
-
-  if (comm != nullptr and *comm != MPI_COMM_NULL) {
-    communicator_ = *comm;
-  } else {
-    communicator_ = MPI_COMM_WORLD;
-  }
+  is_comm_world_ = comm == MPI_COMM_WORLD;
 
   if (is_interop) {
-    MPI_Comm vt_comm;
-    MPI_Comm_dup(communicator_, &vt_comm);
-    communicator_ = vt_comm;
+    MPI_Comm_dup(comm, &comm);
   }
-
-  is_comm_world_ = communicator_ == MPI_COMM_WORLD;
 
   int numNodesLocal = uninitialized_destination;
   int thisNodeLocal = uninitialized_destination;
 
-  MPI_Comm_size(communicator_, &numNodesLocal);
-  MPI_Comm_rank(communicator_, &thisNodeLocal);
+  MPI_Comm_size(comm, &numNodesLocal);
+  MPI_Comm_rank(comm, &thisNodeLocal);
 
+  communicator_ = comm;
   numNodes_ = static_cast<NodeType>(numNodesLocal);
   thisNode_ = static_cast<NodeType>(thisNodeLocal);
 
   setDefaultWorker();
 }
-
-Context::Context(bool const interop, MPI_Comm* comm)
-  : Context(0, nullptr, interop, comm)
-{ }
 
 void Context::setDefaultWorker() {
   setWorker(worker_id_comm_thread);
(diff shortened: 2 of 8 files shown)


src/vt/runtime/runtime.cc
```diff
@@ -145,13 +145,17 @@ Runtime::Runtime(
...
-    MPI_Comm_rank(comm, &rank);
+    // To better honor MPI, force an MPI_Init then MPI_Abort as relevant.
+    MPI_Comm comm = communicator_;
+
+    if (not is_interop_) {
+      MPI_Init(NULL, NULL);
+    }
+
+    int rank = 0;
+    if (comm not_eq MPI_COMM_NULL) {
```
pnstickne, line 156: Having MPI_COMM_NULL passed in should probably be an error in it's own right..

src/vt/runtime/runtime.cc
```diff
@@ -165,7 +169,11 @@ Runtime::Runtime(
           << std::flush;
     }
 
-    MPI_Abort(comm, exit_code);
+    // Even in interop mode, still abort MPI.
+    if (comm not_eq MPI_COMM_NULL) {
+      MPI_Abort(comm, exit_code);
+    }
+
     std::_Exit(exit_code); // no return
```
pnstickne, line 177: Still terminates on invalid args -- albeit not with a random MPI failure.

src/vt/runtime/runtime.cc
```diff
@@ -439,15 +441,14 @@ bool Runtime::finalize(bool const force_now, bool const disable_sig) {
       printShutdownBanner(num_units, coll_epochs);
     }
 
-    finalizeMPI();
```
pnstickne, line 442: MPI is initialized in constructor, finalized in destructor (if not under interop). Previously the MPI_Init was buried in the context constructor which was only called on the Runtime initialize.

Likewise, communicator is duplicated/created in the context constructor and free'd in the context destructor.

src/vt/collective/collective_ops.cc
```diff
@@ -63,9 +63,11 @@ RuntimePtrType CollectiveAnyOps<instance>::initialize(
   using vt::runtime::Runtime;
   using vt::runtime::eRuntimeInstance;
 
+  MPI_Comm resolved_comm = comm not_eq nullptr ? *comm : MPI_COMM_WORLD;
```
pnstickne, line 66: vt::initialize, CollectiveOps, maintains external signature.. Runtime ctor only deals with a non-ptr MPI_Comm.

lifflander:
Is this preempted by #1093?

pnstickne:
> Is this preempted by #1093?

I was hoping to get this merged, closing the story (focused on making VT not crash with invalid args in interop..), then follow with the other story/PR.

src/vt/runtime/runtime.cc
```diff
@@ -127,30 +125,28 @@ Runtime::Runtime(
...
 
-  // MPI_Init 'should' be called first on the original arguments,
-  // with the justification that in some environments in addition to removing
-  // special MPI arguments, it can actually ADD arguments not from argv.
-  // That is not done here and doing so moves up parts of 'initialize' logic.
+  // Always initialize MPI first, before handling arguments.
+  // This also allows MPI a first-stab at dealing with arguments.
+  if (not is_interop_) {
+    MPI_Init(&argc, &argv);
+  }
```
PhilMiller, line 138: Should this carry an 
```
else { 
int flag;
MPI_Initialized(&flag);
vtAbortIf(flag == 0);
}
```

PhilMiller: Actually, should we also test and abort on MPI being initialized in the interop case? So put the flag and test outside the `if`, and then pick the sense of the abort depending on which case we're in?

lifflander: Ah, that's a good point. @pnstickne What do you think about adding this logic? I think it could be useful to detect that error.

lifflander (approved):

lifflander:
This looks great to me. I think @PhilMiller has a point that we might want to add some a check for when is either not initialized and should be (and vice versa).

src/vt/runtime/runtime.cc
pnstickne: I think it's a good idea. Changes in. It tests in both cases and should emit a "useful" message.

src/vt/context/context.cc
```diff
@@ -56,47 +56,32 @@
...
-  } else {
-    communicator_ = MPI_COMM_WORLD;
-  }
+  is_comm_world_ = comm == MPI_COMM_WORLD;
 
   if (is_interop) {
-    MPI_Comm vt_comm;
-    MPI_Comm_dup(communicator_, &vt_comm);
-    communicator_ = vt_comm;
+    MPI_Comm_dup(comm, &comm);
```
PhilMiller, line 69: As a matter of good MPI practice, we should *always* duplicate the supplied comm.

For clarity, if made ubiquitously, this call can thus be rewritten `MPI_Comm_dup(comm, &communicator_)` and the assignment below deleted.

PhilMiller: At that point, the `is_interop` argument isn't relevant, and can be dropped from this constructor.

PhilMiller (approved):

src/vt/context/context.cc
PhilMiller: I see that's handled in #1093.

PR #1107, opened 2020-10-12
1098 Enable assertions during CI when release builds are used

### LINKED ISSUE #1098: Enable assertions during CI when release builds are used
Right now, if you build with cmake in release mode, assertions `vtAssert*()` are disabled. Since the workflows all build in release to reduce memory usage, we don't see if assertions are breaking. I believe the configuration parameter `vt_check_enabled(production)` is used to switch between assertion failures breaking or not in release mode. 

Add a cmake option to control this so we can have assertions in release builds in the workflows.

JacobDomagala:
Added new flag `vt_ci_build` to distinguish CI build from normal one. By default this only affects CI runs, meaning that if VT is built in Release mode locally (via docker containter or cmake) this will still have `production` flag set on.

Also, I think having the option to check for build type in the code can be useful in future (i.e. for unit tests we might want different behaviour in CI and different locally). So now you can check for CI build using `vt_check_enabled(ci_build)`

Fixes #1098

### DIFF
--- a/.github/workflows/dockerimage-clang-3.9-ubuntu-mpich.yml
+++ b/.github/workflows/dockerimage-clang-3.9-ubuntu-mpich.yml
@@ -37,6 +37,7 @@ jobs:
       VT_EXTENDED_TESTS: 1
       VT_UNITY_BUILD: 1
       VT_ZOLTAN: 0
+      VT_CI_BUILD: 1
       CACHE: ~/.local/cache/
 
     steps:
--- a/cmake/define_build_types.cmake
+++ b/cmake/define_build_types.cmake
@@ -164,6 +164,9 @@ else()
   set(vt_feature_cmake_zoltan "0")
 endif()
 
+message(STATUS "CI_BUILD = ${vt_ci_build}")
+set(vt_feature_cmake_ci_build ${vt_ci_build})
+
 set(vt_feature_cmake_no_feature "0")
 set(vt_feature_cmake_production "0")
 
--- a/cmake/define_build_types.cmake
+++ b/cmake/define_build_types.cmake
@@ -212,8 +212,8 @@ foreach(loop_build_type ${VT_CONFIG_TYPES})
     ${cmake_vt_debug_modes_${loop_build_type}}
   )
 
-  # assume production mode for everything except debug
-  if (loop_build_type STREQUAL "debug")
+  # assume production mode for everything except debug or CI build
+  if (loop_build_type STREQUAL "debug" OR ${vt_ci_build})
     set(vt_feature_cmake_production "0")
   else()
     set(vt_feature_cmake_production "1")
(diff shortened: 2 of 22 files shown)


.github/workflows/dockerimage-clang-3.9-ubuntu-mpich.yml
```diff
@@ -37,6 +37,7 @@ jobs:
       VT_EXTENDED_TESTS: 1
       VT_UNITY_BUILD: 1
       VT_ZOLTAN: 0
+      VT_CI_BUILD: 1
```
lifflander, line 40: The workflow files are generated. Instead of editing them, you should edit `vt/scripts/workflows.ini` and `vt/scripts/workflow-template.yml`. Then, you can run this command to generate:

```bash
./scripts/generate-workflow.pl ./scripts/workflow-template.yml ./scripts/workflows.ini
````

**Review by lifflander — CHANGES_REQUESTED** [review]

File: .github/workflows/dockerimage-clang-3.9-ubuntu-mpich.yml (hunk 1 above)
Comment (JacobDomagala) [reply .github/workflows/dockerimage-clang-3.9-ubuntu-mpich.yml:40]: It even says on top of the workflow files that there're generated. My bad.

Comment (lifflander) [reply .github/workflows/dockerimage-clang-3.9-ubuntu-mpich.yml:40]: No problem... Looks good!

**Review by lifflander — APPROVED** [review]
Looks good now! :)

File: cmake/define_build_types.cmake [hunk 2]
```diff
@@ -212,8 +215,8 @@ foreach(loop_build_type ${VT_CONFIG_TYPES})
     ${cmake_vt_debug_modes_${loop_build_type}}
   )
 
-  # assume production mode for everything except debug
-  if (loop_build_type STREQUAL "debug")
+  # assume production mode for everything except debug or CI build
+  if (loop_build_type STREQUAL "debug" OR ${vt_ci_build})
```
PhilMiller, line 219: This seems to have broken manual builds outside the CI setting, with an error as follows:

```
CMake Error at cmake/define_build_types.cmake:219 (if):
  if given arguments:

    "loop_build_type" "STREQUAL" "debug" "OR"

  Unknown arguments specified
  src/CMakeLists.txt:275 (include)
```

The variable expansion `${vt_ci_build}` seems to be an empty string, causing a CMake syntax error.

PhilMiller: I was building with the command
> `cmake /projects/empire/users/pbmille/repos/vt/ -DCMAKE_BUILD_TYPE=RelWithDebInfo`

to get that error.

When I run
> `cmake /projects/empire/users/pbmille/repos/vt/ -DCMAKE_BUILD_TYPE=RelWithDebInfo -Dvt_ci_build=False`

It succeeds

JacobDomagala: Addressed the issue in https://github.com/DARMA-tasking/vt/pull/1109

PR #1109, opened 2020-10-13
1098: Set default value of vt_ci_build to '0' inside CMake

### LINKED ISSUE #1098: Enable assertions during CI when release builds are used

JacobDomagala:
Fixes #1098


JacobDomagala:
Hotfix for discovered issue (https://github.com/DARMA-tasking/vt/pull/1107#discussion_r504063326)

jstrzebonski (approved):
Seems fine.

PhilMiller (approved):

PR #1119, opened 2020-10-20
1112 Callback: Add sendMsg() which takes MsgPtrThief

### LINKED ISSUE #1112: Callbacks do not use MsgThief or accept MsgSharedPtr
**What Needs to be Done?**

The callback API `cb.send(msg)` does not take a MsgSharedPtr or MsgThief (whichever is the right approach?). Add that to the overload.

JacobDomagala:
Fixes #1112


cz4rs (approved):
looks good to me :+1:

src/vt/pipe/callback/cb_union/cb_raw_base.h
```diff
@@ -128,9 +128,15 @@ struct CallbackRawBaseSingle {
   bool null()  const { return cb_.null();  }
   bool valid() const { return cb_.valid(); }
 
+  template <typename MsgT, typename... Args>
+  void send(Args... args);
+
   template <typename MsgT>
   void send(MsgT* msg);
```
PhilMiller, line 135: As elsewhere, do we want to make this overload as deprecated?

src/vt/pipe/callback/cb_union/cb_raw_base.h
```diff
@@ -232,12 +238,22 @@ struct CallbackTyped : CallbackRawBaseSingle {
     cb_   = std::move(other.cb_);
   }
 
+  template <typename MsgU, typename... Args>
+  void send(Args... args) {
+    static_assert(std::is_same<MsgT, MsgU>::value, "Required exact type match");
```
PhilMiller, line 243: I could see us wanting to loosen this to derivation, or even convertibility. No need to do that here and now, though.

tutorial/tutorial_1g.h
```diff
@@ -147,10 +147,8 @@ static void getCallbackHandler(MsgWithCallback* msg) {
   auto const cur_node = ::vt::theContext()->getNode();
   ::fmt::print("getCallbackHandler: triggered on node={}\n", cur_node);
 
-  // Create a msg to trigger to callback
-  auto data_msg = ::vt::makeMessage<DataMsg>();
   // Send the callback with the message
```
PhilMiller, line 150: This comment is now a bit wrong - "Send the callback a message" perhaps? Or "an empty message"?

PhilMiller (approved):

src/vt/pipe/callback/cb_union/cb_raw_base.h
```diff
@@ -128,9 +128,15 @@ struct CallbackRawBaseSingle {
   bool null()  const { return cb_.null();  }
   bool valid() const { return cb_.valid(); }
 
+  template <typename MsgT, typename... Args>
```
lifflander, line 131: Now that I'm looking at this, is there some way to avoid having to specify `MsgT` when the type is known by the (e.g, callback `Callback<MyMsg>`)?

src/vt/pipe/callback/cb_union/cb_raw_base.h
JacobDomagala: I would assume it's widely used in EMPIRE, not sure if we want to flood it with warnings. We could add doxygen `\deprecated` command though. 
Can you comment on that @lifflander?

tutorial/tutorial_1g.h
JacobDomagala: Fixed.

PR #1130, opened 2020-10-28
1079: Rewrote sucessor to parent epoch in all associated files

### LINKED ISSUE #1079: Rewrite 'Successor Epoch' to 'Parent Epoch' as appropriate
In various parts of the termination detection (TD) code, there's logic to handle the creation of an epoch 'C' as an event occurring during another epoch 'P'. Thus, events occurring during 'C' logically also occur during 'P'. So, 'P' cannot terminate until at least the transitive closure of those events completes, which is precisely when 'C' terminates. Right now, this is described as P being a (the) 'successor' of C. This terminology is misleading, because C necessarily is created after P, and many additional epochs D, E, and F may all have P as their successor. 

The terminology that Jonathan and I agree would be preferable is that P is the 'parent' of C, a 'child'. This reflects that epochs and their associated events form a tree. So, the `SuccessorEpochCapture` and variables associated with that type or usage should be renamed or otherwise restructured to reflect that preferred terminology.

bradybray:
fixes #1079.

### DIFF
--- a/src/vt/termination/epoch_tags.cc
+++ b/src/vt/termination/epoch_tags.cc
@@ -48,7 +48,7 @@
 
 namespace vt { namespace term {
 
-ChildEpochCapture::ChildEpochCapture()
+ParentEpochCapture::ParentEpochCapture()
   : epoch_(
       (theMsg()->getEpoch() != no_epoch and
        theMsg()->getEpoch() != term::any_epoch_sentinel) ?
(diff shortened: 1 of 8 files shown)


PhilMiller (changes requested):
So, I realize that the description in the issue was not entirely explicit. The captured 'Successor' epoch is the **parent**, not the child. The epochs make a tree, with all children terminating before a parent terminates. That's what equates parent<->successor.

PhilMiller (approved):

bradybray:
> 
> 
> Looks good to me!

@PhilMiller @lifflander There appears to be a problem with the python version used for brew on OS X. Do either of you know anything about this?

lifflander (approved):

PR #1132, opened 2020-10-29
1051: Add local invoke to proxies that trace and record LB

### LINKED ISSUE #1051: Add local invoke to proxies that trace and record LB
**What Needs to be Done?**

For objgroups, collections, (and maybe bare handlers), add a local proxy invoke. Right now, for instance, you can do `proxy.send<MyCol, &MyCol::handler>()` which always goes through the scheduler even if it's local. So if you want to execute right away, you have run the handler manually. Add a `proxy.invoke<...>()` that does the local invocation inline. It should error if the element does not exist where the user tried to invoke it.

JacobDomagala:
Fixes #1051 

- [x] virtual collection proxy
- [x] obj group proxy
- [x] unittests

### DIFF
--- a/src/vt/trace/trace_constants.h
+++ b/src/vt/trace/trace_constants.h
@@ -83,7 +83,8 @@ enum class eTraceConstants : int32_t {
   UserStat                  = 32,
   BeginUserEventPair        = 98,
   EndUserEventPair          = 99,
-  UserEventPair             = 100
+  UserEventPair             = 100,
+  LocalInvoke               = 101
 };
 
 
--- a/src/vt/trace/trace_constants.h
+++ b/src/vt/trace/trace_constants.h
@@ -83,8 +83,7 @@ enum class eTraceConstants : int32_t {
   UserStat                  = 32,
   BeginUserEventPair        = 98,
   EndUserEventPair          = 99,
-  UserEventPair             = 100,
-  LocalInvoke               = 101
+  UserEventPair             = 100
 };
 
 
--- a/src/vt/vrt/collection/invoke/invokable.h
+++ b/src/vt/vrt/collection/invoke/invokable.h
@@ -2,7 +2,7 @@
 //@HEADER
 // *****************************************************************************
 //
-//                              proxy_elm_traits.h
+//                                 invokable.h
 //                           DARMA Toolkit v. 1.0.0
 //                       DARMA/vt => Virtual Transport
 //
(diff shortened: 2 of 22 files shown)


src/vt/trace/trace_constants.h
```diff
@@ -83,7 +83,8 @@ enum class eTraceConstants : int32_t {
   UserStat                  = 32,
   BeginUserEventPair        = 98,
   EndUserEventPair          = 99,
-  UserEventPair             = 100
+  UserEventPair             = 100,
+  LocalInvoke               = 101
```
JacobDomagala, line 87: Not sure whether we want (or can as this probably won't work with Projections tool) to add new trace type for invoke or should we use existing one? At the end the output will be the same as for Creation event type.

lifflander:
Per @PhilMiller suggestion, we should create a `makeTraceCreationSend(..)` for both starting the local invocation and on the receiving side after the invocation returns. So it will looks like a message was sent to start the local invoke and then a message was sent back to the same node to continue the work after the local invocation.

src/vt/trace/trace_constants.h
lifflander: Let's make this just a regular `Creation` event.

lifflander: Projections doesn't do anything special for these.

src/vt/vrt/collection/invoke/invokable.h
```diff
@@ -0,0 +1,78 @@
...
+
+#if !defined INCLUDED_VRT_COLLECTION_INVOKE_INVOKABLE_H
+#define INCLUDED_VRT_COLLECTION_INVOKE_INVOKABLE_H
+
+#include "vt/config.h"
+#include "vt/vrt/collection/active/active_funcs.h"
+
+namespace vt { namespace vrt { namespace collection {
+
+template <typename ColT, typename IndexT, typename BaseProxyT>
```
lifflander, line 53: Can you add doxygen for these new methods here?

lifflander (approved):
Overall, this looks great and it's nearly ready to merge I think. Just a little more doxygen to be added to the proxy interface for virtual collections. Also, we should just use the standard trace event.

JacobDomagala:
@PhilMiller @lifflander  Currently invoke produces following traces:

> 1. Creation event (let's say with trace_id == 2)
> 2. Begin Processing (trace_id == 2)
> _execute user's function_
> 3. End Processing (trace_id == 2)


So you're proposing we should add another Creation event like following?

> 1. Creation event (let's say with trace_id == 2)
> 2. Begin Processing (trace_id == 2)
> _execute user's function_
> 3. End Processing (trace_id == 2)
> 4. Creation event (**trace_id == 3**)

Or do you have something else in mind?

JacobDomagala:
PR updated

lifflander:
> PR updated

Looks good I think this is ready to rebase and merge

PR #1133, opened 2020-10-29
1125 scheduler: template runInEpoch* functions 

### LINKED ISSUE #1125: runInEpoch* do not take a universal reference (they take a r-val ref)
**What Needs to be Done?**

Due to this, you can pass an existing std::function to them. Template them and move to the header file. Something like this would work:

```C++
template <typename Callable>
void runInEpochCollective(Callable&& fn) {
  //...
}
```

cz4rs:
Fixes #1125 

- [x] implementation
- [x] unit tests

### DIFF
--- a/src/vt/scheduler/scheduler.impl.h
+++ b/src/vt/scheduler/scheduler.impl.h
@@ -46,6 +46,36 @@
 #define INCLUDED_VT_SCHEDULER_SCHEDULER_IMPL_H
 
 #include "vt/config.h"
+#include "vt/messaging/active.h"
+#include "vt/termination/termination.h"
+
+namespace vt {
+
+template <typename Callable>
+void runInEpochCollective(Callable&& fn) {
+  theSched()->triggerEvent(sched::SchedulerEvent::PendingSchedulerLoop);
+
+  auto ep = theTerm()->makeEpochCollective();
+  theMsg()->pushEpoch(ep);
+  fn();
+  theMsg()->popEpoch(ep);
+  theTerm()->finishedEpoch(ep);
+  runSchedulerThrough(ep);
+}
+
+template <typename Callable>
+void runInEpochRooted(Callable&& fn) {
+  theSched()->triggerEvent(sched::SchedulerEvent::PendingSchedulerLoop);
+
+  auto ep = theTerm()->makeEpochRooted();
+  theMsg()->pushEpoch(ep);
+  fn();
+  theMsg()->popEpoch(ep);
+  theTerm()->finishedEpoch(ep);
+  runSchedulerThrough(ep);
+}
+
+} /* end namespace vt */
 
 namespace vt { namespace sched {
 
--- a/src/vt/scheduler/scheduler.impl.h
+++ b/src/vt/scheduler/scheduler.impl.h
@@ -52,10 +52,9 @@
 namespace vt {
 
 template <typename Callable>
-void runInEpochCollective(Callable&& fn) {
+void runInEpoch(EpochType ep, Callable&& fn) {
   theSched()->triggerEvent(sched::SchedulerEvent::PendingSchedulerLoop);
 
-  auto ep = theTerm()->makeEpochCollective();
   theMsg()->pushEpoch(ep);
   fn();
   theMsg()->popEpoch(ep);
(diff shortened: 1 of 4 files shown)


src/vt/scheduler/scheduler.impl.h
```diff
@@ -45,7 +45,39 @@
 #if !defined INCLUDED_VT_SCHEDULER_SCHEDULER_IMPL_H
 #define INCLUDED_VT_SCHEDULER_SCHEDULER_IMPL_H
 
+#include "vt/scheduler/scheduler.h"
```
PhilMiller, line 48: This is a circular inclusion, and should be unnecessary

cz4rs: removed :+1:

tests/unit/scheduler/test_high_level_calls.cc
```diff
@@ -0,0 +1,90 @@
...
+#include <gtest/gtest.h>
+
+#include "vt/transport.h"
+#include "test_parallel_harness.h"
+
+namespace vt { namespace tests { namespace unit {
+
+struct TestHighLevelCalls : TestParallelHarness {
+  bool called = false;
+};
```
PhilMiller, line 54: Just as a check, this gets instantiated separately for each of the tests below, correct?

PhilMiller: Assuming that's affirmed, I'm +1 on this

src/vt/scheduler/scheduler.impl.h
```diff
@@ -46,6 +46,36 @@
...
+void runInEpochRooted(Callable&& fn) {
+  theSched()->triggerEvent(sched::SchedulerEvent::PendingSchedulerLoop);
+
+  auto ep = theTerm()->makeEpochRooted();
+  theMsg()->pushEpoch(ep);
+  fn();
+  theMsg()->popEpoch(ep);
+  theTerm()->finishedEpoch(ep);
+  runSchedulerThrough(ep);
+}
```
jstrzebonski, line 76: Those two functions are almost identical. Maybe we would like to do something like this?
```
template <EpochType (term::TerminationDetector::*memFun)()>
struct runInEpoch {
  template <typename Callable>
  static void run(Callable&& fn) {
    theSched()->triggerEvent(sched::SchedulerEvent::PendingSchedulerLoop);

    auto ep = (theTerm()->*memFun)();
    theMsg()->pushEpoch(ep);
    fn();
    theMsg()->popEpoch(ep);
    theTerm()->finishedEpoch(ep);
    runSchedulerThrough(ep);
  }
};

template <typename Callable>
void runInEpochCollective(Callable&& fn) {
  runInEpoch<&term::TerminationDetector::makeEpochCollective>::run(
    std::forward<Callable>(fn)
  );
}

template <typename Callable>
void runInEpochRooted(Callable&& fn) {
  runInEpoch<&term::TerminationDetector::makeEpochRooted>::run(
    std::forward<Callable>(fn)
  );
}
```

tests/unit/scheduler/test_high_level_calls.cc
cz4rs: yes, it's a separate instance for each `TEST_F`

src/vt/scheduler/scheduler.impl.h
cz4rs: I guess simply passing an epoch to a generalized function (like runInEpoch) would work? I will have a look

lifflander (approved):
This looks great to me

src/vt/scheduler/scheduler.impl.h
```diff
@@ -46,6 +46,35 @@
...
+
+template <typename Callable>
+void runInEpochCollective(Callable&& fn) {
+  auto ep = theTerm()->makeEpochCollective();
+  runInEpoch(ep, fn);
```
jstrzebonski, line 68: I guess `runInEpoch(ep, std::forward(fn));?

jstrzebonski (approved):
Seems to be OK.

JacobDomagala (approved):

cz4rs:
(re-running to get Codacy/PR Quality Review to pass)

Issue #1111, opened 2020-10-13
Active messenger fails when sending over 2 GiB of data due to MPI limitations


lifflander:
**Describe the bug**
Due to limitations in MPI, the code fails when sending a single payload over 2 GiB due to overflow.

lifflander:
for release, branch `1111-try-port-to-release` is squashed and rebased on 1.0.0

PR #1135, opened 2020-11-03
1134 Fix tests not running

### LINKED ISSUE #1134: Make all tests run properly and fix bugs in tests that were unintentionally disabled
**Describe the bug**
A bug in googletest cmake is causing some tests not to run, which have bugs in them now.

lifflander:
Fixes #1134 

Pulled out from work stemming from #1111, which was interleaved

### DIFF
--- a/cmake-modules/GoogleTest.cmake
+++ b/cmake-modules/GoogleTest.cmake
@@ -353,8 +353,6 @@ function(gtest_add_tests)
         continue()
       endif()
 
-      message(STATUS "Test ${test_type} NAME=${gtest_test_name}")
-
       # Make sure tests disabled in GTest get disabled in CTest
       if(gtest_test_name MATCHES "(^|\\.)DISABLED_")
         # Add the disabled test if CMake is new enough
--- a/cmake-modules/GoogleTest.cmake
+++ b/cmake-modules/GoogleTest.cmake
@@ -5,6 +5,8 @@
 GoogleTest
 ----------
 
+.. versionadded:: 3.9
+
 This module defines functions to help use the Google Test infrastructure.  Two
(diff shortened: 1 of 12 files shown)


docker-compose.yml
```diff
@@ -256,6 +257,7 @@ services:
     environment:
       <<: *ccache
       <<: *vtopts
+      OMPI_MCA_btl: "^vader"
```
PhilMiller, line 260: Would it be possible to instead add the relevant `ptrace()` permission mentioned on the OMPI issue, so that our testing more readily reflects well-configured production environments?

lifflander: The problem with that is I believe I can't do that universally across all launches. I could add it in the Github workflow script when launches docker, but if you're running locally, I'm not sure I can add it easily. This is the option you are referring to:

`docker run --cap-add=SYS_PTRACE ...`

https://github.com/open-mpi/ompi/issues/4948

PhilMiller: Ok, that's a good point.

PhilMiller (approved):

cmake-modules/GoogleTest.cmake
```diff
@@ -298,14 +329,15 @@ function(gtest_add_tests)
   unset(testList)
 
   set(gtest_case_name_regex ".*\\( *([A-Za-z_0-9]+) *, *([A-Za-z_0-9]+) *\\).*")
-  set(gtest_test_type_regex "(TYPED_TEST|TEST_?[FP]?)")
+  #set(gtest_test_type_regex "(TYPED_TEST|TEST_?[FP]?)")
```
jstrzebonski, line 332: Commented out code

lifflander: Removed

Issue #1110, opened 2020-10-13
NodeStats test failing on develop for Intel 18.03


lifflander:
**Describe the bug**
Here's the output from CI (https://github.com/DARMA-tasking/vt/runs/1249034540)

```
        Start  71: vt:*/TestNodeStatsDumper.test_node_stats_dumping_with_interval/*_proc_2
2797
 71/263 Test  #71: vt:*/TestNodeStatsDumper.test_node_stats_dumping_with_interval/*_proc_2 ............***Failed  Required regular expression not found.Regex=[PASSED
2798
]  0.36 sec
2799
Note: Google Test filter = */TestNodeStatsDumper.test_node_stats_dumping_with_interval/*
2800
[==========] Running 10 tests from 1 test suite.
2801
[----------] Global test environment set-up.
2802
[----------] 10 tests from NodeStatsDumperExplode/TestNodeStatsDumper
2803
[ RUN      ] NodeStatsDumperExplode/TestNodeStatsDumper.test_node_stats_dumping_with_interval/0
2804
Note: Google Test filter = */TestNodeStatsDumper.test_node_stats_dumping_with_interval/*
2805
[==========] Running 10 tests from 1 test suite.
2806
[----------] Global test environment set-up.
2807
[----------] 10 tests from NodeStatsDumperExplode/TestNodeStatsDumper
2808
[ RUN      ] NodeStatsDumperExplode/TestNodeStatsDumper.test_node_stats_dumping_with_interval/0
2809
vt: Runtime Initializing: interop=true: mode: single-thread per rank
2810
vt: Program: collection_extended (./collection_extended)
2811
vt: Running on: 2 nodes
2812
vt: Machine Hostname: f927140265be
2813
vt: MPI Version: 3.1
2814
vt: MPI Max tag: 268435455
2815
vt: Build SHA: a1be253a849a3db00a895df3f2f002413827e49c
2816
vt: Build Ref: refs/heads/develop
2817
vt: Description: heads/develop-0-ga1be253a84 
2818
vt: Compile-time Features Enabled:
2819
vt: 	Tracing Projections
2820
vt: 	C++ Trait Detector
2821
vt: 	Load Balancing for Collections
2822
vt: 	OpenMP Threading
2823
vt: 	Message priorities
2824
vt: 	MPI access guards
2825
vt: Runtime Configuration:
2826
vt: 	Option: flag --vt_sched_num_progress on: Running MPI progress 2 times each invocation
2827
[...]

PR #1137, opened 2020-11-04
1122 elementstats is leaking

### LINKED ISSUE #1122: ElementStats is leaking memory from previous phases
**Describe the bug**
`ElementStats` inside collection elements is leaking memory.

nlslatt:
The vectors holding stats in ElementStats grow without bound.  This PR erases stats data for phases older than the lookback.

Fixes #1122

### DIFF
--- a/src/vt/vrt/collection/balance/elm_stats.cc
+++ b/src/vt/vrt/collection/balance/elm_stats.cc
@@ -113,17 +113,6 @@ void ElementStats::recvToNode(
   recvComm(key, bytes);
 }
 
-void ElementStats::setModelWeight(TimeType const& time) {
-  cur_time_started_ = false;
-  addTime(time);
-
-  vt_debug_print(
-    lb, node,
-    "ElementStats: setModelWeight: time={}, cur_load={}\n",
-    time, phase_timings_.at(cur_phase_)
-  );
-}
-
 void ElementStats::addTime(TimeType const& time) {
   phase_timings_.resize(cur_phase_ + 1);
   phase_timings_.at(cur_phase_) += time;
--- a/src/vt/vrt/collection/balance/elm_stats.cc
+++ b/src/vt/vrt/collection/balance/elm_stats.cc
@@ -132,6 +132,13 @@ void ElementStats::updatePhase(PhaseType const& inc) {
   );
 
   cur_phase_ += inc;
+
+  // Access all table entries for current phase, to ensure presence even
+  // if they're left empty
+  phase_timings_[cur_phase_];
+  subphase_timings_[cur_phase_];
+  phase_comm_[cur_phase_];
+  subphase_comm_[cur_phase_];
 }
 
 PhaseType ElementStats::getPhase() const {
--- a/src/vt/vrt/collection/balance/elm_stats.cc
+++ b/src/vt/vrt/collection/balance/elm_stats.cc
@@ -224,6 +224,22 @@ SubphaseType ElementStats::getSubPhase() const {
   return cur_subphase_;
 }
 
+std::size_t ElementStats::getLoadPhaseCount() const {
+  return phase_timings_.size();
+}
+
+std::size_t ElementStats::getCommPhaseCount() const {
+  return comm_.size();
+}
+
+std::size_t ElementStats::getSubphaseLoadPhaseCount() const {
+  return subphase_timings_.size();
+}
+
+std::size_t ElementStats::getSubphaseCommPhaseCount() const {
+  return subphase_comm_.size();
+}
+
 /*static*/
 void ElementStats::setFocusedSubPhase(VirtualProxyType collection, SubphaseType subphase) {
   focused_subphase_[collection] = subphase;
--- a/src/vt/vrt/collection/balance/elm_stats.cc
+++ b/src/vt/vrt/collection/balance/elm_stats.cc
@@ -82,7 +82,7 @@ void ElementStats::stopTime() {
 void ElementStats::recvComm(
   LBCommKey key, double bytes
 ) {
-  comm_[cur_phase_][key].receiveMsg(bytes);
+  phase_comm_[cur_phase_][key].receiveMsg(bytes);
   subphase_comm_[cur_phase_].resize(cur_subphase_ + 1);
   subphase_comm_[cur_phase_].at(cur_subphase_)[key].receiveMsg(bytes);
 }
@@ -170,7 +170,7 @@ TimeType ElementStats::getLoad(PhaseType phase, SubphaseType subphase) const {
 
 CommMapType const&
 ElementStats::getComm(PhaseType const& phase) {
-  auto const& phase_comm = comm_[phase];
+  auto const& phase_comm = phase_comm_[phase];
 
   vt_debug_print(
     lb, node,
@@ -206,7 +206,7 @@ void ElementStats::releaseStatsFromUnneededPhases(PhaseType phase, unsigned int 
   if (phase >= look_back) {
     phase_timings_.erase(phase - look_back);
     subphase_timings_.erase(phase - look_back);
-    comm_.erase(phase - look_back);
+    phase_comm_.erase(phase - look_back);
     subphase_comm_.erase(phase - look_back);
   }
 }
@@ -216,7 +216,7 @@ std::size_t ElementStats::getLoadPhaseCount() const {
 }
 
 std::size_t ElementStats::getCommPhaseCount() const {
-  return comm_.size();
+  return phase_comm_.size();
 }
 
 std::size_t ElementStats::getSubphaseLoadPhaseCount() const {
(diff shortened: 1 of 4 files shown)


src/vt/vrt/collection/balance/elm_stats.h
```diff
@@ -93,6 +92,17 @@ struct ElementStats {
   void setSubPhase(SubphaseType subphase);
   SubphaseType getSubPhase() const;
 
+  /**
+   * \internal \brief Cleanup after LB runs
+   */
+  void startIterCleanup(PhaseType phase, unsigned int look_back);
```
jstrzebonski, line 98: I kind of don't like the name. Does it 'start' the cleaning up (so it gets started and continues), or just cleans up and move on?

Maybe simple `void cleanUp(...)` would be sufficient?

src/vt/vrt/collection/balance/elm_stats.h
```diff
@@ -93,6 +92,17 @@ struct ElementStats {
...
+
+  // these are just for unit testing
+  std::size_t getLoadPhaseCount() const;
+  std::size_t getCommPhaseCount() const;
+  std::size_t getSubphaseLoadPhaseCount() const;
+  std::size_t getSubphaseCommPhaseCount() const;
```
jstrzebonski, line 104: IMHO that's smelly code. We're adding some functions only to do unit tests. I would say that it might be better to make it possible to somehow inject ElementStats-ish class into `Migratable`. That way we could write a struct deriving from ElementStats with added test functionalities such as those, and have it explicitly for unit testing.

Such refactoring might get big, so maybe a separate issue for doing this would be a good idea.

src/vt/vrt/collection/balance/elm_stats.h
```diff
@@ -110,12 +120,12 @@ struct ElementStats {
   bool cur_time_started_ = false;
   TimeType cur_time_ = 0.0;
   PhaseType cur_phase_ = fst_lb_phase;
-  std::vector<TimeType> phase_timings_ = {};
-  std::vector<CommMapType> comm_ = {};
+  std::unordered_map<PhaseType, TimeType> phase_timings_ = {};
+  std::unordered_map<PhaseType, CommMapType> comm_ = {};
```
jstrzebonski, line 124: We have `phase_timings`, `subphase_timings_` and `subphase_comm_`. Could we rename this (for consistency) to `phase_comm_`?

src/vt/vrt/collection/balance/elm_stats.h
```diff
@@ -110,12 +120,12 @@ struct ElementStats {
...
 
   SubphaseType cur_subphase_ = 0;
-  std::vector<std::vector<TimeType>> subphase_timings_ = {};
-  std::vector<std::vector<CommMapType>> subphase_comm_ = {};
+  std::unordered_map<PhaseType, std::vector<TimeType>> subphase_timings_ = {};
+  std::unordered_map<PhaseType, std::vector<CommMapType>> subphase_comm_ = {};
```
jstrzebonski, line 128: This might not make much sense (performance/memory overhead?), but if we change this to `unordered_map<PhaseType, unordered_map<SubphaseType, ...>>` we might simplify code by getting rid of all `resize()`s and `at()`s.

src/vt/vrt/collection/balance/elm_stats.h
nlslatt: The name was set to match what's being used in `NodeStats`, but you're right that it needs a better name.  I'm calling it `releaseStatsFromUnneededPhases` now.

src/vt/vrt/collection/balance/elm_stats.h
nlslatt: Agreed.  Making the change.

src/vt/vrt/collection/balance/elm_stats.h
nlslatt: I understand your complaint, but I don't fully understand your suggested solution.  Can you please elaborate?

tests/unit/collection/test_lbstats_retention.cc
```diff
@@ -0,0 +1,226 @@
...
+
+    #if vt_check_enabled(lblite)
+      if (phase > phases_needed) {
+        // syncNextPhase will have caused timings to be added for the
+        // next phase already
+        EXPECT_EQ(load_phase_count, phases_needed + 1);
+        EXPECT_EQ(sp_load_phase_count, phases_needed + 1);
+
+        EXPECT_EQ(comm_phase_count, phases_needed);
+        EXPECT_EQ(sp_comm_phase_count, phases_needed);
```
PhilMiller, line 85: The offset between the timing and communication structures after phase 0 is a bit bothersome. Is there a reasonable adjustment in the way these are kept that could equalize them?

PhilMiller (approved):
This all looks good to me. Your choice whether to address the mismatch I noted.

tests/unit/collection/test_lbstats_retention.cc
nlslatt: The current phase as tracked by `ElementStats` is incremented by a call to `updatePhase` made from `syncNextPhase`.  That causes the timings for the calls to `syncNextPhase` to be added to the loads for the next phase, which is appropriate because `syncNextPhase` already consumed the loads for the current phase.  I haven't thought of any way to equalize them without breaking the correctness.

src/vt/vrt/collection/balance/elm_stats.h
jstrzebonski: That's far from the solution, just a suggestion to consider (probably) for the future.

Generally, I was thinking if some sort of dependency injection might be possible (and might make sense) in this case, so that all those functions could be moved to another class, let's say, `TestElementStats : public ElementStats {...};`, and that class could be then used for unit testing.

But again - I don't think that's an easy change, but rather sizeable refactoring. So if that sounds reasonable it might be worth to write an issue for it. And if not, that's OK with me.

PhilMiller:
The failure of `141 - vt:*/TestNodeStatsDumper.test_node_stats_dumping_with_interval/*_proc_4` on clang-8/macosx/mpich looks like it could be real?

src/vt/vrt/collection/balance/elm_stats.cc
```diff
@@ -147,17 +132,13 @@ void ElementStats::updatePhase(PhaseType const& inc) {
   );
 
   cur_phase_ += inc;
```
PhilMiller, line 134: Would ensuring equality of all of the structure sizes be as simple as accessing each of them here:

```c++
// Access all table entries for current phase, to ensure presence even if they're left empty
phase_timings_[cur_phase_];
subphase_timings_[cur_phase_];
phase_comm_[cur_phase_];
subphase_comm_[cur_phase_];
```

nlslatt: Oh, I think I misunderstood one of your previous comments.  Is your goal to eliminate the difference in sizes between the comm-based stats data and non-comm-based?  I interpreted what you said before as being concerned about the sizes of the non-comm-based ones jumping by 2 only at the end of the first phase.  I can make your suggested change.

jstrzebonski:
> 
> 
> The failure of `141 - vt:*/TestNodeStatsDumper.test_node_stats_dumping_with_interval/*_proc_4` on clang-8/macosx/mpich looks like it could be real?

Don't want to say yes or now, but given that it failed only on one build it might be just flaky :'- (
https://github.com/DARMA-tasking/vt/issues/1110

src/vt/vrt/collection/balance/elm_stats.cc
PhilMiller: Yeah, I was bothered by seeing the tests check for `phases_needed + 1` on some structures, and `phases_needed` on others.

nlslatt: Fixed.

tests/unit/collection/test_lbstats_retention.cc
nlslatt: Fixed, now that I understand what you meant.

nlslatt:
> > The failure of `141 - vt:*/TestNodeStatsDumper.test_node_stats_dumping_with_interval/*_proc_4` on clang-8/macosx/mpich looks like it could be real?
> 
> Don't want to say yes or now, but given that it failed only on one build it might be just flaky :'- (
> There's an issue for this test -> #1110

It definitely looks like a legitimate failure, but I'm not sure yet if it's related to my changes.

PhilMiller (approved):

src/vt/vrt/collection/balance/elm_stats.h
nlslatt: I'll leave it to @lifflander to make this call.

lifflander: I think in this case, it's not worth doing any heavy refactor. These are const methods and shouldn't hurt anything. In general, we have this issue elsewhere so we could discuss a broader solution in separate work.

lifflander (approved):
Looks good to me

PR #1140, opened 2020-11-06
1139 Insert checkpoint documentation

### LINKED ISSUE #1139: Insert documentation from checkpoint
**What Needs to be Done?**

- [x] Add documentation directory from checkpoint
- [x] Update the navigation bar

uhetmaniuk:
Closes #1139


uhetmaniuk:
WARNING!! This pull request depends on merging the [checkpoint pull-request #153](https://github.com/DARMA-tasking/checkpoint/pull/153)  (which adds a documentation to checkpoint).

lifflander:
> @lifflander
> I made two commits via Github web-interface where I forgot to include the issue number.
> It appears that there are also unsigned (?) commits.
> 
> How can I fix these problems to pass all the checks?

- For signing the commits, you need to [set up GPG](https://git-scm.com/book/en/v2/Git-Tools-Signing-Your-Work) with git so it automatically signs them and register your key with GitHub.
- Rewrite all the commits using `git rebase -i` to amend the commits with the issue number. This will add the signature also if you have that working before you rebase.

Also, it looks like your rebase got really screwed up since there are commits that shouldn't be here. Are you rebasing or merging when the branch gets behind. You should always rebase on develop to keep the history clean.

PR #1141, opened 2020-11-06
1128 stats files will contain tempids

### LINKED ISSUE #1128: Comm/Subphase data in stats files will have object temporary IDs
**Describe the bug**

`NodeStats::startIterCleanup` rewrites its stats storage phase by phase to use permanent IDs instead of temporary IDs so that they're consistent in the dumped stats file regardless of where objects have migrated. This conversion is missing for subphase and communication data, and so those entries in stats files will exist in terms of temporary IDs.

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

nlslatt:
This PR converts temporary IDs to permanent IDs in subphase timings that are dumped to stats files.  It also addresses memory growth in the subphase comm stats.

Fixes #1128

### DIFF
--- a/src/vt/vrt/collection/balance/node_stats.cc
+++ b/src/vt/vrt/collection/balance/node_stats.cc
@@ -160,6 +160,7 @@ void NodeStats::startIterCleanup(PhaseType phase, unsigned int look_back) {
     node_data_.erase(phase - look_back);
     node_subphase_data_.erase(phase - look_back);
     node_comm_.erase(phase - look_back);
+    node_subphase_comm_.erase(phase - look_back);
   }
 
   // Create migrate lambdas and temp to perm map since LB is complete
--- a/src/vt/vrt/collection/balance/node_stats.cc
+++ b/src/vt/vrt/collection/balance/node_stats.cc
@@ -132,9 +132,7 @@ void NodeStats::clearStats() {
 }
 
 void NodeStats::startIterCleanup(PhaseType phase, unsigned int look_back) {
-  // TODO: Add in subphase support here too
-
-  // Convert the temp ID node_data_ for the last iteration into perm ID for
+  // Convert the temp ID node_data_ for the last phase into perm ID for
   // stats output
   auto const prev_data = std::move(node_data_[phase]);
   std::unordered_map<ElementIDType,TimeType> new_data;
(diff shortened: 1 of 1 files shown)


lifflander (approved):
This looks correct to me. We could (maybe?) refactor this into a method.. but I think it's fine

nlslatt:
> This looks correct to me. We could (maybe?) refactor this into a method.. but I think it's fine

I had started down that road when I thought that all four stats structures would need the same treatment, but wasn't sure it was worth it just for the two, especially because there are plans to get rid of temporary IDs.  I'm open to it, however.

PhilMiller (approved):

JacobDomagala (approved):

src/vt/vrt/collection/balance/node_stats.cc
```diff
@@ -146,10 +144,23 @@ void NodeStats::startIterCleanup(PhaseType phase, unsigned int look_back) {
   }
   node_data_[phase] = std::move(new_data);
 
+  // Convert the temp ID node_subphase_data_ for the last phase into perm ID for
+  // stats output
+  auto const prev_subphase_data = std::move(node_subphase_data_[phase]);
```
jstrzebonski, line 149: I'm not sure what's the intention here, but IMHO you're creating a copy of a `SubphaseLoadMapType` (which is `unordered_map` of `vectors`), so that might become resource-consuming someday.

Wouldn't simple `auto const& prev_subphase_data = node_subphase_data_[phase];` be sufficient?

src/vt/vrt/collection/balance/node_stats.cc
```diff
@@ -146,10 +144,23 @@ void NodeStats::startIterCleanup(PhaseType phase, unsigned int look_back) {
...
+  std::unordered_map<ElementIDType, std::vector<TimeType>> new_subphase_data;
+  for (auto& elm : prev_subphase_data) {
```
jstrzebonski, line 151: `elm` is IMO const already, because `prev_subphase_data` is const, but still, you might want to add it for readability.

src/vt/vrt/collection/balance/node_stats.cc
jstrzebonski: Ah, it's the move assignment. Just ignore this comment ;- )

jstrzebonski (approved):

PhilMiller (approved):

nlslatt:
The second commit on this was already merged with develop from #1155.  This PR may get closed without merging due to #1153.

nlslatt:
This is replaced with #1418 because the issue no longer exists on develop.

PR #1146, opened 2020-11-14
#973: implement default proxy

### LINKED ISSUE #973: Implement NodeProxy as default objgroup for bare handlers
**What Needs to be Done?**

Currently, objgroups go through a different API and always have an object associated with them. While for many users objgroups are ideal sometimes a bare handler is needed or more useful. Much of *vt* is implemented using bare handlers.

Create a special objgroup proxy instance (and maybe a specialization with separate logic?) that sends messages to bare handlers using a symmetrical proxy interface.

jstrzebonski:
Fixes #973

### DIFF
--- a/examples/callback/callback_context.cc
+++ b/examples/callback/callback_context.cc
@@ -108,8 +108,9 @@ int main(int argc, char** argv) {
     auto cb = vt::theCB()->makeFunc<DataMsg,MyContext>(
       vt::pipe::LifetimeEnum::Once, &my_global_ctx, callbackFn
     );
-    auto msg = vt::makeMessage<CallbackMsg>(cb);
-    vt::theMsg()->sendMsg<CallbackMsg,handler>(1, msg);
+
+    auto const default_proxy = vt::theObjGroup()->getDefault();
+    default_proxy[1].send<CallbackMsg, handler>(cb);
   }
 
   vt::finalize();
--- a/src/vt/objgroup/manager.cc
+++ b/src/vt/objgroup/manager.cc
@@ -52,6 +52,10 @@
 
 namespace vt { namespace objgroup {
 
+proxy::DefaultProxyType ObjGroupManager::getDefault() const {
+  return proxy::DefaultProxyType{};
+}
+
 ObjGroupProxyType ObjGroupManager::getProxy(ObjGroupProxyType proxy) {
   return proxy;
 }
--- a/src/vt/objgroup/manager.h
+++ b/src/vt/objgroup/manager.h
@@ -111,6 +111,14 @@ struct ObjGroupManager : runtime::component::Component<ObjGroupManager> {
    */
 
   /**
+   * \brief Construct a special proxy instance that allows sending, broadcasting
+   * and reducing without actual object group.
+   *
+   * \return proxy to the object group
+   */
+  proxy::DefaultProxyType getDefault() const;
+
+  /**
    * \brief Collectively construct a new object group. Allocates and constructs
    * the object on each node by forwarding constructor arguments.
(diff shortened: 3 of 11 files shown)


jstrzebonski:
@lifflander that's only a draft, but is it something you had in mind?

examples/callback/callback_context.cc
```diff
@@ -108,8 +140,9 @@ int main(int argc, char** argv) {
     auto cb = vt::theCB()->makeFunc<DataMsg,MyContext>(
       vt::pipe::LifetimeEnum::Once, &my_global_ctx, callbackFn
     );
-    auto msg = vt::makeMessage<CallbackMsg>(cb);
-    vt::theMsg()->sendMsg<CallbackMsg,handler>(1, msg);
+
+    auto const& default_proxy = DefaultProxy::get();
```
jstrzebonski, line 144: Could be simple var creation as well (`DefaultProxy default_proxy;`) but I thought that it might be more elegant to do that this way.

jstrzebonski:
At first, I misunderstood the description completely, but now I think it's closer to what you had in mind, is it @lifflander ?

Still, some cleanup is needed - move implementations to `.impl` headers, add oxygen comments, and actually use new implementation around the code.

lifflander:
@jstrzebonski Yes, this looks like exactly what I want; I like the `void` overload! Can you add some of the other methods also, like the broadcast and reduce?

jstrzebonski:
@lifflander I've just pushed `broadcast`, but I'm not sure if so much code is needed - I mean, internally I can see four "types" of broadcasts. Also sometimes the message is used outside of broadcast as well **[1]**, so that's why I've added functions `broadcastMsg()` that takes a msg, and `broadcast()`, that takes args and takes care of making massage inside.

But as we spoke yesterday, new API won't be used internally by vt, so I wonder if that makes sense for such variety exposed by default proxy, or maybe just one function **[2]** would be sufficient?

**[1]** example of such modification of message before broadcasting `examples/collection/transpose.cc solveGroupSetup()`:
```
  vt::theGroup()->newGroupCollective(
    is_even_node, [=](vt::GroupType group_id){
      fmt::print("Group is created: id={:x}\n", group_id);
      if (this_node == 1) {
        auto msg = vt::makeMessage<SubSolveMsg>(coll_proxy);
        vt::envelopeSetGroup(msg->env, group_id); 
        vt::theMsg()->broadcastMsg<SubSolveMsg,SubSolveInfo::subSolveHandler>(msg);
      }
    }, true
  );
```

**[2]** the only function that exposing of might be sufficient:
```
  template <typename MsgT, ActiveTypedFnType<MsgT>* f, typename... Args>
  messaging::PendingSend broadcast(Args&&... args) const {
    return broadcastMsg<MsgT, f>(
      makeMessage<MsgT>(std::forward<Args>(args)...)
    );
  }
```

lifflander:
> message

I think that 2 forms of broadcasting should be provided: `broadcastMsg` and `broadcast`. Sometimes it really makes a lot more sense to construct the message before it is used (setting members, etc.)

jstrzebonski:
@jstrzebonski wherever you're able to move implementation to `*.impl.h` files.

lifflander:
Can you rename `makeCollective` to `getDefault`?

jstrzebonski:
> Can you rename `makeCollective` to `getDefault`?

Will do!
Also waiting for #1178 to be merged before merging this one.

lifflander (approved):
This looks good to me!

cz4rs (approved):

JacobDomagala (approved):

PR #1164, opened 2020-11-26
1163: fix initialization with no arguments

### LINKED ISSUE #1163: Calling `vt::initialize` with no arguments silently makes VT fail
**Describe the bug**
A simple program that just calls `vt::initialize()` (without passing `argc, argv`) compiles and seems to do nothing.

cz4rs:
fixes #1163

### DIFF
--- a/tests/unit/runtime/test_initialization.nompi.cc
+++ b/tests/unit/runtime/test_initialization.nompi.cc
@@ -1,0 +1,63 @@
+/*
+//@HEADER
+// *****************************************************************************
+//
+//                        test_initialization.nompi.cc
+//                           DARMA Toolkit v. 1.0.0
+//                       DARMA/vt => Virtual Transport
+//
+// Copyright 2020 National Technology & Engineering Solutions of Sandia, LLC
+// (NTESS). Under the terms of Contract DE-NA0003525 with NTESS, the U.S.
+// Government retains certain rights in this software.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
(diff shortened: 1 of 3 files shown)


cz4rs:
`No space left on device` spotted, I'll check if this happens regularly

Screenshot from 2020-11-27 12-53-16

lifflander (approved):
Looks great! Can you write a unit test for this?

tests/unit/runtime/test_initialization.nompi.cc
```diff
@@ -0,0 +1,63 @@
...
+#include <vt/transport.h>
+
+namespace vt { namespace tests { namespace unit {
+
+using TestInitialization = TestHarness;
+
+TEST_F(TestInitialization, test_initialize_no_args) {
+  vt::initialize();
+
+  EXPECT_EQ(theConfig()->prog_name, "vt_unknown");
```
cz4rs, line 58: not much to check here, errors during initialization will result in `vtAbort()` being called (guaranteeing test failure)

tests/unit/runtime/test_initialization.nompi.cc
```diff
@@ -0,0 +1,63 @@
...
+
+#include <gtest/gtest.h>
+
+#include "test_harness.h"
+
+#include <vt/transport.h>
+
+namespace vt { namespace tests { namespace unit {
+
+using TestInitialization = TestHarness;
```
cz4rs, line 53: `TestHarness` is used instead of `TestParalellHarness` - if this proves to be problematic, `TestParallelHarness` can be used with `SetUp()` method overriden.

PR #1165, opened 2020-11-26
1161 CMake targets not correctly exporting lib

### LINKED ISSUE #1161: The VT targets are not correctly exporting lib (TPL) targets
**Describe the bug**
VT fails when used a separate package (`find_package`).

lifflander:
Fixes #1161


lifflander:
Before this is merged, I need to write the new CI tests that use `vt-sample-project` to test this automatically

lifflander:
So the reason this is failing is because the intel-18 container uses an older version of cmake:

https://stackoverflow.com/questions/34443128/cmake-install-targets-in-subdirectories

Evidently, this was fixed in cmake 3.13, but we only require 3.10. EMPIRE also only requires 3.10 as the min version...

lifflander:
The follow-on issue https://github.com/DARMA-tasking/vt/issues/1174, will test with `vt-sample-project` that the install works correctly.

src/CMakeLists.txt
```diff
@@ -324,10 +325,31 @@ else()
   set(MIMALLOC_LIBRARY_EXPORT "")
 endif()
 
+if (${vt_libfort_enabled})
+  set(FORT_LIBRARY_EXPORT ${FORT_LIBRARY})
+  install(TARGETS ${FORT_LIBRARY} EXPORT ${VIRTUAL_TRANSPORT_LIBRARY})
+else()
+  set(FORT_LIBRARY_EXPORT "")
```
nmm0, line 332: Does this need to be explicitly set to blank if it's not a cache variable? Mostly curious if that caused any bugs if not erased.

nmm0 (approved):

lifflander:
Per conversation with @PhilMiller, I'm going to downgrade to 3.10 but display a warning when the user tries to install

src/CMakeLists.txt
lifflander: Removed as it is spurious

src/CMakeLists.txt
```diff
@@ -312,22 +313,46 @@ install(
   COMPONENT                 runtime
 )
 
-if (${vt_fcontext_enabled})
-  set(FCONTEXT_LIBRARY_EXPORT ${FCONTEXT_LIBRARY})
+if(CMAKE_VERSION VERSION_LESS "3.13.0")
+  message(
+    WARNING "VT: Install targets skipped, effects packages built with a VT install"
```
PhilMiller, line 318: Typo, should be 'affects'

PhilMiller, line 318: As with Checkpoint, this should call out the CMake version dependence, and indicate why

src/CMakeLists.txt
```diff
@@ -312,22 +313,46 @@ install(
...
+
+if (vt_find_detector_dep EQUAL 0 AND cmake_do_install)
+  install(TARGETS detector EXPORT ${VIRTUAL_TRANSPORT_LIBRARY})
+  set(DETECTOR_EXPORT detector)
+endif()
+
+if (vt_find_checkpoint_dep EQUAL 0 AND cmake_do_install)
+  install(TARGETS checkpoint EXPORT ${VIRTUAL_TRANSPORT_LIBRARY})
+  set(CHECKPOINT_EXPORT checkpoint)
+endif()
```
PhilMiller, line 346: Since these are all `AND cmake_do_install`, could you just wrap that around the whole batch of these conditions?

src/CMakeLists.txt
lifflander: Fixed

lifflander: Fixed

src/CMakeLists.txt
lifflander: Done, factored out

PR #1167, opened 2020-11-27
1166 Set proper phase when element is inserted

### LINKED ISSUE #1166: Dynamic insertions fail with LB due to incorrect phase number
**Describe the bug**
When insertions happen it, the `ElementStats` phase needs to match the `PhaseManager`.

@nmm0

lifflander:
Fixes #1166

### DIFF
--- a/src/vt/vrt/collection/manager.impl.h
+++ b/src/vt/vrt/collection/manager.impl.h
@@ -2740,6 +2740,8 @@ void CollectionManager::insert(
          */
         CollectionTypeAttorney::setup(new_vc, num_elms, cur_idx, untyped_proxy);
 
+        new_vc->getStats().updatePhase(thePhase()->getCurrentPhase());
+
         theCollection()->insertCollectionElement<ColT, IndexT>(
           std::move(new_vc), cur_idx, max_idx, map_han, untyped_proxy, false,
           mapped_node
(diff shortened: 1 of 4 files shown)


src/vt/vrt/collection/manager.impl.h
```diff
@@ -2545,6 +2545,8 @@ void CollectionManager::insert(
          */
         CollectionTypeAttorney::setup(new_vc, num_elms, cur_idx, untyped_proxy);
 
+        new_vc->getStats().updatePhase(thePhase()->getCurrentPhase());
```
PhilMiller, line 2548: Rather than doing this, and recording 'current phase' as a member of every element, why not have the collection elements actually just query the phase as needed?

PhilMiller (approved):

lifflander:
Refer to https://github.com/DARMA-tasking/vt/issues/1224

src/vt/vrt/collection/manager.impl.h
lifflander: We addressed this in the meeting from a while ago so I am going to merge this since bvh depends on this change.

lifflander:
Just fixed the final bug: a race condition between the transition to the next phase on multiple nodes and a non-local insert occurring. This ensures the transition is synchronized across nodes so `nextPhaseCollective` doesn't return until **all** ranks have completely transitioned to the next phase.

PR #1170, opened 2020-11-30
1169: CI: allocate gcc-7 build to new agent pool

### LINKED ISSUE #1169: Upgrade to paid provisioned resources in Azure cloud for some of the builds that are compute/memory/storage intensive
**What Needs to be Done?**

Start with gcc-7, due to storage requirements being exceeded. Then, move Intel/NVCC to get those builds to pass more quickly.

lifflander:
Fixes #1169


ci/azure/azure-gcc-7-ubuntu-mpich.yml
```diff
@@ -56,9 +56,15 @@ stages:
...
-      vmImage: 'ubuntu-18.04'
+      name: 'vt-agent-pool'
     timeoutInMinutes: 180
     steps:
+      - task: Bash@3
+        displayName: Install cmake
+        inputs:
+          targetType: 'inline'
+          script: |
+            docker login --username lifflander1 --password 55e8e524-2219-4369-8f3c-71599d6695c5
```
cz4rs, line 67: cleartext password

cz4rs (approved):
looks good as long as the `gcc-7` passes, please squash the commits / rewrite history a bit

cz4rs:
minor note: the gcc-7 build stage seems pretty slow :/

lifflander:
> minor note: the gcc-7 build stage seems pretty slow :/

Yes, I'm planning on sqashing. I just upgraded that build to 4 cores. Now it runs in 30 mins.

ci/azure/azure-gcc-7-ubuntu-mpich.yml
lifflander: Yes, I'm still not sure how to solve this problem.

PhilMiller: Every CI platform has some means of injecting secrets in an environment variable or such without exposing them in the scripts or output.

cz4rs: some random links:
https://stefanscherer.github.io/how-to-use-secrets-in-azure-pipelines/
https://docs.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch#secret-variables

lifflander:
Once I have the password issue fixed, I will merge this. I have to rebuild the base linux image to fix this with the credentials encrypted in it.

ci/azure/azure-gcc-7-ubuntu-mpich.yml
lifflander: Yeah, it's easy to do in Azure pipelines, but not easy with the virtual scale set. I did manage to figure it out the solution though.

lifflander: I've rewritten the history so it no longer exists

PhilMiller: And then changed that password, one presumes?

PR #1175, opened 2020-12-02
1173 omp_lock_t is undefined on Intel serial build

### LINKED ISSUE #1173: omp_lock_t is undefined on Intel serial build
**Describe the bug**
From an EMPIRE dev:

> If I build intel-serial (no openmp), I get the compiler error,
> /pscratch/nroberd/EMPIRE/TPL/vt/src/vt/utils/mutex/omp_mutex.h(57): error: identifier "omp_lock_t" is undefined
>    omp_lock_t omp_lock;
>

cz4rs:
fixes #1173

### DIFF
--- a/ci/azure/azure-intel-19-ubuntu-mpich.yml
+++ b/ci/azure/azure-intel-19-ubuntu-mpich.yml
@@ -1,0 +1,145 @@
+###############################################################################
+############## Warning this is a generated file---do not modify ###############
+###############################################################################
+
+name: PR tests (intel 19, ubuntu, mpich)
+
+trigger:
+  branches:
+    include:
+      - develop
+      - 1.0.0*
+
+pr:
+  branches:
+   include:
+     - '*'
+
+
+resources:
+- repo: self
+
+variables:
+  tag: '$(Build.BuildId)'
+  REPO: lifflander1/vt
+  ARCH: amd64
+  UBUNTU: 18.04
+  COMPILER_TYPE: intel
+  COMPILER: icc-19
+  BUILD_TYPE: release
+  ULIMIT_CORE: 0
+  CODE_COVERAGE: 0
+  VT_LB: 1
+  VT_TRACE: 1
+  VT_TRACE_RT: 0
+  VT_MIMALLOC: 0
+  VT_DOCS: 0
+  VT_ASAN: 0
+  VT_POOL: 0
+  VT_EXTENDED_TESTS: 0
+  VT_UNITY_BUILD: 1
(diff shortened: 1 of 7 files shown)


PhilMiller (changes requested):
I don't think this is suitable. This isn't a problem of an incomplete type, it's a type that's not declared at all. We need to use `#if _OPENMP` appropriately.

cz4rs:
@lifflander I assume that this is introduced by footprinting PR - in that case this one-liner should help


note: the description says:
> If I build intel-serial **(no openmp)**, I get the compiler error,

there is:
```cpp
#if vt_check_enabled(openmp)
```
covering all of `omp_mutex.h` content, so a build without openmp should skip this completely (?)

cz4rs:
> I don't think this is suitable. This isn't a problem of an incomplete type, it's a type that's not declared at all. We need to use `#if _OPENMP` appropriately.

@PhilMiller I was still writing when you posted this - true, I'm trying to see what's wrong with `#if vt_check_enabled(openmp)`

cz4rs:
ok, just to keep the discussion here: we have `clang-8, macosx` build which correctly detects the absence of OpenMP:
```
-- Could NOT find OpenMP_C (missing: OpenMP_C_FLAGS OpenMP_C_LIB_NAMES) 
-- Could NOT find OpenMP_CXX (missing: OpenMP_CXX_FLAGS OpenMP_CXX_LIB_NAMES) 
-- Could NOT find OpenMP (missing: OpenMP_C_FOUND OpenMP_CXX_FOUND) 
OpenMP not found: using std::thread for workers
```

it looks like this should be fixed on the CMake side rather then in C++ code

@lifflander can you ask the reporter for the CMake output / complete build log?

lifflander:
> ok, just to keep the discussion here: we have `clang-8, macosx` build which correctly detects the absence of OpenMP:
> 
> ```
> -- Could NOT find OpenMP_C (missing: OpenMP_C_FLAGS OpenMP_C_LIB_NAMES) 
> -- Could NOT find OpenMP_CXX (missing: OpenMP_CXX_FLAGS OpenMP_CXX_LIB_NAMES) 
> -- Could NOT find OpenMP (missing: OpenMP_C_FOUND OpenMP_CXX_FOUND) 
> OpenMP not found: using std::thread for workers
> ```
> 
> it looks like this should be fixed on the CMake side rather then in C++ code
> 
> @lifflander can you ask the reporter for the CMake output / complete build log?

I'm trying to get a full report. The dev is offline right now.. so I'm not sure when it will come

PR #1179, opened 2020-12-07
#1156: check if ccache exists before calling it

### LINKED ISSUE #1156: Check for ccache existence before assuming it exists in build_cpp.sh
**What Needs to be Done?**

On some machines it doesn't exist and causes an error if running manually.

jstrzebonski:
Fixes #1156


lifflander (approved):
Looks great

PR #1188, opened 2020-12-16
1183 reset the phase when restoring from file

### LINKED ISSUE #1183: Phase assertion fails during restart
**Describe the bug**
Phases must match ... (stats.getPhase() == msg->getPhase()) ...  (from vrt/collection/balance/elm_stats.impl.h function syncNextPhase).

cz4rs:
fixes #1183 

from discussion with Jonathan:

> If we assume that `theCollection()->restoreFromFile()` always is after a reinitialization of VT we can just set the phase in every element to 0 instead of deserializing it from file and getting the old phase (a true restart)

> The phase in `ElementStats` (embedded in the collection) needs to align with the user’s call to `theCollection()->nextPhase(proxy, phase)` so I think resetting it to zero is fine. Even if a full restart isn’t done, the user should just pass 0 to `nextPhase` since it was restored

### DIFF
--- a/src/vt/vrt/collection/balance/elm_stats.cc
+++ b/src/vt/vrt/collection/balance/elm_stats.cc
@@ -141,6 +141,10 @@ void ElementStats::updatePhase(PhaseType const& inc) {
   subphase_comm_[cur_phase_];
 }
 
+void ElementStats::resetPhase() {
+  cur_phase_ = fst_lb_phase;
+}
+
 PhaseType ElementStats::getPhase() const {
   return cur_phase_;
 }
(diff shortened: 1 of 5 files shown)


cz4rs:
waiting for the CI to see if this breaks anything

this should be covered by:
https://github.com/DARMA-tasking/vt/blob/8acd899aee752d2916966c0b275ef769c2a2d808/tests/unit/collection/test_checkpoint.extended.cc#L180-L205
but there is no assertion done on phase after deserialization

lifflander (approved):

lifflander:
This looks good to me!

lifflander:
> waiting for the CI to see if this breaks anything
> 
> this should be covered by:
> 
> https://github.com/DARMA-tasking/vt/blob/8acd899aee752d2916966c0b275ef769c2a2d808/tests/unit/collection/test_checkpoint.extended.cc#L180-L205
> 
> 
> but there is no assertion done on phase after deserialization

You might want to add a `nextPhaseCollective` to this so it is actually using phases to check for correctness

PR #1196, opened 2020-12-28
1195: passing argc and argv by value causes incorrect argument parsing

### LINKED ISSUE #1195: passing argc and argv by value causes incorrect argument parsing
Discovered by @PhilMiller:
> in https://github.com/DARMA-tasking/vt/pull/1164 the runtime initialization was changed to take `argc` and `argv` by value instead of reference. Way down the stack in `ArgConfig::parse` on the release branch, the code relied on that reference passing to strip vt arguments out of the command line that the application would see. One of EMPIRE’s unit tests now fails with the new release because it adds `--vt_no_terminate` to the command line arguments before initializing vt

Note: the exact commit introducing this is 24deca82bd7af4c9a3343d767e6aa824311b1aeb.

**Expected behavior**
Allow modifying command line arguments and add unit tests covering initialization **with** arguments (currently there is only one tests covering `vt::initialize` call without any arguments).

cz4rs:
fixes #1195

### DIFF
--- a/tests/unit/runtime/test_initialization.cc
+++ b/tests/unit/runtime/test_initialization.cc
@@ -70,8 +70,9 @@ struct TestInitialization : TestParallelHarness {
     custom_args.emplace_back(&prog_name[0]);
     custom_args.emplace_back(&cli_argument[0]);
     custom_args.emplace_back(&vt_no_terminate[0]);
+    custom_args.emplace_back(nullptr);
 
-    custom_argc = custom_args.size();
+    custom_argc = custom_args.size() - 1;
     custom_argv = custom_args.data();
     EXPECT_EQ(custom_argc, 3);
 
@@ -90,6 +91,7 @@ TEST_F(TestInitialization, test_initialize_with_args) {
   EXPECT_EQ(custom_argc, 2);
   EXPECT_STREQ(custom_argv[0], "vt_program");
   EXPECT_STREQ(custom_argv[1], "--cli_argument=100");
+  EXPECT_EQ(custom_argv[2], nullptr);
 }
 
 }}} // end namespace vt::tests::unit
--- a/tests/unit/runtime/test_initialization.cc
+++ b/tests/unit/runtime/test_initialization.cc
@@ -1,0 +1,95 @@
+/*
+//@HEADER
+// *****************************************************************************
+//
+//                           test_initialization.cc
+//                           DARMA Toolkit v. 1.0.0
+//                       DARMA/vt => Virtual Transport
+//
+// Copyright 2020 National Technology & Engineering Solutions of Sandia, LLC
+// (NTESS). Under the terms of Contract DE-NA0003525 with NTESS, the U.S.
+// Government retains certain rights in this software.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+// * Redistributions of source code must retain the above copyright notice,
+//   this list of conditions and the following disclaimer.
+//
+// * Redistributions in binary form must reproduce the above copyright notice,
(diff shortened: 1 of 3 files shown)


PhilMiller:
This seems to pass the full range of EMPIRE tests, including the ones that I missed in the pre-release review.

I'd be fine with merging this as is, but were you planning to add a unit test for the case with arguments?

cz4rs:
> This seems to pass the full range of EMPIRE tests, including the ones that I missed in the pre-release review.
> 
> I'd be fine with merging this as is, but were you planning to add a unit test for the case with arguments?

@PhilMiller  thanks for running the tests!
I couldn't get back to it yesterday, but I've just added a unit test for this. As far as this bug is concerned, it should be sufficient
(it's not pretty with all the borrowed from TestParallelHarness, but it fails as expected when the fix is removed).

- this is targeted at `develop`, how should I create a hotfix for `beta.10.3` - should I create a PR targeting `1.0.0-beta.10.3-proposed-update` branch?
- if more coverage for actual argument parsing is required I can either stuff more custom arguments here or create a separate issue to cover `ArgConfig::parse` only

PhilMiller:
Let's just make it beta.10.4

PhilMiller (approved):

tests/unit/runtime/test_initialization.cc
```diff
@@ -0,0 +1,96 @@
...
+};
+
+TEST_F(TestInitialization, test_initialize_with_args) {
+  EXPECT_EQ(theConfig()->prog_name, "vt_program");
+  EXPECT_EQ(theConfig()->vt_no_terminate, true);
+
+  EXPECT_EQ(custom_argc, 2);
+  EXPECT_STREQ(custom_argv[0], "vt_program");
+  EXPECT_STREQ(custom_argv[1], "--cli_argument=100");
+}
```
PhilMiller, line 94: If we want to be stringent here, there should be this as well
`EXPECT_EQ(custom_argv[2], nullptr);`

PhilMiller: And if that fails right now, we should fix it.

cz4rs: done

PhilMiller:
Looks good. Thanks!

PhilMiller:
Please update the stable branch version of this as well.

cz4rs:
side note: `TestParallelHarness` also requires `nullptr` treatment, I will create a separate issue for this
https://github.com/DARMA-tasking/vt/blob/3f7d76192af9da195abac31ee48d1e359f59c4e4/tests/unit/test_parallel_harness.h#L152-L184

```bash
> git grep addAdditionalArgs ./tests/
../tests/unit/active/test_active_send_large.cc:107:  void addAdditionalArgs() override {
../tests/unit/collection/test_lb.extended.cc:132:  virtual void addAdditionalArgs() override {
../tests/unit/runnable/test_invoke.cc:56:  virtual void addAdditionalArgs() override {
../tests/unit/test_parallel_harness.h:158:    addAdditionalArgs();
../tests/unit/test_parallel_harness.h:175:   *   virtual void addAdditionalArgs() override {
../tests/unit/test_parallel_harness.h:184:  virtual void addAdditionalArgs() {}
```

cz4rs:
@PhilMiller
it seems like Codacy job is still broken (it is marked as required) so I will not be able to merge this myself 
I will ping you once the nvidia builds pass

PhilMiller:
> side note: `TestParallelHarness` also requires `nullptr` treatment, I will create a separate issue for this
> 
> https://github.com/DARMA-tasking/vt/blob/3f7d76192af9da195abac31ee48d1e359f59c4e4/tests/unit/test_parallel_harness.h#L152-L184

Thanks for checking and noticing this.

PR #1113, opened 2020-10-14
1111 Fix for large message sends

lifflander:
Fixes #1111 

- Start with tests that fail for serialized/non-serialized messages

- ~Make a **major** improvement to how tests are being dispatched. Currently, tests are being ignored due to deficiencies in the gtest cmake. Pull in the latest gtest cmake and fix the bugs in that version of the cmake. **This is causing "new" tests to fail on this PR that weren't running before which I am systematically trying to fix**~
  - Addressed & merged in separate PR

- Implement "multi-sends" for serialized and non-serialized messages (now passing!)

- Add a new runtime argument `--vt_max_mpi_send_size=X` that defaults to `1ull << 30`.

- Use the new argument injector to set the size for testing to `16384` bytes

PR #1198, opened 2020-12-29
1197: prepare beta-10.4 release

### LINKED ISSUE #1197: prepare beta-10.4 release
**What Needs to be Done?**
Prepare `beta-10.4` release.
Issues:
- #1195

cz4rs:
fixes #1197

### DIFF
--- a/cmake-modules/CheckLinkerFlag.cmake
+++ b/cmake-modules/CheckLinkerFlag.cmake
@@ -1,0 +1,12 @@
+include(CheckCXXCompilerFlag)
+
+if(POLICY CMP0056)
+  # Honor link flags in try_compile() source-file signature.
+  cmake_policy(SET CMP0056 NEW)
+endif()
+
+# This is written because cmake does not support `check_linker_flag` until 3.18
+function(check_linker_flag flag out_var)
+  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} ${flag}")
+  check_cxx_compiler_flag("" ${out_var})
+endfunction()
--- a/tests/unit/runtime/test_initialization_with_args.cc
+++ b/tests/unit/runtime/test_initialization_with_args.cc
@@ -2,11 +2,11 @@
 //@HEADER
 // *****************************************************************************
 //
-//                                  startup.cc
+//                           test_initialization.cc
 //                           DARMA Toolkit v. 1.0.0
 //                       DARMA/vt => Virtual Transport
 //
-// Copyright 2019 National Technology & Engineering Solutions of Sandia, LLC
+// Copyright 2020 National Technology & Engineering Solutions of Sandia, LLC
 // (NTESS). Under the terms of Contract DE-NA0003525 with NTESS, the U.S.
 // Government retains certain rights in this software.
 //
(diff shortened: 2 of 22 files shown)


tests/unit/runtime/test_initialization_with_args.cc
```diff
@@ -0,0 +1,94 @@
...
+
+    // communicator is duplicated.
+    MPI_Comm comm = mpi_singleton->getComm();
+
+    static char prog_name[]{"vt_program"};
+    static char cli_argument[]{"--cli_argument=100"};
+    static char vt_no_terminate[]{"--vt_no_terminate"};
+    custom_args.emplace_back(&prog_name[0]);
+    custom_args.emplace_back(&cli_argument[0]);
+    custom_args.emplace_back(&vt_no_terminate[0]);
```
PhilMiller, line 72: For strict standards compliance (i.e. if the parser in cli11 relies on it now or in the future), you should push a `nullptr` on the end of this vector.

https://stackoverflow.com/questions/16418932/is-argvargc-equal-to-null-pointer

PhilMiller:
Tested successfully in EMPIRE with gcc. Nvidia and Intel are pending.

PhilMiller:
Intel tests pass. Testing on CUDA is slow.

PhilMiller (approved):

cz4rs:
@PhilMiller
Codacy and Travis jobs are stuck, other tests pass fine. As with the other PR, I cannot merge it myself (no permissions for override + I think merging into `1.0.0` is restricted as well)

if everything is ok, can you please merge this? (I will create the tag afterwards)

cz4rs:
Regarding the extended tests, I have added the build definitions, but it seems like they have not been triggered.

Honestly I don't know if there should be much investigation put into this, since extended tests are failing regularly on `develop` anyways (I believe this was discussed at the last DARMA weekly, but there's no solution so far)
Screenshot from 2020-12-30 15-51-33

PhilMiller:
We're looking at incorporating

- #915 Stacktrace generation
- A fix for missing deletion of https://github.com/DARMA-tasking/vt/pull/1113/files#diff-51dbee743b2faddab4d156c0fbf6707273c54f18e7b48a53480eefb455e0e2f0L100 in the backport onto beta 10.3
- #1107 assertions in CI builds

cmake-modules/CheckLinkerFlag.cmake
```diff
@@ -0,0 +1,14 @@
+include(CheckCXXCompilerFlag)
+
+if(POLICY CMP0056)
+  # Honor link flags in try_compile() source-file signature.
+  cmake_policy(SET CMP0056 NEW)
+endif()
+
+# This is written because cmake does not support `check_linker_flag` until 3.18
```
PhilMiller, line 8: Does this need a version check around it? Was that concern brought up in #915?

cz4rs:
note: #1098 has 2 PRs, but they were both done on `1.0.0-1098-enable-assertions-during-ci` branch

cmake-modules/CheckLinkerFlag.cmake
cz4rs: there's a conversation in this PR regarding the implementation of this, but no mention of version check actually

PhilMiller: @lifflander Is this going to cause issues with very recent CMake versions?

cz4rs: I did some reading and it seems that overriding builtin functions is not recommended / potentially problematic (especially if someone would do the same in EMPIRE) - meaning we should only define it  `if(${CMAKE_VERSION} VERSION_LESS "3.18.0")` (?)

source:
https://crascit.com/2018/09/14/do-not-redefine-cmake-commands/

cz4rs: disclaimer: I don't feel like an expert on this, so a confirmation would be nice before opening a bug (I guess this should be fixed on develop as well)

cz4rs: we use `CMake` 3.18.4 in all `vt` containers:
```bash
ci/docker/ubuntu-18.04-clang-cpp.dockerfile:39:RUN ./cmake.sh 3.18.4
```

```bash
root@9d78a3e16284:/# cmake --version
cmake version 3.18.4
```

after removing / "ifdefing" our implementation of `check_linker_flag`, I am getting a following error:
```
root@9d78a3e16284:/# /vt/ci/build_cpp.sh /vt /build
(...)
CMake Error at cmake/set_linker_flags.cmake:3 (check_linker_flag):
  Unknown CMake command "check_linker_flag".
  src/CMakeLists.txt:294 (include)

-- Configuring incomplete, errors occurred!
See also "/build/vt/CMakeFiles/CMakeOutput.log".
See also "/build/vt/CMakeFiles/CMakeError.log".
```

PhilMiller:
This passes the full EMPIRE test suite in the `Debug` mode builds that unexpectedly failed. I'm re-running the test in other modes to not repeat my earlier mess.

PhilMiller:
Manual and automatic testing of Release mode EMPIRE builds also passes. I'm re-running Intel 19 tests as well

PhilMiller:
Intel 19 tests all pass as well. I think this is good to go.

cz4rs:
> Intel 19 tests all pass as well. I think this is good to go.

:+1: 
@PhilMiller @lifflander 
If the CMake conversation above can be resolved as a non-issue, please merge this PR (I cannot do this myself - this is targeting release branch + Codacy is stuck again). I will create a tag & release notes afterwards.

lifflander:
I think Travis CI needs to be removed for merging on release

cmake-modules/CheckLinkerFlag.cmake
lifflander: I think this is fine because the policy is being invoked properly.

lifflander (approved):
Looks good to me. Let's merge.

PR #1208, opened 2021-01-07
1207 Fix Mac CI

### LINKED ISSUE #1207: Fix Mac CI target failure due to brew linking python
**Describe the bug**
Fails to link python during setup phase

[This seems to be the solution](https://support.circleci.com/hc/en-us/articles/360043542933--Error-The-brew-link-step-did-not-complete-successfully-When-Installing-Python)

lifflander:
Fixes #1207

### DIFF
--- a/.github/workflows/macosx-clang-mpich.yml
+++ b/.github/workflows/macosx-clang-mpich.yml
@@ -34,7 +34,7 @@ jobs:
           ${{ runner.os }}-macosx-clang-8-ccache-${{ secrets.GH_ACTIONS_CACHE_VERSION }}
     - name: Install Dependencies
       shell: bash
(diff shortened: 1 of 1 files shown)


lifflander:
Turns out that the Mac CI fix is easily than I thought--stop upgrading it before installing the packages.

cz4rs (approved):
Great!

PhilMiller:
I'm fine with this.

Does Homebrew keep older versions available indefinitely, so that whatever is referenced from the index in the image will install nicely? If not, this is going to break with some regularity and require manual updates. Still better than being continuously broken right now.

lifflander:
> I'm fine with this.
> 
> Does Homebrew keep older versions available indefinitely, so that whatever is referenced from the index in the image will install nicely? If not, this is going to break with some regularity and require manual updates. Still better than being continuously broken right now.

Not sure about homebrew... but I don't think they keep old versions on the platform. homebrew comes pre-installed on the image

PR #1209, opened 2021-01-08
1201: Add -Werror for CI builds that are warning clean

### LINKED ISSUE #1201: Add -Werror for CI builds that are warning clean
**What Needs to be Done?**

- [ ] Check which compiler versions in CI builds don't generate any warnings
- [ ] Add `-Werror` to the build configuration / scripts for those builds

The codebase should progressively become more warning-clean, and stay that way for settings that it already achieves that.

**Additional context**

cz4rs:
fixes #1201


cz4rs:
I was overly optimistic with this because I was using `Debug` configuration in local (which passes happily)

I will take a quick look at how much work it would take to make basic gcc and clang builds to pass

PhilMiller:
There's a funny quirk of compiler diagnostics, that some of them only become apparent with optimizations enabled, because the analysis that makes the need for a diagnostic apparent is part of an optimization pass. Those passes are skipped with optimizations disabled for the sake of performance.

src/vt/messaging/send_info.h
```diff
@@ -55,6 +55,7 @@ namespace vt { namespace messaging {
  * \brief Returned from a data send to be used to receive the data
  */
 struct SendInfo {
+  SendInfo() = default;
```
cz4rs, line 58: this was causing a warning for `vt::messaging::MultiMsg` (`active.cc`), which has default ctor

note: it might be better to fix this in some other way instead of allowing for default-construction

cz4rs:
note to self: 

- [ ] update documentation (`build` section) with new configuration variables
- [ ] make sure that builds fail as expected when a warning is introduced

cz4rs:
I have fixed a couple of simple warnings which allowed to enable `-Werror` for all builds except:
- `nvcc` builds (it will not fail even when warnings are present)
- extended builds
- `gcc-5` - raises `defined but not used` for the code which uses `#pragma GCC diagnostic ignored "-Wunused-variable"`
- `gcc-8` - asan build, complains about `vt::runtime::component::detail::DiagnosticValueWrapper` members being potentially used uninitialized (`[-Werror=maybe-uninitialized]`)

cz4rs:
note: this is **disabled** by default, enabled only for selected builds - up for debate if we should turn this around

src/vt/messaging/send_info.h
jstrzebonski: Shouldn't `MultiMsg() = default;` be removed? Is it used anywhere?

src/vt/vrt/collection/balance/stats_restart_reader.cc
```diff
@@ -148,21 +148,22 @@ void StatsRestartReader::inputStatsFile(
...
-        vtAssertExpr(res == 3);
-      } else {
-        // Load detected, create the new element
-        fsetpos (pFile,&pos);
-        if (prevPhaseID != phaseID) {
-          prevPhaseID = phaseID;
-          element_history.push_back(buffer);
-          buffer.clear();
+      fgetpos(pFile, &pos);
+      if (fscanf(pFile, "%c", &separator)) {
```
PhilMiller, line 152: Should this check have an `else` that indicates a malformed record/file?

tests/unit/collection/test_model_linear_model.nompi.cc
```diff
@@ -133,14 +133,14 @@ TEST_F(TestLinearModel, test_model_linear_model_1) {
       {ElementIDType{1}, TimeType{100}}, {ElementIDType{2}, TimeType{10}}},
   };
 
-  std::array<std::pair<TimeType, TimeType>, num_test_interations> expected_data{
+  std::array<std::pair<TimeType, TimeType>, num_test_interations> expected_data{{
```
jstrzebonski, line 136: `warning: suggest braces around initialization of subobject [-Wmissing-braces]` is a bug in clang -> https://bugs.llvm.org/show_bug.cgi?id=21629
I haven't checked it yet, but maybe there's a way to turn that off?

src/vt/messaging/send_info.h
PhilMiller: This exact warning has come up in the release used in EMPIRE, where they also use `-Werror`. I'll open an issue to address this in a 1.0.1

tests/unit/collection/test_model_linear_model.nompi.cc
PhilMiller: Since it sounds like it's only an issue in quite old compiler versions (e.g. clang <6.0), we should probably just add `-Wno-missing-braces` to the command lines used for those builds as additional `CXXFLAGS`, rather than changing the code. One of the comments on that bug noted that adding braces unnecessarily can actually lead to bugs.

src/vt/messaging/send_info.h
PhilMiller: #1211, and I just tested that removing the `MultiMsg` default ctor as @jstrzebonski suggested compiles all the tests and examples. That should be preferred.

tests/unit/collection/test_model_linear_model.nompi.cc
cz4rs: `-Wno-missing-braces` added

lifflander:
> * DiagnosticValueWrapper

For `DiagnosticValueWrapper`, we should change this line `T initial_value_;` to `T initial_value_ = {}`;

.env
```diff
@@ -2,8 +2,8 @@ REPO=lifflander1/vt
 ARCH=amd64
 UBUNTU=18.04
 ULIMIT_CORE=0
-COMPILER=gcc-7
```
lifflander, line 5: Is there a reason you are changing the default compiler in docker to clang-5?

src/vt/vrt/collection/balance/stats_restart_reader.cc
lifflander: Yes, we probably need more error checking in general for this code. Maybe a separate issue?

lifflander (approved):
Overall, looks good. I think the default compiler should probably be changed back unless there was a reason.

.env
cz4rs: nope, I have added this by mistake

cz4rs: reverted, good catch!

src/vt/vrt/collection/balance/stats_restart_reader.cc
cz4rs: #1213 created.

cz4rs:
- `-Werror` enabled for all `gcc` and `clang` builds except for `gcc-5` (which does not respect `pragma GCC diagnostic ignored "-Wunused-variable"` we have in the code)
- regarding `DiagnosticValueWrapper`, it might be better to simply remove default ctor for this class (requires https://github.com/DARMA-tasking/checkpoint/issues/161)
- #1213 created to handle error checking in `StatsRestartReader` properly

jstrzebonski (approved):
Looks good to me

JacobDomagala (approved):

PR #1214, opened 2021-01-13
1173: omp_lock_t is undefined on Intel serial build

### LINKED ISSUE #1173: omp_lock_t is undefined on Intel serial build

cz4rs:
fixes #1173


jstrzebonski:
@cz4rs I found this -> https://community.intel.com/t5/Intel-C-Compiler/Cause-for-remark-11074-Inlining-inhibited-by-limit-max-size/td-p/1073635 regarding Intel's warnings:
```
remark #11074: Inlining inhibited by limit max-size 
remark #11074: Inlining inhibited by limit max-total-size 
remark #11076: To get full report use -qopt-report=4 -qopt-report-phase ipo
```

lib/libfort/CMakeLists.txt
```diff
@@ -52,30 +52,39 @@ else()
...
+            ")
+    endif()
     #Supress
     if("${FORT_COMPILER}" STREQUAL "Intel")
     set(ADDITIONAL_WARNINGS "${ADDITIONAL_WARNINGS} \
         -wd279 \
         -wd2192 \
-        -wd2102")
+        -wd2102 \
+        -diag-disable:remark \
```
cz4rs, line 75: this disables all remarks, but I think it should be acceptable for `libfort` since we don't maintain the lib

cz4rs: reminder, we currently have:
```
remark #11074: Inlining inhibited by limit max-size 
remark #11074: Inlining inhibited by limit max-total-size 
remark #11076: To get full report use -qopt-report=4 -qopt-report-phase ipo
```

thanks @jstrzebonski for the hint

CMakeLists.txt
```diff
@@ -100,7 +98,7 @@ else()
   option(vt_libfort_enabled "Build VT with fort library enabled" OFF)
 endif()
 
-option(vt_fcontext_enabled "Build VT with fcontext (ULT) enabled" OFF)
+option(vt_fcontext_enabled "Build VT with fcontext (ULT) enabled" ON)
```
cz4rs, line 101: FIXME: to be reverted, enabling temporarily to check all builds for errors

cmake/threading_config.cmake
```diff
@@ -28,3 +28,9 @@ function(config_for_std_thread)
   set(vt_feature_cmake_openmp "0" PARENT_SCOPE)
   set(vt_feature_cmake_stdthread "1" PARENT_SCOPE)
 endfunction(config_for_std_thread)
+
+function(config_no_threading)
+  set(vt_feature_cmake_openmp "0" PARENT_SCOPE)
+  set(vt_feature_cmake_stdthread "0" PARENT_SCOPE)
+  option(vt_fcontext_enabled "Build VT with fcontext (ULT) enabled" ON)
```
cz4rs, line 35: cannot use simple `set` because of [CMP0077](https://cmake.org/cmake/help/latest/policy/CMP0077.html#policy:CMP0077)
(changed in CMake 3.13, we still support 3.10)

cz4rs:
- `OpenMP` and `std::thread` are disabled by default
- configuration in CMake should be consistent with the logic in the code now, based on `backend_no_threading` macro in `worker_group.h` / `worker_headers.h`
- `fcontext` remains disabled by default, but it will be enabled if neither `USE_OPENMP` nor `USE_STD_THREAD` is enabled (also, `vt_fcontext_enabled` CMake option remains functional)
- `intel 19` build uses CMake 3.10.2 and triggers following warnings:
```
CMake Warning at src/CMakeLists.txt:318 (message):
  VT: Install targets skipped because CMake version 3.10.2<3.13; affects
  packages built with a VT install.  CMake <3.13 does not support exporting
  targets from subdirectories
```
which would then lead to:
```
CMake Error in src/CMakeLists.txt:
  export called with target "vt" which requires target "fcontext" that is not
  in the export set.

  If the required target is not easy to reference in this call, consider
  using the APPEND option with multiple separate calls.
```
I've had to add abedc72 to make it work (CI seems to be happy with it, does it need any additional verification?)

TODO:
- [ ] update the documentation
- [ ] make sure that all threading configurations are tested in CI (for now everything uses `fcontext`, add `USE_OPENMP` and `USE_STD_THREAD` to some of the builds)
_it seems like we've had a gap in testing where `vt::worker::WorkerSeq` was never used (because of threading auto-detection always picking `OpenMP` or `std::thread`) and it has silently stopped to compile_

cz4rs:
current status:
- `OpenMP` and `std::thread` are disabled by default
- configuration in CMake should be consistent with the logic in the code now, based on `backend_no_threading` macro in `worker_group.h` / `worker_headers.h`
- `fcontext` remains disabled by default, but it will be enabled if neither `USE_OPENMP` nor `USE_STD_THREAD` is enabled (also, `vt_fcontext_enabled` CMake option remains functional)

installation / export:
- `intel 19` build uses CMake 3.10.2 and triggers following warnings / errors:
```
CMake Warning at src/CMakeLists.txt:318 (message):
  VT: Install targets skipped because CMake version 3.10.2<3.13; affects
  packages built with a VT install.  CMake <3.13 does not support exporting
  targets from subdirectories
(...)
CMake Error in src/CMakeLists.txt:
  export called with target "vt" which requires target "fcontext" that is not
  in the export set.

  If the required target is not easy to reference in this call, consider
  using the APPEND option with multiple separate calls.
```
I've had to add abedc72 to make it work (CI seems to be happy with it, does it need any additional verification in terms of installation?)

TODO:
- [x] update the documentation
- [x] make sure that all threading configurations are tested in CI (for now everything uses `fcontext`, add `USE_OPENMP` and `USE_STD_THREAD` to some of the builds)
_it seems like we've had a gap in testing where `vt::worker::WorkerSeq` was never used (because of threading auto-detection always picking `OpenMP` or `std::thread`) and it has silently stopped to compile_

----
tested configurations are:
| compiler | OpenMP | std::thread | fcontext (backend_no_threading) |
--- | --- | --- | ---
gcc | + | - | +
clang | - | + | +
nvidia | - | - | +
intel | - | - | +

(see `workflows-azure.ini` for the details)

- trying to use `OpenMP` with `clang` results in:
```
-- Using OpenMP for worker threading
-- Could NOT find OpenMP_C (missing: OpenMP_C_FLAGS OpenMP_C_LIB_NAMES) 
-- Could NOT find OpenMP_CXX (missing: OpenMP_CXX_FLAGS OpenMP_CXX_LIB_NAMES) 
-- Could NOT find OpenMP (missing: OpenMP_C_FOUND OpenMP_CXX_FOUND) 
-- OpenMP has been found: Linker="", CC FLAGS="NOTFOUND", CXX FLAGS="NOTFOUND"
CMake Error at cmake/load_threading_package.cmake:26 (message):
  requested OpenMP with -DUSE_OPENMP=On, but cannot find valid OpenMP in
  compiler
  cmake/load_packages.cmake:13 (include)
  CMakeLists.txt:61 (include)
```

- trying to use `gcc` with `std::thread` results in https://github.com/DARMA-tasking/vt/pull/1214#issuecomment-767122410

cz4rs:
note: original Intel error and clang error above might boil down to different implementations of OpenMP (`libgomp`, `libomp` and whatever Intel uses)

disabling threading by default fixes the problem as long as OpenMP support is not the goal here, but it might be necessary to create separate issues for desired configurations

lifflander (approved):
Overall, this looks good to me. On some systems fcontext might not be supported at all. Can we have an option to have no threading, including fcontext being shut off?

cz4rs:
@lifflander 
> Overall, this looks good to me. On some systems fcontext might not be supported at all. Can we have an option to have no threading, including fcontext being shut off?

This seems like a non-trivial change and I would need help with the design here. Currently, we always use some kind of `WorkerGroupType` and `WorkerType`, with `WorkerSeq` (based on fcontext) filling the role of "no threading" solution:
https://github.com/DARMA-tasking/vt/blob/468a384de1479ffad73b252a6696574701d10d9c/src/vt/worker/worker_headers.h#L64-L82

Not using `theWorkerGroup` doesn't seem like an option, so another type of worker would be required, resulting in something like this:
```cpp
#if vt_check_enabled(openmp)
  using WorkerType = OMPWorker;
#elif vt_check_enabled(stdthread)
  using WorkerType = StdThreadWorker;
#elif vt_check_enabled(fcontext)
  using WorkerType = WorkerSeq;
#else
  using WorkerType = WorkerNoThreading; // maybe a better name...
#endif
```
with `WorkerNoThreading` fulfilling `vt::worker::WorkerTraits<...>::is_worker` requirements (implementing `spawn()`, `join()` etc.). **Is that the direction to go?**

----
_Note: other references to `fcontext` (in `seq_context.h`, `seq_ult_context.h`, `sequencer.impl.h`) seem to be guarded precisely with `#if vt_check_enabled(fcontext)` - no work required there._

lifflander:
> @lifflander
> 
> with `WorkerNoThreading` fulfilling `vt::worker::WorkerTraits<...>::is_worker` requirements (implementing `spawn()`, `join()` etc.). **Is that the direction to go?**
> 
> _Note: other references to `fcontext` (in `seq_context.h`, `seq_ult_context.h`, `sequencer.impl.h`) seem to be guarded precisely with `#if vt_check_enabled(fcontext)` - no work required there._

I'm thinking that we should completely disable the worker component if no threading is available. We might just want create a dummy one so that things that reference `theWorkerGrp` don't have to be protected with a #if. Go with whatever is the easiest implementation.

cz4rs:
converting to draft as this is definitely still in progress

- I have started with dummy worker to limit the number of changes, but using `#if vt_threading_enabled` in call sites makes it easier to analyze the impact
- the tests are passing ok, but there are some places missing `#else` statement that seem to be reachable - I will mark them directly in review
- no threading is now default option
- configuration variables (Docker, script) for `fcontext` are required (+ some builds in CI)
- #1241 was merged, but `clang` +` OpenMP` in CI has to wait for Docker images to be uploaded

src/vt/pool/pool.cc
```diff
@@ -212,9 +212,11 @@ void Pool::dealloc(void* const buf) {
   );
 
   if (pool_type != ePoolSize::Malloc && alloc_worker != worker) {
+    #if vt_threading_enabled
     theWorkerGrp()->enqueueForWorker(worker, [buf]{
       thePool()->dealloc(buf);
     });
+    #endif
```
cz4rs, line 219: missing `#else`

src/vt/vrt/context/context_vrtinfo.cc
```diff
@@ -80,9 +80,13 @@ void VirtualInfo::setVirtualContextPtr(VirtualPtrType in_vrt_ptr) {
   );
 
   msg_buffer_.attach([this](VirtualMessage* msg){
+    #if vt_threading_enabled
     theWorkerGrp()->enqueueCommThread([this,msg]{
       enqueueWorkUnit(msg);
     });
+    #else
+    (void)this;
```
cz4rs, line 88: silences the unused variable warning, but doesn't provide an actual alternative

cz4rs:
no threading is the default behavior now
`fcontext`, `std::thread`, `OpenMP` and "no threading" configurations are spread out throughout our CI

currently there is no `clang` + `OpenMP` build - #1233 was fixed and is already merged, but update to docker images is required

lifflander (approved):
This looks good to me

cz4rs:
regarding rebuilding docker images when needed: it doesn't seem to be possible with `docker` / `docker-compose` only

an alternative would be to just look at the git diff and detect changes in Dockerfiles
https://stackoverflow.com/questions/47174711/perform-docker-build-only-if-dockerfile-is-changed

Issue #566, opened 2019-11-12
Objgroup not instrumented for LB


lifflander:
**Describe the bug**
Objgroups and other handlers (non-collection) are left out of LB instrumentation.

PhilMiller:
Fixed in #1601, I think

PR #1221, opened 2021-01-18
#1217: Check for zero objects on a MPI rank/node case after LB runs

### LINKED ISSUE #1217: Check for zero objects on a MPI rank/node case after LB runs
**What Needs to be Done?**

After the load balancer runs, it's possible it generates a distributed where a rank/node has zero objects. That configuration might cause VT to hang (which it in the process of being fixed), but also may not be supported for some applications.

We are seeing a hang in EMPIRE after LB that might be that case, but it's impossible to know.

For now, let's check for that case after migrations and print out if that's the case. We could also reject migrations that cause that case to happen.

cz4rs:
fixes #1217

### DIFF
--- a/docs/md/collection.md
+++ b/docs/md/collection.md
@@ -11,9 +11,12 @@ location to manage the location of these elements to efficiently deliver
 messages. It also utilizes the \ref group to build a spanning tree across
 the nodes that the collection is currently mapped to. This group makes
 broadcasts efficient and allows reductions to make progress without waiting for
-nodes that do not have collection elements. The \ref proc-stats component
-stores the statistics for live collections that then passes the instrumented
-data to the \ref lb-manager component to apply load balancing strategies.
+nodes that do not have collection elements.
+
+The \ref proc-stats component stores the statistics for live collections that
+then passes the instrumented data to the \ref lb-manager component to apply load
+balancing strategies. You can use `--vt_lb_keep_last_elm` flag to prohibit load
+balancer from migrating last element in collection.
 
 \section rooted-hello-world-collection Hello World 1D Dense Collection (Rooted)
 \snippet  examples/hello_world/hello_world_collection.cc Hello world collection
--- a/src/vt/configs/arguments/app_config.h
+++ b/src/vt/configs/arguments/app_config.h
@@ -137,6 +137,7 @@ struct AppConfig {
   std::string vt_lb_name      = "NoLB";
   std::string vt_lb_args      = "";
   int32_t vt_lb_interval      = 1;
+  bool vt_lb_keep_last_elm    = false;
   bool vt_lb_stats            = false;
   std::string vt_lb_stats_dir     = "vt_lb_stats";
   std::string vt_lb_stats_file    = "stats";
--- a/src/vt/configs/arguments/args.cc
+++ b/src/vt/configs/arguments/args.cc
@@ -336,6 +336,7 @@ void ArgConfig::addLbArgs(CLI::App& app) {
   auto lb_show_spec  = "Show LB specification during startup";
   auto lb_name       = "Name of the load balancer to use";
   auto lb_interval   = "Load balancing interval";
+  auto lb_keep_last_elm = "Do not migrate last element in collection";
   auto lb_stats      = "Enable load balancing statistics";
   auto lb_stats_dir  = "Load balancing statistics output directory";
   auto lb_stats_file = "Load balancing statistics output file name";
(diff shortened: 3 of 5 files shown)


cz4rs:
@lifflander 

I'm not sure if I'm checking the right thing, but if we want to only emit a warning - could this work?
(also, I was playing around a bit with `TestLoadBalancer` / `LoadBalancerExplode` test, but I couldn't reproduce zero objects case, so this PR only works in theory at the moment)

lifflander:
I'm thinking that maybe we should also have an option to disallow this from happening since it might be a non-recoverable error in some cases.

Right before asking for the last migration on a node we could stop it if its the last object.

@PhilMiller Thoughts?

PhilMiller:
A few thoughts, in order of me thinking of them:
* I'd consider it something of a conceptual design flaw for code to fail in the "some process has no assigned objects" case
* There's a potential performance tradeoff in enforcing object presence on every process, if (e.g.) a communication-aware LB strategy suggests clumping them
* "At least one object on every process" is not generally a guarantee we can make, if there are fewer objects than processes, or objects are marked as not suitable for migration
* Is the system going to be designed to pro-actively fix the lack of objects on some processes if the application has requested this guarantee? What if the initial placement has that condition?
* If the application consciously meets the constraints that would ensure an object should be present on every process (enough objects, covering placement, etc), it does seem like a great convenience to offer the guarantee that we won't break the condition through LB. In that case, perhaps this condition should be flagged at collection creation, and potentially checked *prior* to LB, so we can identify when the application has screwed itself.

PhilMiller:
Anyway, for development/debugging purposes, having a warning that a process is becoming empty which might precede job failure would be useful on request. It could maybe just be a flag value displayed in the post-balancing output from BaseLB, rather than a separate print?

src/vt/vrt/collection/balance/lb_invoke/lb_manager.cc
```diff
@@ -180,6 +180,18 @@ LBManager::runLB(LBProxyType base_proxy, PhaseType phase) {
     strat->applyMigrations(strat->getTransfers());
   });
 
+  runInEpochCollective([=] {
+    auto node_load = theNodeStats()->getNodeLoad();
+    auto load = node_load->find(thePhase()->getCurrentPhase());
+    if (load == node_load->end() or
+        (load != node_load->end() and load->second.empty())) {
```
PhilMiller, line 187: I realize this test may end up being problematic, if we address #566, since it will reflect load on the process distinct from collection objects.

PhilMiller:
I think any test needs to be per-collection, since application logic is likely to work or fail depending on presence of a member of one or more particular collections.

cz4rs:
> Anyway, for development/debugging purposes, having a warning that a process is becoming empty which might precede job failure would be useful on request. It could maybe just be a flag value displayed in the post-balancing output from BaseLB, rather than a separate print?

:+1: for keeping the logs in one place, having `vt_debug_lb` on already produces lots of output

lifflander:
If we want to stop the last collection element from migrating, we can add something like this to `CollectionManager::migrateOut`:

```C++
 if (elm_holder->numElements() == 1) {
     return MigrateStatus::ElementNotLocal;
   }
```

cz4rs:
note: it seems that per-collection information is already printed in `CollectionManager::migrateOut`:
https://github.com/DARMA-tasking/vt/blob/92c2ee9017ed5383b5ae947296c1446e37014f1d/src/vt/vrt/collection/manager.impl.h#L2842-L2846

cz4rs:
- `--vt_lb_keep_last_elm` flag introduced 
- currently the debug print was left unchanged and requires `--vt_debug_vrt_coll` to be enabled (this produces massive amount of output)

running lb tests without the flag, we hit the "0 elements" case multiple times:
```
~   vt  build  mpiexec -n 2 ./tests/collection_extended --vt_debug_vrt_coll --gtest_filter=*TestLoadBalancer* | grep "migrateOut: (after remove) holder numElements=0"
vt: [0] vcc: migrateOut: (after remove) holder numElements=0
vt: [1] vcc: migrateOut: (after remove) holder numElements=0
(...)
vt: [0] vcc: migrateOut: (after remove) holder numElements=0
```
with the flag enabled tests run fine, but no such case is detected:
```
~  vt  build  mpiexec -n 2 ./tests/collection_extended --vt_debug_vrt_coll --vt_lb_keep_last_elm --gtest_filter=*TestLoadBalancer* | grep "migrateOut: (after remove) holder numElements=0"
~  vt  build  0  1   
```

lifflander (approved):
This looks good to me. Can you add a sentence in the docs about this option? Maybe the virtual context collection page?

Issue #1006, opened 2020-08-25
GroupCollective setup races with continuation registration code


lifflander:
**Describe the bug**
This was discovered in #1004, due to a buggy rewrite of addAction #979 in collection callback test code where a different collection is constructed on every node (which was not intended). However, this flooding of collections revealed a bug in how continuations are registered when a new collective group is established in `InfoColl::setupCollective`. Basically, if collective (but non-sync'ed) registration is delayed on one node and it starts sending messages before the continuations are registered, the continuation runs but doesn't have the proper state to make progress. So we either need to make it dependent on `setupCollective` finishing first (enqueue and wait for that) or call `setupCollective` on demand, which might not be possible because all state is not known at that point.

Specifically, the continuation registration calls https://github.com/DARMA-tasking/vt/blob/306d71352f8b74ba7f0090f8d357bd2a3d8eac52/src/vt/group/collective/group_info_collective.cc#L155-L160 have not completed when the handler runs and thus doesn't dispatch properly.

This might point to a broader issue with how collective groups are created. Possibly, this should be refactored to use more epoch-based sync.

lifflander:
The first step in solving this problem should be a write a test that reproduces it. Basically, a collective group construction that delays the call on one of the nodes. It might be a little tricky to get the delay written correctly, but should be very possible.

lifflander:
@cz4rs Let me know if you need to explanation on this issue. We should probably start by trying to reproduce the issue in a test.

PR #1225, opened 2021-01-19
883: Make broadcast deliver to sender (for default group)

### LINKED ISSUE #883: Make broadcasts deliver to the sender
**What Needs to be Done?**

Right now, vt broadcasts over certain sets of recipients exclude the sender from having the handler called (general handler bcast, objgroup bcast, I think?) while broadcasts to collections go to all elements, even if the sender is an element. I think this may be an undesirable semantic inconsistency, depending on what callers do relative to this non-uniformity.

We should go over broadcast call sites, and see how this impacts callers. Do they all just call the handler in-line? Does this represent an issue of processing order, since all other recipients will have the broadcasts ordered by the root? Are they all doing something special, such that they really shouldn't receive the broadcast?

**Additional context**

JacobDomagala:
Fixes #883

### DIFF
--- a/src/vt/collective/barrier/barrier.cc
+++ b/src/vt/collective/barrier/barrier.cc
@@ -60,12 +60,9 @@ Barrier::Barrier() :
 }
 
 /*static*/ void Barrier::barrierDown(BarrierMsg* msg) {
-  auto const is_bcast = envelopeIsBcast(msg->env);
-  auto const dest = envelopeGetDest(msg->env);
-  auto const this_node = theContext()->getNode();
-  const auto is_root = is_bcast && (this_node == dest);
-
-  if (is_root) {
+  // Handler already executed inline after calling bcast
+  // Don't run this function twice
+  if(envelopeIsBcastRoot(msg->env)){
     return;
   }
 
--- a/src/vt/collective/barrier/barrier.cc
+++ b/src/vt/collective/barrier/barrier.cc
@@ -235,7 +235,6 @@ void Barrier::barrierUp(
         "barrierDown: barrier={}\n", barrier
       );
       theMsg()->broadcastMsg<BarrierMsg, barrierDown>(msg);
-      barrierDown(is_named, is_wait, barrier);
     }
   }
 }
--- a/src/vt/collective/barrier/barrier.cc
+++ b/src/vt/collective/barrier/barrier.cc
@@ -60,6 +60,15 @@ Barrier::Barrier() :
 }
 
 /*static*/ void Barrier::barrierDown(BarrierMsg* msg) {
+  auto const is_bcast = envelopeIsBcast(msg->env);
+  auto const dest = envelopeGetDest(msg->env);
+  auto const this_node = theContext()->getNode();
+  const auto is_root = is_bcast && (this_node == dest);
+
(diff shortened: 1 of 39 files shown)


JacobDomagala:
As mentioned on the meeting, if we're ok with not delivering messages marked as termination or few other cases (barriers), maybe we should add the flag to envelope based on which we would decide whether or not to deliver given message.

lifflander:
> As mentioned on the meeting, if we're ok with not delivering messages marked as termination or few other cases (barriers), maybe we should add the flag to envelope based on which we would decide whether or not to deliver given message.

Yes, I think this a good idea. Can you add an extra bit to the default envelope for that?

src/vt/collective/barrier/barrier.cc
```diff
@@ -60,6 +60,12 @@ Barrier::Barrier() :
 }
 
 /*static*/ void Barrier::barrierDown(BarrierMsg* msg) {
+  // Handler already executed inline after calling bcast
```
lifflander, line 63: Is there a reason here we can't just stop calling it inline?

lifflander (commented):
Looking good. Let's add that bit and then merge this

src/vt/collective/barrier/barrier.cc
JacobDomagala: This caused **(op_id != no_op_id)** assertion ( [pipeline example](https://dev.azure.com/DARMA-tasking/DARMA/_build/results?buildId=4444&view=logs&j=3dc8fd7e-4368-5a92-293e-d53cefc8c4b3&t=28db5144-7e5d-5c90-2820-8676d630d9d2) ) on few pipelines with `tutorial_example` execution, where due to scheduling `barrierDown` collective creation of group was delayed on node 0.

image
 
This resulted in node 0 executing `InfoColl::upHan` handler sent from node 1 inside first barrier (line 111 on the screenshot above) and later during execution of `InfoColl::setupCollective` it would send message to node 1 with **no_op** value . So basically it was reproducing issue https://github.com/DARMA-tasking/vt/issues/1006

lifflander: I think this is a bug in the tutorial.

JacobDomagala: With https://github.com/DARMA-tasking/vt/issues/1261 merged I've changed the barrier to deliver the message to root instead of running it locally.

lifflander (approved):
I think this is ready to merge. I looked at the termination code and there are valid reasons those handlers need to be invoked inline. Also, I was worried about documentation, but it looks like the tutorial was updated with the default proxy---so nothing is out of line.

src/vt/messaging/envelope/envelope_base.h
```diff
@@ -98,6 +98,10 @@ struct ActiveEnvelope {
   /// True iff the message is considered locked.
   /// If locked, changes to the envelope will result in failure.
   bool is_locked : 1;
+
+  // Used only for broadcast to default group
+  // Determines whether message should be send also to sender
```
PhilMiller, line 103: Reword please:
"Determines whether message should also be sent to the sender"

PhilMiller (approved):

PR #1226, opened 2021-01-19
868 remove temporary ids

### LINKED ISSUE #868: Remove distinction between temporary and permanent object IDs in load balancing

nlslatt:
I replaced all instances of a collection element's permanent or temporary ID with a struct called `ElementIDStruct` containing an ID that is unique across ranks, the home rank, and the rank where the element currently resides.  I left the unique ID part of the struct the same as the old permanent ID, which had bits encoding the rank. There shouldn't be any logic remaining that depends on those bits, so any alternative means of making the IDs unique across ranks should work fine.

Closes #868

### DIFF
--- a/src/vt/vrt/collection/balance/lb_common.h
+++ b/src/vt/vrt/collection/balance/lb_common.h
@@ -56,6 +56,26 @@ namespace balance {
 
 using ElementIDType = uint64_t;
 
+struct ElementIDStruct {
+  using isByteCopyable = std::true_type;
+
+  ElementIDType id = 0;
+  NodeType home_node = uninitialized_destination;
+  NodeType curr_node = uninitialized_destination;
+
+  bool operator==(const ElementIDStruct& rhs) const {
+    return id == rhs.id and home_node == rhs.home_node;
+  }
+
+  bool operator<(const ElementIDStruct& rhs) const {
+    return id < rhs.id;
+  }
+};
+
+std::ostream& operator<<(
+  std::ostream& os, const ::vt::vrt::collection::balance::ElementIDStruct& id
+);
+
 static constexpr ElementIDType const no_element_id = 0;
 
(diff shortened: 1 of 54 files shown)


nlslatt:
@lifflander Please update `ZoltanLB` to use the new ID struct.

src/vt/vrt/collection/balance/lb_common.h
```diff
@@ -50,16 +50,37 @@
...
+  NodeType home_node = uninitialized_destination;
+  NodeType curr_node = uninitialized_destination;
+
+  bool operator==(const ElementIDStruct& rhs) const {
+    return id == rhs.id and home_node == rhs.home_node;
+  }
+
+  bool operator<(const ElementIDStruct& rhs) const {
+    return id < rhs.id;
+  }
```
PhilMiller, line 73: I believe for ordered collections like `std::map` and `std::set`, the definition of comparison operators needs to be consistent - i.e. I think this should be

`return id < rhs.id || (id == rhs.id && home_node < rhs.home_node);`

PhilMiller: Given the stated invariant that `id` will continue to be unique across processes, the two definitions are consistent, and just the comparison of `home_node` in `operator==` is redundant.

PhilMiller:
I'll be very happy to see this merged. Everything looks pretty good.

src/vt/vrt/collection/balance/lb_common.h
nlslatt: Redundancy removed.

PhilMiller (approved):

JacobDomagala (approved):

JacobDomagala (approved):
Apart from merge conflicts blocking the workflow runs, it looks good.

PR #1230, opened 2021-01-26
#1174: add vt-sample compilation to CI

### LINKED ISSUE #1174: Write test with vt-sample-project to test VT cmake install
**What Needs to be Done?**

We don't have any good testing for VT's cmake install. Write a CI test that can test an installed version of VT to see if installs all files and exports all the correct cmake targets.

jstrzebonski:
Fixes #1174

### DIFF
--- a/cmake/load_zoltan_package.cmake
+++ b/cmake/load_zoltan_package.cmake
@@ -1,16 +1,7 @@
-
-
 set(vt_zoltan_found "0")
 
-# optional directory for this package
-optional_pkg_directory(zoltan "Zoltan library" 1)
-# find the optional packages locally if identified
-if (${zoltan_DIR_FOUND})
-  find_package_local(zoltan "${zoltan_DIR}" Zoltan)
-  if(NOT projHasParent)
-    if (NOT ${zoltan_FOUND})
-      message(FATAL_ERROR "Zoltan library not found")
-    endif()
-  endif()
+if(vt_zoltan_enabled)
+  find_package(Zoltan REQUIRED)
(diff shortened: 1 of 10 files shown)


cmake/load_zoltan_package.cmake
```diff
@@ -1,16 +1,7 @@
...
-optional_pkg_directory(zoltan "Zoltan library" 1)
-# find the optional packages locally if identified
-if (${zoltan_DIR_FOUND})
-  find_package_local(zoltan "${zoltan_DIR}" Zoltan)
-  if(NOT projHasParent)
-    if (NOT ${zoltan_FOUND})
-      message(FATAL_ERROR "Zoltan library not found")
-    endif()
-  endif()
+if(vt_zoltan_enabled)
```
lifflander, line 3: How have you been testing Zoltan? Does this work with Trillions builds of Zoltan which builds it as a sub-package? I think it must because you are using CI which installs Zoltan with Trilinos.

lifflander: Overall, I like the approach of find_package a lot better if this works!

lifflander (approved):

cmake/load_zoltan_package.cmake
jstrzebonski: I tested it on our dockers, where it's built and installed from the source.

lifflander: Great, then we should be good to go

cz4rs (approved):
looks good to me!

JacobDomagala (approved):

PR #1236, opened 2021-01-27
1232 Implement Storable collection trait in support of CUDA streams

### LINKED ISSUE #1232: Associate data with collection elements w/ user getter/setters
Support storing CUDA streams and kokkos devices.

lifflander:
Fixes #1232

Implement a general mechanism to store data with a collection element. Implemented with a key/value store that detects whether the value needs serialization or not. The string-based key could add some overhead, but I think it's fine for the use cases we have.

### DIFF
--- a/src/vt/vrt/collection/migrate/migrate_handlers.impl.h
+++ b/src/vt/vrt/collection/migrate/migrate_handlers.impl.h
@@ -87,7 +87,7 @@ template <typename ColT, typename IndexT>
   // auto vc_raw_ptr = ::checkpoint::deserialize<ColT>(
   //   buf, col_t
   // );
-  auto vc_elm_ptr = std::make_unique<ColT>(std::move(*msg->elm_));
+  auto vc_elm_ptr = std::unique_ptr<ColT>(msg->elm_);
 
   auto const& migrate_status =
     CollectionElmAttorney<ColT,IndexT>::migrateIn(
--- a/src/vt/vrt/collection/migrate/migrate_msg.h
+++ b/src/vt/vrt/collection/migrate/migrate_msg.h
@@ -66,12 +66,6 @@ struct MigrateMsg final : ::vt::Message {
       range_(in_range), elm_(in_elm)
   { }
 
-  ~MigrateMsg() {
-    if (elm_ and owns_elm_) {
-      delete elm_;
-    }
-  }
-
   VrtElmProxy<ColT, IndexT> getElementProxy() const { return elm_proxy_; }
   NodeType getFromNode() const { return from_; }
   NodeType getToNode() const { return to_; }
@@ -84,7 +78,6 @@ struct MigrateMsg final : ::vt::Message {
     s | elm_proxy_ | from_ | to_ | map_fn_ | range_;
     if (s.isUnpacking()) {
       elm_ = new ColT{};
-      owns_elm_ = true;
     }
     s | *elm_;
   }
@@ -95,7 +88,6 @@ private:
   NodeType to_ = uninitialized_destination;
   HandlerType map_fn_ = uninitialized_handler;
   IndexT range_;
-  bool owns_elm_ = false;
 public:
   ColT* elm_ = nullptr;
 };
--- a/src/vt/vrt/collection/migrate/migrate_msg.h
+++ b/src/vt/vrt/collection/migrate/migrate_msg.h
@@ -77,6 +77,7 @@ struct MigrateMsg final : ::vt::Message {
     MessageParentType::serialize(s);
     s | elm_proxy_ | from_ | to_ | map_fn_ | range_;
     if (s.isUnpacking()) {
+      // handler always takes ownership of this by constructing a unique_ptr
       elm_ = new ColT{};
     }
     s | *elm_;
--- a/src/vt/vrt/collection/types/storage/storable.cc
+++ b/src/vt/vrt/collection/types/storage/storable.cc
@@ -2,7 +2,7 @@
 //@HEADER
 // *****************************************************************************
 //
-//                              migratable.impl.h
+//                                 storable.cc
 //                           DARMA Toolkit v. 1.0.0
 //                       DARMA/vt => Virtual Transport
 //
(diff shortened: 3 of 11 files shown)


src/vt/vrt/collection/types/storage/storable.cc
```diff
@@ -0,0 +1,65 @@
...
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+//
+// Questions? Contact darma@sandia.gov
+//
+// *****************************************************************************
+//@HEADER
+*/
+
+#if !defined INCLUDED_VT_VRT_COLLECTION_TYPES_STORAGE_STORABLE_CC
```
cz4rs, line 45: why the include guard in .cc file?

src/vt/vrt/collection/types/storage/store_elm.h
```diff
@@ -0,0 +1,260 @@
...
+/**
+ * \struct StoreElm
+ *
+ * \brief Typed value specialized for byte-copyable stored values
+ */
+template <typename T>
+struct StoreElm<
+  T,
+  typename std::enable_if_t<
+    not ::checkpoint::SerializableTraits<T>::is_serializable
```
cz4rs, line 204: is this condition enough to guarantee byte-copyable value? (is it possible that we get `T` that is neither serializable nor byte-copyable?)

tests/unit/collection/test_storage.cc
```diff
@@ -0,0 +1,127 @@
...
+
+  int64_t a = 0;
+  float b = 0.;
+};
+
+struct TestCol : Collection<TestCol, Index1D> {
+
+  struct TestMsg : CollectionMessage<TestCol> { };
+
+  void testHandler(TestMsg* msg) {
```
cz4rs, line 69: codecov complains about `valRemove` not being tested - maybe it can be added here?

cz4rs (approved):
looks good overall :+1: please have a look at questions posted

src/vt/vrt/collection/types/storage/storable.cc
lifflander: Oops, accidentally generated it

src/vt/vrt/collection/types/storage/store_elm.h
lifflander: No, it's not. I'm planning to add some `static_assert` to check for some errors here. I might need to check for the trait here.

src/vt/vrt/collection/migrate/migrate_handlers.impl.h
```diff
@@ -87,7 +87,7 @@ template <typename ColT, typename IndexT>
   // auto vc_raw_ptr = ::checkpoint::deserialize<ColT>(
   //   buf, col_t
   // );
-  auto vc_elm_ptr = std::make_unique<ColT>(std::move(*msg->elm_));
+  auto vc_elm_ptr = std::unique_ptr<ColT>(msg->elm_);
```
PhilMiller, line 90: I'm happy to see this change finally being made.

src/vt/vrt/collection/migrate/migrate_msg.h
```diff
@@ -95,7 +88,6 @@ struct MigrateMsg final : ::vt::Message {
   NodeType to_ = uninitialized_destination;
   HandlerType map_fn_ = uninitialized_handler;
   IndexT range_;
-  bool owns_elm_ = false;
 public:
   ColT* elm_ = nullptr;
```
PhilMiller, line 92: Should this actually be a `unique_ptr<ColT>` that the handler moves from? Otherwise, there should be a comment here that the handler always takes ownership of the element.

src/vt/vrt/collection/migrate/migrate_msg.h
```diff
@@ -84,7 +78,6 @@ struct MigrateMsg final : ::vt::Message {
     s | elm_proxy_ | from_ | to_ | map_fn_ | range_;
     if (s.isUnpacking()) {
       elm_ = new ColT{};
```
PhilMiller, line 80: Not exactly relevant to this commit, but should this be a reconstruction call through checkpoint? Allowing the same customized logic as for any other serializable type?

src/vt/vrt/collection/types/storage/store_elm.h
```diff
@@ -0,0 +1,258 @@
...
+  explicit StoreElm(U&& u)
+    : elm_(std::forward<U>(u))
+  { }
+
+  /**
+   * \brief Serialization re-constructor
+   *
+   * \param[in] SERIALIZE_CONSTRUCT_TAG tag
+   */
+  explicit StoreElm(checkpoint::SERIALIZE_CONSTRUCT_TAG) {}
```
PhilMiller, line 134: Should this pass through to `elm_`? More generally, is there a right way to do pass-through reconstruction, for types like this that contain and must serialize arbitrary other types?

PhilMiller:
If we wanted to not use string-based keys, we could use tag types with a registry instance.

PhilMiller (approved):
I'll be happy to have this for the CUDA support work. Looks great.

src/vt/vrt/collection/types/storage/storable.cc
lifflander: Fixed

tests/unit/collection/test_storage.cc
lifflander: added to test

src/vt/vrt/collection/migrate/migrate_msg.h
lifflander: There are some reasons why the unique_ptr doesn't work in the message due to how ownership is transferred. I think for now I'll just add a comment.

src/vt/vrt/collection/migrate/migrate_msg.h
lifflander: Yes, I think so. It should be made uniform.

lifflander: https://github.com/DARMA-tasking/vt/issues/1244

src/vt/vrt/collection/types/storage/store_elm.h
lifflander: That's a really good point. I was thinking about that when I wrote the code, but don't know if we have a way to do that cleanly. Right now, `checkpoint::Reconstructor` doesn't have the a good interface for that.

lifflander: Opened an issue to address this broadly: https://github.com/DARMA-tasking/checkpoint/issues/177

src/vt/vrt/collection/types/storage/store_elm.h
lifflander: I've added a necessary but not sufficient condition, which will catch some errors. We could require the byte copy trait, but many types may not have it and still be correct. So for now, I'm going to leave it with the new check.

PR #1238, opened 2021-01-28
1237 Implement AsyncOp for general polling epoch tasks

### LINKED ISSUE #1237: Implement async, polling task that registers with an epoch
**What Needs to be Done?**

Intended to be used for CUDA events or other async MPI calls that is polled by the system until completion.

lifflander:
Fixes #1237

### DIFF
--- a/src/vt/messaging/active.cc
+++ b/src/vt/messaging/active.cc
@@ -66,8 +66,10 @@ ActiveMessenger::ActiveMessenger()
   trace_isend(trace::registerEventCollective("MPI_Isend")),
   trace_irecv_polling_am(trace::registerEventCollective("IRecv: Active Msg poll")),
   trace_irecv_polling_dm(trace::registerEventCollective("IRecv: Data Msg poll")),
+  trace_asyncop(trace::registerEventCollective("AsyncOP: poll")),
   in_progress_active_msg_irecv(trace_irecv_polling_am),
   in_progress_data_irecv(trace_irecv_polling_dm),
+  in_progress_ops(trace_asyncop),
 # endif
   this_node_(theContext()->getNode())
 {
--- a/src/vt/messaging/active.cc
+++ b/src/vt/messaging/active.cc
@@ -1225,15 +1225,27 @@ bool ActiveMessenger::testPendingDataMsgAsyncRecv() {
   return ret;
 }
 
+bool ActiveMessenger::testPendingAsyncOps() {
+  int num_tests = 0;
+  bool const ret = in_progress_ops.testAll(
+    [](AsyncOpWrapper* op) {
+      op->done();
+    },
+    num_tests
+  );
+  return ret;
+}
+
 int ActiveMessenger::progress() {
   bool const started_irecv_active_msg = tryProcessIncomingActiveMsg();
   bool const started_irecv_data_msg = tryProcessDataMsgRecv();
   processMaybeReadyHanTag();
   bool const received_active_msg = testPendingActiveMsgAsyncRecv();
   bool const received_data_msg = testPendingDataMsgAsyncRecv();
+  bool const general_async = testPendingAsyncOps();
 
   return started_irecv_active_msg or started_irecv_data_msg or
-         received_active_msg or received_data_msg;
+         received_active_msg or received_data_msg or general_async;
 }
 
 void ActiveMessenger::processMaybeReadyHanTag() {
--- a/src/vt/messaging/active.cc
+++ b/src/vt/messaging/active.cc
@@ -1229,13 +1229,12 @@ bool ActiveMessenger::testPendingDataMsgAsyncRecv() {
 
 bool ActiveMessenger::testPendingAsyncOps() {
   int num_tests = 0;
-  bool const ret = in_progress_ops.testAll(
+  return in_progress_ops.testAll(
     [](AsyncOpWrapper* op) {
       op->done();
     },
     num_tests
   );
-  return ret;
 }
 
 int ActiveMessenger::progress() {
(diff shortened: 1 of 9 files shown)


src/vt/messaging/async_op.h
```diff
@@ -0,0 +1,103 @@
...
+// Questions? Contact darma@sandia.gov
+//
+// *****************************************************************************
+//@HEADER
+*/
+
+#if !defined INCLUDED_VT_MESSAGING_ASYNC_OP_H
+#define INCLUDED_VT_MESSAGING_ASYNC_OP_H
+
+#include "vt/config.h"
```
cz4rs, line 48: `vt/configs/types/types_headers.h` should be enough here (although I'm not sure if this is big enough improvement?)

cz4rs (approved):
looks good to me

src/vt/messaging/active.cc
```diff
@@ -1225,15 +1225,27 @@ bool ActiveMessenger::testPendingDataMsgAsyncRecv() {
   return ret;
 }
 
+bool ActiveMessenger::testPendingAsyncOps() {
+  int num_tests = 0;
+  bool const ret = in_progress_ops.testAll(
```
PhilMiller, line 1232: Why initialize the variable when it's returned in the next statement?

src/vt/messaging/active.cc
```diff
@@ -1225,15 +1225,27 @@ bool ActiveMessenger::testPendingDataMsgAsyncRecv() {
...
+  return ret;
+}
+
 int ActiveMessenger::progress() {
   bool const started_irecv_active_msg = tryProcessIncomingActiveMsg();
   bool const started_irecv_data_msg = tryProcessDataMsgRecv();
   processMaybeReadyHanTag();
   bool const received_active_msg = testPendingActiveMsgAsyncRecv();
   bool const received_data_msg = testPendingDataMsgAsyncRecv();
+  bool const general_async = testPendingAsyncOps();
```
PhilMiller, line 1247: Does this belong in `ActiveMessenger`, or in the scheduler's overall polling loop?

src/vt/messaging/active.h
```diff
@@ -1771,6 +1799,7 @@ struct ActiveMessenger : runtime::component::PollableComponent<ActiveMessenger>
   std::vector<ListenerType> send_listen_                  = {};
   IRecvHolder<InProgressIRecv> in_progress_active_msg_irecv;
   IRecvHolder<InProgressDataIRecv> in_progress_data_irecv;
+  IRecvHolder<AsyncOpWrapper> in_progress_ops;
```
PhilMiller, line 1804: This suggests that `IrecvHolder` really ought to be named something more general like `RequestHolder`

src/vt/messaging/active.impl.h
```diff
@@ -596,6 +596,12 @@ inline EpochType ActiveMessenger::setupEpochMsg(MsgSharedPtr<MsgT> const& msg) {
   return setupEpochMsg(msg.get());
 }
 
+template <typename T>
+void ActiveMessenger::registerAsyncOp(std::unique_ptr<T> in) {
+ std::unique_ptr<AsyncOp> op(static_cast<AsyncOp*>(in.release()));
+  in_progress_ops.emplace(AsyncOpWrapper{std::move(op)});
```
PhilMiller, line 602: If the cast is valid, shouldn't it be possible to directly construct from `move(in)`? I.e. with no need for the `op` variable at all?

src/vt/messaging/async_op.cc
```diff
@@ -0,0 +1,64 @@
...
+
+namespace vt { namespace messaging {
+
+AsyncOp::AsyncOp() {
+  cur_epoch_ = theMsg()->getEpoch();
+  theTerm()->produce(cur_epoch_);
+}
+
+AsyncOp::AsyncOp(AsyncOp&& in) {
+  std::swap(cur_epoch_, in.cur_epoch_);
```
PhilMiller, line 55: This is right about `in.cur_epoch_` ending in holding `no_epoch`, but with a lot of subtlety - the default initialization of `this->cur_epoch_`, over in the header file, and the lack of initializer list on this constructor.

For ease of reading locally, could this just be written

```c++
cur_epoch_ = in.cur_epoch_;
in.cur_epoch_ = no_epoch;
```

src/vt/messaging/active.cc
PhilMiller: Seeing the entire PR, the implementation does seem to be easier here.

PhilMiller (approved):
This all looks good, though there's some refactoring that would be nice for clarity.

src/vt/messaging/async_op.h
lifflander: Fixed

src/vt/messaging/active.cc
lifflander: Good point.. fixed

src/vt/messaging/active.cc
lifflander: Yeah, I had the same thought initially, but determined that implementing in the `ActiveMessenger` would be easier since it has all the data structures

src/vt/messaging/active.h
lifflander: Renamed to `RequestHolder`

src/vt/messaging/async_op.cc
lifflander: Yes, the logic is a bit tricky. Change made

src/vt/messaging/active.impl.h
lifflander: I thought there was an issue of inherited types, but I think you are right that constructing from the move will work. Seems to pass the test

PR #1245, opened 2021-02-01
1126: Add collective calls for VT examples

### LINKED ISSUE #1126: Use collective epochs and phases in Jacobi stencil examples
**What Needs to be Done?**

Make the examples, shown in the manual, `jacobi1d_vt`, `jacobi2d_vt`, and `reduce_integral` more idiomatic to VT by uses collective epochs for each iteration and called `nextPhaseCollective` for each phase. Also, use collective broadcasts where possible.

JacobDomagala:
Fixes #1126

### DIFF
--- a/examples/collection/jacobi1d_vt.cc
+++ b/examples/collection/jacobi1d_vt.cc
@@ -62,6 +62,7 @@
 
 
 #include <vt/transport.h>
+#include <vt/runnable/invoke.h>
 
 #include <cstdlib>
 #include <cassert>
@@ -70,7 +71,15 @@
 static constexpr std::size_t const default_nrow_object = 8;
 static constexpr std::size_t const default_num_objs = 4;
 static constexpr double const default_tol = 1.0e-02;
-static bool is_done = false;
+
+struct NodeObj {
+  bool is_finished_ = false;
+  struct WorkFinishedMsg : vt::Message {};
+
+  void workFinishedHandler(WorkFinishedMsg*) { is_finished_ = true; }
+  bool isWorkFinished() { return is_finished_; }
+};
+using NodeObjProxy = vt::objgroup::proxy::Proxy<NodeObj>;
 
 struct LinearPb1DJacobi : vt::Collection<LinearPb1DJacobi,vt::Index1D> {
 
@@ -83,6 +92,7 @@ private:
   size_t numObjs_ = 1;
   size_t numRowsPerObject_ = 1;
   size_t maxIter_ = 8;
+  NodeObjProxy objProxy_;
 
 public:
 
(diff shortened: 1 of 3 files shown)


examples/collection/jacobi1d_vt.cc
```diff
@@ -366,21 +352,32 @@ int main(int argc, char** argv) {
...
-  if (this_node == 0) {
+  // Create the decomposition into objects
+  using BaseIndexType = typename vt::Index1D::DenseIndexType;
+  auto range = vt::Index1D(static_cast<BaseIndexType>(num_objs));
 
-    // Create the decomposition into objects
-    using BaseIndexType = typename vt::Index1D::DenseIndexType;
-    auto range = vt::Index1D(static_cast<BaseIndexType>(num_objs));
+  auto proxy =
+    vt::theCollection()->constructCollective<LinearPb1DJacobi>(range);
```
JacobDomagala, line 360: Should this also be wrapped with `runInEpochCollective`? We use it both ways in our codebase:
- https://github.com/DARMA-tasking/vt/blob/f7ce5215ea4647a22b0fb6c8dbc9527308bf057e/tests/unit/collection/test_lb.extended.cc#L103-L118

- https://github.com/DARMA-tasking/vt/blob/f7ce5215ea4647a22b0fb6c8dbc9527308bf057e/examples/collection/lb_iter.cc#L121-L139

lifflander: Let's construct the collection without its own epoch.

examples/collection/jacobi1d_vt.cc
```diff
@@ -366,21 +352,32 @@ int main(int argc, char** argv) {
...
 
-    auto proxy = vt::theCollection()->construct<LinearPb1DJacobi>(range);
-    proxy.broadcast<LinearPb1DJacobi::LPMsg, &LinearPb1DJacobi::solve>(
+  vt::runInEpochCollective([proxy, num_objs, numRowsPerObject, maxIter]{
+    proxy.broadcastCollective<LinearPb1DJacobi::LPMsg, &LinearPb1DJacobi::init>(
       num_objs, numRowsPerObject, maxIter
     );
+  });
+
+  while(!is_done){
```
lifflander, line 368: I would prefer not to set a static variable when this is done, since that's not very idiomatic. Maybe we can create another collective epoch that invokes `checkCompleteCB` here, with a callback that tells this code whether it is done or not. Not sure the best approach here...

JacobDomagala: Let me know what you think about the updated version of jacobi1D (https://github.com/DARMA-tasking/vt/pull/1245/commits/f16eed29aee14300a22938e93cf3d75d993abc80). When the collection is finished, the "root" element (collection element with ID == 0) will broadcast the message to all nodes that computation is finished. 

If we want this example to be entirely about using virtual collection, then i guess I could change it so that each node would query one of its elements, if the work is done or not.

lifflander:
I like this new approach with an obj group!

lifflander (approved):
Let's merge once you apply it to both examples

lifflander:
Is this ready to merge?

PR #1246, opened 2021-02-02
1240: Add function to set zlib incremental flush mode for trace file

### LINKED ISSUE #1240: Add argument for more expensive trace flushing that doesn't leave corrupted logs when app crashes
**What Needs to be Done?**

Basically we want a runtime argument that sets `Trace::incremental_flush_mode` to `Z_FINISH`:

https://github.com/DARMA-tasking/vt/blob/468a384de1479ffad73b252a6696574701d10d9c/src/vt/trace/trace.cc#L92-L116

JacobDomagala:
Fixes #1240


lifflander:
Can you add an argument that sets this to `Z_FINISH`?

lifflander (approved):
This looks great to me

Issue #1233, opened 2021-01-26
clang builds fail to compile with OpenMP


cz4rs:
**Describe the bug**
Trying to use `clang` with `OpenMP` results in:
```
-- Using OpenMP for worker threading
-- Could NOT find OpenMP_C (missing: OpenMP_C_FLAGS OpenMP_C_LIB_NAMES) 
-- Could NOT find OpenMP_CXX (missing: OpenMP_CXX_FLAGS OpenMP_CXX_LIB_NAMES) 
-- Could NOT find OpenMP (missing: OpenMP_C_FOUND OpenMP_CXX_FOUND) 
-- OpenMP has been found: Linker="", CC FLAGS="NOTFOUND", CXX FLAGS="NOTFOUND"
CMake Error at cmake/load_threading_package.cmake:26 (message):
  requested OpenMP with -DUSE_OPENMP=On, but cannot find valid OpenMP in
  compiler
Call Stack (most recent call first):
  cmake/load_packages.cmake:13 (include)
  CMakeLists.txt:61 (include)
```

This could possibly be fixed with installing `libomp5-dev` in clang container.

**To Reproduce**
Use `VT_USE_OPENMP` / `USE_OPENMP` configuration variables.

**Expected behavior**
Make sure that CI passes with one of `clang` builds using `OpenMP`.

See #1214 for recent changes in the area.

PR #1251, opened 2021-02-03
1250: rebuild containers in CI when dockerfiles change

### LINKED ISSUE #1250: rebuild containers in CI when dockerfiles change
Currently we always use `docker-compose pull` before building `vt` in CI:
https://github.com/DARMA-tasking/vt/blob/f7ce5215ea4647a22b0fb6c8dbc9527308bf057e/scripts/azure-workflow-template.yml#L96-L109

If there are significant changes in Dockerfiles, manual build and push to dockerhub needs to be performed.

**What Needs to be Done?**
Make sure that CI builds the container (instead of pulling) when PR changes Dockerfiles

cz4rs:
fixes #1250


cz4rs:
note: 72d5254 tests the mechanism for only one build, update to `azure-workflow-template.yml` is required

lifflander:
This looks great! Good idea on the implementation.

cz4rs:
> This looks great! Good idea on the implementation.

Thanks! This is almost ready, but there are still some things to clarify/consider:
- we only check the content of last PR merged (`develop`) or current PR (branches) to see if dockerfiles were made
- on `develop` this can cause problems if we don't update the images in registry (what if penultimate PR triggered rebuild, but the last one didn't? back to `docker-compose pull`)
- luckily we already have `Push container to registry` task and it should work as intended after this PR is merged

note: **`Push container to registry` doesn't seem to work on `develop` at the moment** - we (always) pull, and then push the same image we've just pulled
- `clang-5.0` image was correctly built and pushed from this branch (see recently added `libomp-dev` on the screenshot below)
image

TODO:
- [ ] we should probably limit pushing to `develop` branch (any others? release branches require different tags probably)
- [ ] all images are built if at least one dockerfile is changed - we could add more fine-grained detection / control, but honestly I don't think that added complexity if worth it (we don't change dockerfiles often)

ci/azure/azure-clang-5.0-ubuntu-mpich.yml
```diff
@@ -40,8 +40,8 @@ variables:
   VT_EXTENDED_TESTS: 1
   VT_UNITY_BUILD: 1
   VT_FCONTEXT: 0
-  VT_USE_OPENMP: 0
```
lifflander, line 43: Why is this change included here?

lifflander (approved):
Looks great. I just had a question about the threading change here. Was that just to test that this is working? Isn't that what the removal of whitespace did?

jstrzebonski (approved):
Looks good to me.

ci/azure/azure-clang-5.0-ubuntu-mpich.yml
cz4rs: whitespace change was indeed enough to test this PR
`VT_USE_OPENMP` here is a leftover of #1233, but I thought I would enable it without creating a separate issue

cz4rs:
ff7031b makes sure that we only push from `develop` branch.
I will merge this in the morning (CET) if the builds are green.

ci/azure/azure-clang-5.0-ubuntu-mpich.yml
lifflander: Cool, sounds good to me.

PR #1255, opened 2021-02-10
1248 Subphase distributed label management

### LINKED ISSUE #1248: Give LB subphases string names
**What Needs to be Done?**

Makes it easier to align subphases; LoadModels easier to target subphases

lifflander:
Fixes #1248


src/vt/phase/subphase/subphase_bits.h
```diff
@@ -0,0 +1,95 @@
...
+  SubphaseSeq       = 0,        /**< Sequence identifier */
+  SubphaseNode      = 31,       /**< Only exists for rooted epochs */
+  SubphaseIsRooted  = 63        /**< Whether this is rooted or collective */
+};
+
+/// Length of pure ID in subphase for rooted subphases
+static constexpr SubphaseType const seq_len_rooted = 31;
+
+/// Length of pure ID in subphase for collective subphases
+static constexpr SubphaseType const seq_len_collective = 63;
```
PhilMiller, line 68: Given the relationship involved, I think these should sensibly use the values from the enum?

src/vt/phase/phase_manager.h
```diff
@@ -80,7 +81,9 @@ struct NextMsg;
  * rooted; collective hooks are invoked in the order in which they are
  * registered and are always run in a collective epoch.
  */
-struct PhaseManager : runtime::component::Component<PhaseManager> {
+struct PhaseManager
+  : runtime::component::Component<PhaseManager>, subphase::SubphaseManager
```
PhilMiller, line 85: Is there a specific reason to inherit here, rather than just composing with `SubphaseManager` as a member? Is it to get the handlers or something?

lifflander: It's so the users can directly call: `thePhase()->registerCollectiveSubphase`. I could just write some forwarding functions, but is there a reason why inheritance here isn't a good idea?

PR #1257, opened 2021-02-10
1256:  Create vt_trace_only option before creating trace-only CMake target

### LINKED ISSUE #1256: Move option setup for vt_trace_only before trace-only target is created
**What Needs to be Done?**
Currently the vt_trace_only option `option(vt_trace_only "Build VT with trace-only mode enabled" ON)` is added after `VT_TRACE_LIB` target is created:
image

This works fine as long as you add cmake flag `-Dvt_trace_only` to enable/disable trace-only target. Without the flag the 'vt_trace_only' option is set (by default) to ON and later CMake tries to use the non-existent target which results in CMake Error:
image

JacobDomagala:
Fixes #1256 

With this fix, it's no longer required to add -Dvt_trace_only

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -110,6 +110,11 @@ include(cmake/load_bundled_libraries.cmake)
 
 include(cmake/nvcc_no_deprecated_gpu_targets.cmake)
 
+option(vt_trace_only "Build VT with trace-only mode enabled" ON)
+if (vt_trace_only)
+  message(STATUS "Building additional target for VT in trace-only mode")
+endif()
+
 # Primary VT build
 add_subdirectory(src)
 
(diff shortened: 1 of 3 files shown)


cz4rs (approved):
looks good to me

jstrzebonski (approved):

lifflander (commented):
Do you need to update the docs for this?

Looks good to me

CMakeLists.txt
```diff
@@ -110,6 +110,11 @@ include(cmake/load_bundled_libraries.cmake)
 
 include(cmake/nvcc_no_deprecated_gpu_targets.cmake)
 
+option(vt_trace_only "Build VT with trace-only mode enabled" OFF)
```
JacobDomagala, line 113: I've also changed it to be disabled by default. I think it's better to only build it on demand.

JacobDomagala:
@lifflander  Docs are still valid. This change fixes the small bug when building VT with cmake, without using `--Dvt_trace` flag. Also I made it not build by default, so if user wants to use trace-only, he has to use the cmake flag (which is described in the docs).

lifflander (approved):

PR #1263, opened 2021-02-15
1258: add missing fields in 'serialize' methods

### LINKED ISSUE #1258: fields omitted during serialization
**Describe the bug**
Multiple fields are not serialized during either footprinting or regular serialization. Add missing fields to `serialize` methods.
Note: this should be safe for all of footprinting, caution is required for serializers used in other modes.

**To Reproduce**
See [serialization sanitizer run over `vt`](https://github.com/DARMA-tasking/serialization-sanitizer/actions?query=workflow%3A%22PR+tests+%28vt%2C+clang-5.0%2C+ubuntu%29%22).

cz4rs:
fixes #1258

### DIFF
--- a/src/vt/pipe/callback/cb_union/cb_raw.h
+++ b/src/vt/pipe/callback/cb_union/cb_raw.h
@@ -222,7 +222,10 @@ struct GeneralCallback {
     using EnumDataType = typename std::underlying_type<CallbackEnum>::type;
     EnumDataType val = static_cast<EnumDataType>(active_);
     s | val;
+    s.skip(active_);
+
     active_ = static_cast<CallbackEnum>(val);
+    s.skip(u_); // only serialize actual content of the union
     switch (active_) {
     case CallbackEnum::AnonCB:
       s | u_.anon_cb_;
@@ -255,9 +258,6 @@ struct GeneralCallback {
       vtAssert(0, "Should be unreachable");
       break;
     }
-
-    s.skip(active_);
-    s.skip(u_);
   }
 
   bool null()  const { return active_ == CallbackEnum::NoCB; }
--- a/src/vt/pipe/callback/cb_union/cb_raw.h
+++ b/src/vt/pipe/callback/cb_union/cb_raw.h
@@ -223,9 +223,10 @@ struct GeneralCallback {
 
     // serialize actual content of the union and account for leftovers and
     // padding bytes when footprinting
+    auto union_size = sizeof(this->u_);
     auto ser = [&](auto& cb){
       s | cb;
-      s.addBytes(sizeof(this->u_) - sizeof(cb));
+      s.addBytes(union_size - sizeof(cb));
     };
     s.skip(u_);
     switch (active_) {
--- a/src/vt/pipe/callback/cb_union/cb_raw.h
+++ b/src/vt/pipe/callback/cb_union/cb_raw.h
@@ -255,6 +255,9 @@ struct GeneralCallback {
       vtAssert(0, "Should be unreachable");
       break;
     }
+
+    s.skip(active_);
+    s.skip(u_);
   }
 
   bool null()  const { return active_ == CallbackEnum::NoCB; }
--- a/src/vt/pipe/callback/cb_union/cb_raw.h
+++ b/src/vt/pipe/callback/cb_union/cb_raw.h
@@ -219,40 +219,43 @@ struct GeneralCallback {
 
   template <typename SerializerT>
   void serialize(SerializerT& s) {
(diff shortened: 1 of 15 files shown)


cz4rs:
This is based on the experimental reports (after reducing the number of false positives).
- add a bunch of fields completely missed for footprinting purposes
- change `countBytes` to `|` operator - possible after https://github.com/DARMA-tasking/checkpoint/pull/157 was merged (footprinting is available for all non-serializable types `T` using `sizeof(T)` behind the scenes)
- add missing fields for regular serialization in tests and some simple cases


cases still reported:

- [ ] `vt::messaging::MsgSharedPtr::impl_`
- [ ] `vt::messaging::MsgSharedPtr::ptr_ `
- [ ] `t::pipe::callback::cbunion::GeneralCallback::active_`
- [ ] `vt::pipe::callback::cbunion::GeneralCallback::u_ `
- [ ] `vt::pipe::signal::Signal::data_ptr_`
- [ ] `vt::rdma::BaseTypedHandle::actions_ `
- [ ] `vt::rdma::BaseTypedHandle::lock_ `
- [ ] `vt::rdma::BaseTypedHandle::user_buffer_ `
- [ ] `vt::rdma::Manager::holder_footprint_ ` - custom serialization for static holder
- [ ] `vt::term::interval::IntegralSetBase::hint_ ` - custom unpacking
- [ ] `vt::termination::graph::EpochGraph::successors_ ` - custom unpacking
- [ ] `vt::tests::unit::MyCol::op6_msgs_ ` - custom behavior
- [ ] `vt::trace::Trace::log_file_` - definition unavailable, `error: invalid application of ‘sizeof’ to incomplete type ‘vt::trace::vt_gzFile’`
- [ ] `vt::tutorial::ParticleMsg::a ` - custom serialization of raw pointer
- [ ] `vt::vrt::collection::MigrateMsg::owns_elm_ `

src/vt/pipe/callback/cb_union/cb_raw.h
```diff
@@ -255,6 +255,9 @@ struct GeneralCallback {
       vtAssert(0, "Should be unreachable");
       break;
     }
+
+    s.skip(active_);
```
PhilMiller, line 259: Stylistically, I'd prefer having `skip` calls directly alongside the block of code that takes care of the variable in question.

cz4rs: moved close to the `switch` case

src/vt/pipe/callback/cb_union/cb_raw.h
```diff
@@ -222,7 +222,10 @@ struct GeneralCallback {
     using EnumDataType = typename std::underlying_type<CallbackEnum>::type;
     EnumDataType val = static_cast<EnumDataType>(active_);
     s | val;
+    s.skip(active_);
+
     active_ = static_cast<CallbackEnum>(val);
+    s.skip(u_); // only serialize actual content of the union
```
PhilMiller, line 228: On second look, I realize that specifically in the footprinting case, serializing just the active member hypothetically undercounts the size of the callback itself, even as the payload of the callback is likely to be substantially larger.

A rigorous solution would be something like `s.addBytes(sizeof(u_) - sizeof(u_.active_member_))` I think - i.e. round up the leftover and padding bytes and add that.

As with everything else, the case of concern is if we happen to have many instances of a type accumulating unexpectedly, and blowing up memory, and that's the one we really want footprinting to illuminate.

src/vt/pipe/callback/cb_union/cb_raw.h
```diff
@@ -219,37 +219,43 @@ struct GeneralCallback {
...
   void serialize(SerializerT& s) {
-    using EnumDataType = typename std::underlying_type<CallbackEnum>::type;
-    EnumDataType val = static_cast<EnumDataType>(active_);
-    s | val;
-    active_ = static_cast<CallbackEnum>(val);
+    s | active_;
+
+    // serialize actual content of the union and account for leftovers and
+    // padding bytes when footprinting
+    auto ser = [&s, this](auto& cb){
```
cz4rs, line 226: `nvcc` would complain without explicit capture of `this`:
```
/vt/src/vt/pipe/callback/cb_union/cb_raw.h:228:26: error: base operand of '->' is not a pointer
       s.addBytes(sizeof(this->u_) - sizeof(cb));
                          ^~
```

lifflander (approved):

src/vt/pipe/callback/cb_union/cb_raw.h
cz4rs: :+1: for noticing this, fixed it along with `NoCB` case

PR #1280, opened 2021-02-23
1266: Use checkpoint::reconstructPointedToObjectIfNeeded for reconstructing migrated collection element

### LINKED ISSUE #1266: Collection migration should use checkpoint reconstructor instead of requiring default construction
**What Needs to be Done?**

Currently, it seems that collections must have a default constructor. We want to loosen that requirement to just being reconstructible according to the checkpoint serializable trait

JacobDomagala:
Fixes #1266

### DIFF
--- a/examples/collection/migrate_collection.cc
+++ b/examples/collection/migrate_collection.cc
@@ -48,13 +48,14 @@
 static constexpr int32_t const default_num_elms = 16;
 
 struct Hello : vt::Collection<Hello, vt::Index1D> {
-  Hello() = default;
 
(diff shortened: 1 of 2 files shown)


cz4rs (approved):
looks good to me
(note that this is changing the order of variables being serialized, but I don't think that it will be a problem)

PhilMiller (approved):

lifflander (approved):

JacobDomagala:
Updated PR with the use case

cz4rs (approved):

PhilMiller (approved):

PR #1288, opened 2021-02-25
1229: Explore enabling debug prints in release builds and assertions

### LINKED ISSUE #1229: Explore enabling debug prints in release builds and assertions
Create multiple levels of debug prints. First level can be enabled in release.

Terse, normal, verbose

cz4rs:
fixes #1229


cmake/define_build_types.cmake
```diff
@@ -19,13 +19,6 @@ else()
 endif()
 list(REMOVE_DUPLICATES VT_CONFIG_TYPES)
 
-# to speedup compiling, we can disable vt_debug_print
-if (${VT_DEBUG_FAST})
```
cz4rs, line 23: this was only ever used in `build_vt.pl`

cmake/define_build_types.cmake
```diff
@@ -235,50 +228,18 @@ endif()
...
-    set(cmake_config_debug_enabled_${loop_build_type} 0)
-  endif()
-
-  # use the debug_print modes specified for this build type before the loop
-  set(
-    cmake_vt_debug_modes
-    ${cmake_vt_debug_modes_${loop_build_type}}
-  )
-
   # assume production mode for everything except debug or CI build
```
cz4rs, line 236: the only remaining use for this would be `Production Build` information in the banner:
 ```cpp
#if vt_check_enabled(production)
  features.push_back(vt_feature_str_production);
#endif
 ```

scripts/build_vt.pl
```diff
@@ -101,11 +101,6 @@ sub mk {
 
 my $source_base_dir = "../$vt";
 
-my $fast_str = "";
```
cz4rs, line 104: after taking a closer look, I have decided to remove the script completely

src/vt/configs/debug/debug_masterconfig.h
```diff
@@ -52,38 +52,13 @@
...
 
-#define vt_debug_enabled cmake_config_debug_enabled
 #define vt_debug_force_enabled 0
 
 #include "vt/configs/debug/debug_config.h"
 #include "vt/configs/debug/debug_print.h"
 
 namespace vt { namespace config {
 
-#if !vt_debug_enabled
```
cz4rs, line 63: this no longer depends on build type - we could introduce explicit configuration variable instead (`vt_debug_print_enabled`?)

tests/unit/test_parallel_harness.h
```diff
@@ -155,8 +155,8 @@ struct TestParallelHarnessAny : TestHarnessAny<TestBase> {
 private:
   std::pair<int, char**>
   injectAdditionalArgs(int old_argc, char** old_argv) {
-    std::copy(
```
cz4rs, line 158: random fix: binary name wasn't placed correctly

src/vt/configs/features/features_featureswitch.h
```diff
@@ -45,29 +45,28 @@
...
-
 /*
  * Strings for various vt features
  */
 
 #define vt_feature_str_bit_check_overflow "Check bitfield overflow"
+#define vt_feature_str_cons_multi_idx     "Collection Constructor Positional"
 #define vt_feature_str_detector           "C++ Trait Detector"
+#define vt_feature_str_diagnostics        "Performance analysis diagnostics"
+#define vt_feature_str_fcontext           "fcontext Threading"
```
cz4rs, line 56: added missing string for `fcontext` (and sorted all the strings for clarity)

cz4rs:
@lifflander 
This was mostly about removing code so far (no categorization done for `Terse, normal, verbose`). I am looking at simplifying things even further if we decide to stick with `vt_debug_print` always enabled, but I need your feedback / thoughts on general direction for logging.
- I haven't measured the impact on build time precisely - looking at the intermediate CI builds, there is some hit, but I have not a comparison with caching disabled yet.
- Also, how often is the configuration below actually used? I'm not sure if keeping it in cmake source is preferential to having it directly in C++ code.
https://github.com/DARMA-tasking/vt/blob/0a4bacbce7788ae72e049ec94691abc4dd515d80/cmake/define_build_types.cmake#L29-L59

src/vt/configs/error/config_assert.h
```diff
@@ -105,75 +105,56 @@
 #define vt_make_list_strings(a) #a,
 #define argsToString(...) vt_applyxn(vt_make_list_strings,__VA_ARGS__)
 
-#if vt_check_enabled(production)
```
cz4rs, line 108: enabling assertions for release build

note:
There is `#if vt_check_enabled(assert_no_fail)` conditional below that I have not removed yet, but `vt_feature_cmake_assert_no_fail` is not defined anywhere. Should we always enforce failing or create a config option for `assert_no_fail`?

cmake/define_build_types.cmake
```diff
@@ -19,13 +19,6 @@ else()
...
-# to speedup compiling, we can disable vt_debug_print
-if (${VT_DEBUG_FAST})
-  set(VT_DEBUG_MODE_ON 0)
-else()
-  set(VT_DEBUG_MODE_ON 1)
-endif()
-
 # all vt_debug_print and vt_print categories we can potentially compile in
 set(
   cmake_vt_debug_modes_all
```
cz4rs, line 24: compared to `vt::config::CatEnum` we are missing following categories here:
```cpp
  event        = 1ull<<7,
  rdma_state   = 1ull<<13,
  sequence_vrt = 1ull<<19,
  vrt          = 1ull<<24,
  worker       = 1ull<<26,
```

cmake/define_build_types.cmake
lifflander: We can delete `build_vt.pl` and `build_vt.sh`..they are both old scripts that aren't used anymore AFAIK.

cmake/define_build_types.cmake
lifflander: I think we can remove it then, unless there is another use for it.

cmake/define_build_types.cmake
lifflander: That was intensional for the most part, but `event` should have probably been included. There's a question of whether we need that distinction now though.

docs/md/building.md
```diff
@@ -54,14 +54,13 @@ build configuration:
...
 | `vt_zoltan_enabled`              | 0               | Build with Zoltan enabled for `ZoltanLB` support |
 | `vt_mpi_guards`                  | 0               | Guards against mis-use of MPI calls in code using *vt* |
-| `vt_fcontext_enabled`            | 0               | Enable user-level threads through boost fcontext |
 | `vt_priorities_enabled`          | 1               | Enable prioritization of work (adds bits in envelope) |
 | `vt_diagnostics_enabled`         | 1               | Enable VT component diagnostics for performance analysis |
 | `vt_diagnostics_runtime_enabled` | 0               | Enable VT component diagnostics at runtime by default |
 | `vt_priority_bits_per_level`     | 3               | Number of bits per level of priority in envelope |
 | `vt_build_extended_tests`        | 1               | Build with full, extended testing |
 | `vt_unity_build_enabled`         | 0               | Build with Unity/Jumbo mode enabled (requires CMake >= 3.16) |
-| `vt_fcontext_enabled`            | 1               | Force use of fcontext for threading |
```
lifflander, line 64: I'm a little confused on the wording here that was changed in a different PR? Does that flag enable fcontext *and* force its use in worker or just enable it?

scripts/build_vt.pl
lifflander: With this change, we should probably delete the `vt-auto-build` repository also since it's the one that calls this script. Now that checkpoint and detector can be put into the `vt/lib` directory, there's no reason for the builder script I think.

src/vt/configs/debug/debug_masterconfig.h
lifflander: I think we still need a mode where all prints are compile-time removed. I think it shouldn't be the default though for release.

src/vt/configs/error/config_assert.h
lifflander: I think that assertions, if enabled, should always cause the program to fail in general. The reason that option exists is the rare occasion that you want the code to run past a bug to observe the behavior. I think we could add a runtime argument for doing just that---it doesn't really make sense at compile-time.

lifflander: I still think there should be a mode where assertions are disabled, but that should be a cmake configuration variable that is opt-in.

tests/unit/test_parallel_harness.h
lifflander: Good catch

cmake/define_build_types.cmake
lifflander: We could say that a production build disables all debug prints at compile time and disables assertions. What do you think about that idea?

lifflander (changes requested):

lifflander:
Overall I like the direction this is headed. But, I think we should preserve the mode where debug printing is completely compiled out and assertions are disabled. Maybe we can all that `production`, with an opt-in flag at compile-time to enable it.

I would like to get others feedback on this issue: @PhilMiller @nlslatt

src/vt/configs/error/config_assert.h
nlslatt: I agree about an opt-in disable of assertions and debug prints, so that we can optimize fully for production runs.

docs/md/building.md
cz4rs: It does both, enables fcontext and forces its use in worker.
There is only one configuration variable defined that is used both in `worker::WorkerSeq` and `seq::SeqULTContext` - should this be controlled independently?

cz4rs: + should the description be changed back to `Enable user-level threads through boost fcontext`?

lifflander: I think we can leave it alone for now--enabling and forcing the use. Later on, depending on @PhilMiller work with ULTs, we might want to have seperate options available.

cmake/define_build_types.cmake
cz4rs: Sounds good to me.
- this would replace replace `vt_debug_enabled` (which was deduced from build type) with explicit, opt-in **`vt_production_build`** configuration variable (as discussed in the assertions thread)
- do you think it is worth it to have separate configuration variables for debug prints and assertions?

src/vt/configs/error/config_assert.h
cz4rs: Introduced `vt_production_build_enabled` CMake configuration variable (default off).

cmake/define_build_types.cmake
cz4rs: So is this fine-grained control mechanism needed? We have the runtime flags as a basic tool to control which debug prints will be displayed. Now we will also have the possibility to disable assertions and debug prints completely with the opt-in.
We could:
- remove `CatEnum` mechanism completely and simplify debug printing logic 
- at least move this list directly into C++ code (as some `constexpr`) to avoid having it separated from actual debug code (this is definitely "advanced" control and requires rebuilding anyways)

cz4rs:
`vt_production_build_enabled=1` working as expected:

```
hello_world  --vt_debug_runtime
```
image

src/vt/configs/error/config_assert.h
cz4rs: + runtime argument `--vt_no_assert_fail`

cmake/define_build_types.cmake
cz4rs: `cmake_vt_debug_modes_all` was removed, compile-time policy is now based on `vt_production_build` config

cz4rs:
remaining issues:
- `macosx` build fails with following error:
```
 vt: [1] ------------------------------------------- Runtime Error: System Aborting! --------------------------------------------
 vt: [1] ------------------------------------------------ Fatal Error on Node 1 -------------------------------------------------
 vt: [1] ------------------------------------------------------------------------------------------------------------------------
 vt: [1] 
 vt: [1]              Reason: Must have valid op
 vt: [1]    Assertion failed: (op_id != no_op_id)
 vt: [1]                Node: 1
 vt: [1]           Num Nodes: 2
 vt: [1]                File: /Users/runner/work/vt/vt/src/vt/group/collective/group_info_collective.cc
 vt: [1]                Line: 587
 vt: [1]            Function: tree
(...)
vt: [1] 5   18  0x109db95a1   vt::group::InfoColl::tree(vt::group::GroupMsg<vt::messaging::ActiveMsg<vt::messaging::EpochTagActiveEnvelope> >*) + 273
```
This is hard to debug because I don't have access to any mac machine, currently suspecting that for wrong ctor for `GroupMsg` is used.

- alpine build - had segmentation faults during compilation and some other errors randomly, trying to reproduce the issue consistently

PhilMiller:
For compilation segfaults try using less parallelism during the build. That's often the result of memory exhaustion.

cz4rs:
> I believe this is a known bug. Refer to #1006

That's right - a329082 works around this but a long term solution for #1006 is required.
As far as this PR goes, the only remaining issue is alpine not respecting `--vt_no_assert_fail` flag.

lifflander (approved):
Looks great to me!

cz4rs:
```
The following tests FAILED:
	  8 - vt_example:insertable_collection_2 (Failed)
```
This is a recently reported bug - #1317. I will see if there is any quick fix for the memory leak possible.

PR #1298, opened 2021-03-02
1296: Avoid running LB on warm-up phase

### LINKED ISSUE #1296: Never run load balancer on first phase
**What Needs to be Done?**

nlslatt:
Unless we're running LB every phase, skip running LB on phase 0, which is a warm-up/app initialization phase, unless an LB spec file is used.  Now LB at intervals happens on `phase % interval == 1`.  I made this PR directly on the release branch because we already have a very similar change on develop.

Closes #1296


nlslatt:
@bathmatt FYI

PhilMiller (approved):

lifflander (approved):

JacobDomagala (approved):

PR #1301, opened 2021-03-05
1300 Make collection element insertion idempotent

### LINKED ISSUE #1300: Collection insertions are non-idempotent after migration
**What Needs to be Done?**

bvh does multiple insertions of the same index expecting them to be idempotent. This works out if the load balancer doesn't run because it does a local check to ensure not to insert multiple times. However, with the load balancer, two copies of the element end up existing causing a crash.

The solution is to query the home node to ensure that the insertion has not happened already, even if the insertion is off home (the user is allowed to insert wherever they desire).

lifflander:
Fixes #1300

### DIFF
--- a/src/vt/topos/location/location.impl.h
+++ b/src/vt/topos/location/location.impl.h
@@ -181,6 +181,35 @@ void EntityLocationCoord<EntityID>::registerEntity(
 }
 
 template <typename EntityID>
+void EntityLocationCoord<EntityID>::registerEntityRemote(
+  EntityID const& id, NodeType const& home, NodeType const create_node,
+  LocMsgActionType msg_action
+) {
+  auto reg_iter = local_registered_.find(id);
+  vtAssert(
+    reg_iter == local_registered_.end(),
+    "EntityLocationCoord entity should not already be registered"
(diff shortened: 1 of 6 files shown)


src/vt/topos/location/location.impl.h
```diff
@@ -180,6 +180,35 @@ void EntityLocationCoord<EntityID>::registerEntity(
   }
 }
 
+template <typename EntityID>
+void EntityLocationCoord<EntityID>::registerEntityRemote(
+  EntityID const& id, NodeType const& home, NodeType const create_node,
+  LocMsgActionType msg_action
+) {
+  auto reg_iter = local_registered_.find(id);
```
PhilMiller, line 188: Should we be asserting that `home == this_node`?

PhilMiller (approved):
Looks good

src/vt/topos/location/location.impl.h
lifflander: Yes, good point. I'll add that assertion.

lifflander: Added

PR #1305, opened 2021-03-08
1272: Update the usage of scheduler in code

### LINKED ISSUE #1272: Rename runScheduler to runSchedulerForTesting
VT's `runScheduler` should not be called directly by users. `runSchedulerWhile` is the correct call for applications. There may be cases where `runScheduler` is useful for testing. Rename it appropriately. Update documentation if needed.

JacobDomagala:
Fixes #1272 

For the most part I've changed the uses of `while() runScheduler()` to `theSched()->runSchedulerWhile()`. The exceptions were:
https://github.com/DARMA-tasking/vt/blob/4a0f2cd17d11fc2099e93bbe850acf17f02eec3e/tests/perf/memory_checker.cc#L75-L86
https://github.com/DARMA-tasking/vt/blob/4a0f2cd17d11fc2099e93bbe850acf17f02eec3e/tests/unit/scheduler/test_scheduler_loop.cc#L100-L105

</br>
Since `runScheduler` was just a wrapper for `theSched()->scheduler()` and it's meant to be used only by our internal code, I think it's better to remove `runScheduler` altogether and use  `theSched()->scheduler()` instead.

### DIFF
--- a/src/vt/runtime/runtime.h
+++ b/src/vt/runtime/runtime.h
@@ -191,7 +191,7 @@ struct Runtime {
   void computeAndPrintDiagnostics();
 
   /**
-   * \brief Run the scheduler once
+   * \internal \brief Run the scheduler once
    */
   void runScheduler();
 
--- a/src/vt/runtime/runtime.h
+++ b/src/vt/runtime/runtime.h
@@ -191,11 +191,6 @@ struct Runtime {
   void computeAndPrintDiagnostics();
 
   /**
-   * \internal \brief Run the scheduler once
-   */
-  void runScheduler();
-
-  /**
    * \brief Abort--die immediately after spitting out error message
    *
    * \param[in] abort_str the error message
--- a/tests/unit/active/test_pending_send.cc
+++ b/tests/unit/active/test_pending_send.cc
@@ -98,8 +98,8 @@ TEST_F(TestPendingSend, test_pending_send_hold) {
   // break out
   int k = 0;
 
-  theSched()->runSchedulerWhile([&k, ep]() mutable {
-    return theTerm()->isEpochTerminated(ep) || (k++ > 10);
+  theSched()->runSchedulerWhile([&k, ep] {
+    return !theTerm()->isEpochTerminated(ep) && (++k <= 10);
   });
 
   // Epoch should not end with a valid pending send created in an live epoch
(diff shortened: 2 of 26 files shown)


src/vt/runtime/runtime.h
```diff
@@ -190,11 +190,6 @@ struct Runtime {
    */
   void computeAndPrintDiagnostics();
 
-  /**
-   * \brief Run the scheduler once
-   */
-  void runScheduler();
```
JacobDomagala, line 196: Is this being used by EMPIRE? It was unused by our code so unless it's being used there I think we should remove it.

lifflander: Maybe we should deprecate it?

nlslatt: It's not being used by EMPIRE.

tests/unit/active/test_pending_send.cc
```diff
@@ -97,13 +97,10 @@ TEST_F(TestPendingSend, test_pending_send_hold) {
...
-  while (!theTerm()->isEpochTerminated(ep)) {
-    k++;
-    vt::runScheduler();
-    if (k > 10) {
-      break;
-    }
-  }
+
+  theSched()->runSchedulerWhile([&k, ep]() mutable {
+    return theTerm()->isEpochTerminated(ep) || (k++ > 10);
```
nlslatt, line 102: I think the equivalent of the previous code would use `(++k > 10)` because the increment used to complete before the comparison.

nlslatt: Wait, this is a while loop, I think this still isn't right.  This currently runs while terminated instead of until terminated, right?

lifflander (approved):
Looking good. Let's completely remove `runScheduler` and change the name for `theSched()->scheduler()` to `theSched()->runSchedulerImpl(int num_times = 1)` and mark it `\internal` on doxygen and document that it shouldn't be used by users.

cz4rs (approved):
Looks good to me

JacobDomagala:
> change the name for `theSched()->scheduler()` to `theSched()->runSchedulerImpl(int num_times = 1)`

@lifflander Do we want to add new param to this function? If we want to emphasise that it will spin the scheduler only once, then I guess we could just rename it to `runSchedulerOnceImpl`?

lifflander:
> > change the name for `theSched()->scheduler()` to `theSched()->runSchedulerImpl(int num_times = 1)`
> 
> @lifflander Do we want to add new param to this function? If we want to emphasise that it will spin the scheduler only once, then I guess we could just rename it to `runSchedulerOnceImpl`?

Let's rename it to `runSchedulerOnceImpl` with no arguments.

lifflander:
@JacobDomagala When you're ready for this to be merged, let me (or @jstrzebonski or @PhilMiller) know. We can override the Mac build to get this merged.

PR #1313, opened 2021-03-09
#1311: Fix cmake to not add duplicate -std=c++14 flag that bothers nvcc

PhilMiller:


cz4rs (approved):
Looks good to me.

`target_compile_options` and `set_property(TARGET ...)` are "moder cmake" reccomended, but it would probably require much more work - we could add an issue to explore that later.

cmake-modules/SetCXXCompilerFlags.cmake
```diff
@@ -3,28 +3,26 @@
 
 macro(set_darma_compiler_flags)
 
-set(CXX_STANDARD_FLAGS)
+set_property(DIRECTORY PROPERTY CXX_STANDARD 14)
```
cz4rs, line 6: on a second thought: after reading about the "decay" behavior of `CXX_STANDARD`, do we need to set [CXX_STANDARD_REQUIRED](https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD_REQUIRED.html#prop_tgt:CXX_STANDARD_REQUIRED) to `ON`?

lifflander (approved):

PR #1322, opened 2021-03-12
413 Implement user-level threads in scheduler, add explicit context, implement suspend/resume API

### LINKED ISSUE #413: Routine to run function/action/lambda in suspendable thread
For the expedient sake of blocking reductions as currently used in EMPIRE, it would be really useful if a handler could call some function like

`void runInThread(std::function<void()> F)`

That would create a context thread and call the function `F` within it. We need ancillary routines to
* get a reference to the running thread
* suspend the running thread (error if it wasn't one created as above)
* resume a thread

In the blocking reduction use case, the VT handler would call the above routine, passing an argument of the current code that calls reductions. The reduction implementation would add an action to resume the current thread when the reduction completes, submit its contribution, and then suspend.

lifflander:
Fixes #413

Tasks:
- [x] dust off ULTs (fcontext) and see if tests/examples build---if not fix them
- [x] refactor all context into new context classes
- [x] make Runnable a instance that can be transferred to the scheduler
- [x] change how tasks are enqueued in general to be after the runnable is created
- [x] start suspending/resuming threads and build user interface for this

Left to do:
- [x] Make stack size a command-line argument and start with a sane default
- [x] `ThreadManager` is a static class right now. Promote to instance, put in scheduler?
- [x] Write a couple more tests that check that epoch stack is preserved after suspend and is correctly resumed

### DIFF
--- a/src/vt/context/context.cc
+++ b/src/vt/context/context.cc
@@ -46,9 +46,10 @@
 #include "vt/context/runnable_context/from_node.h"
 #include "vt/runnable/runnable.h"
 
-#if !vt_check_enabled(trace_only)
-# include "vt/runtime/runtime.h"
+#if vt_check_enabled(trace_only)
 # include "vt/trace/trace_common.h"
+#else
+# include "vt/runtime/runtime.h"
 #endif
 
 #include <string>
--- a/src/vt/context/context.cc
+++ b/src/vt/context/context.cc
@@ -122,7 +122,10 @@ NodeType Context::getFromNodeCurrentTask() const {
 trace::TraceEventIDType Context::getTraceEventCurrentTask() const {
 #if !vt_check_enabled(trace_only)
   if (getTask() != nullptr) {
-    return theContext()->getTask()->get<ctx::Trace>()->getEvent();
+    auto t = getTask()->get<ctx::Trace>();
+    if (t != nullptr) {
+      return t->getEvent();
+    }
   }
 #endif
   return trace::no_trace_event;
--- a/src/vt/context/context.cc
+++ b/src/vt/context/context.cc
@@ -46,10 +46,12 @@
 #include "vt/context/runnable_context/from_node.h"
 #include "vt/runnable/runnable.h"
 
-#if vt_check_enabled(trace_only)
-# include "vt/trace/trace_common.h"
-#else
+#if !vt_check_enabled(trace_only)
 # include "vt/runtime/runtime.h"
+#endif
+
+#if vt_check_enabled(trace_enabled)
+# include "vt/trace/trace_common.h"
 #endif
 
 #include <string>
--- a/src/vt/context/context.cc
+++ b/src/vt/context/context.cc
@@ -44,9 +44,11 @@
 
 #include "vt/context/context.h"
 #include "vt/context/runnable_context/from_node.h"
+#include "vt/runnable/runnable.h"
 
 #if !vt_check_enabled(trace_only)
 # include "vt/runtime/runtime.h"
+# include "vt/trace/trace_common.h"
 #endif
 
 #include <string>
@@ -106,6 +108,16 @@ NodeType Context::getFromNodeCurrentTask() const {
   }
   return getNode();
 }
+
+#if vt_check_enabled(trace_enabled)
+trace::TraceEventIDType Context::getTraceEventCurrentTask() const {
+  if (getTask() != nullptr) {
+    return theContext()->getTask()->get<ctx::Trace>()->getEvent();
+  } else {
+    return trace::no_trace_event;
+  }
+}
+#endif /* vt_check_enabled(trace_enabled) */
 
 }}  // end namespace vt::ctx
 
--- a/src/vt/context/context.cc
+++ b/src/vt/context/context.cc
@@ -44,10 +44,14 @@
 
 #include "vt/context/context.h"
 #include "vt/context/runnable_context/from_node.h"
-#include "vt/runnable/runnable.h"
 
-#if !vt_check_enabled(trace_only)
+#if vt_check_enabled(trace_only)
+namespace vt { namespace runnable {
+struct RunnableNew {};
+}} /* end namespact vt::runnable */
+#else
 # include "vt/runtime/runtime.h"
+# include "vt/runnable/runnable.h"
 #endif
 
 #if vt_check_enabled(trace_enabled)
@@ -103,22 +107,25 @@ void Context::setTask(runnable::RunnableNew* in_task) {
 }
 
 NodeType Context::getFromNodeCurrentTask() const {
+#if !vt_check_enabled(trace_only)
   if (getTask() != nullptr) {
     auto from = getTask()->get<ctx::FromNode>();
     if (from != nullptr) {
       return from->get();
     }
   }
+#endif
   return getNode();
 }
 
 #if vt_check_enabled(trace_enabled)
 trace::TraceEventIDType Context::getTraceEventCurrentTask() const {
+#if !vt_check_enabled(trace_only)
   if (getTask() != nullptr) {
     return theContext()->getTask()->get<ctx::Trace>()->getEvent();
-  } else {
-    return trace::no_trace_event;
   }
+#endif
+  return trace::no_trace_event;
 }
 #endif /* vt_check_enabled(trace_enabled) */
 
--- a/src/vt/context/context.cc
+++ b/src/vt/context/context.cc
@@ -43,6 +43,7 @@
 */
 
 #include "vt/context/context.h"
+#include "vt/context/runnable_context/from_node.h"
 
 #if !vt_check_enabled(trace_only)
 # include "vt/runtime/runtime.h"
@@ -94,6 +95,16 @@ DeclareClassOutsideInitTLS(Context, WorkerIDType, thisWorker_, no_worker_id)
 
 void Context::setTask(runnable::RunnableNew* in_task) {
   cur_task_ = in_task;
+}
+
+NodeType Context::getFromNodeCurrentTask() const {
+  if (getTask() != nullptr) {
+    auto from = getTask()->get<ctx::FromNode>();
+    if (from != nullptr) {
+      return from->get();
+    }
+  }
+  return getNode();
 }
 
 }}  // end namespace vt::ctx
--- a/src/vt/context/runnable_context/base.h
+++ b/src/vt/context/runnable_context/base.h
@@ -45,6 +45,8 @@
 #if !defined INCLUDED_VT_CONTEXT_RUNNABLE_CONTEXT_BASE_H
 #define INCLUDED_VT_CONTEXT_RUNNABLE_CONTEXT_BASE_H
 
+#include "vt/configs/types/types_type.h"
+
 namespace vt { namespace ctx {
 
 /**
--- a/src/vt/context/runnable_context/collection.h
+++ b/src/vt/context/runnable_context/collection.h
@@ -93,6 +93,4 @@ private:
 
 }} /* end namespace vt::ctx */
 
-#include "vt/context/runnable_context/collection.impl.h"
-
 #endif /*INCLUDED_VT_CONTEXT_RUNNABLE_CONTEXT_COLLECTION_H*/
--- a/src/vt/context/runnable_context/lb_stats.cc
+++ b/src/vt/context/runnable_context/lb_stats.cc
@@ -48,12 +48,6 @@
 namespace vt { namespace ctx {
 
 void LBStats::begin() {
-  // Save current element ID context
-  prev_elm_id_ = theCollection()->getCurrentContext();
-
-  // Set the current element context for communication stats
-  theCollection()->setCurrentContext(cur_elm_id_);
-
   // record start time
   if (should_instrument_) {
     stats_->startTime();
@@ -61,9 +55,6 @@ void LBStats::begin() {
 }
 
 void LBStats::end() {
-  // Set the element ID context back the previous element
-  theCollection()->setCurrentContext(prev_elm_id_);
-
   // record end time
   if (should_instrument_) {
     stats_->stopTime();
@@ -80,6 +71,10 @@ void LBStats::suspend() {
 
 void LBStats::resume() {
   begin();
+}
+
+typename LBStats::ElementIDStruct const& LBStats::getCurrentElementID() const {
+  return cur_elm_id_;
 }
 
 }} /* end namespace vt::ctx */
--- a/src/vt/context/runnable_context/set_context.cc
+++ b/src/vt/context/runnable_context/set_context.cc
@@ -52,9 +52,21 @@ void SetContext::begin() {
   // previous context (pop) and set the new task (push)
   prev_task_ = theContext()->getTask();
   ContextAttorney::setTask(cur_task_);
+
+  vt_debug_print(
+    context, node,
+    "{}: begin(): prev={}, task={}\n",
+    print_ptr(this), print_ptr(prev_task_.get()), print_ptr(cur_task_.get())
+  );
 }
 
 void SetContext::end() {
+  vt_debug_print(
+    context, node,
+    "{}: end(): prev={}, task={}\n",
+    print_ptr(this), print_ptr(prev_task_.get()), print_ptr(theContext()->getTask())
+  );
+
   vtAssert(
     theContext()->getTask() == cur_task_, "Must be correct task"
   );
@@ -62,10 +74,22 @@ void SetContext::end() {
 }
 
 void SetContext::suspend() {
+  vt_debug_print(
+    context, node,
+    "{}: suspend(): prev={}, task={}\n",
+    print_ptr(this), print_ptr(prev_task_.get()), print_ptr(cur_task_.get())
+  );
+
   end();
 }
 
 void SetContext::resume() {
+  vt_debug_print(
+    context, node,
+    "{}: resume(): prev={}, task={}\n",
+    print_ptr(this), print_ptr(prev_task_.get()), print_ptr(cur_task_.get())
+  );
+
   begin();
 }
 
--- a/src/vt/context/runnable_context/set_context.cc
+++ b/src/vt/context/runnable_context/set_context.cc
@@ -50,15 +50,15 @@ namespace vt { namespace ctx {
 void SetContext::begin() {
   // we have to handle the ugly handler-inside-handler case.. preserve the
   // previous context (pop) and set the new task (push)
-  nonowning_prev_task_ = theContext()->getTask();
-  ContextAttorney::setTask(nonowning_cur_task_);
+  prev_task_ = theContext()->getTask();
+  ContextAttorney::setTask(cur_task_);
 }
 
 void SetContext::end() {
   vtAssert(
-    theContext()->getTask() == nonowning_cur_task_, "Must be correct task"
+    theContext()->getTask() == cur_task_, "Must be correct task"
   );
-  ContextAttorney::setTask(nonowning_prev_task_);
+  ContextAttorney::setTask(prev_task_);
 }
 
 void SetContext::suspend() {
--- a/src/vt/context/runnable_context/set_context.cc
+++ b/src/vt/context/runnable_context/set_context.cc
@@ -2,7 +2,7 @@
 //@HEADER
 // *****************************************************************************
 //
-//                                set_context.h
+//                                set_context.cc
 //                           DARMA Toolkit v. 1.0.0
 //                       DARMA/vt => Virtual Transport
 //
@@ -42,52 +42,31 @@
 //@HEADER
 */
 
-#if !defined INCLUDED_VT_CONTEXT_RUNNABLE_CONTEXT_SET_CONTEXT_H
-#define INCLUDED_VT_CONTEXT_RUNNABLE_CONTEXT_SET_CONTEXT_H
-
-#include "vt/context/runnable_context/base.h"
-#include "vt/runnable/runnable.fwd.h"
+#include "vt/context/runnable_context/set_context.h"
+#include "vt/context/context_attorney.h"
 
 namespace vt { namespace ctx {
 
-/**
- * \struct SetContext
- *
- * \brief Set the context of the current running task for query by other
- * components or users.
- */
-struct SetContext final : Base {
+void SetContext::begin() {
+  // we have to handle the ugly handler-inside-handler case.. preserve the
+  // previous context (pop) and set the new task (push)
+  nonowning_prev_task_ = theContext()->getTask();
+  ContextAttorney::setTask(nonowning_cur_task_);
+}
 
-  /**
-   * \brief Construct a \c SetContext
-   *
-   * \param[in] in_nonowning_cur_task the current task (non-owning ptr held)
-   */
-  explicit SetContext(runnable::RunnableNew* in_nonowning_cur_task)
-    : nonowning_cur_task_(in_nonowning_cur_task)
-  {}
+void SetContext::end() {
+  vtAssert(
+    theContext()->getTask() == nonowning_cur_task_, "Must be correct task"
+  );
+  ContextAttorney::setTask(nonowning_prev_task_);
+}
 
-  /**
-   * \brief Preserve the existing task and replace with a new one
-   */
-  void begin() final override;
+void SetContext::suspend() {
+  end();
+}
 
-  /**
-   * \brief Restore the previous existing task to the context (if there was one)
-   */
-  void end() final override;
-
-  void suspend() final override;
-
-  void resume() final override;
-
-private:
-  /// The previous runnable that was in the context
-  runnable::RunnableNew* nonowning_prev_task_ = nullptr;
-  /// The new runnable that is replacing it
-  runnable::RunnableNew* nonowning_cur_task_ = nullptr;
-};
+void SetContext::resume() {
+  begin();
+}
 
 }} /* end namespace vt::ctx */
-
-#endif /*INCLUDED_VT_CONTEXT_RUNNABLE_CONTEXT_SET_CONTEXT_H*/
--- a/src/vt/context/runnable_context/td.h
+++ b/src/vt/context/runnable_context/td.h
@@ -57,24 +57,26 @@ namespace vt { namespace ctx {
  * \struct TD
  *
  * \brief Context for termination detection to be preserved with a task. Manages
- * the epoch stack associated with running tasks.
+ * the epoch stack  associated with running tasks. Produces and consumes in the
+ * constructor and destructor to ensure termination is not detected early.
  */
 struct TD final : Base {
 
   /**
-   * \brief Construct with a given epoch
+   * \brief Construct with a given epoch; produce on that epoch.
    *
    * \param[in] in_ep the epoch
    */
   explicit TD(EpochType in_ep);
 
   /**
-   * \brief On destructor consume the epoch
+   * \brief When destroyed, consume the epoch held by the context.
    */
   virtual ~TD();
 
   /**
-   * \brief Construct with a message to extract the epoch
+   * \brief Construct with a message to extract the epoch; produce on that
+   * epoch.
    *
    * \param[in] msg the message to extract the epoch from
    */
@@ -82,13 +84,13 @@ struct TD final : Base {
   explicit TD(MsgPtrT msg);
 
   /**
-   * \brief During begin \c TD will produce on the epoch and push it on the
-   * epoch stack.
+   * \brief During begin \c TD will capture the epoch stack size and push \c ep_
    */
   void begin() final override;
 
   /**
-   * \brief During end \c TD will consume on the epoch and pop it off the stack
+   * \brief During end \c TD will pop all epochs off of the stack down to the
+   * size in captured in \c begin()
    */
   void end() final override;
 
--- a/src/vt/context/runnable_context/trace.h
+++ b/src/vt/context/runnable_context/trace.h
@@ -136,7 +136,7 @@ private:
 struct Trace : Base {
 
   template <typename... Args>
-  Trace(Args... args) {}
+  Trace(Args&&... args) {}
 
 };
 
--- a/src/vt/messaging/active.cc
+++ b/src/vt/messaging/active.cc
@@ -414,10 +414,7 @@ EventType ActiveMessenger::sendMsgBytes(
   }
 
   if (theContext()->getTask() != nullptr) {
-    auto lb = theContext()->getTask()->get<ctx::LBStats>();
-    if (lb != nullptr) {
-      lb->send(dest, msg_size, is_bcast);
-    }
+    theContext()->getTask()->send(dest, msg_size, is_bcast);
   }
 
   return event_id;
@@ -536,10 +533,7 @@ SendInfo ActiveMessenger::sendData(
   theTerm()->hangDetectSend();
 
   if (theContext()->getTask() != nullptr) {
-    auto lb = theContext()->getTask()->get<ctx::LBStats>();
-    if (lb != nullptr) {
-      lb->send(dest, num_bytes, false);
-    }
+    theContext()->getTask()->send(dest, num_bytes, false);
   }
 
   return SendInfo{event_id, send_tag, num};
--- a/src/vt/messaging/active.cc
+++ b/src/vt/messaging/active.cc
@@ -55,13 +55,7 @@
 #include "vt/utils/mpi_limits/mpi_max_tag.h"
 #include "vt/runtime/mpi_access.h"
 #include "vt/scheduler/scheduler.h"
-#include "vt/runnable/runnable.h"
-#include "vt/context/runnable_context/td.h"
-#include "vt/context/runnable_context/continuation.h"
-#include "vt/context/runnable_context/trace.h"
-#include "vt/context/runnable_context/from_node.h"
-#include "vt/context/runnable_context/set_context.h"
-#include "vt/context/runnable_context/lb_stats.h"
+#include "vt/runnable/make_runnable.h"
 
 namespace vt { namespace messaging {
 
@@ -951,23 +945,16 @@ bool ActiveMessenger::prepareActiveMsgToRun(
   }
 
   if (has_handler) {
-    auto r = std::make_unique<runnable::RunnableNew>(base, not is_term);
-    if (not is_term) {
-      r->template addContext<ctx::TD>(base);
-    }
-    r->template addContext<ctx::Continuation>(cont);
-    r->template addContext<ctx::Trace>(
-      base, handler, from_node, auto_registry::RegistryTypeEnum::RegGeneral
-    );
-    r->template addContext<ctx::FromNode>(from_node);
+    runnable::makeRunnable(base, not is_term, handler, from_node)
+      .withContinuation(cont)
+      .withTag(tag)
+      .withTDMsg(is_term)
+      .enqueue();
 
     if (is_term) {
       tdRecvCount.increment(1);
     }
     amHandlerCount.increment(1);
-
-    r->setupHandler(handler, false, tag);
-    theSched()->enqueue(base, std::move(r));
 
     if (not is_term) {
       theTerm()->consume(epoch,1,from_node);
--- a/src/vt/messaging/active.cc
+++ b/src/vt/messaging/active.cc
@@ -881,16 +881,6 @@ bool ActiveMessenger::recvDataMsg(
   );
 }
 
-NodeType ActiveMessenger::getFromNodeCurrentHandler() const {
-  if (theContext()->getTask() != nullptr) {
-    auto from = theContext()->getTask()->get<ctx::FromNode>();
-    if (from != nullptr) {
-      return from->get();
-    }
-  }
-  return theContext()->getNode();
-}
-
 bool ActiveMessenger::processActiveMsg(
   MsgSharedPtr<BaseMsgType> const& base, NodeType const& from,
   MsgSizeType const& size, bool insert, ActionType cont
--- a/src/vt/messaging/active.cc
+++ b/src/vt/messaging/active.cc
@@ -983,7 +983,7 @@ bool ActiveMessenger::prepareActiveMsgToRun(
     }
     amHandlerCount.increment(1);
 
-    r->setupHandler(RunnableEnum::Active, handler, from_node, tag);
+    r->setupHandler(handler, from_node, false, tag);
     theSched()->enqueue(base, std::move(r));
 
     if (not is_term) {
--- a/src/vt/messaging/active.cc
+++ b/src/vt/messaging/active.cc
@@ -429,12 +429,6 @@ EventType ActiveMessenger::sendMsgBytes(
   return event_id;
 }
 
-#if vt_check_enabled(trace_enabled)
-trace::TraceEventIDType ActiveMessenger::getCurrentTraceEvent() const {
-  return theContext()->getTask()->get<ctx::Trace>()->getEvent();
-}
-#endif
-
 EventType ActiveMessenger::doMessageSend(
   MsgSharedPtr<BaseMsgType>& base, MsgSizeType msg_size
 ) {
--- a/src/vt/messaging/active.cc
+++ b/src/vt/messaging/active.cc
@@ -953,7 +953,7 @@ bool ActiveMessenger::prepareActiveMsgToRun(
   if (has_handler) {
     auto r = std::make_unique<runnable::RunnableNew>(base, not is_term);
     if (not is_term) {
-      r->template addContext<ctx::TD>(epoch);
+      r->template addContext<ctx::TD>(base);
     }
     r->template addContext<ctx::Continuation>(cont);
     r->template addContext<ctx::Trace>(
--- a/src/vt/messaging/active.cc
+++ b/src/vt/messaging/active.cc
@@ -942,7 +942,7 @@ bool ActiveMessenger::prepareActiveMsgToRun(
     runnable::makeRunnable(base, not is_term, handler, from_node)
       .withContinuation(cont)
       .withTag(tag)
-      .withTDMsg(is_term)
+      .withTDEpochFromMsg(is_term)
       .enqueue();
 
     if (is_term) {
--- a/src/vt/messaging/active.cc
+++ b/src/vt/messaging/active.cc
@@ -960,7 +960,6 @@ bool ActiveMessenger::prepareActiveMsgToRun(
       base, handler, from_node, auto_registry::RegistryTypeEnum::RegGeneral
     );
     r->template addContext<ctx::FromNode>(from_node);
-    r->template addContext<ctx::SetContext>(r.get());
 
     if (is_term) {
       tdRecvCount.increment(1);
--- a/src/vt/messaging/active.cc
+++ b/src/vt/messaging/active.cc
@@ -967,7 +967,7 @@ bool ActiveMessenger::prepareActiveMsgToRun(
     }
     amHandlerCount.increment(1);
 
-    r->setupHandler(handler, from_node, false, tag);
+    r->setupHandler(handler, false, tag);
     theSched()->enqueue(base, std::move(r));
 
     if (not is_term) {
--- a/src/vt/messaging/active.h
+++ b/src/vt/messaging/active.h
@@ -1597,6 +1597,18 @@ struct ActiveMessenger : runtime::component::PollableComponent<ActiveMessenger> 
   template <typename T>
   void registerAsyncOp(std::unique_ptr<T> op);
 
+  /**
+   * \brief Block the current task's execution on an pollable async operation
+   * until it completes
+   *
+   * \note This may only be safely called if the current task is running in a
+   * user-level thread.
+   *
+   * \param[in] op the async operation to block on
+   */
+  template <typename T>
+  void blockOnAsyncOp(std::unique_ptr<T> op);
+
   template <typename SerializerT>
   void serialize(SerializerT& s) {
     s | maybe_ready_tag_han_
--- a/src/vt/messaging/active.h
+++ b/src/vt/messaging/active.h
@@ -1381,16 +1381,6 @@ struct ActiveMessenger : runtime::component::PollableComponent<ActiveMessenger> 
     ActiveClosureFnType fn, TagType const& tag = no_tag
   );
 
-  #if vt_check_enabled(trace_enabled)
-    /**
-     * \internal
-     * \brief Get the trace event on the handler running
-     *
-     * \return the trace event on the message that triggered the current handler
-     */
-    trace::TraceEventIDType getCurrentTraceEvent() const;
-  #endif
-
   /**
    * \internal
    * \brief Process an incoming active message
--- a/src/vt/messaging/active.h
+++ b/src/vt/messaging/active.h
@@ -1381,20 +1381,6 @@ struct ActiveMessenger : runtime::component::PollableComponent<ActiveMessenger> 
     ActiveClosureFnType fn, TagType const& tag = no_tag
   );
 
-  /**
-   * \internal
-   * \brief Get the from node for the current running handler (valid only while
-   * running a handler)
-   *
-   * For the current handler that is executing, get the node that sent the
-   * message that caused this handler to run. Note, for collection handlers this
-   * will not be the logical node that sent the message. It will be the node
-   * that last forwarded the message during location discovery.
-   *
-   * \return the node that sent the message that triggered the current handler
-   */
-  NodeType getFromNodeCurrentHandler() const;
-
   #if vt_check_enabled(trace_enabled)
     /**
      * \internal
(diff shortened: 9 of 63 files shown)


examples/hello_world/hello_world.cc
```diff
@@ -43,6 +43,9 @@
 */
 
 #include <vt/transport.h>
+#include <vt/scheduler/thread_action.h>
```
lifflander, line 46: This won't be merged..just playing with this interface.

src/vt/scheduler/thread_action.h
```diff
@@ -0,0 +1,127 @@
...
+#define INCLUDED_VT_SCHEDULER_THREAD_ACTION_H
+
+#include "vt/config.h"
+#include <context/fcontext.h>
+
+namespace vt { namespace scheduler {
+
+struct ThreadAction {
+
+  explicit ThreadAction(ActionType in_action)
```
lifflander, line 55: Need to pick up epoch here to produce and consume on suspend and resume

PhilMiller:
@jstrzebonski It looks like the testing is reporting as `Testing - passed` even when the build didn't complete.

src/vt/scheduler/thread_action.cc
```diff
@@ -0,0 +1,69 @@
...
+  : action_(in_action),
+    cur_epoch_(theMsg()->getEpoch()),
+    stack(create_fcontext_stack())
+{
+  theTerm()->produce(cur_epoch_);
+}
+
+ThreadAction::~ThreadAction() {
+  theTerm()->consume(cur_epoch_);
+  destroy_fcontext_stack(stack);
```
PhilMiller, line 62: I think these need to be in `runFnImpl` rather than here, since that accurately captures the active life of the thread.

PhilMiller:
This looks fine for what it is. The scheduler bits are where it will start to get interesting.

lifflander:
> This looks fine for what it is. The scheduler bits are where it will start to get interesting.

Cool, I'm working on the next bit. I'm thinking about tying the suspend to `AsyncOpWrapper`.

PhilMiller:
I think you're making the TD interaction too complicated, and in a way that could actually be wrong. Produce when a thread starts, consume when it's done, and don't mess around at suspend or resume.

PhilMiller:
Candidates for context classes to feed into `Runnable`:
* Trace events
* LB timing
* TD
* collection index/proxy

src/vt/context/runnable_context/base.h
```diff
@@ -0,0 +1,93 @@
...
+#define INCLUDED_VT_CONTEXT_RUNNABLE_CONTEXT_BASE_H
+
+namespace vt { namespace ctx {
+
+/**
+ * \struct Base
+ *
+ * \brief Base context for runnable tasks.
+ *
+ * \c ctx::Base is used to create contexts that are associated with tasks
```
lifflander, line 55: Lots of typos here...

src/vt/context/context.cc
```diff
@@ -43,8 +43,10 @@
 */
 
 #include "vt/context/context.h"
+#include "vt/runnable/runnable.h"
```
PhilMiller, line 46: This isn't needed - the forward declaration from the context header is sufficient to take and assign a pointer

src/vt/context/runnable_context/base.h
PhilMiller: s/build/built/ in at least a few instances

src/vt/context/runnable_context/collection.h
```diff
@@ -0,0 +1,98 @@
...
+   */
+  void begin() override;
+
+  /**
+   * \brief Remove the collection context
+   */
+  void end() override;
+
+  void suspend() override;
+  void resume() override;
```
PhilMiller, line 87: It's probably worth marking all of these `final`, since the calls to them are likely to be inlined, and that would facilitate devirtualization.

src/vt/context/runnable_context/collection.impl.h
```diff
@@ -0,0 +1,84 @@
...
+template <typename ColT>
+/*explicit*/ Collection<IndexT>::Collection(
+  vrt::collection::CollectionBase<ColT, IndexT>* elm
+) : idx_(elm->getIndex()),
+    proxy_(elm->getProxy())
+{ }
+
+template <typename IndexT>
+void Collection<IndexT>::begin() {
+  vrt::collection::InsertContextHolder<IndexT>::set(&idx_, proxy_);
```
PhilMiller, line 64: Is this context class only used for element insertions, or is the name of the thing it's calling a historical misnomer?

src/vt/context/runnable_context/lb_stats.impl.h
```diff
@@ -42,64 +42,79 @@
...
-      idx1, idx2, idx3, idx4
+  // Add the listener for the active messenger
+  std::unique_ptr<messaging::Listener> listener =
+    std::make_unique<vrt::collection::balance::LBListener>(
+      [&](NodeType dest, MsgSizeType size, bool bcast){
+        auto& stats = elm_->getStats();
+        stats.recvToNode(dest, elm_id, size, bcast);
+      }
     );
+  theMsg()->addSendListener(std::move(listener));
```
PhilMiller, line 84: This method of instrumenting sends seems rather overwrought, in that it involves allocating and inserting the listener object in a data structure.

There's an opportunity here - can we just add a `send` method to the context classes, for this use?

src/vt/context/runnable_context/td.h
```diff
@@ -0,0 +1,103 @@
...
+  /**
+   * \brief When suspended, \c TD will preserve any epochs pushed on the stack
+   * after begin and restore the stack back to the state before begin was
+   * invoked
+   */
+  void suspend() override;
+
+  /**
+   * \brief When resumed, \c TD will restore the stack back from when it was
+   * suspended
```
PhilMiller, line 91: Subtle note here - it will restore the stack values at and below the one in which the task is running. It's not going to touch the 'ambient' contents of the stack, which may differ between `suspend()` and `resume()` (safely, I believe)

src/vt/messaging/active.cc
```diff
@@ -870,7 +875,13 @@ bool ActiveMessenger::recvDataMsg(
...
 
 NodeType ActiveMessenger::getFromNodeCurrentHandler() const {
-  return current_node_context_;
+  if (theContext()->getTask() != nullptr) {
+    auto from = theContext()->getTask()->get<ctx::FromNode>();
+    if (from != nullptr) {
+      return from->get();
+    }
+  }
+  return theContext()->getNode();
```
PhilMiller, line 884: It seems like this could be refactored to be a method of `theContext` instead

PhilMiller: Or even a `static` member of `ctx::FromNode`, if `theContext` shouldn't have references to concrete classes it may be holding.

PhilMiller: The same applies for `getCurrentTraceEvent` above.

src/vt/messaging/active.impl.h
```diff
@@ -478,9 +481,7 @@ inline void ActiveMessenger::epochEpilogHandler(
   vtWarnInfo(
     ep_stack_size == cur_stack_size, "Stack must be same size",
     ep_stack_size, cur_stack_size,
-    cur_stack_size > 0 ? epoch_stack_.top() : no_epoch,
-    current_handler_context_, current_epoch_context_, current_node_context_,
-    cur_epoch
+    cur_stack_size > 0 ? epoch_stack_.top() : no_epoch
```
PhilMiller, line 484: Can't we delete `epochPreludeHandler` and `epochEpilogHandler` now?

src/vt/messaging/active.h
```diff
@@ -78,6 +78,10 @@
 
 namespace vt {
 
+namespace runnable {
+struct RunnableNew;
+}
+
```
PhilMiller, line 84: I don't see this referenced in the header or the impl.h - move down to the cc file?

src/vt/pipe/callback/handler_send/callback_send.impl.h
```diff
@@ -89,7 +92,12 @@ CallbackSend<MsgT>::triggerDispatch(SignalDataType* data, PipeType const& pid) {
     this_node, send_node_
   );
   if (this_node == send_node_) {
-    runnable::RunnableVoid::run(handler_,this_node);
+    auto r = std::make_unique<runnable::RunnableNew>(true);
+    r->template addContext<ctx::TD>(theMsg()->getEpoch());
+    r->template addContext<ctx::FromNode>(this_node);
+    r->template addContext<ctx::SetContext>(r.get());
+    r->setupHandler(RunnableEnum::Void, handler_, this_node);
+    r->run();
```
PhilMiller, line 100: Not exactly entailed by the present issue, but should this be calling `run` inline, rather than `theSched()->enqueue`? There may be re-entrancy issues relative to when the caller would expect a 'sent message' to run, especially if the send ends up being an object sending a message to itself.

PhilMiller: Or, rather, shouldn't it call into whatever call path would handle a just-received message, and let that do the context setup and such?

src/vt/runnable/runnable.cc
```diff
@@ -0,0 +1,190 @@
...
+  } else {
+    vtAbort("Invalid runnable type---should be unreachable");
+  }
+}
+
+void RunnableNew::setupHandlerElement(
+  vrt::collection::UntypedCollection* elm, RunnableEnum run_type,
+  HandlerType handler, NodeType from_node
+) {
+  if (run_type == RunnableEnum::Collection) {
```
PhilMiller, line 137: Could this check and abort be a `vtAssert()` instead?

src/vt/runnable/runnable.cc
```diff
@@ -0,0 +1,190 @@
...
+    bool is_functor = HandlerManagerType::isHandlerFunctor(handler);
+
+    ActiveFnPtrType func = nullptr;
+
+    if (is_auto && is_functor) {
+      func = auto_registry::getAutoHandlerFunctor(handler);
+    } else if (is_auto) {
+      func = auto_registry::getAutoHandler(handler);
+    } else {
+      vtAssert(false, "Must be auto handler");
```
PhilMiller, line 123: Instead of `else`, maybe `vtAssert(func != nullptr)`?

PhilMiller: And regardless,  `vtAssert(false, ...)` should just be `vtAbort()`

src/vt/runnable/runnable.cc
```diff
@@ -0,0 +1,190 @@
...
+#include "vt/runnable/runnable.h"
+#include "vt/objgroup/manager.h"
+#include "vt/messaging/envelope.h"
+#include "vt/vrt/context/context_vrt.h"
+#include "vt/vrt/collection/types/untyped.h"
+
+namespace vt { namespace runnable {
+
+RunnableNew::~RunnableNew() {
+  consumeEpochMsg();
```
PhilMiller, line 54: Could this call be moved to `end()` instead?

More generally, should the produce/consume pair here instead be handled by an `enqueue()` method implemented on `ctx::TD` (as a new `virtual void Base::enqueue()`), with `RunnableNew::enqueue` being called by the scheduler when asked to enqueue one of these?

I can also see potential performance analysis uses for such a method, to record a time, and then see how long the work spent in the queue.

src/vt/runnable/runnable.h
```diff
@@ -0,0 +1,251 @@
...
+public:
+  /**
+   * \brief Begin all contexts
+   */
+  void begin();
+
+  /**
+   * \brief End all contexts
+   */
+  void end();
```
PhilMiller, line 202: I don't think `begin()` and `end()` need to be `public`, since they're being called by `run()`

src/vt/runnable/runnable.cc
PhilMiller: With that change, I think any concern about TD and epochs here in `RunnableNew` could just go away. The `produce` call in `TD::begin` would move to `TD::enqueue`, and then nothing would be necessary here.

PhilMiller: The epoch determination logic probably belongs in `ctx::TD` as well.

src/vt/scheduler/priority_queue.h
```diff
@@ -62,9 +62,12 @@ struct PriorityQueue {
 
   void emplace(T&& elm) { impl_.emplace(std::forward<T>(elm)); }
 
-  T pop() { auto elm = impl_.top(); impl_.pop(); return elm; }
-
-  T const& top() { return impl_.top(); }
+  T pop() {
+    // pull out the BIG guns:
+    T elm = std::move(const_cast<T&>(impl_.top()));
```
PhilMiller, line 67: I'd had some sense that the interface defined by `std::priority_queue` was not great, but this really drives it home.

src/vt/runnable/runnable.cc
PhilMiller: Or, instead of an `enqueue()` method, should the `produce()` call in `ctx::TD` just be in its constructor instead of its `begin()` method?

src/vt/serialization/messaging/serialized_messenger.impl.h
```diff
@@ -405,7 +434,17 @@ template <typename MsgT, typename BaseT>
...
+          r->template addContext<ctx::TD>(theMsg()->getEpoch());
+          r->template addContext<ctx::Trace>(
+            msg, typed_handler, node,
+            auto_registry::RegistryTypeEnum::RegGeneral
+          );
+          r->template addContext<ctx::FromNode>(node);
+          r->template addContext<ctx::SetContext>(r.get());
+
+          r->setupHandler(RunnableEnum::Active, typed_handler, node);
+          theSched()->enqueue(msg, std::move(r));
```
PhilMiller, line 447: Similar to my thought about callbacks, could this just be a call to the method that would be called if this message were just received?

src/vt/context/runnable_context/set_context.h
```diff
@@ -0,0 +1,107 @@
...
+#include "vt/context/runnable_context/base.h"
+#include "vt/runnable/runnable.fwd.h"
+
+namespace vt { namespace ctx {
+
+/**
+ * \struct SetContext
+ *
+ * \brief Set the context of the current running task for query by other
+ * components or users.
```
PhilMiller, line 57: I have a rough sense that this class's behavior should probably be integral to `RunnableNew` itself, or perhaps it should always add it itself, rather than callers doing so.

PhilMiller:
I really like how this is coming along. We should talk on Tuesday about some of the more involved changes this makes.

src/vt/context/runnable_context/collection.impl.h
lifflander: It's a historical misnomer. Before it was just used for insertions, but then I repurposed it more generally for the request Matt made to be able to access the current index/proxy. I will rename `InsertContextHolder` appropriately.

src/vt/context/runnable_context/lb_stats.impl.h
lifflander: I really like that idea. I will rework it.

src/vt/messaging/active.cc
lifflander: Agreed, will do

src/vt/messaging/active.h
lifflander: Ah, yes that's a remnant of the past

src/vt/messaging/active.impl.h
lifflander: Yes, good point

src/vt/pipe/callback/handler_send/callback_send.impl.h
lifflander: Yes, I think I chose the wrong path on this one. It should enqueue itself for the callback path.

src/vt/runnable/runnable.cc
lifflander: I think I can remove the `RunnableEnum` with the refactoring I did. All collections runnable should go through this overload for `setupHandlerElement`

src/vt/scheduler/priority_queue.h
lifflander: Unfortunately, `std::priority_queue` is very limited and does not support move-only types. This is a "safe" workaround given that I've encapsulated it here. I think C++17 or 20 fixes this.

src/vt/serialization/messaging/serialized_messenger.impl.h
lifflander: Yes, we probably could abstract that out into a callsite. It's not the quite same as an active messenger message coming in, due to how the handlers are passed

src/vt/scheduler/thread_action.h
lifflander: Resolved with TD

src/vt/context/context.cc
lifflander: Removed

src/vt/context/runnable_context/collection.h
lifflander: Done, marked the methods and those classes final

src/vt/context/runnable_context/base.h
lifflander: Fixed

src/vt/context/runnable_context/collection.impl.h
lifflander: Renamed

src/vt/context/runnable_context/lb_stats.impl.h
lifflander: Ok, I've done this refactoring. I was actually able to completely de-template `LBStats` also with this change.

src/vt/messaging/active.impl.h
lifflander: Removed

src/vt/messaging/active.h
lifflander: Removed

src/vt/pipe/callback/handler_send/callback_send.impl.h
lifflander: Fixed

src/vt/context/runnable_context/set_context.h
lifflander: Initially, I was thinking that was the better design. However, I really like the encapsulation this provides. Otherwise, we would have to implement all this behavior in `RunnableNew`, including tracking suspensions and the previous task that was set (along with restoring it back). I think having it as part of the context and having the logic isolated here is actually more natural.

So, for now, I'm going to leave it as is and maybe we can discuss it later when we delve into the broader impacts this has on VT.

src/vt/runnable/runnable.h
lifflander: Fixed

src/vt/runnable/runnable.cc
lifflander: Completely removed RunnableEnum which cleaned up the code quite a bit!

src/vt/runnable/runnable.cc
lifflander: Done, make it a `vtAbort`

src/vt/messaging/active.cc
lifflander: Done!

src/vt/runnable/runnable.cc
lifflander: Done, cleaned this up a ton!

src/vt/context/context.cc
```diff
@@ -91,6 +96,30 @@ void Context::setDefaultWorker() {
...
+      return from->get();
+    }
+  }
+  return getNode();
+}
+
+#if vt_check_enabled(trace_enabled)
+trace::TraceEventIDType Context::getTraceEventCurrentTask() const {
+  if (getTask() != nullptr) {
+    return theContext()->getTask()->get<ctx::Trace>()->getEvent();
```
PhilMiller, line 116: Are we going to uphold an invariant that every task will be traced? I could easily see wanting to exclude some unimportant things from traces entirely, to limit overhead and trace size.

I think this should be like `FromNode`, where it checks whether there's a `Trace` context or not.

src/vt/context/runnable_context/base.h
```diff
@@ -0,0 +1,102 @@
...
+  virtual void resume() {}
+
+  /**
+   * \brief Invoked when a message is sent to any node
+   *
+   * \param[in] dest the destination of the message
+   * \param[in] size the size of the message
+   * \param[in] bcast whether the message is being broadcast or sent
+   */
+  virtual void send(NodeType dest, MsgSizeType size, bool bcast) { }
```
PhilMiller, line 97: Rather than passing the message size, would it make sense to pass the message pointer itself, so that a more sophisticated implementation could extract any info it wants for itself?

lifflander: You can't template this if it's virtual. If it's virtual, you can only pass `MsgSharedPtr<BaseMsgType>` so you won't have the actual message type and `sizeof(MsgT)` isn't valid.

lifflander: And if the message is serializable, you can't serialize without the actual type.

src/vt/context/runnable_context/lb_stats.cc
```diff
@@ -0,0 +1,85 @@
...
+#include "vt/vrt/collection/manager.h"
+
+namespace vt { namespace ctx {
+
+void LBStats::begin() {
+  // Save current element ID context
+  prev_elm_id_ = theCollection()->getCurrentContext();
+
+  // Set the current element context for communication stats
+  theCollection()->setCurrentContext(cur_elm_id_);
```
PhilMiller, line 55: It would be additional refactoring, but can we just do away with that entirely, and ask for this?

src/vt/context/runnable_context/collection.impl.h
```diff
@@ -0,0 +1,84 @@
...
+template <typename ColT>
+/*explicit*/ Collection<IndexT>::Collection(
+  vrt::collection::CollectionBase<ColT, IndexT>* elm
+) : idx_(elm->getIndex()),
+    proxy_(elm->getProxy())
+{ }
+
+template <typename IndexT>
+void Collection<IndexT>::begin() {
+  vrt::collection::CollectionContextHolder<IndexT>::set(&idx_, proxy_);
```
PhilMiller, line 64: Would it be feasible to just keep things right here, and eliminate that entirely?

src/vt/context/runnable_context/set_context.cc
```diff
@@ -0,0 +1,72 @@
...
+#include "vt/context/runnable_context/set_context.h"
+#include "vt/context/context_attorney.h"
+
+namespace vt { namespace ctx {
+
+void SetContext::begin() {
+  // we have to handle the ugly handler-inside-handler case.. preserve the
+  // previous context (pop) and set the new task (push)
+  nonowning_prev_task_ = theContext()->getTask();
+  ContextAttorney::setTask(nonowning_cur_task_);
```
PhilMiller, line 54: Could we implement and use a simple `observer_ptr` wrapper, and not stick that information in the member name? I almost got confused below by the similarity between `nonowning_cur_task_` and `nonowning_prev_task_`

src/vt/context/runnable_context/td.h
```diff
@@ -0,0 +1,118 @@
...
+  /**
+   * \brief Construct with a message to extract the epoch
+   *
+   * \param[in] msg the message to extract the epoch from
+   */
+  template <typename MsgPtrT>
+  explicit TD(MsgPtrT msg);
+
+  /**
+   * \brief During begin \c TD will produce on the epoch and push it on the
```
PhilMiller, line 85: The `produce` call is happening in the constructor now, not `begin()`

src/vt/context/runnable_context/td.h
```diff
@@ -0,0 +1,118 @@
...
+   * epoch stack.
+   */
+  void begin() final override;
+
+  /**
+   * \brief During end \c TD will consume on the epoch and pop it off the stack
```
PhilMiller, line 91: Similarly, `consume` is in the dtor

src/vt/context/runnable_context/trace.h
```diff
@@ -0,0 +1,149 @@
...
+  /// The open processing tag
+  trace::TraceProcessingTag processing_tag_;
+};
+
+#else
+
+struct Trace : Base {
+
+  template <typename... Args>
+  Trace(Args... args) {}
```
PhilMiller, line 139: Want to take these by const or universal reference, so we're not copying `MsgT`?

src/vt/context/context.cc
lifflander: FIxed

src/vt/context/runnable_context/base.h
PhilMiller: Right, I forgot about those constraints.

PhilMiller: Would it make sense to move the calls to get this info to the interface to `MsgSharedPtr`, rather than having callers pass them in here?

src/vt/context/runnable_context/td.h
lifflander: Fixed

src/vt/context/runnable_context/td.h
lifflander: Fixed

src/vt/context/runnable_context/trace.h
lifflander: Done

src/vt/runnable/runnable.cc
```diff
@@ -0,0 +1,156 @@
...
+#include "vt/runnable/runnable.h"
+#include "vt/objgroup/manager.h"
+#include "vt/messaging/envelope.h"
+#include "vt/vrt/context/context_vrt.h"
+#include "vt/vrt/collection/types/untyped.h"
+
+namespace vt { namespace runnable {
+
+void RunnableNew::setupHandler(
+  HandlerType handler, NodeType from_node, bool is_void, TagType tag
```
PhilMiller, line 54: The `from_node` argument is never being used, in any of these methods. Is it meant to be?

src/vt/runnable/runnable.cc
```diff
@@ -0,0 +1,156 @@
...
+
+void RunnableNew::setupHandlerElement(
+  vrt::VirtualContext* elm, HandlerType handler, NodeType from_node
+) {
+  auto const func = auto_registry::getAutoHandlerVC(handler);
+  task_ = [=]{ func(msg_.get(), elm); };
+}
+
+void RunnableNew::run() {
+  begin();
```
PhilMiller, line 127: My remark about `RunnableNew` handling the context pointer itself was suggesting adding a simple call here:
`addContext<SetContext>(this)`

PhilMiller: Before `begin()`, obviously.

src/vt/serialization/messaging/serialized_messenger.impl.h
```diff
@@ -133,14 +147,21 @@ template <typename UserMsgT>
...
+        r->template addContext<ctx::TD>(epoch);
       }
-      runnable::Runnable<UserMsgT>::run(handler, nullptr, msg.get(), node);
-      action();
+      r->template addContext<ctx::Trace>(
+        msg, handler, node, auto_registry::RegistryTypeEnum::RegGeneral
+      );
+      r->template addContext<ctx::FromNode>(node);
+      r->template addContext<ctx::SetContext>(r.get());
+      r->template addContext<ctx::Continuation>(action);
```
PhilMiller, line 159: Would it make sense to `std::move(action)` here? Just in case it's captured something heavy?

src/vt/serialization/messaging/serialized_messenger.impl.h
```diff
@@ -133,14 +147,21 @@ template <typename UserMsgT>
         handler, recv_tag, envelopeGetEpoch(msg->env)
       );
 
+      auto r = std::make_unique<runnable::RunnableNew>(msg, true);
       if (is_valid_epoch) {
```
PhilMiller, line 151: It would be nice if we could create and capture the `ctx::TD` or the entire `RunnableNew` outside this lambda, and just do away with the `is_valid_epoch` and associated `produce/consume` calls.

src/vt/runnable/runnable.cc
lifflander: Good point, I think that's a hold over from when that method handled trace stuff. I will remove it!

src/vt/context/runnable_context/set_context.cc
lifflander: Done, check out the implementation let me know what you think: https://github.com/DARMA-tasking/vt/pull/1322/commits/356a0090fc5596a574506cdc58c9e8f9e9231468

src/vt/runnable/runnable.cc
lifflander: Good point.. done

src/vt/runnable/runnable.cc
lifflander: Removed

src/vt/context/runnable_context/set_context.h
lifflander: Resolved per later comment

src/vt/messaging/active.cc
```diff
@@ -939,101 +926,48 @@ bool ActiveMessenger::deliverActiveMsg(
...
-
-    #if vt_check_enabled(trace_enabled)
-      current_trace_context_  = envelopeGetTraceEvent(msg->env);
-    #endif
-
-    if (has_epoch) {
-      ep_stack_size = epochPreludeHandler(cur_epoch);
+    auto r = std::make_unique<runnable::RunnableNew>(base, not is_term);
+    if (not is_term) {
+      r->template addContext<ctx::TD>(base);
```
PhilMiller, line 956: Could `ctx::TD` do the `is_term` check itself? It doesn't look like there's any type dependence there.

src/vt/pipe/callback/handler_send/callback_send.impl.h
```diff
@@ -89,7 +93,11 @@ CallbackSend<MsgT>::triggerDispatch(SignalDataType* data, PipeType const& pid) {
     this_node, send_node_
   );
   if (this_node == send_node_) {
-    runnable::RunnableVoid::run(handler_,this_node);
+    auto r = std::make_unique<runnable::RunnableNew>(true);
+    r->template addContext<ctx::TD>(theMsg()->getEpoch());
+    r->template addContext<ctx::FromNode>(this_node);
+    r->setupHandler(handler_, true);
+    theSched()->enqueue(std::move(r));
```
PhilMiller, line 100: This calls to attention that a no-message handler callback won't `Trace` back to the event that triggered it. Would it make sense to add one to the current event?

src/vt/context/runnable_context/lb_stats.cc
lifflander: Yep, working on that right now.

src/vt/context/runnable_context/collection.impl.h
lifflander: Same with this.

src/vt/context/runnable_context/lb_stats.cc
lifflander: Cleaned this up a ton!

src/vt/context/runnable_context/collection.impl.h
lifflander: I was able to refactor the `getCurrentContext`/`setCurrentContext` out for `LBStats`. For this one, there are times when the context is set outside of `RunnableNew`, specifically when the collection is being set up (and a handler is not running). So I don't think I can refactor this out quite yet. We might want to switch some of that setup to happen inside a `RunnableNew`, like the collection insertion and collection constructor. I think I'm going to leave that for later so I can start focusing on more of the threaded stuff.

src/vt/pipe/callback/handler_send/callback_send.impl.h
lifflander: That's a good point. I'm not sure if I understand your solution.

src/vt/messaging/active.cc
lifflander: Actually, it does this check already: https://github.com/DARMA-tasking/vt/blob/fe1b7b4bb7503b5d5611987859924f74570fcd84/src/vt/context/runnable_context/td.impl.h#L63-L83

However the way `TD::begin()` and `TD::end()` are written they would push/pop `no_epoch` on the stack, which would be very confusing. I think it's better to just avoid having anything happen with respect to `ctx::TD` when it's a termination message as it's not valid to create an epoch context when it's a termination message. In essence, a termination message has no epoch and should never change any state related to epochs.

src/vt/serialization/messaging/serialized_messenger.impl.h
lifflander: Outside of this lambda we don't have `msg` because it hasn't arrived from `theMsg()->recvDataDirect(..)` yet to then be deserialized. Thus, we can't really create the runnable outside of it. The issue is that `recvDataDirect` does not know about epochs so we have to produce/consume to make sure TD doesn't happen early. It might be a good idea to have it produce/consume when it's called based on whatever is on the top of the epoch stack when it is called? Not sure if there are edge cases there were it's called outside of a handler though (not in this case, but other use cases). I think I would prefer to leave this as is for now. Also, it's probably possible to remove `is_valid_epoch` (because it should always be valid). We should sanity check the codebase that it's impossible to send a message with `no_epoch` because at the very least messages should have `term::any_epoch_sentinel` (which we should probably get rid of eventually). Anyway, I'm going to leave these things as an exercise for the future since they are only tangentially related to the matters this PR addresses.

src/vt/context/runnable_context/base.h
lifflander: Are you thinking that when `MsgSharedPtr<T>` is created (before `T` is erased) that we should capture the size for when it is possibly type-erased in the future? I like that idea. Not sure if that should go with this PR

src/vt/serialization/messaging/serialized_messenger.impl.h
lifflander: OK--I think I have a pretty nice solution with `makeRunnable`.

src/vt/runnable/make_runnable.h
```diff
@@ -0,0 +1,319 @@
...
+    }
+    return std::move(*this);
+  }
+
+  /**
+   * \brief Add a termination message to extract an epoch
+   *
+   * \param[in] is_term whether it's a termination message
+   */
+  RunnableMaker&& withTDMsg(bool is_term = false) {
```
PhilMiller, line 126: Is this really `withTDEpochFromMessage`, essentially? While I understand that some messages are or aren't `EpochMsg`, that's not really relevant to the naming of this function.

lifflander: Right, yes that is a more accurate name

src/vt/runnable/runnable.cc
```diff
@@ -0,0 +1,171 @@
...
+  task_ = [=]{ func(msg_.get(), elm); };
+}
+
+void RunnableNew::run() {
+  vt_debug_print(
+    context, node,
+    "start running task={}\n", print_ptr(this)
+  );
+
+  addContext<ctx::SetContext>(this);
```
PhilMiller, line 133: Given the builder now in use, does it make sense to move this call there?

src/vt/runnable/make_runnable.h
```diff
@@ -0,0 +1,319 @@
...
+    impl_->setupHandlerElement(elm, handler_);
+    return std::move(*this);
+  }
+
+  /**
+   * \brief Add a collection; sets up the handler and collection contexts
+   *
+   * \param[in] elm the collection element pointer
+   */
+  template <typename IdxT, typename ElmT>
```
lifflander, line 150: I can actually extract the `IdxT` from `ElmT`. No reason to have that manually specified here.

src/vt/runnable/runnable.h
```diff
@@ -0,0 +1,206 @@
...
+
+  /**
+   * \brief Run the task!
+   *
+   * \note If the runnable is non-threaded, \c isDone() will be \c true after
+   * this runs. If it is threaded, it might suspend and the runnable might have
+   * more work to complete.
+   */
+  void run();
+
```
PhilMiller, line 151: Obvious next step here is that this needs `suspend` and `resume` methods that call the associated methods of the contexts.

src/vt/scheduler/base_unit.cc
```diff
@@ -0,0 +1,65 @@
...
+
+void BaseUnit::execute() {
+  if (r_) {
+    r_->run();
+    if (not r_->isDone()) {
+      //
+      // @todo: do something!
+      //
+      // preserve the lifetime of this runnable so the
+      // allocation doesn't disappear under the stack
```
PhilMiller, line 58: Should we just have a container (`unordered_set`, maybe?) of suspended tasks, from which it can be moved back into the queue when it's resumed?

src/vt/runnable/make_runnable.h
lifflander: Done, fixed

src/vt/context/runnable_context/collection.impl.h
lifflander: reference issue: https://github.com/DARMA-tasking/vt/issues/1339

src/vt/serialization/messaging/serialized_messenger.impl.h
lifflander: refer to issue: https://github.com/DARMA-tasking/vt/issues/1340

src/vt/serialization/messaging/serialized_messenger.impl.h
lifflander: refer to issue: https://github.com/DARMA-tasking/vt/issues/1341

src/vt/messaging/active.cc
lifflander: refer to issue: https://github.com/DARMA-tasking/vt/issues/1341

src/vt/pipe/callback/handler_send/callback_send.impl.h
lifflander: refer to issue: https://github.com/DARMA-tasking/vt/issues/1342

src/vt/runnable/runnable.cc
lifflander: Yes, I was thinking the same. Will do

lifflander: Done

src/vt/scheduler/suspended_units.cc
```diff
@@ -0,0 +1,74 @@
...
+    std::forward_as_tuple(tid),
+    std::forward_as_tuple(
+      detail::SuspendedRunnable{std::move(runnable), p}
+    )
+  );
+}
+
+void SuspendedUnits::resumeRunnable(ThreadIDType tid) {
+  auto iter = units_.find(tid);
+  vtAbortIf(iter == units_.end(), "Must have valid thread ID to suspend");
```
PhilMiller, line 67: To resume, presumably

src/vt/messaging/async_op_wrapper.cc
```diff
@@ -0,0 +1,66 @@
...
+  num_tests++;
+  return is_done;
+}
+
+void AsyncOpWrapper::done() {
+  op_->done();
+  op_ = nullptr;
+
+  if (tid_ != no_thread_id) {
+    theSched()->suspended_.resumeRunnable(tid_);
```
PhilMiller, line 62: Why not add a method on theSched to do this, rather than making the member accessible?

src/vt/messaging/active.cc
```diff
@@ -413,19 +413,16 @@ EventType ActiveMessenger::sendMsgBytes(
     theTerm()->hangDetectSend();
   }
 
-  for (auto&& l : send_listen_) {
-    l->send(dest, msg_size, is_bcast);
+  if (theContext()->getTask() != nullptr) {
+    auto lb = theContext()->getTask()->get<ctx::LBStats>();
```
PhilMiller, line 417: This should probably be a `send` method of context itself, with the loop over context instances, like `begin`, `end`, etc. Otherwise, why is there a `virtual Base::send`?

I could potentially see `Trace` and `TD` wanting to implement the method as well, for instance. `TD` could perhaps automatically call `produce` there, and maybe the `hangDetectSend` call seen above as well..

src/vt/vrt/collection/manager.impl.h
```diff
@@ -374,33 +365,27 @@ CollectionManager::collectionAutoMsgDeliver(
...
 
-  // Clear the current index context
-  IdxContextHolder::clear();
+  runnable::makeRunnable(user_msg, true, han, from, reg)
+    .withTDEpoch(theMsg()->getEpochContextMsg(msg))
+    .withCollection(base)
+    .withTraceIndex(event, idx1, idx2, idx3, idx4)
+#if vt_check_enabled(lblite)
+    .withLBStats(base, msg)
+#endif
```
PhilMiller, line 379: Can this `#if` be inside `withLBStats`, so it doesn't need to be repeated at call sites?

src/vt/context/runnable_context/continuation.h
```diff
@@ -0,0 +1,83 @@
...
+  explicit Continuation(ActionType in_cont)
+    : cont_(in_cont)
+  { }
+
+  /**
+   * \brief After the task runs, invoke the continuation if non-null
+   */
+  void end() final override {
+    if (cont_) {
+      cont_();
```
PhilMiller, line 73: This may need to be `theSched()->enquqeue(cont_)`, so that it doesn't execute in some inconsistent environment created by the other context `end()` methods being called in an unknown order.

PhilMiller:
OK, I think that's pretty much all I'm going to have to say about this set of changes. It would be nice to get the move of `AsyncOpMPI` from the tests to the source in, too.

src/vt/vrt/collection/manager.impl.h
lifflander: Done

src/vt/scheduler/scheduler.h
```diff
@@ -351,6 +364,11 @@ struct Scheduler : runtime::component::Component<Scheduler> {
   template <typename Callable>
   friend void vt::runInEpochCollective(Callable&& fn);
 
+  // For access to suspended_ to move runnables if they don't complete
+  friend struct BaseUnit;
+  // For access to suspended_ to re-enqueue runnables when they complete
+  friend struct messaging::AsyncOpWrapper;
+
```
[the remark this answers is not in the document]
lifflander: Done

src/vt/messaging/async_op_wrapper.cc
lifflander: Done

src/vt/scheduler/suspended_units.cc
lifflander: Fixed

src/vt/messaging/active.cc
lifflander: Good point

lifflander: Done

src/vt/context/runnable_context/continuation.h
lifflander: For now I'm going to leave this as is, because the continuation is typically some cleanup action like a deleter.

src/vt/scheduler/base_unit.cc
lifflander: Done

src/vt/runnable/make_runnable.h
lifflander: Renamed

src/vt/scheduler/thread_action.cc
lifflander: I can't move that there because it has a jump at the end and we must deallocate the stack *afrer* that jump.

src/vt/context/runnable_context/base.h
lifflander: refer to: https://github.com/DARMA-tasking/vt/issues/1344

lifflander:
> OK, I think that's pretty much all I'm going to have to say about this set of changes. It would be nice to get the move of `AsyncOpMPI` from the tests to the source in, too.

Done, added to the main code base.

src/vt/context/runnable_context/continuation.h
PhilMiller: In that case, want to mention in the class comments that any continuation should not depend on other context?

src/vt/scheduler/thread_action.h
```diff
@@ -0,0 +1,165 @@
...
+
+  ~ThreadAction();
+
+  /**
+   * \brief Run the thread and the action
+   */
+  void run();
+
+  /**
+   * \brief Resume the thread were left offn
```
PhilMiller, line 95: Typos
- where
- off

src/vt/scheduler/thread_action.h
```diff
@@ -0,0 +1,165 @@
...
+   */
+  void resume();
+
+  /**
+   * \brief Keep yielding back to thread (resuming after suspension) unit the
```
PhilMiller, line 100: Typo unit -> until

Also,  I think this is only suitable to call for testing purposes? When else would a just-suspended thread want to be resumed immediately?

src/vt/scheduler/thread_action.h
```diff
@@ -0,0 +1,165 @@
...
+   *
+   * \param[in] t the transfer
+   */
+  static void runFnImpl(fcontext_transfer_t t);
+
+private:
+  /// The currently running \c ThreadAction
+  static ThreadAction* cur_running_;
+
+private:
```
PhilMiller, line 152: Any reason for the mixed comment styles and multiple `private:` blocks here?

src/vt/runnable/make_runnable.h
```diff
@@ -0,0 +1,325 @@
...
+namespace vt { namespace runnable {
+
+/**
+ * \struct RunnableMaker
+ *
+ * \brief Convenience builder class for setting up a \c RunnableNew before
+ * enqueuing it or running it.
+ */
+template <typename MsgT>
+struct RunnableMaker {
```
PhilMiller, line 67: As written, this is going to pull in a ton of headers all over the place. Would it make sense to have a `RunnableMakerBase` that's not templated, containing the implementation of most of these?

Alternately, does the whole class actually need to be templated over MsgT, or can that safely be type erased at construction?

src/vt/scheduler/thread_action.h
lifflander: Yes, just for testing.

src/vt/scheduler/thread_action.h
lifflander: Fixed

src/vt/scheduler/thread_action.h
lifflander: Fixed

PhilMiller (approved):

src/vt/runnable/make_runnable.h
lifflander: This looks like a lot of inclusions, but most of the runnable context headers have fairly minimal includes (which I spent time minimizing). `lb_stats.impl.h` is probably the heaviest one because it has to include `elm_stats.h`. To remediate this dependency problem I don't include the `lb_stats.impl.h` in `lb_stats.h`, instead including it in `vrt/collection/manager.h` at the very end. Same with `collection.impl.h`, which has the heaver inclusions.

lifflander: The one reason I can't type-erase `MsgT` is due to `withTraceIndex` which requires the non-type-erased message type to properly trace collection messages. I could have that pass in the message again, instead of holding on to it, but that would require some parameter duplication at those callsites. Do you think it's work it?

@PhilMiller

PhilMiller: If that's the only reason we can't type-erase this, I think the parameter duplication is worthwhile.

src/vt/runtime/runtime_banner.cc
```diff
@@ -762,6 +768,29 @@ void Runtime::printStartupBanner() {
...
+    );
+    auto f13 = fmt::format(
+      "User-level stack size is {} {}",
+      std::get<1>(ret), std::get<0>(ret)
+    );
+    auto f14 = opt_on("--vt_ult_stack_size", f13);
+    fmt::print("{}\t{}{}", vt_pre, f14, reset);
+  } else {
+    auto f11 = fmt::format("Handlers running in user-level threads are disabled");
+    auto f12 = opt_remove("--vt_diag_disable", f11);
```
PhilMiller, line 789: Shouldn't this be `--vt_ult_disable_`?

PhilMiller (changes requested):
See the option message comment

src/vt/runtime/runtime_banner.cc
lifflander: Oops, good catch

lifflander: Fixed

lifflander:
Note that two of the CI targets have `fcontext` enabled so CI will be testing all the threading across one clang build and one GNU build.

PhilMiller (approved):
I'm happy with this. Your call whether you want to do the type erasure on `RunnableMaker` or not.

PR #1306, opened 2021-03-08
1297: Add CMake variable to set max number of ranks/nodes for tests and examples.

JacobDomagala:
Fixes #1297

PR #1329, opened 2021-03-16
1328: Change CMake to also run tests on single node.

### LINKED ISSUE #1328: Run tests on single Node
Currently the minimum amount of nodes used for tests is 2 (on the machine with 4 processors we run on 2/4 nodes, on the machine with 2 processesors, such as our pipelines, we run only on 2 nodes). 
Change CMake to also run them on single node and disable/skip the ones that require at least 2 nodes. 

This is part of https://github.com/DARMA-tasking/vt/issues/1297

JacobDomagala:
Fixes #1328 

1. Add helper macros to skip tests in runtime based on number of nodes
2. Change CMake to also run the tests on single rank
3. Skip the tests that aren't meant to be run on less than 2 nodes


tests/unit/sequencer/test_sequencer_extensive.extended.cc
```diff
@@ -221,6 +222,20 @@ static constexpr CountType const max_seq_depth = 8;
...
+   * empty or test is skipped (using GTEST_SKIP). This is probably due to the
+   * number of test cases generated by this function (672 to be exact).
+   *
+   * So in this case we want to disable the parametrized generation when
+   * we oversubscribe
+  */
+  if(isOversubscribed()){
+    return ParamContainerType{};
+  }
+
```
JacobDomagala, line 238: This one is related to https://github.com/DARMA-tasking/vt/pull/1306

tests/CMakeLists.txt
```diff
@@ -99,13 +99,15 @@ function(add_unit_test unit_test_name unit_test_files uses_mpi)
...
         EXECUTE_COMMAND        ${MPI_RUN_COMMAND} ${MPI_EXTRA_FLAGS} ${MPI_NUMPROC_FLAG} ${PROC}
+        EXTRA_ARGS             --MPI_TEST
       )
 
       set_tests_properties(
         ${${CUR_TEST_LIST}}
         PROPERTIES TIMEOUT 60
         FAIL_REGULAR_EXPRESSION "FAILED;should be deleted but never is;Segmentation fault"
         PASS_REGULAR_EXPRESSION "PASSED"
+        SKIP_REGULAR_EXPRESSION "SKIPPED"
```
JacobDomagala, line 110: Skipped tests are displayed in the run summary like so:
image


I will test whether it interferes with our 'comment-pr' action when test fails.

JacobDomagala: Looks like it doesn't break anything:</br>
image

tests/unit/main.cc
```diff
@@ -57,10 +57,36 @@ std::unique_ptr<MPISingletonMultiTest> mpi_singleton = nullptr;
...
+    "Last argument should be either --MPI_TEST or --NO_MPI_TEST"
+  );
+
+  // 'Remove' the argument so we don't pass it further to vt
+  --argc;
+
+  if (test_arg == "--MPI_TEST") {
+    mpi_singleton = std::make_unique<MPISingletonMultiTest>(argc, argv);
+  }
+}
```
JacobDomagala, line 78: While this works for CI and running tests via `/vt/ci/test_cpp.sh` it could be problematic when running single test by hand. I should probably change that

lifflander (approved):
Looks good and ready to me!

JacobDomagala:
After rebasing the branch with the latest changes from develop, some tests started to fail while running on the single node. 
I'm looking into that

src/vt/pipe/callback/handler_send/callback_send_tl.impl.h
```diff
@@ -73,7 +73,7 @@ void CallbackSendTypeless::trigger(MsgT* msg, PipeType const& pipe) {
   auto pmsg = promoteMsg(msg);
   if (this_node == send_node_) {
     runnable::makeRunnable(pmsg, true, handler_, this_node)
-      .withTDEpochFromMsg()
+      .withTDEpoch(theMsg()->getEpoch())
       .enqueue();
   } else {
     theMsg()->sendMsg<MsgT>(send_node_, handler_, pmsg);
```
JacobDomagala, line 79: While [this](https://github.com/DARMA-tasking/vt/pull/1329/commits/cf18bb0345bd8aabd4284e342c0ea10093f1389c) fixes the issue with failing tests on single node, I think we should change `withTDEpochFromMsg` to fallback to current epoch on the stack (i.e `theMsg()->getEpoch()`) instead of `term::any_epoch_sentinel`

image

JacobDomagala: And just to explain the issue, if the Runnable is added to the scheduler queue using `withTDEpochFromMsg` and its message is not using any epoch, it means it won't produce work for rooted/collective epoch. </br>
So for example: 
image
Here the `cb.send` would call `runnable::makeRunnable` which would then add the Runnable to the sheduler's queue, but since it didn't produce any work for rooted epoch, the epoch would finish right away.

src/vt/pipe/callback/handler_send/callback_send.impl.h
```diff
@@ -112,7 +112,7 @@ CallbackSend<MsgT>::triggerDispatch(SignalDataType* data, PipeType const& pid) {
     auto msg = reinterpret_cast<ShortMessage*>(data);
     auto m = promoteMsg(msg);
     runnable::makeRunnable(m, true, handler_, this_node)
-      .withTDEpochFromMsg()
+      .withTDEpoch(theMsg()->getEpoch())
```
PhilMiller, line 115: As discussed on today's call, the message should get an epoch set on it, and this code should go back to calling `withTDEpochFromMsg()`

The same probably applies to the other callback and reduce cases

lifflander (approved):
Looks good to me. The new changes for callbacks look correct.

Issue #276, opened 2019-02-19
Automatically synthesize message types for sends to arbitrary handlers


PhilMiller:
Following on #272, I think it would not be all that difficult to write template code that would let applications make a `send` call with arguments matching those of some handler, and the template code would take care of describing a suitable message type and actual target handler to unwrap the message and make the call.

The pieces I see playing into this are roughly as follows:

* A method (probably easy enough to write with [`std::forward_as_tuple`](https://en.cppreference.com/w/cpp/utility/tuple/forward_as_tuple)
`template <void (*Func)(Args...), typename Args...> sendMagic(Args&& ...args)`
* A very generic message type 
`template <typename Args...> struct MagicMsg { std::tuple<...Args> a;};`
* A very generic handler function to call the actual handler via [`std::apply`](https://en.cppreference.com/w/cpp/utility/apply)
`template <void (*Func)(Args...), typename Args...> receiveMagic(MagicMsg<...Args>* m)`

Once the simplest cases are working, we should probably apply a type trait to each of the argument types to represent how they should be represented in the message / on the wire, in cases where they shouldn't be literal. In particular, automatic handling of anything array-like on the sending side coming out to whatever array-like thing the receiver expects would be very convenient.

PhilMiller:
More than a year later, Nvidia is still the laggard in implementing P0127 support. I'm going to de-assign this, just to take it off the radar.

lifflander:
@PhilMiller Sounds like a plan. Should be have a label for back burner type tasks? Is pri:low enough to signify that?

PhilMiller:
Yeah, I think so.

PhilMiller:
And, as a note, Nvidia's compiler generally says that it doesn't implement partial standards, so we won't have any C++17 support from them until they offer complete C++17 support. We should maybe see if this code would work there in spite of their lacking support, possibly by it not parsing code that it's not going to need to generate anything for itself.

PhilMiller:
As of nvcc 11.0, C++17 support is available.

PR #1330, opened 2021-03-16
1260: Fix active messaging functor inference using the FunctorExtractor

### LINKED ISSUE #1260: Active messaging functor inference using the FunctorExtractor is broken?
**Describe the bug**
Sending a message to a functor with the `ActiveMessenger` requires the message type, which should be inferred. This might be related to changes to the message thief.

Write a test that replicates this problem and solve it.

JacobDomagala:
Fixes #1260

### DIFF
--- a/examples/hello_world/hello_world_functor.cc
+++ b/examples/hello_world/hello_world_functor.cc
@@ -52,9 +52,21 @@ struct HelloMsg : vt::Message {
   { }
 };
 
+struct AnotherMsg : vt::Message{};
+
 struct HelloWorld {
   void operator()(HelloMsg* msg) const {
-    fmt::print("{}: Hello from node {}\n", vt::theContext()->getNode(), msg->from);
+    fmt::print("{}: HelloWorld -> Hello from node {}\n", vt::theContext()->getNode(), msg->from);
+  }
+};
+
+struct MultipleFunctions {
+  void operator()(HelloMsg* msg) const {
+    fmt::print("{}: MultipleFunctions -> Hello from node {}\n", vt::theContext()->getNode(), msg->from);
+  }
+
+  void operator()(AnotherMsg* msg) const {
+    fmt::print("{}: MultipleFunctions with AnotherMsg\n", vt::theContext()->getNode());
   }
 };
 
@@ -70,7 +82,21 @@ int main(int argc, char** argv) {
 
   if (this_node == 0) {
     auto msg = vt::makeMessage<HelloMsg>(this_node);
+
+    // 'HelloWorld' functor has only single 'operator()' declared
+    // so we can call send/braodcast without specyfing message type
     vt::theMsg()->broadcastMsg<HelloWorld>(msg);
+
+    msg = vt::makeMessage<HelloMsg>(this_node);
+    vt::theMsg()->sendMsg<HelloWorld>(1, msg);
+
+    // 'MultipleFunctions' functor declares more than one 'operator()'
+    // so we have to specify the type of the message, as it can't be deduced
+    auto new_msg = vt::makeMessage<AnotherMsg>();
+    vt::theMsg()->broadcastMsg<MultipleFunctions, AnotherMsg>(new_msg);
+
+    new_msg = vt::makeMessage<AnotherMsg>();
+    vt::theMsg()->sendMsg<MultipleFunctions, AnotherMsg>(1, new_msg);
   }
 
   vt::finalize();
--- a/examples/hello_world/hello_world_functor.cc
+++ b/examples/hello_world/hello_world_functor.cc
@@ -70,7 +70,7 @@ int main(int argc, char** argv) {
 
   if (this_node == 0) {
     auto msg = vt::makeMessage<HelloMsg>(this_node);
-    vt::theMsg()->broadcastMsg<HelloWorld,HelloMsg>(msg);
+    vt::theMsg()->broadcastMsg<HelloWorld>(msg);
   }
 
   vt::finalize();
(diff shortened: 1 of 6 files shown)


tutorial/tutorial_1b.h
```diff
@@ -154,7 +154,7 @@ static void msgHandlerA(MyMsg* msg) {
   NodeType const to_node = 0;
   auto msg2 = ::vt::makeMessage<MyMsg>(10,20);
 
-  ::vt::theMsg()->sendMsg<MsgHandlerB, MyMsg>(to_node, msg2);
+  ::vt::theMsg()->sendMsg<MsgHandlerB>(to_node, msg2);
```
JacobDomagala, line 157: This reproduces the issue

JacobDomagala: Also as a more isolated example https://cppinsights.io/s/0a6836aa.

src/vt/messaging/active.h
```diff
@@ -904,7 +904,7 @@ struct ActiveMessenger : runtime::component::PollableComponent<ActiveMessenger>
     typename MsgT = typename util::FunctorExtractor<FunctorT>::MessageType
   >
   PendingSendType broadcastMsg(
-    MsgPtrThief<MsgT> msg,
+    MsgPtrThief<typename util::FunctorExtractor<FunctorT>::MessageType> msg,
```
JacobDomagala, line 907: This ugly fix makes MsgT not needed anymore. For vt's sake we could remove it from the interface but I can assume either EMPIRE or NimbleSM could be using send/braodcast with both explicit template arguments.

CMakeLists.txt
```diff
@@ -101,9 +101,9 @@ option(vt_werror_enabled "Build VT with -Werror enabled" OFF)
 if (vt_werror_enabled)
   enable_cxx_compiler_flag_if_supported("-Werror")
 endif()
-include(cmake/load_bundled_libraries.cmake)
 
 include(cmake/nvcc_no_deprecated_gpu_targets.cmake)
+include(cmake/load_bundled_libraries.cmake)
```
jstrzebonski, line 106: Nice

src/vt/messaging/active.h
PhilMiller: Considering this in isolation, I think it might be worthwhile to take that breaking change. It actually partially addresses the goal of #276, of not having to type the message type at every send call site.

PhilMiller: Removing it from the interface may not be worth much, though, since it won't actually change the number of template instantiations at all.

PhilMiller:
I *really* like what this does. I want to request an extension to cover updating the codebase as broadly as possible, but that's asking for merge conflicts. Let's get this integrated, and then clean things up across the board in a follow-on PR.

src/vt/messaging/active.h
PhilMiller: It may make sense in all of these methods to have something along the lines of 

`static_assert(std::is_same<MsgT, typename util::FunctorExtractor<FunctorT>::MessageType>::value, "Message type must match type of handler argument");`

so that mixups get caught

PhilMiller:
Could the same thing be done to the interfaces on objgroup and collection proxies?

JacobDomagala:
This PR is merely just a bugfix for Active send/broadcast when used with Functor handler. For any arbitrary handler we're limited by C++14's lack of using the object as template argument, without specifying its type first. 

If we really would want to get rid of explicit message type, I think the best we can do currently is:
```c++
template <typename HandlerType, HandlerType handler>
send(MsgPtrThief<typename FunctionExtractor<HandlerType >::MsgType >msg, ...)
```

Then we could provide helper macro like so:
```c++ 
#define VT_HANDLER(handler) decltype(&handler), &handler
```

which would make the call to `send` look like this:
```c++
send<VT_HANDLER(myHandler)>(my_msg);
```

Example: https://godbolt.org/z/fK6xvjWa8

lifflander (approved):
Look good

JacobDomagala:
I just realized that this change enforces the user to have a Functor with only single `operator()` defined. Otherwise it won't compile. I think I will have to update this PR

examples/hello_world/hello_world_functor.cc
```diff
@@ -70,7 +82,21 @@ int main(int argc, char** argv) {
 
   if (this_node == 0) {
     auto msg = vt::makeMessage<HelloMsg>(this_node);
-    vt::theMsg()->broadcastMsg<HelloWorld,HelloMsg>(msg);
+
+    // 'HelloWorld' functor has only single 'operator()' declared
+    // so we can call send/braodcast without specyfing message type
```
PhilMiller, line 87: Typo here 'specifying'

PhilMiller: And 'broadcast'

examples/hello_world/hello_world_functor.cc
```diff
@@ -70,7 +82,21 @@ int main(int argc, char** argv) {
...
+    msg = vt::makeMessage<HelloMsg>(this_node);
+    vt::theMsg()->sendMsg<HelloWorld>(1, msg);
+
+    // 'MultipleFunctions' functor declares more than one 'operator()'
+    // so we have to specify the type of the message, as it can't be deduced
+    auto new_msg = vt::makeMessage<AnotherMsg>();
+    vt::theMsg()->broadcastMsg<MultipleFunctions, AnotherMsg>(new_msg);
+
+    new_msg = vt::makeMessage<AnotherMsg>();
+    vt::theMsg()->sendMsg<MultipleFunctions, AnotherMsg>(1, new_msg);
```
PhilMiller, line 99: Is one of these intended to send `HelloMsg` to illustrate use of both overloads of `MultupleFunctions::operator()`?

src/vt/utils/static_checks/functor.h
```diff
@@ -47,6 +47,7 @@
 
 #include "vt/config.h"
 
+
```
PhilMiller, line 50: Stray newline added here

examples/hello_world/hello_world_functor.cc
JacobDomagala: Wasn't intended, but I think it's actually good idea to add it.

PhilMiller (approved):

PR #1335, opened 2021-03-23
1333: CI: Generate Build Stats wiki page

### LINKED ISSUE #1333: Change the way we graph the build times
Current approach of using https://github.com/DARMA-tasking/vt/actions/workflows/pushdockerimage.yml as an indicator of VT's build time has proven to be unreliable. 

For example, if we take a look at one of the latest runs (https://github.com/DARMA-tasking/vt/actions/runs/665352181)
The actual runtime for the Job was 51min (this includes build, tests and push to dockerhub)
image

</br>
But for some reason, for the next 10 mins this workflow was still running, and so the final time for this workflow was 1 hour.

image

**What Needs to be Done?**

Ideally we would like to measure only the VT's build time (without tests and push to dockerhub), so we should rewrite `pushdockerimage.yml` action to export the execution time of `/vt/ci/build_cpp.sh /vt /build` and then use that value to generate build time badge and use it in our graph.

### LINKED ISSUE #1352: Create separate workflow for VT diagnostics
TODO:
- create separate docker file with clang9+ that will use https://github.com/aras-p/ClangBuildAnalyzer 
- Move graph generation to this workflow
- Post the results to new wiki page (vt_build_diagnostics) (?)

JacobDomagala:
Fixes #1333 Fixes #1352


.github/workflows/pushdockerimage_test.yml
```diff
@@ -0,0 +1,58 @@
+name: Docker Image CI and graph
```
JacobDomagala, line 1: File and workflow name changed temporarily so it doesn't interfere with our current implementation

lifflander (approved):
This is looking good!

JacobDomagala:
With changes from [updated github action](https://github.com/DARMA-tasking/graph-build-times/pull/8) now following is generated:
- Updated the build graph,now it also includes vt-lib and tests&examples as separate graphs (alongside the total build time)

- Most expensive template instantiations (click on image to see original size)
[](https://github.com/DARMA-tasking/vt/wiki/build_info.png)

- The [Full ClangBuildAnalyzer raport](https://github.com/DARMA-tasking/vt/wiki/build_stats/build_result.txt) is also stored in our wiki page so if we want we can parse it and generate additional data

lifflander (approved):

PR #1346, opened 2021-03-25
1345: context: register stack with valgrind

### LINKED ISSUE #1345: Register ULT stacks with valgrind if headers are present while compiling
**What Needs to be Done?**

Call `VALGRIND_STACK_REGISTER`, etc.

lifflander:
Fixes #1345

### DIFF
--- a/lib/context/check_valgrind.cmake
+++ b/lib/context/check_valgrind.cmake
@@ -1,0 +1,42 @@
+
+function(check_compile VAR CC_FILE)
+  if(NOT DEFINED ${VAR})
+    set(check_cc_result FALSE)
+    set(check_cc_output "")
+
+    try_compile(
+      check_cc_result
+      ${CMAKE_CURRENT_BINARY_DIR}
+      ${CMAKE_CURRENT_SOURCE_DIR}/${CC_FILE}
+      OUTPUT_VARIABLE check_cc_output
+      CXX_STANDARD ${CMAKE_CXX_STANDARD}
+      CXX_STANDARD_REQUIRED ON
+      CXX_EXTENSIONS OFF
+    )
+
+    ## message(STATUS ${check_cc_output})
+
+    mark_as_advanced(${VAR})
+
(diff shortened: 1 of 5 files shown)


PhilMiller:
Code looks fine. Confirm that at least one of the CI builds has valgrind installed with the headers, and so is testing this code?

PhilMiller:
And maybe add some separate test to the script that runs something simple under valgrind.

For bonus points, write a "must fail" test to run under valgrind and affirm that it detects an error.

lifflander:
> And maybe add some separate test to the script that runs something simple under valgrind.
> 
> For bonus points, write a "must fail" test to run under valgrind and affirm that it detects an error.

I could even make the macro call to make sure it exists in the same form?

lifflander:
@cz4rs @PhilMiller @JacobDomagala @jstrzebonski Could I get a review on this PR?

PhilMiller (approved):
It's still not clear that this is getting tested, but the code looks exactly as expected.

lib/context/check_valgrind.cmake
```diff
@@ -0,0 +1,42 @@
...
+  valgrind_stack_register.cmake.cc
+)
+
+if (context_valgrind_stack_compiled)
+  message(
+    STATUS "Successfully compiled with valgrind stack registration"
+  )
+else()
+  message(
+    STATUS "Failed to compile with valgrind stack registration"
```
JacobDomagala, line 40: Should we compile vt even if this fails? I think we could use `FATAL_ERROR` instead of `STATUS` here

JacobDomagala (approved):

lib/context/check_valgrind.cmake
PhilMiller: We definitely don't want to require that valgrind be installed to build vt. It's purely a debugging tool

lifflander: Agreed

Issue #1265, opened 2021-02-16
Implement general test driver for LB testing


lifflander:
**What Needs to be Done?**

Read in stats and recreate collection with input LB stats to test/debug LB strategies

PhilMiller:
Once #1583 is merged, we'll have readily usable functions to form a load model from a set of load data, compute statistics over it, run LB strategies over it, and compute statistics over the results, all without actually instantiating or migrating any real objects.

PR #1348, opened 2021-03-26
1279 gossiplb inform and decide have bugs

### LINKED ISSUE #1279: GossipLB inform and decide have bugs
**Describe the bug**
Imbalances produced by GossipLB are inferior to those predicted by LBAF due to bugs in both the inform and decide steps.

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

nlslatt:
Includes many bug fixes and improvements to GossipLB.  The following LB args were added:

- `trials` (default `3`): how many times to repeat the requested number of iterations, hoping to find a better imbalance; helps if it’s easy to get stuck in a local minimum
- `deterministic` (default false): for debugging purposes, make the migration decision deterministic assuming deterministic loads (for testing on deterministic loads, consider using the driver in development under #1265)
- `inform` (default SyncInform): choice of gossiping approach
  - SyncInform (`0`): synchronous propagates after all recvs for a round, but has sync cost (matches LBAF approach)
  - AsyncInform (`1`): asynchronous propagates after the first recv for a round, but avoids sync cost
- `ordering` (default Marginal): order in which to evaluate local objects for migration
  - Arbitrary (`0`): use the unordered_map iteration order
  - ElmID (`1`): order by ascending element ID
  - Marginal (`2`): order by descending load starting with the object of marginal load, then order ascending for larger loads
- `cmf` (default NormByMax): the algorithm used for computing the CMF
  - Original (`0`): remove processors from the CMF as soon as they exceed the target (e.g., processor-average) load; use a CMF factor of 1.0/x, where x is the target load
  - NormByMax (`1`): do not remove processors from the CMF that exceed the target load until the next iteration; use a CMF factor of 1.0x, where x is the maximum of the target load and the most loaded processor in the CMF
  - NormBySelf (`2`): do not remove processors from the CMF that exceed the target load until the next iteration; use a CMF factor of 1.0x, where x is the load of the processor that is computing the CMF
- `rollback` (default true): whether to roll back to an earlier iteration if it had the best imbalance
- `targetpole` (default false): whether to replace the processor-average load with the max of that and the maximum object load, effectively redefining overloaded/underloaded based on the longest pole load when it exceeds the processor-average load

Note that I have a both a develop-based branch (this PR; `1279-gossiplb-inform-and-decide-have-bugs`) and a release-based branch (`1279-gossiplb-inform-and-decide-have-bugs-release`).

Closes #1279

### DIFF
--- a/src/vt/vrt/collection/balance/gossiplb/gossip_msg.h
+++ b/src/vt/vrt/collection/balance/gossiplb/gossip_msg.h
@@ -85,6 +85,31 @@ private:
   NodeLoadType node_load_ = {};
 };
 
+struct GossipMsgAsync : GossipMsg {
+  using MessageParentType = GossipMsg;
+  vt_msg_serialize_if_needed_by_parent();
+
+  GossipMsgAsync() = default;
+  GossipMsgAsync(
+    NodeType in_from_node, NodeLoadType const& in_node_load, int round
+  )
+    : GossipMsg(in_from_node, in_node_load), round_(round)
+  { }
+
+  uint8_t getRound() const {
+    return round_;
+  }
+
+  template <typename SerializerT>
+  void serialize(SerializerT& s) {
+    MessageParentType::serialize(s);
+    s | round_;
+  }
+
+private:
+  int round_;
+};
+
 struct LazyMigrationMsg : SerializeRequired<
   vt::Message,
   LazyMigrationMsg
--- a/src/vt/vrt/collection/balance/gossiplb/gossip_msg.h
+++ b/src/vt/vrt/collection/balance/gossiplb/gossip_msg.h
@@ -159,13 +159,27 @@ struct RejectionStats {
   int n_transfers_ = 0;
 };
 
-struct GossipRejectionStatsMsg : collective::ReduceTMsg<RejectionStats> {
+static_assert(
+  vt::messaging::is_byte_copyable_t<RejectionStats>::value,
+  "Must be trivially copyable to avoid serialization."
+);
+
+struct GossipRejectionStatsMsg : NonSerialized<
+  collective::ReduceTMsg<RejectionStats>,
+  GossipRejectionStatsMsg
+>
+{
+  using MessageParentType = NonSerialized<
+    collective::ReduceTMsg<RejectionStats>,
+    GossipRejectionStatsMsg
+  >;
+
   GossipRejectionStatsMsg() = default;
   GossipRejectionStatsMsg(int n_rejected, int n_transfers)
-    : ReduceTMsg<RejectionStats>(RejectionStats(n_rejected, n_transfers))
+    : MessageParentType(RejectionStats(n_rejected, n_transfers))
   { }
   GossipRejectionStatsMsg(RejectionStats&& rs)
-    : ReduceTMsg<RejectionStats>(std::move(rs))
+    : MessageParentType(std::move(rs))
   { }
 };
 
--- a/src/vt/vrt/collection/balance/gossiplb/gossip_msg.h
+++ b/src/vt/vrt/collection/balance/gossiplb/gossip_msg.h
@@ -143,6 +143,32 @@ private:
   ObjsType objs_  = {};
 };
 
+struct RejectionStats {
+  RejectionStats() = default;
+  RejectionStats(int n_rejected, int n_transfers)
+    : n_rejected_(n_rejected), n_transfers_(n_transfers) { }
+
+  friend RejectionStats operator+(RejectionStats a1, RejectionStats const& a2) {
+    a1.n_rejected_ += a2.n_rejected_;
+    a1.n_transfers_ += a2.n_transfers_;
+
+    return a1;
+  }
+
+  int n_rejected_ = 0;
+  int n_transfers_ = 0;
+};
+
+struct GossipRejectionStatsMsg : collective::ReduceTMsg<RejectionStats> {
+  GossipRejectionStatsMsg() = default;
+  GossipRejectionStatsMsg(int n_rejected, int n_transfers)
+    : ReduceTMsg<RejectionStats>(RejectionStats(n_rejected, n_transfers))
+  { }
+  GossipRejectionStatsMsg(RejectionStats&& rs)
+    : ReduceTMsg<RejectionStats>(std::move(rs))
+  { }
+};
+
 }}}} /* end namespace vt::vrt::collection::balance */
 
 #endif /*INCLUDED_VT_VRT_COLLECTION_BALANCE_GOSSIPLB_GOSSIP_MSG_H*/
--- a/src/vt/vrt/collection/balance/gossiplb/gossiplb.cc
+++ b/src/vt/vrt/collection/balance/gossiplb/gossiplb.cc
@@ -305,7 +305,7 @@ void GossipLB::informAsync() {
   uint8_t k_cur_async = 0;
 
   vt_debug_print(
-    normal, gossiplb, node,
+    normal, gossiplb,
     "GossipLB::informAsync: starting inform phase: trial={}, iter={}, "
     "k_max={}, k_cur={}, is_underloaded={}, is_overloaded={}, load={}\n",
     trial_, iter_, k_max_, k_cur_, is_underloaded_, is_overloaded_, this_new_load_
@@ -330,7 +330,7 @@ void GossipLB::informAsync() {
 
   // Underloaded start the round
   if (is_underloaded_) {
-    propagateRoundAsync(k_cur_async, propagate_epoch);
+    propagateRound(k_cur_async, false, propagate_epoch);
   }
 
   theTerm()->finishedEpoch(propagate_epoch);
@@ -388,7 +388,7 @@ void GossipLB::informSync() {
     // Underloaded start the first round; ranks that received on some round
     // start subsequent rounds
     if (propagate_this_round) {
-      propagateRoundSync(propagate_epoch);
+      propagateRound(k_cur_, propagate_epoch, true);
     }
 
     theTerm()->finishedEpoch(propagate_epoch);
@@ -421,11 +421,11 @@ void GossipLB::setupDone(ReduceMsgType* msg) {
   setup_done_ = true;
 }
 
-void GossipLB::propagateRoundAsync(uint8_t k_cur_async, EpochType epoch) {
+void GossipLB::propagateRound(uint8_t k_cur, bool sync, EpochType epoch) {
   vt_debug_print(
     normal, gossiplb,
-    "GossipLB::propagateRoundAsync: trial={}, iter={}, k_max={}, k_cur={}\n",
-    trial_, iter_, k_max_, k_cur_
+    "GossipLB::propagateRound: trial={}, iter={}, k_max={}, k_cur={}\n",
+    trial_, iter_, k_max_, k_cur
   );
 
   auto const this_node = theContext()->getNode();
@@ -446,9 +446,9 @@ void GossipLB::propagateRoundAsync(uint8_t k_cur_async, EpochType epoch) {
 
   vt_debug_print(
     verbose, gossiplb,
-    "GossipLB::propagateRoundAsync: trial={}, iter={}, k_max={}, k_cur={}, "
+    "GossipLB::propagateRound: trial={}, iter={}, k_max={}, k_cur={}, "
     "selected.size()={}, fanout={}\n",
-    trial_, iter_, k_max_, k_cur_, selected.size(), fanout
+    trial_, iter_, k_max_, k_cur, selected.size(), fanout
   );
 
   for (int i = 0; i < fanout; i++) {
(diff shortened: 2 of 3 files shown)


src/vt/vrt/collection/balance/gossiplb/gossiplb.h
```diff
@@ -52,16 +52,59 @@
 
 #include <random>
 #include <unordered_map>
+#include <map>
 #include <unordered_set>
 #include <vector>
 
 namespace vt { namespace vrt { namespace collection { namespace lb {
 
+// gossiping approach
```
lifflander, line 61: It would be nice to convert all these to doxygen style comments so they will actually show up on the generated documentation

src/vt/vrt/collection/balance/gossiplb/gossiplb.h
```diff
@@ -103,18 +151,43 @@ struct GossipLB : BaseLB {
   uint8_t k_max_                                    = 4;
   uint8_t k_cur_                                    = 0;
   uint16_t iter_                                    = 0;
+  uint16_t trial_                                   = 0;
   uint16_t num_iters_                               = 4;
+  // how many times to repeat the requested number of iterations, hoping to
```
lifflander, line 156: Same with these comments. I'm fine if we just do this on develop since that's where documentation is generated.

lifflander (approved):
Looks good. I think it's fine if we merge this now. For develop, we could improve the documentation better (doxygen comments) and need to fix the prints since that API was refactored last week.

src/vt/vrt/collection/balance/gossiplb/gossiplb.cc
```diff
@@ -106,59 +108,185 @@ void GossipLB::runLB() {
...
+    }
 
-    vt_debug_print(
-      gossiplb, node,
-      "GossipLB::doLBStages: (after) running iter_={}, num_iters_={}, load={}, new_load={}\n",
-      iter_, num_iters_, this_load, this_new_load_
-    );
+    if (new_imbalance_ <= start_imb && new_imbalance_ < best_imb) {
+      best_load = this_new_load_;
+      best_objs = cur_objs_;
```
PhilMiller: It would probably be good to do `best_objs = std::move(cur_objs_);` to avoid copying. Probably not really critical in current use cases.

src/vt/vrt/collection/balance/gossiplb/gossiplb.cc
```diff
@@ -464,6 +596,15 @@ void GossipLB::decide() {
   theTerm()->finishedEpoch(lazy_epoch);
 
   vt::runSchedulerThrough(lazy_epoch);
+
+  runInEpochCollective([=] {
+    using ReduceOp = collective::PlusOp<RejectionStats>;
+    auto cb = vt::theCB()->makeBcast<
```
PhilMiller: Why is this broadcast, and then the results only used on node 0? Why not just send it solely to rank 0, and skip the conditional in the handler?

src/vt/vrt/collection/balance/gossiplb/gossiplb.h
```diff
@@ -57,6 +57,32 @@
...
+    a1.n_transfers_ += a2.n_transfers_;
+
+    return a1;
+  }
+
+  int n_rejected_ = 0;
+  int n_transfers_ = 0;
+};
+
+struct GossipRejectionStatsMsg : collective::ReduceTMsg<RejectionStats> {
```
PhilMiller: @lifflander Would it be possible to avoid the need to define types like this, by just passing through constructor arguments to the encapsulated type `T` in `ReduceTMsg<T>`?

src/vt/vrt/collection/balance/gossiplb/gossiplb.cc
```diff
@@ -365,28 +368,31 @@ void GossipLB::propagateRound(EpochType epoch) {
...
-    auto msg = makeMessage<GossipMsg>(this_node, load_info_);
+    auto msg = makeMessage<GossipMsgAsync>(this_node, load_info_, k_cur_async);
     if (epoch != no_epoch) {
       envelopeSetEpoch(msg->env, epoch);
     }
     msg->addNodeLoad(this_node, this_new_load_);
-    proxy_[random_node].sendMsg<GossipMsg, &GossipLB::propagateIncoming>(msg.get());
+    proxy_[random_node].sendMsg<
+      GossipMsgAsync, &GossipLB::propagateIncomingAsync
+    >(msg.get());
```
PhilMiller: I think there's no longer a need to call `msg.get()` rather than just passing `msg`?

src/vt/vrt/collection/balance/gossiplb/gossiplb.cc
```diff
@@ -385,6 +465,66 @@ void GossipLB::propagateRoundAsync(uint8_t k_cur_async, EpochType epoch) {
...
+      gossiplb, node,
+      "GossipLB::propagateRoundSync: k_max_={}, k_cur_={}, sending={}\n",
+      k_max_, k_cur_, random_node
+    );
+
+    // Send message with load
+    auto msg = makeMessage<GossipMsgSync>(this_node, load_info_);
+    if (epoch != no_epoch) {
+      envelopeSetEpoch(msg->env, epoch);
+    }
```
PhilMiller, line 520: It's not clear why this code does any manual epoch handling, rather than just calling `propagateRoundSync` under `runInEpochCollective`, and letting the machinery do its thing. I don't see anything that would behave differently from the normal path.

src/vt/vrt/collection/balance/gossiplb/gossiplb.cc
```diff
@@ -692,9 +704,71 @@ void GossipLB::decide() {
...
+          // (incorrectly) reflect the total load, which will not be a problem
+          auto marginal = this_new_load_;
+          for (auto &obj : cur_objs_) {
+            // the object stats are in seconds but the processor stats are in
+            // milliseconds; for now, convert the object loads to milliseconds
+            auto obj_load_ms = loadMilli(obj.second);
+            if (obj_load_ms > over_avg && obj_load_ms < marginal) {
+              marginal = obj_load_ms;
+            }
+          }
```
PhilMiller: I think this is implementing something different from what I'd meant by 'marginal object load'. It seems to have been beneficial to apply, so we shouldn't just outright replace it. I'll write up a PR that implements my original idea as another potential ordering.

src/vt/vrt/collection/balance/gossiplb/gossiplb.cc
```diff
@@ -754,7 +827,8 @@ void GossipLB::decide() {
           this_new_load_ -= obj_load_ms;
           selected_load += obj_load_ms;
 
-          iter = cur_objs_.erase(iter);
+          iter = ordered_obj_ids.erase(iter);
```
PhilMiller, line 841: I think there may not be a need to do this erasure, since we're just going to keep iterating forward through `ordered_obj_ids`, and then it goes out of scope.

src/vt/vrt/collection/balance/gossiplb/gossiplb.cc
```diff
@@ -695,6 +695,75 @@ GossipLB::selectObject(
   }
 }
 
+std::vector<GossipLB::ObjIDType> GossipLB::orderObjects() {
```
PhilMiller, line 695: For hypothetical testability, would it make sense to refactor this to take `cur_objs_` and `obj_ordering_` as arguments, rather than accessing the members? It could then be a static or free-standing function.

src/vt/vrt/collection/balance/gossiplb/gossiplb.cc
```diff
@@ -61,109 +61,273 @@ namespace vt { namespace vrt { namespace collection { namespace lb {
...
+          auto msg = makeMessage<StatsMsgType>(Statistic::P_l, this_new_load_);
+          this->proxy_.template reduce<ReduceOp>(msg,cb);
+        });
+      }
+
+      if (rollback_ || (iter_ == num_iters_ - 1)) {
+        // if known, save the best iteration within any trial so we can roll back
+        if (new_imbalance_ < best_imb && new_imbalance_ <= start_imb) {
+          best_load = this_new_load_;
+          best_objs = cur_objs_;
```
PhilMiller, line 239: This should be `best_objs_ = std::move(cur_objs_);` to avoid the cost of copying. Probably not a big deal right now, but may be with larger object counts.

src/vt/vrt/collection/balance/gossiplb/gossiplb.cc
PhilMiller: PXL_20210407_171006242
Given the objects on a node sorted in order of load, I think this illustrates the difference. I believe my definition always identifies a smaller marginal load, except in the degenerate case where there's a single smallest object that accounts for the entire overload of the node.

nlslatt: I'm not sure if I'm understanding.  Are you saying that you would have picked the object that is sliced by the average load when ordered descending by load?

src/vt/vrt/collection/balance/gossiplb/gossiplb.cc
```diff
@@ -61,109 +61,273 @@ namespace vt { namespace vrt { namespace collection { namespace lb {
...
+    load_info_.clear();
+    is_overloaded_ = is_underloaded_ = false;
+
+    TimeType best_imb_this_trial = start_imb + 10;
+
+    for (iter_ = 0; iter_ < num_iters_; iter_++) {
+      bool first_iter = iter_ == 0;
+
+      vt_debug_print_verbose(
+        gossiplb, node,
```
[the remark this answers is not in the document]
nlslatt: No, this PR is for develop, so the change does need to be made.

src/vt/vrt/collection/balance/gossiplb/gossiplb.cc
nlslatt: `cur_objs_` will continue being modified on the next iteration, so I can't just move it.

src/vt/vrt/collection/balance/gossiplb/gossiplb.cc
nlslatt: In the past, messages were sent in the global epoch.  Is that no longer true?

src/vt/vrt/collection/balance/gossiplb/gossiplb.cc
nlslatt: It also accesses the members `this_new_load_` and `target_max_load_`.  We could pass all four as arguments for the sake of testability.

nlslatt:
@lifflander @PhilMiller Because the release branch equivalent of this has already been released, should I make only minimal fixes here (i.e., the debug prints) and make other recommended changes as a separate PR so that they can be applied to both develop and 1.0.2?

src/vt/vrt/collection/balance/gossiplb/gossiplb.cc
nlslatt: You're right that the erase on `ordered_obj_ids` is not necessary.

src/vt/vrt/collection/balance/gossiplb/gossiplb.cc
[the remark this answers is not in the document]
nlslatt: It's going to take me a long time to rebase and resolve all these new conflicts.

PhilMiller:
> @lifflander @PhilMiller Because the release branch equivalent of this has already been released, should I make only minimal fixes here (i.e., the debug prints) and make other recommended changes as a separate PR so that they can be applied to both develop and 1.0.2?

I'd support that approach.

src/vt/vrt/collection/balance/gossiplb/gossiplb.cc
PhilMiller: Yes, exactly. If we want to pick the objects that will have the easiest time finding new hosts, that will be the hardest-to-place object that nevertheless needs to be relocated.

Also, that should probably have been 'target load' rather than 'P_avg', for consistency with that reasoning elsewhere.

src/vt/vrt/collection/balance/gossiplb/gossiplb.cc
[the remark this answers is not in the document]
nlslatt: Ok, I figured out a more efficient way to handle it given that debug print changes are the only conflicts.  Should be done soon.

src/vt/vrt/collection/balance/gossiplb/gossip_msg.h
```diff
@@ -85,6 +85,31 @@ struct GossipMsg : vt::Message {
...
+  template <typename SerializerT>
+  void serialize(SerializerT& s) {
+    MessageParentType::serialize(s);
+    s | round_;
+  }
+
+private:
+  int round_;
+};
+
```
PhilMiller, line 112: Is it worth adding another whole message type, with associated template instantiations, rather than just adding the one extra `int round_;` to the base `GossipMsg` that may go unused under some configurations?

src/vt/vrt/collection/balance/gossiplb/gossiplb.cc
```diff
@@ -61,109 +61,273 @@ namespace vt { namespace vrt { namespace collection { namespace lb {
...
 bool GossipLB::isUnderloaded(LoadType load) const {
-  auto const avg  = stats.at(lb::Statistic::P_l).at(lb::StatisticQuantity::avg);
-  return load < avg * gossip_threshold;
+  return load < target_max_load_ * gossip_threshold;
 }
 
 bool GossipLB::isOverloaded(LoadType load) const {
-  auto const avg  = stats.at(lb::Statistic::P_l).at(lb::StatisticQuantity::avg);
-  return load > avg * gossip_threshold;
+  return load > target_max_load_ * gossip_threshold;
```
PhilMiller, line 74: This isn't new to this PR, but the symmetric use of `gossip_threshold` between `isUnderloaded` and `isOverloaded` seems wrong to me. That sort of threshold would seem to be most useful for narrowing the scope of which processes are involved in migration activity. As applied here, it just shifts a line, while leaving every process part of the overloaded/underloaded dichotomy.

src/vt/vrt/collection/balance/gossiplb/gossiplb.cc
```diff
@@ -188,54 +345,138 @@ void GossipLB::inform() {
 
   theSched()->runSchedulerWhile([this]{ return not setup_done_; });
```
PhilMiller, line 346: I think the whole `setup_done_ / setupDone` thing could be simplified to `runInEpochCollective` around a reduction to a null callback, or a named barrier. We're not getting any information out of the handler being called with any result from the reduction that I can see.

PhilMiller: Again, not new to this PR, just a potential improvement I noticed as I'm reading through.

src/vt/vrt/collection/balance/gossiplb/gossiplb.cc
PhilMiller: They'll be sent in whatever epoch is top of the stack from `pushEpoch()` calls. In a handler, it's from the instigating message, and in `runInEpochCollective`, it's the epoch created by the call.

@lifflander could you make sure I'm not missing something here?

PhilMiller (approved):
For the sake of not needlessly diverging from the release branch code, I'd support merging this as is (modulo the `debug_print` conflict resolution), and addressing all the comments in a subsequent PR

src/vt/vrt/collection/balance/gossiplb/gossiplb.cc
nlslatt: How should we distinguish (in terms of a naming convention) between your approach and mine?

PhilMiller: I realized that would be an issue. I don't have a good set of replacement names in mind.

nlslatt: @PhilMiller @lifflander I think the main difference is that the objects smaller than Phil's marginal object also contribute to the amount of overload, whereas my marginal object (if a large enough object exists) is the only contributor.  Perhaps mine should be renamed to something like `Overload` ordering?  I hate to do that after we've released, but the `ordering` LB arg uses an integer so really only the enum within the source code would change.  We should make this change and re-release ASAP if that's the way we want to go.

nlslatt: We decided today that my approach is best described as `LeastMigrations` and Phil's is `SmallestObjects`.

src/vt/vrt/collection/balance/gossiplb/gossiplb.cc
```diff
@@ -61,109 +61,273 @@ namespace vt { namespace vrt { namespace collection { namespace lb {
...
 
 bool GossipLB::isOverloaded(LoadType load) const {
-  auto const avg  = stats.at(lb::Statistic::P_l).at(lb::StatisticQuantity::avg);
-  return load > avg * gossip_threshold;
+  return load > target_max_load_ * gossip_threshold;
 }
 
 void GossipLB::inputParams(balance::SpecEntry* spec) {
-  std::vector<std::string> allowed{"f", "k", "i", "c"};
+  std::vector<std::string> allowed{
```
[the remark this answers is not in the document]
nlslatt: Deferred until #1368.

src/vt/vrt/collection/balance/gossiplb/gossiplb.cc
nlslatt: Deferred until #1368.

src/vt/vrt/collection/balance/gossiplb/gossiplb.cc
nlslatt: Deferred until #1368.

src/vt/vrt/collection/balance/gossiplb/gossip_msg.h
nlslatt: Deferred until #1368.

src/vt/vrt/collection/balance/gossiplb/gossiplb.cc
nlslatt: Deferred until #1368.

src/vt/vrt/collection/balance/gossiplb/gossiplb.cc
nlslatt: Deferred until #1368.

src/vt/vrt/collection/balance/gossiplb/gossiplb.cc
nlslatt: Deferred until #1368.

src/vt/vrt/collection/balance/gossiplb/gossiplb.h
nlslatt: Converted, but not tested.

src/vt/vrt/collection/balance/gossiplb/gossiplb.h
nlslatt: Converted, but not tested.

src/vt/vrt/collection/balance/gossiplb/gossiplb.cc
nlslatt: Deferred until #1369.

src/vt/vrt/collection/balance/gossiplb/gossiplb.cc
nlslatt: Correction: implemented in #1369

PR #1360, opened 2021-04-05
1350: move cmake config options to new file and include before MPI wrapper generation

### LINKED ISSUE #1350: MPI wrappers not generated on first cmake run because vt_mpi_guards is not set yet
**Describe the bug**
`vt_mpi_guards` is only set with a default value via `option()` after it is queried as part of a conditional on whether or not to generate MPI wrappers.

**To Reproduce**
Use a cmake/find_package build setup. Don't set `vt_mpi_guards` manually as a parameter. Build fails at linking because of missing generated wrappers.

**Expected behavior**
Build should succeed.

 - Mac OS X + llvm/clang 9.0.0

nmm0:
Fixes #1350 
Any cmake `option()` commands should now be added to `configure_options.cmake` rather than `define_build_types.cmake`


cz4rs (approved):
looks good to me :+1:

JacobDomagala (approved):

lifflander (approved):
Looks good to me.

PhilMiller (approved):

PhilMiller:
This took care of #1408, and thus obviates #1409

PR #1364, opened 2021-04-06
1359: LB: write docs for args to LBs

### LINKED ISSUE #1359: Provide help/documentation for LB arguments
**What Needs to be Done?**

Give some output to help users out here.

lifflander:
Fixes #1359

### DIFF
--- a/src/vt/vrt/collection/balance/greedylb/greedylb.cc
+++ b/src/vt/vrt/collection/balance/greedylb/greedylb.cc
@@ -81,11 +81,10 @@ GreedyLB::getInputKeysWithHelp() {
 Values: <double>
 Default: 0.8
 Description:
-  The load threshold for objects to consider on each node. The default value of
-  0.8 will consider 80% of the average load for re-balancing on each node. The
-  order these will be selected is based on the value passed to "strategy". If
-  the parameter "auto" is set to "true", this will be the minimum threshold;
-  otherwise, it sets the threshold directly.
+  The load threshold of objects to consider for potential migration on each
+  rank. All objects over threshold * average_load on each rank will
+  considered. If the parameter "auto" is set to "true", this will be the minimum
+  threshold; otherwise, it sets the threshold directly.
 )"
     },
     {
--- a/src/vt/vrt/collection/balance/greedylb/greedylb.cc
+++ b/src/vt/vrt/collection/balance/greedylb/greedylb.cc
@@ -82,7 +82,7 @@ Values: <double>
 Default: 0.8
 Description:
   The load threshold of objects to consider for potential migration on each
-  rank. All objects over threshold * average_load on each rank will
+  rank. All objects over threshold * average_load on each rank will be
   considered. If the parameter "auto" is set to "true", this will be the minimum
   threshold; otherwise, it sets the threshold directly.
 )"
--- a/src/vt/vrt/collection/balance/greedylb/greedylb.cc
+++ b/src/vt/vrt/collection/balance/greedylb/greedylb.cc
@@ -112,6 +112,10 @@ Description:
     {
       "strategy",
       R"(
+Values: {scatter, bcast, pt2pt}
+Default: scatter
+Description:
+  How to distribute the data after the centralized LB makes a decision
 )"
     }
   };
--- a/src/vt/vrt/collection/balance/greedylb/greedylb.cc
+++ b/src/vt/vrt/collection/balance/greedylb/greedylb.cc
@@ -72,8 +72,59 @@ void GreedyLB::init(objgroup::proxy::Proxy<GreedyLB> in_proxy) {
   proxy = scatter_proxy = in_proxy;
 }
 
+/*static*/ std::unordered_map<std::string, std::string>
+GreedyLB::getInputKeysWithHelp() {
+  std::unordered_map<std::string, std::string> const keys_help = {
+    {
+      "min",
+      R"(
+Values: <double>
+Default: 0.8
+Description:
+  The load threshold for objects to consider on each node. The default value of
+  0.8 will consider 80% of the average load for re-balancing on each node. The
+  order these will be selected is based on the value passed to "strategy". If
+  the parameter "auto" is set to "true", this will be the minimum threshold;
+  otherwise, it sets the threshold directly.
+)"
+    },
+    {
+      "max",
+      R"(
+Values: <double>
+Default: 1.004
+Description:
+  The maximum load threshold for objects to consider on each node which is only
+  used if "auto" is "true".
+)"
+    },
+    {
+      "auto",
+      R"(
+Values: {true, false}
+Default: true
+Description:
+  Automatically determine the threshold between "min" and "max" using
+  calculated I (imbalance metric) with the formula
+  min(max(1-I, min), max).
+)"
+    },
+    {
+      "strategy",
+      R"(
+)"
+    }
+  };
+  return keys_help;
+}
+
 void GreedyLB::inputParams(balance::SpecEntry* spec) {
-  std::vector<std::string> allowed{"min", "max", "auto", "strategy"};
+  auto keys_help = getInputKeysWithHelp();
+
+  std::vector<std::string> allowed;
+  for (auto&& elm : keys_help) {
+    allowed.push_back(elm.first);
+  }
   spec->checkAllowedKeys(allowed);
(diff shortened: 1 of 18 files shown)


lifflander:
@nlslatt : Once https://github.com/DARMA-tasking/vt/pull/1348 is merged we should add the documentation for the new arguments here.

I still need to hook this into a VT command-line argument that outputs them.

nlslatt:
@lifflander We had talked about adding support for short-form argument names, like `f` for `fanout` and `k` for rounds in `TemperedLB`. I suggest we merge this and then add that support later, considering how long it's taken for me to get around documenting what's already there.  How does that sound?

nlslatt:
Is there some special treatment being given to `--vt_lb` and `--vt_lb_quiet` so that they don't need to explicitly be set to true or false?  I tried adding an argument to trigger printing the LB args documentation (not committed), but it errors out at runtime if I give the argument but don't explicitly set it to `true` or `1`.

What do we want to call this new argument?  We could do `--vt_lb_args_help`, `--vt_help_lb_args`, or...

nlslatt:
> Is there some special treatment being given to `--vt_lb` and `--vt_lb_quiet` so that they don't need to explicitly be set to true or false? I tried adding an argument to trigger printing the LB args documentation (not committed), but it errors out at runtime if I give the argument but don't explicitly set it to `true` or `1`.
> 
> What do we want to call this new argument? We could do `--vt_lb_args_help`, `--vt_help_lb_args`, or...

Oh, I failed to notice that some are `add_flag` while others are `add_option`.  I'm going to go with `--vt_help_lb_args` for now to group help-related options, but can change that later if someone thinks it's better to group it with the lb options.

lifflander:
I can't review this since I'm an author, but it looks good to me. I think it might be ready to merge.

lifflander:
@nlslatt Can you verify that I addressed all the LBs with my latest changes?

lifflander:
It seems to be printing the LB help message twice with 2 ranks. I think we want only rank 0 to print.

src/vt/vrt/collection/balance/lb_invoke/lb_manager.h
```diff
@@ -133,6 +133,22 @@ struct LBManager : runtime::component::Component<LBManager> {
...
+   *
+   * \param[in] lb the load balancer to query
+   */
+  static void printLBArgsHelp(LBType lb);
+
+  /**
+   * \internal
+   * \brief Print documentation for LB args for all LBs or the one specified
+   *
+   * \param[in] lb the load balancer in use
```
PhilMiller, line 148: Comment update - "load balancer to query", as above

src/vt/vrt/collection/balance/lb_invoke/lb_manager.cc
```diff
@@ -262,6 +262,71 @@ void LBManager::startLB(PhaseType phase, LBType lb) {
...
+    // deliberately skip retrieving arguments
+    break;
+  default:
+    fmt::print("Documentation has not been provided for this LB.\n\n");
+    return;
+    break;
+  }
+
+  if (help.size() > 0) {
+    for (auto &arg_help : help) {
```
PhilMiller, line 305: Should there be an `if (rank == 0)` around this?

src/vt/vrt/collection/balance/greedylb/greedylb.cc
```diff
@@ -72,8 +72,63 @@ void GreedyLB::init(objgroup::proxy::Proxy<GreedyLB> in_proxy) {
...
+Values: {true, false}
+Default: true
+Description:
+  Automatically determine the threshold between "min" and "max" using
+  calculated I (imbalance metric) with the formula
+  min(max(1-I, min), max).
+)"
+    },
+    {
+      "strategy",
```
PhilMiller, line 113: 'strategy' is a terrible name for this parameter. Also, it's not one I think we should have at all. Just pick one good way to distribute the decision, and go with that.

src/vt/vrt/collection/balance/greedylb/greedylb.cc
```diff
@@ -72,8 +72,63 @@ void GreedyLB::init(objgroup::proxy::Proxy<GreedyLB> in_proxy) {
   proxy = scatter_proxy = in_proxy;
 }
 
+/*static*/ std::unordered_map<std::string, std::string>
+GreedyLB::getInputKeysWithHelp() {
```
PhilMiller, line 76: Even knowing  what GreedyLB does, I can't make sense of the descriptions of any of min, max, or auto. Threshold(s) for **what**?

src/vt/vrt/collection/balance/hierarchicallb/hierlb.cc
```diff
@@ -71,8 +71,68 @@ void HierarchicalLB::init(objgroup::proxy::Proxy<HierarchicalLB> in_proxy) {
...
+      "auto",
+      R"(
+Values: {true, false}
+Default: true
+Description:
+  Automatically determine the threshold between "min" and "max" using
+  calculated I (imbalance metric) with the formula
+  min(max(1-I, min), max).
+)"
+    },
```
PhilMiller, line 110: Same exact complaint as for GreedlyLB

src/vt/vrt/collection/balance/hierarchicallb/hierlb.cc
```diff
@@ -71,8 +71,68 @@ void HierarchicalLB::init(objgroup::proxy::Proxy<HierarchicalLB> in_proxy) {
...
+    {
+      "strategy",
```
PhilMiller, line 112: Again, 'strategy' is not descriptive of what's actually being set. Making my complaint about GreedyLB even worse, it means something totally different here, though at least this time it actually has something to do with how load will be redistributed.

src/vt/vrt/collection/balance/lb_invoke/lb_manager.cc
lifflander: I put that in the runtime instead where it calls this.

nlslatt:
> It seems to be printing the LB help message twice with 2 ranks. I think we want only rank 0 to print.

Ah, yes, I refactored near the end and lost my rank protection within the call from the runtime.

src/vt/vrt/collection/balance/lb_invoke/lb_manager.h
nlslatt: This phrasing was deliberate because it will only be passed the load balancer that vt intends to use.  Do you still think it should be changed?

nlslatt:
> @nlslatt Can you verify that I addressed all the LBs with my latest changes?

Looks like you got them all.

nlslatt:
@PhilMiller Given that this issue/PR are for adding documentation to existing options (that you dislike), should we use new issue(s) to address your arg naming-related comments?

src/vt/vrt/collection/balance/lb_invoke/lb_manager.h
PhilMiller: That's fine then.

PhilMiller:
Could we get a review from Roger or one of the other folks representing downstream users on whether they understand the help text?

nlslatt:
> Could we get a review from Roger or one of the other folks representing downstream users on whether they understand the help text?

I started copying the docs into an email to send, but found that I already agree that they need to be clarified.  We could ask his opinion after improving them.  I can look into that.

src/vt/vrt/collection/balance/hierarchicallb/hierlb.cc
nlslatt: I'm trying to improve the descriptions and I have a question.  Does the formula above (which does accurately represent the formula used) imply that the more imbalanced the problem, the less the number of objects allowed to migrate?

nlslatt: Also, given the `1-I` in there, does it ever even make sense to use a maximum threshold greater than 1 (the default is 1.004)?

lifflander: I'm actually completely changing how this works in `1560-fix-lb-args` and getting rid of some of these inputs.

lifflander: But, I haven't pushed it yet.

PhilMiller (approved):

PR #1365, opened 2021-04-07
1247: Generate clang 9/10 workflows for Azure

### LINKED ISSUE #1247: Add clang-9, clang-10 to CI
**What Needs to be Done?**

These builds work---we should add them as CI targets on ubuntu.

JacobDomagala:
Fixes #1247


JacobDomagala:
Both Clang 9 and 10 are using default configuration. Do we want to add anything special for them?

cz4rs (approved):
looks good to me!

jstrzebonski (approved):
Looks good

cz4rs:
> Both Clang 9 and 10 are using default configuration. Do we want to add anything special for them?

I assume that one of these can be used for #1352, but they look just fine at the moment :+1:

lifflander (approved):
Looks good

PR #1374, opened 2021-04-12
1331: Optimize build times - use fmt as a static library

### LINKED ISSUE #1331: Figure out why build/test times increased
**What Needs to be Done?**

Research what happened here to solve the problem.

cz4rs:
experimental PR for #1331 
fixes #1331


PhilMiller:
Could we make the include of the fmt headers via `-isystem` instead of `-I`, so that warnings inside it are suppressed?
Relatedly, build its directory with `-w` for the same reasons

cz4rs:
> Could we make the include of the fmt headers via `-isystem` instead of `-I`, so that warnings inside it are suppressed?
> Relatedly, build its directory with `-w` for the same reasons

I have fixed the warning for now, but `-isystem` and `-w` should be superior - I will have a look at that.

cz4rs:
A simple test with a cleared cache:
`time /usr/bin/cmake --build . --config Release --target install -- -j 10`

**develop**
```
real    9m21,086s
user    70m12,107s
sys     1m22,634s
```
**1331-optimize-fmt-use**
```
real    7m34,897s
user    55m55,678s
sys     1m15,743s
```

cz4rs:
> > Could we make the include of the fmt headers via `-isystem` instead of `-I`, so that warnings inside it are suppressed?
> > Relatedly, build its directory with `-w` for the same reasons
> 
> I have fixed the warning for now, but `-isystem` and `-w` should be superior - I will have a look at that.
> 
> _sidenote: we use `set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${flag}" PARENT_SCOPE)` to set the flags instead of "modern CMake" recommended `target_compile_options` which could also do the trick_

In the end, `os.h` header that contained problematic code was removed completely. I can still go for `-isystem` and `-w` if you think that it provides value.

lib/fmt/CMakeLists.txt
```diff
@@ -1,17 +1,94 @@
...
+  set(headers ${${VAR}})
+  foreach (header ${ARGN})
+    set(headers ${headers} include/fmt/${header})
+  endforeach()
+  set(${VAR} ${headers} PARENT_SCOPE)
+endfunction()
+
+# Define the fmt library, its includes and the needed defines.
+add_headers(FMT_HEADERS chrono.h color.h compile.h core.h format.h format-inl.h
+                        locale.h ostream.h printf.h ranges.h)
```
JacobDomagala, line 74: Do we want to add all the files? I think currently we only use core and format(h/inl.h/cc) and the rest seems optional

cz4rs: Good idea, I'll try to cut out as much as I can.

lib/fmt/CMakeLists.txt
```diff
@@ -1,17 +1,38 @@
...
-#  DESTINATION  include/fmt
-#)
+project(FMT CXX)
+
+if (NOT CMAKE_RUNTIME_OUTPUT_DIRECTORY)
+  set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
+endif ()
+
+set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH}
+  "${CMAKE_CURRENT_SOURCE_DIR}/support/cmake")
```
JacobDomagala, line 8: Do we need these (together with `lib/fmt/support/cmake/fmt-config.cmake.in`)?

JacobDomagala: We're building the static lib so `CMAKE_RUNTIME_OUTPUT_DIRECTORY` shouldn't be used, and `fmt-config.cmake.in` is for (I think) generating packages which we also don't do (with fmt) so I think these should be removed

lifflander: I think you are correct @JacobDomagala, unless we are exporting the configuration. Do we need to be doing that now that it's not just header only for apps that use VT.

lifflander (approved):

lib/fmt/CMakeLists.txt
lifflander: Once we work this out, I think this is ready to merge.

JacobDomagala:
@cz4rs  I think we might want to disable `496 - vt:*TestRDMAHandleCollection/*test_rdma_handle_collection_1_proc_4 (Timeout)`  for now, it's been randomly timing out on OpenMPI lately

cz4rs:
> @cz4rs I think we might want to disable `496 - vt:*TestRDMAHandleCollection/*test_rdma_handle_collection_1_proc_4 (Timeout)` for now, it's been randomly timing out on OpenMPI lately

Disabled, looks like it should be fine now.

lib/fmt/CMakeLists.txt
cz4rs: I have removed it all, no issues with sample project observed :+1:

JacobDomagala (approved):

PR #1384, opened 2021-04-15
1383: Add variants of runInEpoch* that take a string label as the first argument 

### LINKED ISSUE #1383: Add variants of runInEpoch* that take a string label as the first argument
**What Needs to be Done?**

These should match the API in `theTerm()->makeEpochCollective`

cz4rs:
fixes #1383

### DIFF
--- a/src/vt/scheduler/scheduler.h
+++ b/src/vt/scheduler/scheduler.h
@@ -69,7 +69,13 @@ template <typename Callable>
 void runInEpochRooted(Callable&& fn);
 
 template <typename Callable>
+void runInEpochRooted(std::string const& label, Callable&& fn);
+
+template <typename Callable>
 void runInEpochCollective(Callable&& fn);
+
+template <typename Callable>
+void runInEpochCollective(std::string const& label, Callable&& fn);
 
 namespace messaging {
 
--- a/src/vt/scheduler/scheduler.impl.h
+++ b/src/vt/scheduler/scheduler.impl.h
@@ -69,8 +69,20 @@ void runInEpochCollective(Callable&& fn) {
 }
 
 template <typename Callable>
+void runInEpochCollective(std::string const& label, Callable&& fn) {
+  auto ep = theTerm()->makeEpochCollective(label);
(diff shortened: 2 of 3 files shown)


cz4rs:
`makeEpoch*` family has overloads that take `label` argument, but it's pretty much the same as defaulting to an empty string:
https://github.com/DARMA-tasking/vt/blob/d96de206a661eb41507fd738f2592f604829210e/src/vt/termination/termination.cc#L1044-L1048


I have pushed the code that takes string as the first argument (as requested), but we could remove the overloads and just go with:
```cpp
  EpochType makeEpochRooted(
    std::string const& label = {},
    UseDS use_ds = UseDS{false},
    SuccessorEpochCapture successor = SuccessorEpochCapture{}
  );
```
and
```cpp
template <typename Callable>
void runInEpochRooted(Callable&& fn, std::string const& label = {});
```

What do you think @lifflander?

jstrzebonski (approved):
Looks good overall but let's wait for the decision on removing overloads.

lifflander:
> What do you think @lifflander?

I like the idea of reducing overloads as much as possible, but I think having arguments after a potential lambda is not a great interface because it's harder to read the code.

> ```c++
> template <typename Callable>
> void runInEpochRooted(Callable&& fn, std::string const& label = {});
> ```

So you would call this like:

```c++
runInEpochRooted([]{
 // function
}, "my label");
```

I prefer the label going first so that the lambda is the last argument. What do you think @nlslatt @PhilMiller ?

PhilMiller (approved):
Looks good.

I agree that other arguments should generally precede lambdas

Is it possible to query the label of an epoch in the code? Should we test that it matches expectation?

cz4rs:
> Looks good.
> 
> I agree that other arguments should generally precede lambdas
> 
> Is it possible to query the label of an epoch in the code? Should we test that it matches expectation?

`getLabel` is available from `TermState`, but I don't see how to easily access that (since `runInEpoch*` does not expose information about specific epoch used).

PR #1400, opened 2021-04-23
#1399 clean up clang-format config file

### LINKED ISSUE #1399: Clean up .clang-format

jstrzebonski:
Fixes #1399

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,63 +1,92 @@
 Language: Cpp
 ColumnLimit: 80
-
 UseTab: Never
-IndentWidth: 2
+
 AccessModifierOffset: -4
-NamespaceIndentation: None
-CompactNamespaces: true
-AlwaysBreakAfterDefinitionReturnType: None
-AlwaysBreakAfterReturnType: None
 AlignAfterOpenBracket: AlwaysBreak
-ContinuationIndentWidth: 2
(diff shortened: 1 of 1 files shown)


cz4rs (approved):
:+1:

lifflander (approved):
How close it this to our actual style guidelines right now? I remember before I was having trouble specifying it exactly the way I wanted with the older version of clang-format.

jstrzebonski:
It's pretty close, but still, there are some flaws, especially with our bracket-style, but not only. So it cannot be used as format-whole-file tool, but it might be helpful for reformatting parts of source code that were changed (for example moved into or out of if-else block).

Also, it's not really any revolution - I mainly clean up the existing .clang-format, sort the options so that they're aligned with documentation, and also add most of the missing options (from **clang-format 10**), so they're not set into some magic defaults.

---

**Left - original | Right - reformatted**

_not that bad examples:_

image

image

_really ugly one:_

image

jstrzebonski:
Just one example from the top of my head (but there are more of course):

I don't see an option to allow this style of lambda:
`task_ = [=]{ func(msg_.get()); };`

clang-format keeps adding a space just after square bracket:
`task_ = [=] { func(msg_.get()); };`

PR #1404, opened 2021-04-23
1401: documentation fails to build

### LINKED ISSUE #1401: documentation fails to build
**Describe the bug**
After recent changes in how we use `fmt`, documentation fails to build:
```
-- Building without Kokkos kernels library
CMake Warning at lib/checkpoint/src/CMakeLists.txt:147 (message):
  Checkpoint: Install targets skipped because CMake version 3.10.2<3.13;
  affects packages built with a Checkpoint install.  CMake <3.13 does not
  support exporting targets from subdirectories


-- Checkpoint: building as a sub-directory
(...)
-- Performing Test LINKER_SUPPORTS_RDYNAMIC - Success
CMake Warning at src/CMakeLists.txt:308 (message):
  VT: Install targets skipped because CMake version 3.10.2<3.13; affects
  packages built with a VT install.  CMake <3.13 does not support exporting
  targets from subdirectories


-- VT: building examples
(...)
-- Configuring done
CMake Error: install(EXPORT "vt" ...) includes target "vt" which requires target "fmt" that is not in the export set.
-- Generating done
```

Note: documentation build is the only one that uses CMake 3.10.

**To Reproduce**
```
docker-compose run ubuntu-docs
```

**Expected behavior**
Fix the build (either upgrade CMake version or work around this in CMakeLists.txt).

cz4rs:
fixes #1401

### DIFF
--- a/ci/build_cpp.sh
+++ b/ci/build_cpp.sh
@@ -46,17 +46,24 @@ if test -d "${source_dir}/lib/detector"
 then
     { echo "Detector already in lib... not downloading, building, and installing"; } 2>/dev/null
 else
-    git clone -b "${detector_rev}" --depth 1 https://github.com/DARMA-tasking/detector.git
-    export DETECTOR=$PWD/detector
-    export DETECTOR_BUILD=${build_dir}/detector
-    mkdir -p "$DETECTOR_BUILD"
-    cd "$DETECTOR_BUILD"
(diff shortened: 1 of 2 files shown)


JacobDomagala (approved):

lifflander (approved):
Looks good to me!

PR #1405, opened 2021-04-26
 #1357 fix build errors reporting 

### LINKED ISSUE #1357: PR comment doesn't show compilation error in case of gcc-7 "No space left on device" error
Example in PR https://github.com/DARMA-tasking/vt/pull/1356

image

image

jstrzebonski:
Fixes #1357 

So the problem is that if for any reason the build fails without generating logs (from compilation and tests) the script gets empty reports and thinks that it means that everything went super well.

To fix this I pass the pipeline's status to the script, so if the status is not `Succeeded` or `SucceededWithIssues` and the compilation and tests reports are empty then it means that something really bad happened.


JacobDomagala (approved):
🚀

cz4rs (approved):
looks good!

lifflander (approved):
Looks good to me!

PR #1409, opened 2021-04-27
#1408: Add pmpi/generated to list of potential source directories

### LINKED ISSUE #1408: PMPI Wrappers for MPI guards fail to build in embedded setting
**Describe the bug**

A CLANG-DEBUG build with vt as an embedded TPL fails to link like this:

```
TPL/vt/src/libvt-debug.a(runtime.cc.o):runtime.cc:function vt::pmpi::PMPIComponent::startup(): error: undefined reference to 'vt::pmpi::PMPIComponent::registerEventHandlers()'
```

Disabling `vt_mpi_guards` in the CMake configuration averts the error.

The subdirectory listing in `src/CMakeLists.txt` is missing `pmpi/generated`

PhilMiller:
Fixes #1408

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -44,6 +44,8 @@ set(
       pipe/callback/objgroup_bcast
       pipe/callback/anon
       pipe/callback/cb_union
+    pmpi
+      pmpi/generated
     timetrigger
     termination
       termination/dijkstra-scholten


PhilMiller:
Tested in both the disabled and enabled cases, and the code configures, compiles, and links both ways.

JacobDomagala:
Isn't the the issue that we include `include(../cmake/define_build_types.cmake)` (which defines `vt_mpi_guards` build option) after we try to generate `PMPI` wrappers? I could be wrong, but https://github.com/DARMA-tasking/vt/pull/1360 looks like the actual fix for this.

PhilMiller:
Agreed, 1360 may be the proper fix for this.

PhilMiller:
Obviated by #1360, confirmed in EMPIRE that compilation succeeds immediately after the initial CMake configuration, with no second pass to cover for it.

PR #1412, opened 2021-04-28
1407 Fix invoke context for collection element invoke

### LINKED ISSUE #1407: For runnable::invoke make the actual Runnable so context is set properly
**What Needs to be Done?**

- [ ] First do this
- [ ] Then do that.
- [ ] Also this other thing.

**Additional context**

lifflander:
Fixes #1407

### DIFF
--- a/src/vt/context/runnable_context/lb_stats.impl.h
+++ b/src/vt/context/runnable_context/lb_stats.impl.h
@@ -64,6 +64,13 @@ LBStats::LBStats(ElmT* in_elm, MsgT* msg)
   theCollection()->recordStats(in_elm, msg);
 }
 
+template <typename ElmT>
+LBStats::LBStats(ElmT* in_elm)
+  : stats_(&in_elm->getStats()),
+    cur_elm_id_(in_elm->getElmID()),
+    should_instrument_(true)
+{ }
+
 }} /* end namespace vt::ctx */
 
 #endif /*INCLUDED_VT_CONTEXT_RUNNABLE_CONTEXT_LB_STATS_IMPL_H*/
--- a/src/vt/vrt/collection/manager.impl.h
+++ b/src/vt/vrt/collection/manager.impl.h
@@ -687,16 +687,16 @@ util::Copyable<Type> CollectionManager::invoke(
 ) {
   auto ptr = getCollectionPtrForInvoke(proxy);
 
-#if vt_check_enabled(lblite)
-  ptr->getStats().startTime();
-#endif
+  auto const this_node = theContext()->getNode();
+  util::Copyable<Type> result;
 
-  const auto& result =
-    runnable::invoke<Type, f>(ptr, std::forward<Args>(args)...);
-
-#if vt_check_enabled(lblite)
-  ptr->getStats().stopTime();
-#endif
+  runnable::makeRunnableVoid(false, uninitialized_handler, this_node)
+    .withCollection(ptr)
+    .withLBStatsVoidMsg(ptr)
+    .withExplicitTask([&]{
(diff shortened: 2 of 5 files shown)


src/vt/context/runnable_context/lb_stats.impl.h
```diff
@@ -64,6 +64,13 @@ LBStats::LBStats(ElmT* in_elm, MsgT* msg)
   theCollection()->recordStats(in_elm, msg);
 }
 
+template <typename ElmT>
+LBStats::LBStats(ElmT* in_elm)
```
PhilMiller, line 68: Not to be done now, but is there a base type for all `ElmT` that could be passed as an argument to these constructors, to de-template them?

src/vt/vrt/collection/manager.impl.h
```diff
@@ -687,16 +687,16 @@ util::Copyable<Type> CollectionManager::invoke(
...
-#if vt_check_enabled(lblite)
-  ptr->getStats().stopTime();
-#endif
+  runnable::makeRunnableVoid(false, uninitialized_handler, this_node)
+    .withCollection(ptr)
+    .withLBStatsVoidMsg(ptr)
+    .withExplicitTask([&]{
+      result = runnable::invoke<Type, f>(ptr, std::forward<Args>(args)...);
+    })
+    .runOrEnqueue(true);
```
PhilMiller, line 699: Since I don't remember what the `bool` parameter to `runOrEnqueue` means, that suggests it should be an `enum` instead. Your call whether you want to just do that in this PR, or open an issue.

src/vt/vrt/collection/manager.impl.h
```diff
@@ -707,15 +707,17 @@ util::NotCopyable<Type> CollectionManager::invoke(
...
-  auto&& result = runnable::invoke<Type, f>(ptr, std::forward<Args>(args)...);
-
-#if vt_check_enabled(lblite)
-  ptr->getStats().stopTime();
-#endif
+  runnable::makeRunnableVoid(false, uninitialized_handler, this_node)
+    .withCollection(ptr)
+    .withLBStatsVoidMsg(ptr)
+    .withExplicitTask([&]{
+      auto&& ret = runnable::invoke<Type, f>(ptr, std::forward<Args>(args)...);
```
PhilMiller, line 717: The assignment to rvalue reference and subsequent move assignment should be unnecessary. If the second line works, then `result = runnable::invoke...` should work just fine also.

src/vt/vrt/collection/manager.impl.h
```diff
@@ -687,16 +687,16 @@ util::Copyable<Type> CollectionManager::invoke(
...
-    runnable::invoke<Type, f>(ptr, std::forward<Args>(args)...);
-
-#if vt_check_enabled(lblite)
-  ptr->getStats().stopTime();
-#endif
+  runnable::makeRunnableVoid(false, uninitialized_handler, this_node)
+    .withCollection(ptr)
+    .withLBStatsVoidMsg(ptr)
+    .withExplicitTask([&]{
+      result = runnable::invoke<Type, f>(ptr, std::forward<Args>(args)...);
```
PhilMiller, line 697: Since I had to think about it a bit, could `Type` be renamed `FunctionType` or something like that? And those `util::Copyable`, etc helpers `util::ReturnsCopyable`?

PhilMiller: I can't wait for C++17 support where this will become `auto f` and `if constexpr(util::ReturnIsCopy<f>) { }` instead of the SFINAE in return types.

PhilMiller (approved):
None of the comments are crucial to address, or make any functional difference.

PR #1420, opened 2021-04-30
1366 LB strategy name can skip validation

### LINKED ISSUE #1366: LB strategy name can skip validation
**Describe the bug**
The vt banner is printed before apps (like EMPIRE) running on vt can programmatically change vt options.  Because the LB strategy name is validated as the runtime banner is printed, LB strategies specified by apps will not be validated there.

In debug builds, the `vtAssert` in `LBManager::startLB()` (develop) or `LBManager::collectiveImpl()` (release branches) will validate the name, but these are optimized out in release (opt) builds.  In optimized builds, one can specify a non-existent or misspelled LB and vt will not error out or even produce a warning.  Instead, it just runs as if load balancing was not requested.

**To Reproduce**
1. Build EMPIRE in release mode without ZoltanLB compiled in (i.e., use -Dvt_zoltan_enabled=0 or omit)
2. Enable load balancing in the YAML file and specify ZoltanLB as the strategy.
3. Run PIC with the YAML file and note how it runs but does not attempt to load balance.

**Expected behavior**
If the requested LB strategy cannot be used, it should error out.

nlslatt:
Fixes #1366 on the release branch.

### DIFF
--- a/src/vt/vrt/collection/balance/lb_invoke/invoke.cc
+++ b/src/vt/vrt/collection/balance/lb_invoke/invoke.cc
@@ -93,12 +93,22 @@ LBType LBManager::decideLBToRun(PhaseType phase, bool try_file) {
     if (phase % ArgType::vt_lb_interval == 1 || (
       ArgType::vt_lb_interval == 1 && phase != 0
     )) {
+      bool name_match = false;
       for (auto&& elm : lb_names_) {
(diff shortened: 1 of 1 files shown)


cz4rs (approved):
looks good to me :+1:

lifflander (approved):

PR #1422, opened 2021-05-03
1376 improve lbargs for gossiplb

### LINKED ISSUE #1376: Improve LB args for GossipLB
**What Needs to be Done?**

- [ ] Change enum integers into strings.
- [ ] Make longer alternatives to overly terse `k`, `f`, `i`, `c`.
- [ ] Allow fanout and rounds to be set using options like `FullInformation`.

nlslatt:
Makes LB args for GossipLB more human readable, improves error handling, adds the (default) ability to have vt choose an appropriate fanout and number of gossiping rounds, makes asynchronous informs the default, and sets the number of trials to 1 (more lightweight) by default.

Closes #1376.


src/vt/vrt/collection/balance/gossiplb/gossiplb.cc
```diff
@@ -76,57 +76,162 @@ bool GossipLB::isOverloaded(LoadType load) const {
...
+      // set the fanout based on the chosen rounds: f=p^(1/k)
+      k_max_ = spec->getOrDefault<int32_t>("rounds", k_max_);
+      f_ = static_cast<uint16_t>(std::ceil(std::pow(num_nodes, 1.0/k_max_)));
+    } else {
+      // set both the fanout and the rounds
+      k_max_ = static_cast<uint16_t>(
+        std::round(std::sqrt(std::log(num_nodes)/std::log(2.0)))
+      );
+      f_ = static_cast<uint16_t>(std::ceil(std::pow(num_nodes, 1.0/k_max_)));
+    }
```
nlslatt, line 143: @lifflander @PhilMiller Please let me know your thoughts on this in particular.

nlslatt:
@lifflander @PhilMiller This still needs to be reviewed.

tests/unit/collection/test_lb.extended.cc
```diff
@@ -93,12 +93,7 @@ void TestLoadBalancer::runTest() {
     fmt::print("Testing lb {}\n", lb_name);
   }
   if (lb_name.compare("GossipLB") == 0) {
-    auto nproc = vt::theContext()->getNumNodes();
-    int f = std::max(1, std::min(nproc-1, 3));
-    int k = nproc < 4 ? 2 : std::ceil(log(nproc)/log(f));
-    auto lb_args = fmt::format(
-      "f={} k={} trials=1 inform=1 ordering=0 rollback=0 targetpole=0", f, k
-    );
+    auto lb_args = fmt::format("ordering=Arbitrary rollback=false");
```
PhilMiller, line 96: Nitpick, this could just be the string literal now.

src/vt/vrt/collection/balance/gossiplb/gossip_enums.h
```diff
@@ -0,0 +1,281 @@
...
+#include "vt/vrt/collection/balance/read_lb.h"
+#include "vt/vrt/collection/balance/gossiplb/criterion.h"
+
+#include <string>
+#include <unordered_map>
+
+namespace vt { namespace vrt { namespace collection { namespace lb {
+
+template <typename T>
+struct EnumConverter {
```
PhilMiller, line 58: Other than the error messages, could this be a stand-alone utility class, separate from GossipLB?

src/vt/vrt/collection/balance/gossiplb/gossip_enums.h
```diff
@@ -0,0 +1,281 @@
...
+   * rounds = log(num_ranks)/log(fanout).
+   */
+  Log         = 2
+};
+
+}}}} /* end namespace vt::vrt::collection::lb */
+
+namespace std {
+
+using CriterionEnum = ::vt::vrt::collection::lb::CriterionEnum;
```
PhilMiller, line 234: I think it may not be technically allowed to define these names in `namespace std`, as opposed to fully qualifying them in the specializations of `std::hash`

src/vt/vrt/collection/balance/gossiplb/gossip_enums.h
nlslatt: Is there a particular place namespace where it makes sense to put it?  I'm poking around now.

PhilMiller: If there isn't a `vt::utility` namespace somewhere, there probably ought to be.

nlslatt: Making this stand-alone will be #1451.

src/vt/vrt/collection/balance/gossiplb/gossip_enums.h
nlslatt: Fixed for GossipLB, but not for other cases outside the scope of this PR.

nlslatt: Added #1452 to address other cases.

tests/unit/collection/test_lb.extended.cc
nlslatt: Addressed.

src/vt/vrt/collection/balance/gossiplb/gossip_enums.h
```diff
@@ -0,0 +1,271 @@
...
+      str_to_enum_[it.second] = it.first;
+    }
+  }
+
+  virtual ~EnumConverter() = default;
+
+  std::string getString(T e) const {
+    auto it = enum_to_str_.find(e);
+    if (it == enum_to_str_.end()) {
+      auto err = fmt::format(
```
lifflander, line 80: If we feel like being extra careful, we could write tests now for this abort since it maps to a exception now when running tests. I'm personally not super concerned because this logic looks straight forward to me, but code coverage is complaining...

src/vt/vrt/collection/balance/gossiplb/gossip_enums.h
```diff
@@ -0,0 +1,271 @@
...
+        "GossipLB EnumConverter: enum '{}' value '{}' was not found in the map",
+        enum_type_, e
+      );
+      vtAbort(err);
+    }
+    return it->second;
+  }
+
+  T getEnum(std::string s) const {
```
lifflander, line 89: We should pass `std::string` by `const &` here.

src/vt/vrt/collection/balance/gossiplb/gossip_enums.h
```diff
@@ -0,0 +1,271 @@
...
+#include "vt/config.h"
+#include "vt/vrt/collection/balance/read_lb.h"
+#include "vt/vrt/collection/balance/gossiplb/criterion.h"
+
+#include <string>
+#include <unordered_map>
+
+namespace vt { namespace vrt { namespace collection { namespace lb {
+
+template <typename T>
```
lifflander, line 57: I would like some very basic doxygen of what this class does. Especially if we move it to `vt/utils/*` as a general class that's usable.

src/vt/vrt/collection/balance/gossiplb/gossip_enums.h
```diff
@@ -0,0 +1,271 @@
...
+      auto err = fmt::format(
+        "GossipLB: LB argument '{}' value '{}' is not recognized",
+        par_name_, s
+      );
+      vtAbort(err);
+    }
+    return it->second;
+  }
+
+  T getFromSpec(balance::SpecEntry* spec, T def_value) const {
```
lifflander, line 101: I understand why this method is here, but I feel like it makes it quite a bit less general. If we are moving it to utils, we should probably provide this thin helper function elsewhere.

src/vt/vrt/collection/balance/gossiplb/gossip_enums.h
```diff
@@ -0,0 +1,271 @@
...
+};
+
+}}}} /* end namespace vt::vrt::collection::lb */
+
+namespace std {
+
+template <>
+struct hash<::vt::vrt::collection::lb::CriterionEnum> {
+  size_t operator()(::vt::vrt::collection::lb::CriterionEnum const& in) const {
+    return std::hash<uint8_t>()(static_cast<uint8_t>(in));
```
lifflander, line 237: I would prefer using `std::underlying_type<T>` here instead of `uint8_t`. Basically, if we change the type that holds the enum this code must be modified also.

lifflander (changes requested):

src/vt/vrt/collection/balance/gossiplb/gossip_enums.h
nlslatt: It looks to me like throwing on abort is not enabled by default.  Does this mean that the test suite will fail by default on a non-CI build?

src/vt/vrt/collection/balance/gossiplb/gossip_enums.h
nlslatt: Addressed.

src/vt/vrt/collection/balance/gossiplb/gossip_enums.h
nlslatt: Fixed.

src/vt/vrt/collection/balance/gossiplb/gossip_enums.h
JacobDomagala: We could add test, that runs only when 'throw_on_abort' is enabled:
```c++
#if vt_check_enabled(throw_on_abort)
  TEST_F(TestName, test_fail) {
    ASSERT_THROW(
      function_call(),
      std::runtime_error
    );
   }
#endif
```

src/vt/vrt/collection/balance/gossiplb/gossip_enums.h
nlslatt: This class was really meant specifically for handling LB args.  I should have named it differently.  It easily generalizes to use with other LBs, but not trivially to more general purposes.

I could split it into a general enum conversion class and one that extends it with LB-specific functionality, but the error handling will need to be a lot more complicated in order to retain the most meaningful error messages.  It might be better to:

- rename EnumConverter to something that indicates that it's customized for handling enum conversions for LB args
- move that code out of the GossipLB subdirectory to make it available to all LBs, updating the error messages to not refer specifically to GossipLB
- create a new utility by the original name that only includes the generic enum conversion functions

What do you think?

src/vt/vrt/collection/balance/gossiplb/gossip_enums.h
nlslatt: Thanks, @JacobDomagala!

nlslatt: Unfortunately, using this approach kills gtest on my Mac.  It's probably the same issue that makes death tests not work.

nlslatt: Either that or enabling throw_on_abort did not actually result in a thrown exception.  The code in the `#if` wasn't compiling it until I enabled it in cmake, so the feature should be correctly enabled.  Here's the output I'm getting:
```
[ RUN      ] TestGossipLB.test_enum_converter_fail
vt: [-1] 
vt: [-1]              Reason: GossipLB EnumConverter: enum 'DummyEnum' value '0' was not found in the map
vt: [-1]                Type: vtAbort() Invoked
vt: [-1]                Node: -1
vt: [-1]           Num Nodes: -1
vt: [-1]                File: src/vt/vrt/collection/balance/gossiplb/gossip_enums.h
vt: [-1]                Line: 90
vt: [-1]            Function: getString
vt: [-1]                Code: 1
vt: [-1]           Build SHA: 9764b403f30509d11769b4d00ecbae679a431775
vt: [-1]           Build Ref: refs/heads/1376-improve-lbargs-for-gossiplb
vt: [-1]         Description: heads/1376-improve-lbargs-for-gossiplb-0-g9764b403f3
vt: [-1]            GIT Repo: *dirty*
```

lifflander: Actually, maybe we should enable throw_on_abort for all tests regardless of whether we are using CI or not.

src/vt/vrt/collection/balance/gossiplb/gossip_enums.h
lifflander: I think it makes sense to make it a utility class for just LBs, including renaming and moving it to indicate that.

src/vt/vrt/collection/balance/gossiplb/gossip_enums.h
lifflander: @JacobDomagala Would this be easy to do, just add that define for test compilation?

JacobDomagala: @nlslatt  Make sure that this test is not a part of **nompi** test suite, as we need vt::runtime for proper handling of `vtAbort`. I've added the following test:

```c++
#if vt_check_enabled(throw_on_abort)
  TEST_F(TestLBStatsReader, test_fail) {
    using namespace vt::vrt::collection::lb;

    EnumConverter<InformTypeEnum> e("","", {});

    ASSERT_NO_THROW(
      e.getString(InformTypeEnum::AsyncInform)
    );
   }
#endif
```

and the output is:
image

nlslatt: This is a nompi test.  @lifflander and I are discussing how we need to handle this use case (`CollectiveAnyOps::abort`) the same as the runtime does.

lifflander: Refer to the issue I created for addressing this problem: https://github.com/DARMA-tasking/vt/issues/1456

src/vt/vrt/collection/balance/gossiplb/gossip_enums.h
nlslatt: I moved it and added tests of the spec-related functionality.

src/vt/vrt/collection/balance/gossiplb/gossip_enums.h
nlslatt: Code coverage is no longer complaining due to the addition of another test, but I can still add throw tests when #1456 is resolved.

src/vt/vrt/collection/balance/gossiplb/gossiplb.cc
```diff
@@ -76,57 +77,162 @@ bool GossipLB::isOverloaded(LoadType load) const {
...
+      // set the fanout based on the chosen rounds: f=p^(1/k)
+      k_max_ = spec->getOrDefault<int32_t>("rounds", k_max_);
+      f_ = static_cast<uint16_t>(std::ceil(std::pow(num_nodes, 1.0/k_max_)));
+    } else {
+      // set both the fanout and the rounds
+      k_max_ = static_cast<uint16_t>(std::max(1.0,
+        std::round(std::sqrt(std::log(num_nodes)/std::log(2.0)))
+      ));
+      f_ = static_cast<uint16_t>(std::ceil(std::pow(num_nodes, 1.0/k_max_)));
+    }
```
nlslatt, line 144: @lifflander @PhilMiller I'm still not sure if either of you looked over this chunk of code.  Do you think this is reasonable behavior?

src/vt/vrt/collection/balance/gossiplb/gossip_enums.h
JacobDomagala: #1456 is now merged

src/vt/vrt/collection/balance/gossiplb/gossiplb.cc
lifflander: This looks reasonable to me, based on what we discussed.

src/vt/vrt/collection/balance/lb_args_enum_converter.h
```diff
@@ -0,0 +1,153 @@
...
+   * using the reverse of the mapping provided at construction.
+   */
+  T getFromSpec(balance::SpecEntry* spec, T def_value) const {
+    std::string spec_value = spec->getOrDefault<std::string>(
+      par_name_, getString(def_value)
+    );
+    return getEnum(spec_value);
+  }
+
+  EnumToStrMap enum_to_str_;  //< 1 to 1 mapping of enumerated values to strings
```
lifflander, line 145: Should these be private?

lifflander (approved):

src/vt/vrt/collection/balance/lb_args_enum_converter.h
nlslatt: Yes.  Fixed.

nlslatt: Yes. Fixed.

src/vt/vrt/collection/balance/gossiplb/gossip_enums.h
nlslatt: Thanks! I added throw tests.

JacobDomagala (approved):

PR #1424, opened 2021-05-04
1394: Throw exception instead of MPI_Abort on vtAbort

### LINKED ISSUE #1394: Throw exception instead of MPI_Abort on vtAbort (at least for CI/testing with gtest)
To start, let's add a cmake flag that makes vtAbort throw an exception. When CI is running, let's make this the default. We might actually want this for when the test are running by default.

JacobDomagala:
Fixes #1394


cmake/configure_options.cmake
```diff
@@ -134,6 +134,15 @@ else()
     set(vt_feature_cmake_ci_build "0")
 endif()
 
+option(vt_throw_on_abort "Build VT with load balancing enabled" OFF)
```
lifflander, line 137: Looks like this is a bad copy-paste

lifflander (changes requested):

lifflander (approved):
Looks good to me!

PR #1426, opened 2021-05-04
1344 Avoid explicit message size in send() parameters by storing it on MsgSharedPtr

### LINKED ISSUE #1344: Have MsgSharedPtr capture the size of the message before erasing type
**What Needs to be Done?**

Originated from this comment: https://github.com/DARMA-tasking/vt/pull/1322#discussion_r599014853

Should simplify the active messenger quite a bit so the size doesn't have to be passed around everywhere. Need to make sure that `MsgSharedPtr<T>` is constructed from the correct non-erased type everywhere it is used

MikolajZuzek:
Fixes #1344

### DIFF
--- a/src/vt/messaging/active.cc
+++ b/src/vt/messaging/active.cc
@@ -165,6 +165,24 @@ ActiveMessenger::ActiveMessenger()
   vtAssertExpr(epoch_stack_.size() == 0);
 }
 
+trace::TraceEventIDType ActiveMessenger::makeTraceCreationSend(
+  HandlerType const handler, auto_registry::RegistryTypeEnum type,
+  ByteType serialized_msg_size, bool is_bcast
+) {
+  #if vt_check_enabled(trace_enabled)
+    trace::TraceEntryIDType ep = auto_registry::handlerTraceID(handler, type);
+    trace::TraceEventIDType event = trace::no_trace_event;
+    if (not is_bcast) {
+      event = theTrace()->messageCreation(ep, serialized_msg_size);
+    } else {
+      event = theTrace()->messageCreationBcast(ep, serialized_msg_size);
+    }
+    return event;
+  #else
+    return trace::no_trace_event;
+  #endif
+}
+
 void ActiveMessenger::packMsg(
   MessageType* msg, MsgSizeType size, void* ptr, MsgSizeType ptr_bytes
 ) {
@@ -456,8 +474,8 @@ EventType ActiveMessenger::doMessageSend(
       // auto cur_event = envelopeGetTraceEvent(msg->env);
       // if (cur_event == trace::no_trace_event) {
       auto event = makeTraceCreationSend(
-        base, handler, auto_registry::RegistryTypeEnum::RegGeneral,
-        is_bcast
+        handler, auto_registry::RegistryTypeEnum::RegGeneral,
+        base.size(), is_bcast
       );
       envelopeSetTraceEvent(msg->env, event);
     }
--- a/src/vt/messaging/active.cc
+++ b/src/vt/messaging/active.cc
@@ -180,7 +180,7 @@ void ActiveMessenger::packMsg(
 
 EventType ActiveMessenger::sendMsgBytesWithPut(
   NodeType const& dest, MsgSharedPtr<BaseMsgType> const& base,
-  MsgSizeType const& msg_size, TagType const& send_tag
+  TagType const& send_tag
 ) {
   auto msg = base.get();
   auto const& is_term = envelopeIsTerm(msg->env);
@@ -192,7 +192,7 @@ EventType ActiveMessenger::sendMsgBytesWithPut(
     vt_debug_print(
       normal, active,
       "sendMsgBytesWithPut: size={}, dest={}, is_put={}, is_put_packed={}\n",
-      msg_size, dest, print_bool(is_put), print_bool(is_put_packed)
+      base.size(), dest, print_bool(is_put), print_bool(is_put_packed)
     );
   }
 
@@ -201,7 +201,7 @@ EventType ActiveMessenger::sendMsgBytesWithPut(
     "Destination {} should != this node"
   );
 
-  MsgSizeType new_msg_size = msg_size;
+  MsgSizeType new_msg_size = base.size();
 
   if (is_put && !is_put_packed) {
     auto const& put_ptr = envelopeGetPutPtr(msg->env);
@@ -226,12 +226,12 @@ EventType ActiveMessenger::sendMsgBytesWithPut(
         verbose, active,
         "sendMsgBytesWithPut: (put) put_ptr={}, size:[msg={},put={},rem={}],"
         "dest={}, max_pack_size={}, direct_buf_pack={}\n",
-        put_ptr, msg_size, put_size, rem_size, dest, max_pack_direct_size,
+        put_ptr, base.size(), put_size, rem_size, dest, max_pack_direct_size,
         print_bool(direct_buf_pack)
       );
     }
     if (direct_buf_pack) {
-      packMsg(msg, msg_size, put_ptr, put_size);
+      packMsg(msg, base.size(), put_ptr, put_size);
       new_msg_size += put_size;
       envelopeSetPutTag(msg->env, PutPackedTag);
       setPackedPutType(msg->env);
@@ -422,7 +422,7 @@ EventType ActiveMessenger::sendMsgBytes(
 }
 
 EventType ActiveMessenger::doMessageSend(
-  MsgSharedPtr<BaseMsgType>& base, MsgSizeType msg_size
+  MsgSharedPtr<BaseMsgType>& base
 ) {
   auto const& send_tag = static_cast<MPI_TagType>(MPITag::ActiveMsgTag);
 
@@ -457,7 +457,7 @@ EventType ActiveMessenger::doMessageSend(
       // if (cur_event == trace::no_trace_event) {
       auto event = makeTraceCreationSend(
         base, handler, auto_registry::RegistryTypeEnum::RegGeneral,
-        msg_size, is_bcast
+        is_bcast
       );
       envelopeSetTraceEvent(msg->env, event);
     }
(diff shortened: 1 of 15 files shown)


PhilMiller:
Could we get a bit more description/comment on the `forceSize` uses? Everything else besides that looks great.

MikolajZuzek:
Thanks @PhilMiller, I've supplied comments at key change points - where we override default message size (also reduced `forceSize(newSize)` to constructor calls):
* [_vt/messaging/active.impl.h_](https://github.com/DARMA-tasking/vt/blob/1344-avoid-explicit-msg-size-in-send/src/vt/messaging/active.impl.h): we set the size explicitly provided in `ActiveMessenger::sendMsgSz()` and `ActiveMessenger::broadcastMsgSz()`;
* [_vt/messaging/active.cc_](https://github.com/DARMA-tasking/vt/blob/1344-avoid-explicit-msg-size-in-send/src/vt/messaging/active.cc): in `ActiveMessenger::finishPendingDataMsgAsyncRecv()`:
  * at start, we initialize message size with number of bytes detected by `MPI_Irecv()`,
  * at finish, we update the size - reducing it by the put size in case when `put_tag == PutPackedTag`.

There were other size updates in [_vt/messaging/active.cc_](https://github.com/DARMA-tasking/vt/blob/1344-avoid-explicit-msg-size-in-send/src/vt/messaging/active.cc) before `makeTraceCreationSend()` calls (compiled only with `vt_check_enabled(trace_enabled)`) that measured the message with `vt::serialization::MsgSizer<MsgT>::get()`. I don't quite follow why that was done: the new size obtained this way was only sent to `makeTraceCreationSend()` but not used in the followup calls (they operated on the original size). It seemed only like misleading the trace, so I removed those size updates - please let me know if I'm missing out something here.

src/vt/messaging/active.impl.h
```diff
@@ -226,8 +220,8 @@ ActiveMessenger::PendingSendType ActiveMessenger::sendMsgSz(
   ByteType msg_size,
   TagType tag
 ) {
-  MsgSharedPtr<MsgT> msgptr = msg.msg_;
-  return sendMsgImpl<MsgT>(dest, han, msgptr, msg_size, tag);
+  MsgSharedPtr<MsgT> msgptr(msg.msg_, msg_size); // Note: use explicitly proveded message size
```
PhilMiller, line 223: Typo 'provided'

PhilMiller: This is repeated in a few spots (probably copy/paste). Please grep and correct all of them.

src/vt/messaging/active.impl.h
```diff
@@ -106,15 +105,15 @@ void ActiveMessenger::setTagMessage(MsgT* msg, TagType tag) {
 template <typename MsgPtrT>
 trace::TraceEventIDType ActiveMessenger::makeTraceCreationSend(
   MsgPtrT msg, HandlerType const handler, auto_registry::RegistryTypeEnum type,
-  MsgSizeType msg_size, bool is_bcast
+  bool is_bcast
 ) {
   #if vt_check_enabled(trace_enabled)
     trace::TraceEntryIDType ep = auto_registry::handlerTraceID(handler, type);
     trace::TraceEventIDType event = trace::no_trace_event;
     if (not is_bcast) {
```
PhilMiller, line 113: As discussed in today's call, the serialization sizing call will be moved here, and the result used in place of `msg.size()`, since we want to record the amount of data associated with the message to be sent over the network (i.e. with all arrays or other data structures it may point to), not just the size of the message structure itself.

PhilMiller: `auto msg_size = vt::serialization::MsgSizer<MsgT>::get(msgPtr.get());`

PhilMiller: Actually, I want to backtrack on this. Note that the argument `msg` is never used in this function. Thus, we should preferentially remove that `msg` argument rather than `msg_size`, de-template this function, and let the caller pass in the size based on that expression.

MikolajZuzek: On b00e3bc87a8c1a4648d4cf648793bcc3ef22df99 I removed `msg` argument from `ActiveMessenger::makeTraceCreationSend()`, reverted `msg_size` renaming it to `serialized_msg_size` to clarify the semantics and de-templated the method (moving the implementation to _active.cc_).

src/vt/messaging/active.cc
```diff
@@ -455,9 +473,11 @@ EventType ActiveMessenger::doMessageSend(
...
       // auto cur_event = envelopeGetTraceEvent(msg->env);
       // if (cur_event == trace::no_trace_event) {
+      auto serialized_msg_size = vt::serialization::MsgSizer<BaseMsgType>::get(msg);
+        
       auto event = makeTraceCreationSend(
-        base, handler, auto_registry::RegistryTypeEnum::RegGeneral,
-        is_bcast
+        handler, auto_registry::RegistryTypeEnum::RegGeneral,
+        serialized_msg_size, is_bcast
       );
```
MikolajZuzek: **Note:** We were sending `msg_size` with `sizeof<MsgT>` (passed here to `doMessageSend()` by `PendingSend`) to `makeTraceCreationSend()`, now we send serialized size obtained with `MsgSizer` (see the original [_vt/messaging/active.cc_](https://github.com/DARMA-tasking/vt/blob/4d5eac4e8b714e0239310ac508ca1351d0668a7c/src/vt/messaging/active.cc#L458-L461)).

src/vt/vrt/collection/manager.impl.h
```diff
@@ -833,10 +833,9 @@ void CollectionManager::invokeMsgImpl(
   auto reg_type = HandlerManager::isHandlerMember(han) ?
     auto_registry::RegistryTypeEnum::RegVrtCollectionMember :
     auto_registry::RegistryTypeEnum::RegVrtCollection;
-  auto msg_size = vt::serialization::MsgSizer<MsgT>::get(msgPtr.get());
 
   trace_event = theMsg()->makeTraceCreationSend(
-    msgPtr, han, reg_type, msg_size, false
+    msgPtr, han, reg_type, false
```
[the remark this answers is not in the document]
lifflander: This is where things get tricky. There needs to be a SFINAE overload (which I believe exists, but needs to be modified) that checks for a `serialize` method. If that exists, it should call the checkpoint sizer to get the size (which it does already). However, **if the serializer is not available**, it should just get the size from the `MsgSharedPtr`, like is done here.

src/vt/messaging/active.cc
MikolajZuzek: I'm rolling this one back: calling `MsgSizer<BaseMsgType>` here on erased type would use `serialize()` from `BaseMsgType` instead of actual message type (since it's not virtual) and calculate incorrect size.  
If `makeTraceCreationSend()` is supposed to get correct serialized size for serializable messages here in `ActiveMessenger::doMessageSend()`, I suggest we create separate issue for that.

src/vt/vrt/collection/manager.impl.h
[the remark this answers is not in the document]
MikolajZuzek: It seems like for `vt::serialization::MsgSizer` that is called here in [vt/vrt/collection/manager.impl.h](https://github.com/DARMA-tasking/vt/blob/develop/src/vt/vrt/collection/manager.impl.h), we have specializations for serializable and non-serializable messages that get serialized size and `sizeof(MsgT)` respectively (see in [_vt/serialization/sizer.h_](https://github.com/DARMA-tasking/vt/blob/9cd17e9700a2155fb3f4371f8df56e7f7e71ea0b/src/vt/serialization/sizer.h)).

We could change `MsgSizer::get(MsgT* msg)` to take `MsgSharedPtr` instead of the raw pointer and return the actual size carried by it for non-serializable messages - instead of `sizeof(MsgT)`.

src/vt/messaging/active.cc
MikolajZuzek: There is no need to call `MsgSizer` here, as any serializable message would have been already handled and replaced with `SerialPayloadMsg` or `SerialEagerPayloadMsg`. Seems like no changes needed and no bug to be reported.
always-serializable

src/vt/messaging/active.impl.h
MikolajZuzek: Fixed

src/vt/vrt/collection/manager.impl.h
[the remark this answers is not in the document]
PhilMiller: I think that's a good catch, and probably actually independent of this PR. There's no real connection to `makeTraceCreationSend()`, except for the discussion of it leading you to discover this.

If you build and run your proposed test case on the code in `develop`, does it fail as expected? If so, please open an issue and start a PR with that test case added, and we'll address it separately.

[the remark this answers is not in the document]
MikolajZuzek: Created #1466 to address custom-sized message handling and #1471 to address message size logged with `makeTraceCreationSend()`. Let me know if you can think of any other issues.

PhilMiller (approved):

PhilMiller (changes requested):
the boolean flags noted at https://github.com/DARMA-tasking/vt/pull/1473#pullrequestreview-684259541

src/vt/vrt/collection/manager.impl.h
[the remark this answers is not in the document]
MikolajZuzek: Fixed accidentally changed `is_bcast` flags in `makeTraceCreationSend()` calls, see https://github.com/DARMA-tasking/vt/pull/1473#discussion_r652007079

MikolajZuzek:
> 
> 
> the boolean flags noted at [#1473 (review)](https://github.com/DARMA-tasking/vt/pull/1473#pullrequestreview-684259541)

Fixed!

PhilMiller (approved):

PhilMiller:
> **PR tests (clang-9, ubuntu, mpich)**
> 
> Build for [6e7944c](https://github.com/DARMA-tasking/vt/commit/6e7944cb78f2c8985ae16b2d663b7b0263f670c3)
> 
> ```
> 
> 
> The following tests FAILED:
>   155 - vt:*/TestNodeStatsDumper.test_node_stats_dumping_with_interval/*_proc_2 (Timeout)
> ```
> 
> [Build log](https://dev.azure.com/DARMA-tasking/DARMA/_build/results?buildId=18322&view=logs&j=3dc8fd7e-4368-5a92-293e-d53cefc8c4b3&t=28db5144-7e5d-5c90-2820-8676d630d9d2)

This is a sporadic failure, unrelated to this PR. Thus, I'm going ahead with merging.

PR #1432, opened 2021-05-05
1423 Make message ref count limit configurable, print max when enabled

### LINKED ISSUE #1423: Message ref-count limit triggers in simple cases
**Describe the bug**
When running `lb_iter`, I'm getting this assertion when I set the number of elements high enough:

```
vt: [0]              Reason: Bad ref-count on message ref-increment. Message ref-count must never be negative and cannnot exceed limit (100).
vt: [0]    Assertion failed: (envp->ref >= 0 and envp->ref < 100)
vt: [0]           Num Nodes: 2
vt: [0]                File: /Users/nlslatt/work/vt/vt-tmp/src/vt/messaging/envelope/envelope_ref.impl.h
vt: [0]                Line: 68
vt: [0]            Function: envelopeRef
```

I've seen this with GossipLB and RandomLB.  I haven't tried other LBs.

**To Reproduce**

`mpirun -np 2 ./examples/collection/lb_iter 200 1 5 --vt_lb --vt_lb_interval=1 --vt_lb_name=RandomLB`

or

`mpirun -np 2 ./examples/collection/lb_iter 200 1 5 --vt_lb --vt_lb_interval=1 --vt_lb_name=GossipLB`

Reducing the number of elements from 200 to 100 does not trigger the assertion.  This is on my Mac.

lifflander:
Fixes #1423

@nmm0 Let me know if this satisfies your needs.

Added two new options:
- `--vt_vt_msg_ref_count_max` which allows the user to set the ref count limit where the assert triggers
- `--vt_print_max_ref_count`, which prints the message type every time a new maximum ref count is observed

The second looks like this when enabled:

Screen Shot 2021-05-05 at 16 18 22

### DIFF
--- a/src/vt/messaging/message/refs.impl.h
+++ b/src/vt/messaging/message/refs.impl.h
@@ -53,6 +53,33 @@ namespace vt {
 
 template <typename MsgT>
 void messageRef(MsgT* msg) {
+
+  vtAssertInfo(
+    envelopeGetRef(msg->env) >= 0 and
+    (theConfig()->vt_msg_ref_count_max == 0 or
+     envelopeGetRef(msg->env) + 1 < theConfig()->vt_msg_ref_count_max),
+    "Bad ref-count on message ref-increment. "
+    "Message ref-count must never be negative and cannot exceed limit.",
+    static_cast<RefType>(envelopeGetRef(msg->env)),
+    theConfig()->vt_msg_ref_count_max
+  );
+
+  if (theConfig()->vt_print_max_ref_count) {
+    auto const cur_max = theContext()->getMaxObservedRefCount();
+    auto const ref = envelopeGetRef(msg->env);
+    if (ref + 1 > cur_max) {
+      theContext()->setMaxObservedRefCount(ref + 1);
+
+      auto debug_str = fmt::format(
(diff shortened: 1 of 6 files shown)


nmm0 (approved):
Looks good, working well in my debugging

src/vt/messaging/message/refs.impl.h
```diff
@@ -53,6 +53,33 @@ namespace vt {
...
+      auto debug_str = fmt::format(
+        "Message {}{}{} reached new max ref count {}{}{}\n",
+        debug::green(), typeid(MsgT).name(), debug::reset(),
+        debug::magenta(), ref+1, debug::reset()
+      );
+
+      vt_print(active, debug_str);
+    }
+  }
+
```
PhilMiller, line 82: This should be compiled out in release builds, at the very least. This is a lot of code to have on a hot path.

PhilMiller:
If I understand the discussion that led up to this correctly, it's something of a band-aid, since the code is incrementing the ref-count per-recipient in broadcast deliveries. I think it's worth thinking harder about the issue of encoding "is this message ref-count value reasonable?" before adopting something like this.

PR #1438, opened 2021-05-14
1294: Add performance tests to CI 

### LINKED ISSUE #1294: Develop a set of performance tests that we can graph to track VT performance
**What Needs to be Done?**

Let's discuss benchmarks we would like to see as part of a suite of performance tests that we run on VT develop after each merge

JacobDomagala:
Fixes #1294 

## Changes:
- Make CMake to actually compile performance tests that reside under `vt/tests/perf/` directory
- Add labels to tests, to distinguish unit tests from performance ( `ctest -L unit_test` will run unit tests, while `ctest -L perf_test` will run the performance )
- Add initial test_harness for performance tests with simple usage like:
```c++
#include <vt/tests/common/test_harness.h>

using namespace vt::tests::perf::common;

struct MyTest : PerfTestHarness {};

VT_PERF_TEST(MyTest, test_case_name){
  // Test code
}

VT_PERF_TEST_MAIN()
```
**Note:** With the current implementation, it's limited to one test per file, meaning you can't have multiple `VT_PERF_TEST` in a single test file. 
- By default, the test function (created with `VT_PERF_TEST`) will be run 50 times (this value can be changed, using the runtime flag `--vt_perf_num_runs`)

### Example 
```c++

#include "common/test_harness.h"

using namespace vt::tests::perf::common;

std::vector<int32_t*> ptrs;

void allocateAndTouch(std::size_t num_bytes) {
  auto ptr = new int32_t[num_bytes / 4];
  for (std::size_t i = 0; i < num_bytes / 4; i++) {
    ptr[i] = 20 * i + 23;
  }
  ptrs.push_back(ptr);
}

void deallocate() {
  if (ptrs.size() > 0) {
    auto ptr = ptrs.back();
    ptrs.pop_back();
    delete[] ptr;
  }
}

struct MyTest : PerfTestHarness { };

VT_PERF_TEST(MyTest, test_name) {
  constexpr auto num_iters = 32;

  StartTimer("Allocate");
  for (int i = 0; i < num_iters; i++) {
    allocateAndTouch(1024 * 1024 * 64);
    GetMemoryUsage();
  }
  StopTimer("Allocate");

  StartTimer("Deallocate");
  for (int i = 0; i < num_iters; i++) {
    deallocate();
    GetMemoryUsage();
  }
  StopTimer("Deallocate");
}

VT_PERF_TEST_MAIN()
```
### Command:
`mpirun -n 2 /build/vt/tests/example` or `ctest -L perf_test`

### Output:
image

## Generating graphs locally
Graphs can be generated locally using [Build Stats](https://github.com/DARMA-tasking/build-stats/pull/10) repository. 
1. Run performance test with `--vt_perf_gen_file` runtime flag, which will generate two CSV files (one for memory usage and one containing timers data)
2. (Optionally) Go to **Build Stats** directory and create some temporary folder, where the graphs will reside.
3. Run Build Stats' python script `generate_graph.py` like so: 
`python generate_graph.py -time /path/to/file/test_name_time.csv -mem /path/to/file/test_name_mem.csv` 

This will generate:
- time execution graph for 'test_name' (entire function created with **VT_PERF_TEST**)
- separate graph for each unique timer (iterative timers, i.e. 'xxx Bytes' or 'xxx Phase' will be aggregated into single graph)
- memory usage for 'test_name'

4. The above was for the new tests that weren't run on CI before, or if you don't want to compare the execution time for given test with previous CI runs. Otherwise you should run Build Stats' python script `generate_graph.py`  with additional flag `-wiki` that specifies the path to [vt.wiki](https://github.com/DARMA-tasking/vt.wiki.git) repository (all resutls of CI runs are stored there, so in order to graph historic data, we need path to it)

Example:
`python generate_graph.py -time /path/test_name_time.csv -mem /path/test_name_mem.csv -wiki /some/path/vt.wiki`

This will generate:
- time execution graph for 'test_name' (including historic data)
- separate graph for each unique timer (same as before)
- memory usage (same as before)

## Markdown results for perf-tests

For changes on CI side, see [this Pull Request](https://github.com/DARMA-tasking/build-stats/pull/10 )


tests/perf/common/scoped_timer.h
```diff
@@ -0,0 +1,86 @@
...
+#include <chrono>
+
+namespace vt { namespace tests { namespace perf { namespace common {
+
+using TimePoint = std::chrono::time_point<std::chrono::steady_clock>;
+using DurationMicro = std::chrono::duration<double, std::micro>;
+using DurationMilli = std::chrono::duration<double, std::milli>;
+using DurationSec = std::chrono::duration<double>;
+
+struct StopWatch {
```
lifflander, line 58: I like the direction this is headed. I wonder if this might even be more general purpose and useful in other places in the system. Maybe in src/vt/utils?

lifflander: Or scr/vt/timing

JacobDomagala:
### Node 0
[](https://github.com/DARMA-tasking/vt/wiki/perf_tests/flame_heaptrack_jacobi_alloc_count_0.svg)
[](https://github.com/DARMA-tasking/vt/wiki/perf_tests/flame_heaptrack_jacobi_alloc_size_0.svg)

### Node 1 
[](https://github.com/DARMA-tasking/vt/wiki/perf_tests/flame_heaptrack_jacobi_alloc_count_1.svg)
[](https://github.com/DARMA-tasking/vt/wiki/perf_tests/flame_heaptrack_jacobi_alloc_size_1.svg)

lifflander:
@JacobDomagala I really like your description at the top of this page. When we merge this in, can you add this info about the performance test harness/etc. to the wiki?

lifflander:
Is there a easy way to manually run this whole workflow (with the build-stats graph generation) and generate the graphs/data locally during development?

ci/test_cpp.sh
```diff
@@ -9,7 +9,8 @@ export VT=${source_dir}
 export VT_BUILD=${build_dir}/vt
 pushd "$VT_BUILD"
 
-ctest --output-on-failure | tee cmake-output.log
+# Run unit tests only (use label 'unit_test')
+ctest --output-on-failure -L unit_test | tee cmake-output.log
```
lifflander, line 13: Does this also run the examples? Do we need to put those in a category also for consistency?

src/vt/epoch/epoch_window.cc
```diff
@@ -85,7 +85,8 @@ EpochWindow::EpochWindow(EpochType epoch) {
   // All epochs in a given window start out terminated (thus, reusable).
   terminated_epochs_.insertInterval(interval);
 
-  fmt::print(
+  vt_print(
```
lifflander, line 88: Good catch. I accidentally committed this on develop and have been wanting to fix it!

tests/CMakeLists.txt
```diff
@@ -200,19 +206,51 @@ endforeach()
...
+
 foreach(PERF_TEST ${PROJECT_PERF_TESTS})
-  add_executable(${PERF_TEST} ${PROJECT_TEST_PERF_DIR}/${PERF_TEST}.cc)
-  add_dependencies(perf_tests ${PERF_TEST})
+  if (${PERF_TEST} IN_LIST EXCLUDED_PERF_TEST_FILES)
+    message(STATUS "Test file ${PERF_TEST} is present on EXCLUDED_PERF_TEST_FILES list. Skipping!")
+    continue()
+  endif()
+
+  GET_FILENAME_COMPONENT(
```
lifflander, line 224: Prefer lower case here on `GET_FILENAME_COMPONENT` for consistency with other cmake function calls.

lifflander, line 224:
```suggestion
  get_filename_component(
```

tests/perf/comm_cost_curve.cc
```diff
@@ -41,8 +41,8 @@
 //@HEADER
 */
 
-
-#include <vt/transport.h>
```
lifflander, line 45: We may want to disable this performance test. It was never completed and originally intended to use to build communication weights as inputs to LBAF. I'm not sure of there there is any use for it now. Opinions from anyone?

tests/perf/common/test_harness.cc
```diff
@@ -0,0 +1,393 @@
...
+/**
+ * \brief Prints memory usage in human readable format
+ */
+static std::string GetFormattedMemUsage(std::size_t memory) {
+  auto const& best_mem = util::memory::getBestMemoryUnit(memory);
+
+  return fmt::format(
+    "{}",
+    debug::emph(
+      fmt::format("{:.6g}{}", std::get<1>(best_mem), std::get<0>(best_mem))));
```
lifflander, line 69: Small whitespace change here to follow style guidelines (wrap parans to align with open).
```suggestion
      fmt::format("{:.6g}{}", std::get<1>(best_mem), std::get<0>(best_mem))
    )
  );
```

tests/perf/common/test_harness.cc
```diff
@@ -0,0 +1,393 @@
...
+ *  1. Final test result
+ *  2. Test result for given test run
+ *  3. Number of test runs
+ *
+ * \return Vector with final test results (mean/min/max)
+ */
+template <typename InputT, typename OutputT>
+std::vector<OutputT> ProcessInput(
+  std::vector<std::vector<InputT>> const& input_vec,
+  std::function<void(OutputT&, InputT const&, std::size_t const)> populate) {
```
lifflander, line 91: Another style suggestion:
```suggestion
  std::function<void(OutputT&, InputT const&, std::size_t const)> populate
) {
```

tests/perf/common/test_harness.h
```diff
@@ -0,0 +1,193 @@
...
+#include <string>
+#include <vector>
+
+namespace vt { namespace tests { namespace perf { namespace common {
+
+struct TestMsg;
+
+template <typename T>
+struct TestResultHolder{
+  T mean_ = {};
```
lifflander, line 62: I think we should add standard deviation or variance to these stats.

tests/perf/common/test_harness.h
```diff
@@ -0,0 +1,193 @@
...
+   */
+  static void SpinScheduler();
+
+  /**
+   * \brief Helper function used for tracking memory usage.
+   * Should be called each iteration. Uses \c StatM to track memory usage
+   */
+  void GetMemoryUsage();
+
+  private:
```
lifflander, line 163:
```suggestion
private:
```

tests/perf/common/test_harness.h
```diff
@@ -0,0 +1,193 @@
...
+  std::string OutputMemoryUse() const;
+  std::string OutputTimeResults() const;
+
+  protected:
```
lifflander, line 167:
```suggestion
protected:
```

tests/perf/common/test_harness.h
```diff
@@ -0,0 +1,193 @@
...
+  bool gen_file_ = false;
+  bool verbose_ = false;
+  uint32_t num_runs_ = 50;
+  uint32_t current_run_ = 0;
+  std::vector<char*> custom_args_ = {};
+
+  NodeType my_node_ = {};
```
lifflander, line 174: Initialize these with the type sentinels:
```suggestion
  NodeType my_node_ = uninitialized_destination;
```

tests/perf/common/test_harness.h
```diff
@@ -0,0 +1,193 @@
...
+  NodeType num_nodes_ = {};
```
lifflander, line 175:
```suggestion
  NodeType num_nodes_ = uninitialized_destination;
```

tests/perf/common/test_harness.h
```diff
@@ -0,0 +1,193 @@
...
+  // Memory usage (in bytes) per iteration
+  MemoryUsage memory_use_ = {};
+  vt::util::memory::StatM mem_tracker_ = {};
+
+  // Local (per node) timings.
+  TestResults timings_ = {};
+  std::unordered_map<std::string, StopWatch> timers_ = {};
+
+  // Combined timings from all nodes, that are stored on the root node
+  static CombinedResults combined_timings_;
```
lifflander, line 187: It's preferable to avoid `static` variables here if possible. I think you might be able to turn this class into an VT objgroup, and then make the handlers non-static.

lifflander: You can look at a simple example here: https://github.com/DARMA-tasking/vt/blob/develop/examples/hello_world/objgroup.cc

You can allocate the harness yourself and then call `vt::theObjGroup()->makeCollective(this)` in the constructor for `PerfTestHarness`

tests/perf/common/timers.h
```diff
@@ -0,0 +1,103 @@
...
+  template <typename Duration = DurationMilli>
+  TimeType Stop() {
+    auto const now = std::chrono::steady_clock::now();
+    auto const delta = Duration{now - cur_time_};
+    cur_time_ = now;
+
+    return delta.count();
+  }
+
+  private:
```
lifflander, line 97:
```suggestion
private:
```

tests/perf/common/test_harness_macros.h
```diff
@@ -0,0 +1,123 @@
...
+                                                                               \
+      TestType test;                                                           \
+      test.Initialize(argc, argv);                                             \
+      auto const num_runs = test.GetNumRuns();                                 \
+                                                                               \
+      StopWatch timer;                                                         \
+                                                                               \
+      if (rank == 0) {                                                         \
+        fmt::print(                                                            \
+          "\033[1mRunning test:\033[00m \033[32m{}\033[00m (Number of runs = " \
```
lifflander, line 95: Maybe you can use the color functions here instead of using the control characters yourself.

lifflander (changes requested):
Overall this looks great. I have some suggestions for improvement along with some style suggestions.

tests/perf/comm_cost_curve.cc
JacobDomagala: Both `comm_cost_curve` and `memory_checker` are disabled (see `EXCLUDED_PERF_TEST_FILES` vt/tests/CMakeListst.txt)
https://github.com/DARMA-tasking/vt/blob/71efd92973e1f947f9980e0f058dd972fccf5a48/tests/CMakeLists.txt#L209-L214

tests/CMakeLists.txt
JacobDomagala: Updated all occurrences of this function call in our codebase, as it was mixture of uppercase and lowercase.

ci/test_cpp.sh
JacobDomagala: Good point, examples were skipped. I've added the label for them, and included them in test_cpp.sh

tests/perf/common/test_harness_macros.h
JacobDomagala: Good point

tests/CMakeLists.txt
lifflander: Great, thanks!

nlslatt:
This is a great addition! The description in this PR is very useful.

Some of the files have lines far exceeding 80 columns. Can you please reformat?

lifflander (approved):
It looks like you have addressed my comments. This looks good to me!

nlslatt (approved):
Looks good

PR #1440, opened 2021-05-18
1437: Label all uses of runInEpochCollective in source code

### LINKED ISSUE #1437: Label all epochs created by vt
**What Needs to be Done?**

Now that we can pass a string label to `runInEpochCollective` pass descriptive label strings to all those callsites in VT. The key ones that are very useful (that we added manually during debugging) is in `src/vt/vrt/collection/balance/lb_invoke/lb_manager.cc` in the `runLB` function where several collective epochs are created for running the LB strategy, migrating, etc.

JacobDomagala:
Fixes #1437

### DIFF
--- a/src/vt/phase/phase_manager.cc
+++ b/src/vt/phase/phase_manager.cc
@@ -207,7 +207,7 @@ void PhaseManager::runHooks(PhaseHook type) {
     if (iter != collective_hooks_.end()) {
       // note, this second is a map, so they are ordered across nodes
       for (auto&& fn : iter->second) {
-        runInEpochCollective([=]{ fn.second(); });
+        runInEpochCollective("PhaseManager::runHooks", [=]{ fn.second(); });
       }
     }
   }
--- a/src/vt/phase/phase_manager.cc
+++ b/src/vt/phase/phase_manager.cc
@@ -196,7 +196,7 @@ void PhaseManager::runHooks(PhaseHook type) {
     auto iter = rooted_hooks_.find(type_bits);
     if (iter != rooted_hooks_.end()) {
       for (auto&& fn : iter->second) {
-        runInEpochRooted([=]{ fn.second(); });
+        runInEpochRooted("PhaseManager::runHooks", [=]{ fn.second(); });
       }
     }
   }
--- a/src/vt/scheduler/scheduler.impl.h
+++ b/src/vt/scheduler/scheduler.impl.h
@@ -75,8 +75,7 @@ void runInEpochCollective(std::string const& label, Callable&& fn) {
 
 template <typename Callable>
 void runInEpochRooted(Callable&& fn) {
-  auto ep = theTerm()->makeEpochRooted();
-  runInEpoch(ep, std::forward<Callable>(fn));
+  runInEpochRooted("", std::forward<Callable>(fn));
 }
 
 template <typename Callable>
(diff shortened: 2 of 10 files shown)


src/vt/phase/phase_manager.cc
```diff
@@ -207,7 +207,7 @@ void PhaseManager::runHooks(PhaseHook type) {
     if (iter != collective_hooks_.end()) {
       // note, this second is a map, so they are ordered across nodes
       for (auto&& fn : iter->second) {
-        runInEpochCollective([=]{ fn.second(); });
```
PhilMiller, line 210: Do we want to label the call to `runInEpochRooted` just above?

src/vt/scheduler/scheduler.impl.h
```diff
@@ -64,8 +64,7 @@ void runInEpoch(EpochType ep, Callable&& fn) {
 
 template <typename Callable>
 void runInEpochCollective(Callable&& fn) {
-  auto ep = theTerm()->makeEpochCollective();
-  runInEpoch(ep, std::forward<Callable>(fn));
+  runInEpochCollective("", std::forward<Callable>(fn));
```
PhilMiller, line 67: Could this be an explicit `"UNLABELED"`?

PhilMiller:
Looks good overall.

I'm mildly surprised that this small set of calls is the extent of unlabeled usage. How much more is there if you address `runInEpochRooted` too?

JacobDomagala:
> Looks good overall.
> 
> I'm mildly surprised that this small set of calls is the extent of unlabeled usage. How much more is there if you address `runInEpochRooted` too?

@PhilMiller  Looks like this was the only `runInEpochRooted` call. I guess most of the collective epochs are handled via `PhaseManager::nextPhaseCollective()`

src/vt/scheduler/scheduler.impl.h
JacobDomagala: Updated. Also applied this change to rooted epoch

src/vt/scheduler/scheduler.impl.h
```diff
@@ -64,8 +64,7 @@ void runInEpoch(EpochType ep, Callable&& fn) {
 
 template <typename Callable>
 void runInEpochCollective(Callable&& fn) {
-  auto ep = theTerm()->makeEpochCollective();
-  runInEpoch(ep, std::forward<Callable>(fn));
+  runInEpochCollective("UNLABELED", std::forward<Callable>(fn));
```
lifflander, line 67: It would be cool if we could automatically extract the function that calls this.. but I don't think that's easy. Maybe for the longer term if we made this a macro or something... @PhilMiller What do you think?

lifflander (approved):
Looks good, if you can expand to the other `TerminationDetector` calls that create epochs that would be great!

src/vt/scheduler/scheduler.impl.h
PhilMiller: If it were a macro, it could embed `__FUNC__` in the call to the underlying function, yeah. That seems unappealing to me. If a caller knows that it wants to do that, it can just write `__FUNC__` as the argument and be lazy.

We could be tricksy and use our backtrace capability to look up the stack, if we really wanted to.

I think any of these approaches is overkill for now.

lifflander: Ok, I'm fine leaving this alone for now. I was initially thinking `__FUNCTION__` or whatever, but the user can type that if they want.

PhilMiller (approved):

Issue #1425, opened 2021-05-04
prepare `1.0.2` release


cz4rs:
**What Needs to be Done?**
Prepare `1.0.2` release.

- check the [issues included](https://github.com/DARMA-tasking/vt/issues?q=is%3Aissue+label%3A1.0.2) and make sure that they are present both on release and on `develop` branch
- remember to update wiki: https://github.com/DARMA-tasking/vt/wiki#releases

PR #1441, opened 2021-05-18
1425: prepare 1.0.2 release

cz4rs:
Preparation for #1425.


cz4rs:
Failing test:
```
/vt/tests/unit/collection/test_lb_extended.cc:199: Failure
Expected equality of these values:
  readPhases.size()
    Which is: 0
  num_phases
    Which is: 10
```

Related code:
https://github.com/DARMA-tasking/vt/blob/5d87d0829eb72bcf6d800e7a7f2de70f1111dd95/src/vt/vrt/collection/balance/lb_invoke/invoke.cc#L240-L242

cz4rs:
DARMA weekly conclusion: `vtAbort` when user tries to dump stats, since it's not supported on release (source code modification required)

cz4rs:
Output for 5feaf96:
```
vt: Runtime Configuration:
vt: 	Option: flag --vt_sched_num_progress on: Running MPI progress 2 times each invocation
vt: 	Option: flag --vt_sched_progress_han on: Running MPI progress function at least every 0 handler(s) executed
vt: 	Option: flag --vt_lb_stats on: Load balancing statistics collection
vt: [0] ------------------------------------------------------------------------------------------------------------------------
vt: [0] ------------------------------------------- Runtime Error: System Aborting! --------------------------------------------
vt: [0] ------------------------------------------------ Fatal Error on Node 0 -------------------------------------------------
vt: [0] ------------------------------------------------------------------------------------------------------------------------
vt: [0] 
vt: [0] Message: FATAL ERROR: --vt_lb_stats was used, but load balancing statistics collection is disabled in this release.

vt: [0] 
vt: [0] ------------------------------------------------------------------------------------------------------------------------

```

tests/unit/collection/test_lb_extended.cc
```diff
@@ -0,0 +1,286 @@
...
+#   if vt_check_enabled(zoltan)
+    , "ZoltanLB"
+#   endif
+);
+
+INSTANTIATE_TEST_SUITE_P(
+  LoadBalancerExplode, TestLoadBalancer, balancers
+);
+
+/**
```
lifflander, line 141: I think we should delete this rather than just commenting it out.

lifflander (approved):
This looks good to me. I think we should just remove that code that is commented out.

PR #1448, opened 2021-05-20
1447 Fix bug in epoch transfer for serialized messages

### LINKED ISSUE #1447: Epoch fails to transfer on locally delivered serialized message
**Describe the bug**
When we made changes to enqueue runnables, serialized messages that are delivered on the same rank as sent from, do not get their epoch set in the envelope. This causes early termination of the epoch. Inside `sendMsgSerializableImpl`, we need to call `setupEpochMsg` to ensure it gets sets properly.

https://github.com/DARMA-tasking/vt/blob/1a14c551f1c306502247ae5236b7c832ece985d4/src/vt/messaging/active.impl.h#L125-L158

lifflander:
Fixes #1447 

 TODO:
- @cz4rs: Please write a test that exposes this bug so we can confirm it fixes it.
- We have tested bvh/NimbleSM and this fix seems to work.

### DIFF
--- a/src/vt/messaging/active.impl.h
+++ b/src/vt/messaging/active.impl.h
@@ -143,19 +143,17 @@ ActiveMessenger::PendingSendType ActiveMessenger::sendMsgSerializableImpl(
     "Tagged messages serialization not implemented."
   );
 
-  MsgT* rawMsg = msg.get();
-  setupEpochMsg(rawMsg);
-
   // Original message is locked even though it is not the actual message sent.
   // This is for consistency with sending non-serialized messages.
   envelopeSetIsLocked(msg->env, true);
 
+  MsgT* msgp = msg.get();
   if (dest == broadcast_dest) {
     return SerializedMessenger::broadcastSerialMsg<MsgT>(
-      rawMsg, han, envelopeGetDeliverBcast(rawMsg->env)
+      msgp, han, envelopeGetDeliverBcast(msgp->env)
     );
   } else {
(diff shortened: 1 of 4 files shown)


nmm0:
This fixes the test case in bvh; currently checking to see if this impacts the LB problem we've been seeing with NimbleSM. Some of the symptoms are similar so I'm hopeful.

cz4rs:
794d925 temporarily reverts the fix for the sake of reproduction, 072a204 adds a simple test with serialized message (going from node `0` to node `0`). The added test definitely calls `sendMsgSerializableImpl` along the way, but I cannot observe any misbehavior. I need a bit of help here:
- is this even a good direction for reproduction?
- is there any additional condition that must be met (e.g. more work must be performed)?

JacobDomagala:
This happens only for serialized messages that have `size > serialized_msg_eager_size`. For smaller messages, `sendSerialMsg` will eventually end up calling the non-serialized `sendMsgImpl`, which sets the epoch correctly. 
Also, looks like it only happens on a single node execution.

JacobDomagala:
With https://github.com/DARMA-tasking/vt/pull/1448/commits/9d1f383f2224d1c935eb8105543c1b36f3566805 added, the bug can be reproduced by commenting out the line 147 in `active.imlp.h`
https://github.com/DARMA-tasking/vt/blob/0d92ce31c2ecdaacf112b14db5015eaa8f5c2706/src/vt/messaging/active.impl.h#L146-L148

PhilMiller (approved):

PR #1455, opened 2021-05-27
1454 Refactor epoch management into a component, cleanup, and bug fixes

### LINKED ISSUE #1454: Refactor epoch management code into component
**What Needs to be Done?**

I want to strip out the parts of the epoch scope PR that just do the refactoring from PR #1052, issue #1053. I'm trying to get all major refactoring of epoch stuff merged before I start implementing the code for issue #1442.

lifflander:
Fixes #1454 

Addresses several issues:
- Refactor epoch management
- Fix a bug in epochs
- Lots of general cleanup of the interface


src/vt/termination/termination.cc
```diff
@@ -820,7 +804,7 @@ TermStatusEnum TerminationDetector::testEpochTerminated(EpochType epoch) {
   TermStatusEnum status = TermStatusEnum::Pending;
   auto const& is_rooted_epoch = epoch::EpochManip::isRooted(epoch);
 
-  if (getWindow(epoch)->isTerminated(epoch)) {
+  if (theEpoch()->getTerminatedWindow(epoch)->isTerminated(epoch)) {
```
PhilMiller, line 807: Given this repeated motif, there should probably be a method `bool t = theEpoch()->isTerminated(epoch);`, and the windows can be less exposed to code outside the component.

Not necessary to move this PR forward, but would be appreciated down the line.

src/vt/utils/adt/ranged_counter.h
```diff
@@ -0,0 +1,156 @@
...
+    operator--();
+    return copy;
+  }
+
+  /**
+   * \brief Increment the value, keeping it within the range; wraps around if
+   * \c cur_ == hi_
+   */
+  inline void increment() {
+    if (cur_ < hi_) {
```
PhilMiller, line 122: It would be nice if `increment` and `decrement` were implemented in the same terms as the comment, with equality comparisons to the bounds. That would make it much easier to see by inspection that the logic is fine with values like `INT_MAX`

PhilMiller (approved):

PR #1457, opened 2021-05-27
1456: Change vt_throw_on_abort to runtime flag

### LINKED ISSUE #1456: Throw exeception when runtime is not active
**Describe the bug**
Right now, we aren't throwing if the runtime is not active, which occurs in the nompi tests.

Also, if it doesn't already work this way, we want to enable throwing on non-CI builds for all tests. But, we don't want to compile two different versions of the VT code. So this might be a bit tricky.

```C++
template <runtime::RuntimeInstType instance>
void CollectiveAnyOps<instance>::abort(
  std::string const str, ErrorCodeType const code
) {
  auto tls_rt = curRT;
  auto myrt = tls_rt ? tls_rt : ::vt::rt;
  if (myrt) {
    //--- Try to flush most of the traces before aborting
    myrt->theTrace->cleanupTracesFile();
#endif
    myrt->abort(str, code);
  } else {
    std::_Exit(code);
  }

}
```

JacobDomagala:
Fixes #1456


lifflander (approved):
This looks great to me!

nlslatt:
@JacobDomagala Should this line have been removed from `src/vt/configs/features/features_defines.h`?

```
#define vt_feature_throw_on_abort      0 || vt_feature_cmake_throw_on_abort
```

JacobDomagala:
@nlslatt Ah yes, I forgot to remove it. Could you include this change in https://github.com/DARMA-tasking/vt/pull/1422? I don't think it's worth creating separate issue for that.

nlslatt:
> @nlslatt Ah yes, I forgot to remove it. Could you include this change in #1422? I don't think it's worth creating separate issue for that.

Included.

PR #1462, opened 2021-06-02
1442 Make EpochType a strong/named type

### LINKED ISSUE #1442: Make EpochType a strong type
**What Needs to be Done?**

Try to make it a strong type to make pop/push easy to manage.

lifflander:
Fixes #1442


src/vt/termination/termination.cc
```diff
@@ -984,7 +984,7 @@ void TerminationDetector::initializeRootedWaveEpoch(
     terse, term,
     "makeEpochRootedWave: root={}, epoch={:x}, successor={:x},"
     "label={}\n",
-    theContext()->getNode(), epoch, successor, label
+    theContext()->getNode(), epoch, (EpochType)successor, label
```
lifflander, line 987: I should be able to remove these now that I have fmt working.

src/vt/utils/strong/strong_type.h
```diff
@@ -0,0 +1,271 @@
...
+  /**
+   * \brief Pre-decrement operator
+   */
+  ThisType& operator--() {
+    v_--;
+    return *this;
+  }
+
+  /**
+   * \brief Pre-decrement operator
```
PhilMiller, line 201: Post-decrement

src/vt/utils/strong/strong_type.h
```diff
@@ -0,0 +1,271 @@
...
+    operator--();
+    return copy;
+  }
+
+  /**
+   * \brief Convert to the underlying value (const)
+   *
+   * \return the value
+   */
+  operator T const() const { return v_; }
```
PhilMiller, line 214: Do we want this automatic conversion available?

src/vt/utils/strong/strong_type.h
```diff
@@ -0,0 +1,271 @@
...
+  bool operator!=(ThisType const& in) const {
+    return v_ != in.v_;
+  }
+
+  /**
+   * \brief Less-than operator
+   *
+   * \param[in] in the other value
+   */
+  bool operator<(ThisType const& in) const {
```
PhilMiller, line 102: Should these be directly comparable, or just orderable with an overload of `std::less`? I.e. are these comparison operators actually things we want used, or potentially an attractive nuisance like the arithmetic operators?

src/vt/epoch/epoch_window.cc
```diff
@@ -57,29 +57,31 @@ EpochWindow::EpochWindow(EpochType epoch) {
   // easily to incoming epochs to check that they match all the control bit
   // fields. For efficiency, the window relies on the sequentiality of
   // same-typed epochs to create a semi-contiguous window of terminated epochs.
-  EpochManip::setSeq(arch_epoch,0);
+  EpochManip::setSeq(arch_epoch, makeEpochZero());
```
PhilMiller, line 60: Does it make sense for `setSeq` to take `EpochType` for its second argument, rather than `EpochType::ImplType`? The argument is always going to be an explicitly calculated value, not another epoch. I think the interface should change, not the callers.

src/vt/utils/strong/strong_type.h
lifflander: yeah, maybe we should remove this. I'm not sure it's needed anywhere

src/vt/messaging/envelope/envelope_extended.h
```diff
@@ -67,8 +67,8 @@ namespace vt { namespace messaging {
 struct EpochActiveEnvelope {
   using isByteCopyable = std::true_type;
 
-  ActiveEnvelope env;               /**< The basic envelope */
-  EpochType epoch : epoch_num_bits; /**< The epoch */
+  ActiveEnvelope env;                                  /**< Basic envelope */
+  epoch::detail::EpochImplType epoch : epoch_num_bits; /**< Epoch bits */
```
lifflander, line 71: We should always use `epoch::detail::EpochImplType` or `EpochType::ImplType`, instead of mixing these up. @PhilMiller Any preference?

src/vt/epoch/epoch_window.h
```diff
@@ -151,13 +151,13 @@ struct EpochWindow {
   EpochType archetype_epoch_ = no_epoch;
 
   /// The set of epochs terminated
-  vt::IntegralSet<EpochType> terminated_epochs_;
+  vt::IntegralSet<EpochType::ImplType> terminated_epochs_;
```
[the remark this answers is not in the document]
PhilMiller: Or, can that be `template <template <class T> class S> class StrongIntegralSet` whose arguments are all `S<T>`?

src/vt/epoch/epoch_window.cc
lifflander: Fixed

src/vt/epoch/epoch_window.h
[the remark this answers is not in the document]
lifflander: No, that's not going to work. It might have before..but I just removed all the arithmetic operators which means that it doesn't meet the requirements for `T` based on the concept for `vt::Interval`.

src/vt/utils/strong/strong_type.h
lifflander: They actually have to be directly comparable for them to work in all the containers. I tried `std::less` and that is not sufficient.

src/vt/utils/strong/strong_type.h
```diff
@@ -0,0 +1,271 @@
...
+
+/**
+ * \internal \struct Strong
+ *
+ * \brief Used to hoist weak C++ types (like \c vt::VirtualProxyType ) into
+ * strongly typed values that have a unique type to enforce interfaces, inhibit
+ * dangerous conversions, and provide more semantic control.
+ *
+ * All \c Strong types are arithmetic types, thus have operators defined on them
+ */
```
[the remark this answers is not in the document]
lifflander: Removed

src/vt/utils/strong/strong_type.h
lifflander: Removed

src/vt/epoch/epoch_window.h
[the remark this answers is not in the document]
PhilMiller: The idea is that `StrongIntegralSet` members would unwrap `T`, and pass  the underlying value through to the encapsulated `IntegralSet`

src/vt/vrt/collection/manager.impl.h
```diff
@@ -438,7 +438,7 @@ template <typename ColT, typename IndexT, typename MsgT>
         msg->bcast_epoch_, base->cur_bcast_epoch_
       );
       vtAssert(base != nullptr, "Must be valid pointer");
-      base->cur_bcast_epoch_++;
+      *base->cur_bcast_epoch_ = *base->cur_bcast_epoch_ + 1;
```
PhilMiller, line 441: This is not actually a TD epoch, and should not use the `EpochType`

src/vt/epoch/epoch_type.h
```diff
@@ -0,0 +1,170 @@
...
+
+  EpochType() = default;
+
+  /**
+   * \brief Nullptr constructor default constructs strong type to set initial
+   * sentinel value
+   *
+   * \param[in] nullptr_t nullptr
+   */
+  constexpr EpochType(std::nullptr_t) {}
```
PhilMiller, line 75: Why should `EpochType(nullptr)` ever be a valid expression? Is there some technicality that's pushing toward accepting this?

src/vt/utils/strong/strong_type.h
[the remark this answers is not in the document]
PhilMiller: The comment is now out of sync with the code - no more arithmetic operators

PhilMiller (commented):
Minor things as noted

src/vt/epoch/epoch_type.h
lifflander: I was thinking that `nullptr` would be equivalent to `no_epoch`. There no technicality here. If you don't like the look of that, I'm fine removing it.

src/vt/epoch/epoch_window.h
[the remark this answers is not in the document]
lifflander: I think to make this work as expected, I would have to make a container that inherits from `IntegralSet` otherwise waste a lot of time forwarding stuff. 

```C++
template <typename T>
struct StrongIntegralType : IntegralSet<typename T::ImplType> {

};
```

I don't see why `template <template <class T> class S>` is necessary.

PhilMiller:
I'd be happy to approve this, once the remaining minor comments are addressed.

src/vt/epoch/epoch_type.h
lifflander: Removed

src/vt/vrt/collection/manager.impl.h
lifflander: This is removed completely now from another PR

src/vt/epoch/epoch_type.h
```diff
@@ -0,0 +1,165 @@
...
+/// Custom fmt formatter/print for \c EpochType
+template <>
+struct formatter<vt::epoch::EpochType> {
+  /// Presentation format:
+  ///  - 'x' - hex (default)
+  ///  - 'd' - decimal
+  ///  - 'b' - binary
+  char presentation = 'x';
+
+  /// Parses format specifications of the form ['f' | 'e'].
```
PhilMiller, line 124: Comment mismatch with the code.

src/vt/utils/strong/strong_type.h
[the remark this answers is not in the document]
lifflander: Removed

src/vt/epoch/epoch_type.h
lifflander: Fixed

src/vt/epoch/epoch_window.h
```diff
@@ -81,12 +81,12 @@ struct EpochWindow {
...
    * \brief Get the first terminated epoch in the window
    */
-  EpochType getFirst() const { return terminated_epochs_.lower(); }
+  EpochType getFirst() const { return EpochType{terminated_epochs_.lower()}; }
 
   /**
    * \brief Get the last terminated epoch in the window
    */
-  EpochType getLast()  const { return terminated_epochs_.upper(); }
+  EpochType getLast()  const { return EpochType{terminated_epochs_.upper()}; }
```
PhilMiller, line 89: Part of my idea of the `StrongIntegralSet` is that its arguments and returns would be the strong type, and it would handle wrapping/unwrapping.

src/vt/epoch/epoch_type.h
```diff
@@ -0,0 +1,165 @@
...
+
+#if !defined INCLUDED_VT_EPOCH_EPOCH_TYPE_H
+#define INCLUDED_VT_EPOCH_EPOCH_TYPE_H
+
+#include "vt/utils/strong/strong_type.h"
+#include "vt/epoch/epoch_impl_type.h"
+
+namespace vt { namespace epoch {
+
+struct EpochType : Strong<
```
PhilMiller, line 52: Other than `Type` vs `ImplType`, I think this could be boiled down to `using EpochType = Strong<...>`. Any reason not to do that? Compiler message and symbol-name clarity, I guess?

PhilMiller (approved):
Looks fine, though could be a bit tighter. Up to you whether to do that.

src/vt/epoch/epoch_window.h
lifflander: Done

src/vt/epoch/epoch_type.h
lifflander: Yes, I wanted it to be a different symbol.

src/vt/termination/interval/strong_integral_set.h
```diff
@@ -0,0 +1,138 @@
...
+
+namespace vt { namespace term { namespace interval {
+
+template <typename T>
+struct StrongIntegralSet {
+  using ImplType = typename T::ImplType;
+  using SetType = IntegralSet<ImplType>;
+
+  StrongIntegralSet() = default;
+  explicit StrongIntegralSet(ImplType in)
```
PhilMiller, line 57: I'll reiterate my recommendation that all the methods of `StrongIntegralSet` take and return `T` in their signatures, not `T::ImplType`

lifflander: Yes, I'm starting to make that change.

PhilMiller: Looks great

PhilMiller:
I'd be happy to see this merged now

PR #1473, opened 2021-06-11
1466 Custom sized messages are not handled correctly

MikolajZuzek:
Implemented `TestCollectionGroup.test_collection_group_4` test that demonstrates #1466 bug on `Broadcastable<...>::broadcastMsg(messaging::MsgPtrThief<MsgT> msg)` function.


src/vt/vrt/collection/manager.impl.h
```diff
@@ -964,8 +964,9 @@ messaging::PendingSend CollectionManager::broadcastCollectiveMsgImpl(
     auto_registry::RegistryTypeEnum::RegVrtCollectionMember :
     auto_registry::RegistryTypeEnum::RegVrtCollection;
   auto msg_size = vt::serialization::MsgSizer<MsgT>::get(msg.get());
+
   auto event = theMsg()->makeTraceCreationSend(
-    msg, han, reg_type, msg_size, true
+    han, reg_type, msg_size, false
```
PhilMiller, line 969: Ugh, the use of plain `bool` parameters makes it obscure what this is doing. @lifflander we should maybe try to move away from those throughout the system going forward.

PhilMiller: For these calls, could you add `bool is_bcast = ...;` and then pass `is_bcast` as the argument?

Speaking of which, why are two of these calls changing from `true` to `false`?

MikolajZuzek: I broke the flags on original PR #1426 (in a stupid way...) and fixed with 6e7944cb78f2c8985ae16b2d663b7b0263f670c3 (added `is_bcast` as well) - see [the combined changes](https://github.com/DARMA-tasking/vt/pull/1426/files#diff-670f3ee374f27bb87252fdb9df10b4f3eb2abd20ad03b91c4aa003b746d6ab9e) for  _src/vt/vrt/collection/manager.impl.h_.

Issue #1467, opened 2021-06-08
move `vt` version number into a separate file


cz4rs:
Currently every header file in `vt` contains version number like this:
```cpp
//                           DARMA Toolkit v. 1.0.0
```
This is currently not updated after a new release and could pollute the git history when we have frequent releases.

**What Needs to be Done?**
- [x] remove version number from headers
- [x] store it in a single place (`VERSION` file)
- [x] make sure that `find_package(vt)` works correctly

Additional preparation for releases based on `develop`:
- [x] ~describe versioning scheme (`1.1M6 (‘Milestone 6’)` vs `1.1.6`) on wiki (optional: check Sandia requirements for this)~ sticking with semver for now
- [ ] describe new (weekly) release process on wiki
- [x] create a `main` branch (or use `master`?)
- [ ] setup permissions and builds for `main`

Sidenote: as far as release notes go, there is ~3,5k commits added and ~350 issues resolved on `develop` (compared to `1.0.2` release):
```
vt  git log --oneline  master..develop | wc -l
3556
```
```
vt  git log --oneline  master..develop | grep "Merge pull request" | cut -d# -f2- | cut -c 1-4 | sort -u | wc -l
353
```

<details>
  <summary>list of the issues</summary>

#454 
#461 
#463 
#464 
#473 
#479 
#481 
#483 
#485 
#487 
#488 
#495 
#499 
#502 
#504 
#505 
#508 
#513 
#514 
#516 
#518 
#520 
#525 
#528 
#542 
#543 
#544 
#545 
#546 
#547 
#549 
#550 
#555 
#557 
#562 
#565 
#569 
#571 
#573 
#576 
#578 
#579 
#583 
#588 
#590 
#593 
#594 
#598 
#606 
#609 
#613 
#614 
#616 
#618 
#622 
#624 
#626 
#629 
#632 
#638 
#640 
#642 
#644 
#648 
#650 
#652 
#654 
#658 
#661 
#662 
#664 
#669 
#670 
#674 
#675 
#678 
#681 
#683 
#690 
#692 
#693 
#698 
#700 
#701 
#704 
#711 
#714 
#718 
#720 
#722 
#724 
#725 
#727 
#730 
#731 
#733 
#735 
#737 
#739 
#743 
#745 
#747 
#748 
#751 
#753 
#755 
#757 
#766 
#768 
#770 
#773 
#775 
#777 
#779 
#782 
#783 
#784 
#787 
#790 
#792 
#795 
[...]

PR #1475, opened 2021-06-13
1469 Output LB statistics as JSON

### LINKED ISSUE #1469: Explore using JSON for LB stats files for a more consistent, readable format
**What Needs to be Done?**

Choose a new format for the LB stats file.

lifflander:
Fixes #1469 

- [x] Include new json library
- [x] Include new compression library: brotli
- [x] Write C++ wrapper for encoding streaming data using compression library
- [x] Write output adaptor for the JSON library to stream output to compressed buffer
- [x] Write C++ wrapper for decoding? Do we need this? Probably for the stats restart reader
- [x] Adapt tests for new format
- [x] Provide options to users to output JSON uncompressed?

Should we completely remove the old stat output or leave it as an option that the user can select until we decide to remove it?


lifflander:
@PhilMiller @JacobDomagala @jstrzebonski @ppebay @nlslatt 

I would like feedback on this PR!

src/vt/utils/compress/compressor.h
```diff
@@ -0,0 +1,126 @@
...
+//@HEADER
+*/
+
+#if !defined INCLUDED_VT_UTILS_COMPRESS_COMPRESSOR_H
+#define INCLUDED_VT_UTILS_COMPRESS_COMPRESSOR_H
+
+#include <cstdlib>
+#include <memory>
+
+#include <brotli/encode.h>
```
cz4rs, line 51: this could be replaced with a forward declaration of `BrotliEncoderState`

src/vt/vrt/collection/balance/node_stats.cc
```diff
@@ -221,10 +230,11 @@ void NodeStats::outputStatsForPhase(PhaseType phase) {
     return;
   }
 
-  vtAssertExpr(stats_file_ != nullptr);
-
   vt_print(lb, "NodeStats::outputStatsForPhase: phase={}\n", phase);
 
+#if CSV_STAT_FILE
```
cz4rs, line 235: is the CSV format going to be removed before merging this? (if not, runtime option would be preferred)

src/vt/utils/json/output_adaptor.h
```diff
@@ -0,0 +1,117 @@
...
+   * \brief Write a single character to the output
+   *
+   * \param[in] c the character
+   */
+  void write_character(CharType c) override {
+    if (c_) {
+      c_->write(os_, reinterpret_cast<uint8_t const*>(&c), 1);
+    } else {
+      os_.write(&c, 1);
+    }
```
jstrzebonski, line 84: You could just call `write_characters()` here.

src/vt/vrt/collection/balance/node_stats.h
```diff
@@ -54,6 +54,7 @@
 #include "vt/runtime/component/component_pack.h"
 #include "vt/timing/timing.h"
 #include "vt/objgroup/proxy/proxy_objgroup.h"
+#include "vt/utils/json/base_appender.h"
```
cz4rs, line 57: could be replaced with a forward declaration

src/vt/vrt/collection/balance/node_stats.cc
```diff
@@ -221,10 +230,11 @@ void NodeStats::outputStatsForPhase(PhaseType phase) {
...
+  vtAssertExpr(stats_file_ != nullptr);
+
```
jstrzebonski, line 237: Maybe extract those two cases (`CSV_STAT_FILE` and `json`) into separate helper functions for better readability?

src/vt/vrt/collection/balance/node_stats.h
```diff
@@ -270,6 +271,8 @@ struct NodeStats : runtime::component::Component<NodeStats> {
   FILE* stats_file_ = nullptr;
   /// Whether the stats directory has been created
   bool created_dir_ = false;
+  /// The appender for outputting stat files in JSON format
+  std::unique_ptr<util::json::BaseAppender> stat_writer_ = nullptr;
```
jstrzebonski, line 275: `stats_file_` and `stat_writer_` are not inside if-defs.

nlslatt:
@jliffla In response to whether or not we need a wrapper for decoding, we will need one for the LB testing and tuning drivers that I'm working on.  They currently read the old format and will break when we make this change.

src/vt/utils/compress/compressor.impl.h
```diff
@@ -0,0 +1,93 @@
...
+#if !defined INCLUDED_VT_UTILS_COMPRESS_COMPRESSOR_IMPL_H
+#define INCLUDED_VT_UTILS_COMPRESS_COMPRESSOR_IMPL_H
+
+#include "vt/utils/compress/compressor.h"
+
+namespace vt { namespace util { namespace compress {
+
+template <typename StreamLike>
+bool Compressor::write(StreamLike& s, uint8_t const* buffer, std::size_t const size) {
+  vtAssert(enc_, "Must have a valid compressor");
```
nlslatt, line 54: This assert is redundant with the one in `writeImpl()`.

src/vt/utils/compress/compressor.impl.h
```diff
@@ -0,0 +1,93 @@
...
+    }
+    s.write(reinterpret_cast<char*>(out_buf_.get()), buf_size_ - avail_out);
+  }
+
+  return true;
+}
+
+template <typename StreamLike>
+bool Compressor::finish(StreamLike& s) {
+  vtAssert(enc_, "Must have a valid compressor");
```
nlslatt, line 87: This assert is redundant with the one in `writeImpl()`.

nlslatt:
@lifflander Also, as to whether or not we keep the old format around for a while, I think it depends how quickly we'll have the new reader ready in LBAF.

src/vt/utils/compress/compressor.h
lifflander: That's a good point, but the library doesn't provide forward declarations for any of the types. I would have to declare it myself, which might not be best to maintain. I would also have to write the forward decl as a extern "C" since brotli is written in C.

I'm not super worried about this inclusion because `compressor.h` is very minimally included In fact, I think it's only included in one cc file and doesn't propagate very far.

src/vt/utils/json/output_adaptor.h
lifflander: Good point, fixed!

src/vt/vrt/collection/balance/node_stats.h
lifflander: The `BaseAppender` is a forward declaration (essentially). It doesn't include anything. I forgot to actually include that file when I pushed.

src/vt/vrt/collection/balance/node_stats.h
lifflander: Yes, I just have the ifdef as a temporary solution to test this version of the code. I need to decide if we are going to support both modes or just remove the old one.

src/vt/utils/compress/compressor.impl.h
lifflander: FIxed

src/vt/utils/compress/compressor.impl.h
lifflander: Fixed

lifflander:
I've created a related issue in LBAF for @marcinwrobel1986 to work on to support his new format with the Python bindings:
https://github.com/DARMA-tasking/LB-analysis-framework/issues/80

src/vt/utils/compress/compressor.h
cz4rs: Looks good, I didn't think of the impact of C for this.

lifflander:
@PhilMiller @nlslatt @marcinwrobel1986 @JacobDomagala @jstrzebonski @cz4rs 

I've finished implementing this along with adapting all the tests to use the new format! I've tested the input/output workflow with the StatsRestartReader to ensure that everything works properly.

PhilMiller:
In adding in the JSON and Brotli libraries, please make sure they can build into both static and dynamic libraries.

lifflander:
> In adding in the JSON and Brotli libraries, please make sure they can build into both static and dynamic libraries.

Brotli has both options. Which should I made the default? Currently, I have the dynamic ones by default linked with VT.

lifflander:
I'm in the hell right now of getting this to work with Intel/NVCC. I've worked around a lot of the errors, but had to make adjustments to the library's code

PhilMiller:
> > In adding in the JSON and Brotli libraries, please make sure they can build into both static and dynamic libraries.
> 
> Brotli has both options. Which should I made the default? Currently, I have the dynamic ones by default linked with VT.

I'm not too concerned about the defaults. Mostly, I just want to avoid us exacerbating the situation we have where libfort won't build a static library (so it's disabled in EMPIRE), and fcontext won't build a dynamic library (at least on nvcc), and so I have to static link everything when developing there. #1479

lifflander:
I found a patch for the warning on nvcc 11. This does not solve the major compile-time errors with nvcc 10

src/vt/configs/arguments/args.cc
```diff
@@ -743,4 +747,28 @@ std::tuple<int, std::string> ArgConfig::parseArguments(CLI::App& app, int& argc,
...
+    name = name + str_rank;
+  } else {
+    name.replace(rank, 2, str_rank);
+  }
+  return vt_lb_stats_dir + "/" + name;
+}
+
+std::string AppConfig::getLBStatsFileIn() const {
+  auto name = vt_lb_stats_file_in;
+  std::size_t rank = name.find("%p");
```
cz4rs, line 764: duplicated code: this can be extracted into a helper function

src/vt/configs/arguments/args.cc
```diff
@@ -743,4 +747,28 @@ std::tuple<int, std::string> ArgConfig::parseArguments(CLI::App& app, int& argc,
...
+  auto str_rank = std::to_string(theContext()->getNode());
+  if (rank == std::string::npos) {
+    name = name + str_rank;
+  } else {
+    name.replace(rank, 2, str_rank);
+  }
+  return vt_lb_stats_dir_in + "/" + name;
```
cz4rs, line 771: duplicated code: this can be extracted into a helper function

src/vt/utils/compress/compressor.cc
```diff
@@ -0,0 +1,84 @@
+/*
+//@HEADER
+// *****************************************************************************
+//
+//                                compressor.cc
+//                           DARMA Toolkit v. 1.0.0
```
cz4rs, line 6: please remove all `DARMA Toolkit v. 1.0.0` lines (#1467 has been merged)

cz4rs:
@lifflander Can you add a sample JSON output somewhere? Either docs or even a unit test (so we are sure that it is always up to date).

lifflander:
I finally got the fix for nvcc 10.1 implemented. Rewrote half of their meta code before figuring out a somewhat simple fix. By swapping the template parameters of 4 trait classes, the bug goes away!

lifflander:
> @lifflander Can you add a sample JSON output somewhere? Either docs or even a unit test (so we are sure that it is always up to date).

Yes, will do.

src/vt/utils/compress/compressor.cc
lifflander: Ah, we need to fix the script `scripts/add-license-perl.pl` to properly put the template in. My elisp script generates the header using that script automatically when I create a new file.

lifflander: Oh, the license template was updated! When I rebased, I didn't rerun my script to regenerate based on the new template.

src/vt/configs/arguments/args.cc
lifflander: Fixed, extracted to helper function.

lib/brotli/CMakeLists.txt
```diff
@@ -0,0 +1,356 @@
...
+math(EXPR BROTLI_ABI_AGE "${_brotli_abi_info} & 4095")
+math(EXPR BROTLI_ABI_COMPATIBILITY "${BROTLI_ABI_CURRENT} - ${BROTLI_ABI_AGE}")
+mark_as_advanced(BROTLI_ABI_CURRENT BROTLI_ABI_REVISION BROTLI_ABI_AGE BROTLI_ABI_COMPATIBILITY)
+
+if (ENABLE_SANITIZER)
+  set(CMAKE_C_FLAGS " ${CMAKE_C_FLAGS} -fsanitize=${ENABLE_SANITIZER}")
+  set(CMAKE_CXX_FLAGS " ${CMAKE_CXX_FLAGS} -fsanitize=${ENABLE_SANITIZER}")
+  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -fsanitize=${ENABLE_SANITIZER}")
+
+  # By default, brotli depends on undefined behavior, but setting
```
nlslatt, line 105: Does this comment concern you?

lifflander: I've implemented a fix for this to only build brotli in portable mode to avoid the undefined behavior.

lifflander:
I'm still working on the documentation update.

lifflander:
@nlslatt @JacobDomagala @jstrzebonski @cz4rs I would like more feedback on this issue if possible

src/vt/vrt/collection/balance/stats_data.cc
```diff
@@ -0,0 +1,251 @@
...
+            vtAssertExpr(object.is_number());
+
+            NodeType home = uninitialized_destination;
+            if (task["entity"].find("home") != task["entity"].end()) {
+              auto home_json = task["entity"]["home"];
+              vtAssertExpr(home_json.is_number());
+              home = home_json;
+            }
+
+            auto elm = ElementIDStruct{object, home, this_node};
```
nlslatt, line 195: Should this line use `node` (read from the json) instead of `this_node` (the current node, which will typically be the same as `node`)?

nlslatt: I guess allowing them to differ would require extra logic.  Looks like you already checked that they match.

src/vt/vrt/collection/balance/stats_data.cc
```diff
@@ -0,0 +1,251 @@
...
+      sd->node_data_[id];
+
+      if (tasks.is_array()) {
+        for (auto const& task : tasks) {
+          auto node = task["node"];
+          auto time = task["time"];
+          auto etype = task["entity"]["type"];
+          vtAssertExpr(time.is_number());
+          vtAssertExpr(node.is_number());
+          vtAssertExpr(node == theContext()->getNode());
```
nlslatt, line 182: You could use `this_node` here.

tests/unit/collection/test_lb.extended.cc
```diff
@@ -193,15 +196,26 @@ TEST_P(TestNodeStatsDumper, test_node_stats_dumping_with_interval) {
...
+
+    EXPECT_TRUE(json.find("phases") != json.end());
+
+    int len = 0;
+    for (auto&& phase : json["phases"]) {
+      (void)phase;
+      len++;
+    }
+
+    EXPECT_EQ(len, num_phases);
```
JacobDomagala, line 217: This can be simplified to:
```c++
EXPECT_EQ(json["phases"].size(), num_phases);
```

src/vt/utils/compress/decompressor.impl.h
```diff
@@ -0,0 +1,171 @@
...
+  // While we have input data and room to output decompressed data...
+  while (avail_out > 0 and not success) {
+    BrotliDecoderResult res = BrotliDecoderDecompressStream(
+      dec_, &avail_in_, &next_in_, &avail_out, &next_out, nullptr
+    );
+    switch (res) {
+    case BROTLI_DECODER_RESULT_NEEDS_MORE_INPUT:
+    {
+      // We need to read more to continue decoding
+      vtAssert(avail_in_ == 0, "Should not have more input if we need more");
```
nlslatt, line 111: Can you improve the error message here? It took me a minute to understand what this meant. Maybe something like "Brotli asked for more input even though we still had some."

src/vt/utils/compress/decompressor.impl.h
```diff
@@ -0,0 +1,171 @@
...
+      if (not has_more_input) {
+        vtAbort("Needs more input to decompress but the file terminated early");
+      }
+      break;
+    }
+    case BROTLI_DECODER_RESULT_NEEDS_MORE_OUTPUT:
+    {
+      // this is an expected condition
+      vtAssert(
+        avail_out == 0, "Available output should be 0 if we need more output"
```
nlslatt, line 122: Same here. Maybe "Brotli asked for more output even though we still had some."

docs/md/node-stats.md
```diff
@@ -29,23 +29,161 @@ the statistics and mapping.
...
+        }
+    ]
+}
 \endcode
 
 
-Where `<comm-type>` is the type of communication occurred. The type of
-communication lines up the enum `vt::vrt::collection::balance::CommCategory` in
-the code.
+The type of communication lines up the enum
```
nlslatt, line 185: Did you mean `lines up with the enum`?

nlslatt (changes requested):
I haven't reviewed all the code added to `lib/` (just your changes to the libraries), but overall I think this looks great.  Just some minor changes as indicated in the review comments.

JacobDomagala (approved):
Looks good to me.

src/vt/vrt/collection/balance/stats_restart_reader.cc
```diff
@@ -100,17 +102,51 @@ StatsRestartReader::getMigrationList() const {
...
+std::deque<std::set<ElementIDType>> StatsRestartReader::readIntoElementHistory(
+  std::unique_ptr<StatsData> sd
+) {
+  std::deque<std::set<ElementIDType>> element_history;
+  for (PhaseType phase = 0; phase < sd->node_data_.size(); phase++) {
+    std::set<ElementIDType> buffer;
+    for (auto const& obj : sd->node_data_[phase]) {
+      buffer.insert(obj.first.id);
+    }
+    element_history.push_back(buffer);
```
jstrzebonski, line 114: Maybe it's over-optimization, but I think you could do `emplace_back(std::move(buffer))`.

src/vt/vrt/collection/balance/stats_restart_reader.cc
```diff
@@ -100,17 +102,51 @@ StatsRestartReader::getMigrationList() const {
...
 }
 
-void StatsRestartReader::readStats(std::string const& fileName) {
+std::deque<std::set<ElementIDType>> StatsRestartReader::readIntoElementHistory(
+  std::unique_ptr<StatsData> sd
+) {
+  std::deque<std::set<ElementIDType>> element_history;
+  for (PhaseType phase = 0; phase < sd->node_data_.size(); phase++) {
+    std::set<ElementIDType> buffer;
+    for (auto const& obj : sd->node_data_[phase]) {
```
jstrzebonski, line 111: Would you be able to convert this loop into for-each? Maybe something like this:
`for (auto const& nd : sd->node_data_) {`

src/vt/vrt/collection/balance/stats_restart_reader.cc
```diff
@@ -100,17 +102,51 @@ StatsRestartReader::getMigrationList() const {
   return proc_move_list_;
 }
 
-void StatsRestartReader::readStats(std::string const& fileName) {
+std::deque<std::set<ElementIDType>> StatsRestartReader::readIntoElementHistory(
+  std::unique_ptr<StatsData> sd
+) {
+  std::deque<std::set<ElementIDType>> element_history;
+  for (PhaseType phase = 0; phase < sd->node_data_.size(); phase++) {
```
jstrzebonski, line 109: Would you be able to convert this loop into for-each? Maybe something like this:
`for (auto const& nd : sd->node_data_) {`

src/vt/utils/compress/decompressor.impl.h
lifflander: Good point, I definitely need to clarify this better in the messages.

src/vt/vrt/collection/balance/stats_data.cc
lifflander: Right now, the assertion is that they line up. Maybe I should assert that?

nlslatt: I think you already did that on line 182.  I hadn't noticed that when I posted the first comment.

src/vt/utils/compress/compressor.cc
```diff
@@ -0,0 +1,83 @@
...
+    window_bits_ > BROTLI_MAX_WINDOW_BITS
+  ) {
+    window_bits_ = BROTLI_MAX_WINDOW_BITS;
+  }
+  // Set the quality parameter for the encoder
+  BrotliEncoderSetParameter(enc_, BROTLI_PARAM_QUALITY, quality_);
+  // Set the lgwin parameter for the encoder
+  BrotliEncoderSetParameter(enc_, BROTLI_PARAM_LGWIN, window_bits_);
+
+  out_buf_ = std::make_unique<uint8_t[]>(buf_size);
```
jstrzebonski, line 74: Just for consistency with the usage of members in this ctor:
```suggestion
  out_buf_ = std::make_unique<uint8_t[]>(buf_size_);
```

src/vt/utils/compress/compressor.impl.h
```diff
@@ -0,0 +1,90 @@
...
+namespace vt { namespace util { namespace compress {
+
+template <typename StreamLike>
+bool Compressor::write(StreamLike& s, uint8_t const* buffer, std::size_t const size) {
+  return writeImpl(s, buffer, size, false);
+}
+
+template <typename StreamLike>
+bool Compressor::writeImpl(
+  StreamLike& s, uint8_t const* buffer, std::size_t const size, bool finish_
```
jstrzebonski, line 58: I see the point in the usage of `finish_`, but to be honest, I'm not a fan of changing the behaviour of the function with a boolean variable. But from the top of my head, I don't have any clever idea of refactoring this function. So I'm fine with leaving it be.

src/vt/utils/compress/compressor.impl.h
```diff
@@ -0,0 +1,90 @@
...
+#if !defined INCLUDED_VT_UTILS_COMPRESS_COMPRESSOR_IMPL_H
+#define INCLUDED_VT_UTILS_COMPRESS_COMPRESSOR_IMPL_H
+
+#include "vt/utils/compress/compressor.h"
+
+namespace vt { namespace util { namespace compress {
+
+template <typename StreamLike>
+bool Compressor::write(StreamLike& s, uint8_t const* buffer, std::size_t const size) {
+  return writeImpl(s, buffer, size, false);
```
jstrzebonski, line 53: Just for readability would you mind extracting this `false` into a well-named `constexpr` variable?
Maybe simply:
```cpp
constexpr auto finish = false;
```


And the same goes for the `finish()` function.

src/vt/utils/compress/compressor.h
```diff
@@ -0,0 +1,125 @@
...
+   * \param[in] size the length of the buffer to compress
+   * \param[in] finish_ whether to finish writing
+   *
+   * \return whether it was successful or not
+   */
+  template <typename StreamLike>
+  bool writeImpl(
+    StreamLike& s, uint8_t const* buffer, std::size_t const size,
+    bool finish_
+  );
```
jstrzebonski, line 111: BTW, why it's protected? The dtor is not virtual, and I don't see any suggestions to inheriting from this struct.

src/vt/utils/json/decompression_input_container.impl.h
```diff
@@ -0,0 +1,61 @@
...
+namespace vt { namespace util { namespace json {
+
+template <typename StreamLike>
+DecompressionInputContainer::DecompressionInputContainer(
+  AnyStreamTag, StreamLike stream, std::size_t in_chunk_size
+) : chunk_size_(in_chunk_size)
+{
+  d_ = std::make_unique<DecompressorStreamType<StreamLike>>(std::move(stream));
+  output_buf_ = std::make_unique<char[]>(chunk_size_);
+  len_ = d_->read(reinterpret_cast<uint8_t*>(output_buf_.get()), chunk_size_);
```
jstrzebonski, line 56: I see a lot of casting to `uint8_t`. Maybe `output_buf_` could be of type `u_ptr<uint8_t[]>`? That way you would have to do casting to `char` in `getCurrent()`, I don't if anywhere else..

src/vt/utils/json/json_reader.cc
```diff
@@ -0,0 +1,93 @@
...
+        continue;
+      } else {
+        break;
+      }
+    }
+    if (f == '{') {
+      compressed = false;
+    }
+    is.close();
+  }
```
jstrzebonski, line 77: Maybe this block could be extracted to a function `bool isFileCompressed()`?

src/vt/vrt/collection/balance/stats_data.cc
```diff
@@ -0,0 +1,251 @@
...
+              task["entity"].find("collection_id") != task["entity"].end() and
+              task["entity"].find("index") != task["entity"].end()
+            ) {
+              auto cid = task["entity"]["collection_id"];
+              auto idx = task["entity"]["index"];
+              if (cid.is_number() && idx.is_array()) {
+                std::vector<uint64_t> arr;
+                for (auto const& index : idx) {
+                  arr.push_back(static_cast<uint64_t>(index));
+                }
```
jstrzebonski, line 208: Couldn't you do something like `std::vector<uint64_t> arr = idx` or `arr = idx.get<std::vector<>>()`?

jstrzebonski (approved):
I left some minor notes and suggestions, but this looks really great!

src/vt/vrt/collection/balance/stats_data.cc
```diff
@@ -0,0 +1,251 @@
...
+                  );
+                  sd->node_subphase_data_[id][elm][sid] = stime;
+                }
+              }
+            }
+          }
+        }
+      }
+    }
+  }
```
cz4rs, line 235: this makes me think that extracting some helpers could help readability

src/vt/vrt/collection/balance/stats_data.cc
```diff
@@ -0,0 +1,251 @@
...
+      }
+      i++;
+    }
+  }
+
+  return std::make_unique<json>(std::move(j));
+}
+
+/*static*/ std::unique_ptr<StatsData> StatsData::fromJson(
+  std::unique_ptr<nlohmann::json> jin
```
cz4rs, line 160: is `unique_ptr` really needed here? what about simply passing a const reference instead?

src/vt/vrt/collection/balance/stats_data.h
```diff
@@ -0,0 +1,114 @@
...
+  std::unique_ptr<nlohmann::json> toJson(PhaseType phase) const;
+
+  /**
+   * \brief Create \c StatsData from input JSON
+   *
+   * \param[in] j the json that contains the stats
+   *
+   * \return the new \c StatsData
+   */
+  static std::unique_ptr<StatsData> fromJson(std::unique_ptr<nlohmann::json> j);
```
cz4rs, line 92: We could simply return an object instead of `unique_ptr` - copy elision (and NRVO?) should let us avoid unnecessary copies. This would require change in `readIntoElementHistory`, but passing as a const ref should be good enough there.

cz4rs: Another thing is that it could simply become a constructor, but I don't have a strong opinion about this - static method works fine as well.

src/vt/vrt/collection/balance/stats_data.cc
lifflander: Ok, I've fixed this.

src/vt/vrt/collection/balance/stats_data.cc
lifflander: Fixed

src/vt/utils/compress/decompressor.impl.h
lifflander: Fixed

src/vt/utils/compress/decompressor.impl.h
lifflander: Fixed

docs/md/node-stats.md
lifflander: Yes, fixed

tests/unit/collection/test_lb.extended.cc
lifflander: Fixed

src/vt/vrt/collection/balance/stats_restart_reader.cc
lifflander: Done

src/vt/utils/compress/compressor.cc
lifflander: Done

src/vt/utils/compress/compressor.impl.h
lifflander: I initially had two different functions (one for finishing and one for incremental writes). The problem is that I have to duplicate basically this whole function so I combined them.

src/vt/utils/compress/compressor.impl.h
lifflander: Done!

src/vt/utils/compress/compressor.h
lifflander: Good point. I think it should just be `private`

src/vt/vrt/collection/balance/stats_data.cc
lifflander: The reason I have to use `unique_ptr` is to avoid leaking the json headers all around VT. I don't really like using a pointer, but that's the only to avoid header propagation.

src/vt/vrt/collection/balance/stats_data.h
lifflander: Same problem here. Then everything that includes `stats_data.h` will need the full json header which is massive. I only include the forward decls here.

src/vt/utils/json/json_reader.cc
lifflander: Done

src/vt/vrt/collection/balance/stats_data.cc
lifflander: Good idea! Done

src/vt/utils/json/decompression_input_container.impl.h
lifflander: Good idea, changed the type and now there's only one cast.

src/vt/vrt/collection/balance/stats_restart_reader.cc
lifflander: The problem is that `sd->node_data_` is an `unordered_map` but these need to be traversed in the correct order for `element_history` to be correct. So I can't use an iterator.

cz4rs (approved):
Looks good to me!

src/vt/vrt/collection/balance/stats_data.h
cz4rs: `fromJson` can probably be improved a bit, but now I see that returning a pointer with `toJson` is the way to go (to keep forward declarations).

I took a moment to check what can be done, please have a look at 4c112383a4dfbc9b53186f13787e683361e21ac3 and cherry-pick that if you like it.

src/vt/vrt/collection/balance/stats_restart_reader.cc
cz4rs, line 111: While checking the possibility to pass const ref (from the other conversation), I have noticed that this can actually insert an element into node data the phase data is missing for some reason (`sd` is not `const`).

nlslatt: I think the stats restart reader can operate with missing phases (where no LB migrations happened), so it would be a good idea to address this.

nlslatt (approved):
I'm okay with merging this after addressing the point that @cz4rs made about the LB stats reader.

src/vt/vrt/collection/balance/stats_data.h
lifflander: Looks good, I like your changes a lot!

src/vt/vrt/collection/balance/stats_restart_reader.cc
lifflander: @cz4rs's changes address this, which I have cherry-picked.

PR #1491, opened 2021-06-30
1490 Write new collection restore from file in place

### LINKED ISSUE #1490: Implement a restoreFromFileInPlace for application use case
**What Needs to be Done?**

We want to be able to just overwrite the data for an existing collection.

lifflander:
Fixes #1490

### DIFF
--- a/src/vt/vrt/collection/manager.impl.h
+++ b/src/vt/vrt/collection/manager.impl.h
@@ -3160,12 +3160,28 @@ void CollectionManager::checkpointToFile(
   checkpoint::serializeToFile(directory, directory_name);
 }
 
+namespace detail {
+template <typename ColT>
+inline void restoreOffHomeElement(CollectionManager::RestoreMigrateColMsg<ColT>* msg, ColT*) {
+  auto idx = msg->idx_;
+  auto node = msg->to_node_;
+  auto proxy = msg->proxy_;
+  theCollection()->migrate(proxy(idx), node);
+}
+} /* end namespace detail */
+
 template <typename ColT>
 /*static*/ void CollectionManager::restoreHandler(RestoreMigrateMsg<ColT>* msg) {
   auto idx = msg->idx_;
   auto node = msg->to_node_;
   auto proxy = msg->proxy_;
-  theCollection()->migrate(proxy(idx), node);
+  if (proxy(idx).tryGetLocalPtr() != nullptr) {
+    theCollection()->migrate(proxy(idx), node);
+  } else {
+    proxy(idx).template send<
+      RestoreMigrateColMsg<ColT>, detail::restoreOffHomeElement<ColT>
+    >(node, idx, proxy);
+  }
 }
 
 template <typename ColT>
--- a/src/vt/vrt/collection/manager.impl.h
+++ b/src/vt/vrt/collection/manager.impl.h
@@ -3173,7 +3173,9 @@ inline void restoreOffHomeElement(
 } /* end namespace detail */
 
 template <typename ColT>
-/*static*/ void CollectionManager::restoreHandler(RestoreMigrateMsg<ColT>* msg) {
+/*static*/ void CollectionManager::migrateToRestoreLocation(
+  RestoreMigrateMsg<ColT>* msg
+) {
   auto idx = msg->idx_;
   auto node = msg->to_node_;
   auto proxy = msg->proxy_;
@@ -3224,9 +3226,11 @@ void CollectionManager::restoreFromFileInPlace(
         using MsgType = RestoreMigrateMsg<ColT>;
         auto msg = makeMessage<MsgType>(this_node, idx, proxy);
         if (mapped_node != this_node) {
-          theMsg()->sendMsg<MsgType, restoreHandler<ColT>>(mapped_node, msg);
+          theMsg()->sendMsg<MsgType, migrateToRestoreLocation<ColT>>(
+            mapped_node, msg
+          );
         } else {
-          restoreHandler<ColT>(msg.get());
+          migrateToRestoreLocation<ColT>(msg.get());
         }
       }
     }
--- a/src/vt/vrt/collection/manager.impl.h
+++ b/src/vt/vrt/collection/manager.impl.h
@@ -3223,20 +3223,9 @@ void CollectionManager::restoreFromFileInPlace(
     auto const elm_exists = elm_holder->exists(idx);
     vtAssertExpr(elm_exists);
 
-#if 0
-    auto col_ptr = checkpoint::deserializeFromFile<ColT>(file_name);
-    col_ptr->stats_.resetPhase();
-
-    auto map_han = UniversalIndexHolder<>::getMap(proxy_bits);
-    elm_holder->remove(idx);
-    elm_holder->insert(idx, typename Holder<ColT, IndexType>::InnerHolder{
-      std::move(col_ptr), map_han, range
-    });
-#else
     auto ptr = elm_holder->lookup(idx).getCollection();
     checkpoint::deserializeInPlaceFromFile<ColT>(file_name, static_cast<ColT*>(ptr));
     ptr->stats_.resetPhase();
-#endif
   }
 }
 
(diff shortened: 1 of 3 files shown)


src/vt/vrt/collection/manager.impl.h
```diff
@@ -3148,6 +3148,86 @@ void CollectionManager::checkpointToFile(
...
+#if 0
+    auto col_ptr = checkpoint::deserializeFromFile<ColT>(file_name);
+    col_ptr->stats_.resetPhase();
+
+    auto map_han = UniversalIndexHolder<>::getMap(proxy_bits);
+    elm_holder->remove(idx);
+    elm_holder->insert(idx, typename Holder<ColT, IndexType>::InnerHolder{
+      std::move(col_ptr), map_han, range
+    });
+#else
```
nlslatt, line 3223: What's the motivation for retaining the code that's compiled out?

tests/unit/collection/test_checkpoint.extended.cc
```diff
@@ -223,7 +223,67 @@ TEST_F(TestCheckpoint, test_checkpoint_1) {
     // Ensure that all elements were properly destroyed
     EXPECT_EQ(counter, 0);
   }
+}
+
+TEST_F(TestCheckpoint, test_checkpoint_in_place_2) {
```
nlslatt, line 228: It doesn't look like this test exercises the migration feature.  Can you add another test?

src/vt/vrt/collection/manager.impl.h
lifflander: Oh, I meant to make this a parameter to the function for the user to choice the behavior they want.

tests/unit/collection/test_checkpoint.extended.cc
lifflander: Sure

src/vt/vrt/collection/manager.impl.h
```diff
@@ -3148,6 +3148,75 @@ void CollectionManager::checkpointToFile(
...
+  std::string const& file_base
+) {
+  using IndexType = typename ColT::IndexType;
+  using DirectoryType = CollectionDirectory<IndexType>;
+
+  auto proxy_bits = proxy.getProxy();
+
+  auto metadata_file_name = makeMetaFilename<IndexType>(file_base, false);
+
+  if (access(metadata_file_name.c_str(), F_OK) == -1) {
```
PhilMiller, line 3171: For various reasons, it's considered better practice to attempt the intended access, and address its failure if one occurred, rather than calling access() and interpreting its results

src/vt/vrt/collection/manager.impl.h
```diff
@@ -3148,6 +3148,75 @@ void CollectionManager::checkpointToFile(
...
+          RestoreMigrateMsg<ColT>, restoreHandler<ColT>
+        >(mapped_node, msg);
+      }
+    }
+  });
+
+  for (auto&& elm : directory->elements_) {
+    auto idx = elm.idx_;
+    auto file_name = elm.file_name_;
+    vtAssertExpr(proxy(idx).tryGetLocalPtr() != nullptr);
```
PhilMiller, line 3205: This asserted condition is the subject of an if on 3189. Does something change between the two?

src/vt/vrt/collection/manager.impl.h
```diff
@@ -3148,6 +3148,75 @@ void CollectionManager::checkpointToFile(
...
+    auto holder = findColHolder<ColT, IndexType>(proxy_bits);
+    vtAssertExpr(holder != nullptr);
+
+    auto elm_holder = findElmHolder<ColT,IndexType>(proxy_bits);
+    auto const elm_exists = elm_holder->exists(idx);
+    vtAssertExpr(elm_exists);
+
+    auto ptr = elm_holder->lookup(idx).getCollection();
+    checkpoint::deserializeInPlaceFromFile<ColT>(file_name, static_cast<ColT*>(ptr));
+    ptr->stats_.resetPhase();
```
PhilMiller, line 3216: Ideally, there would be a predicate isDeserializing that the element serialize() method would check to do this. We want that anyway, but we don't want to hold up this pr to get it in place

src/vt/vrt/collection/manager.impl.h
```diff
@@ -3148,6 +3148,75 @@ void CollectionManager::checkpointToFile(
   checkpoint::serializeToFile(directory, directory_name);
 }
 
+template <typename ColT>
+/*static*/ void CollectionManager::restoreHandler(RestoreMigrateMsg<ColT>* msg) {
```
PhilMiller, line 3152: If I'm reading this correctly, this is updating the apparent location of the indicated element to the node restoring it. If that's right, could this handler method be named accordingly? Something like updateRestoredLocation?

nlslatt: It's actually moving the collection element from the home rank to the rank where it was when the checkpoint was written, right?  I'm not sure if this is what you meant, @PhilMiller.  Maybe call it `migrateToRestoreLocation`?

PhilMiller, line 3152: If I'm reading this correctly, this is updating the apparent location of the indicated element to the node restoring it. If that's right, could this handler method be named accordingly? Something like updateRestoredLocation?

PhilMiller:
Like the description of this pr, some of the commit messages say "restoreToFile" instead of From. The code is right, though.

Besides the comments I made, this looks good

src/vt/vrt/collection/manager.impl.h
PhilMiller: Rather, that should be called `isRestoring()`

src/vt/vrt/collection/manager.impl.h
lifflander: This is how I'm doing it in the other case so I'm going to leave it this way for now.

src/vt/vrt/collection/manager.impl.h
lifflander: Yes, it's been migrated after that if statement back to where it belongs.

src/vt/vrt/collection/manager.impl.h
lifflander: Refer to: https://github.com/DARMA-tasking/vt/issues/1504

src/vt/vrt/collection/manager.impl.h
lifflander: Renamed

src/vt/vrt/collection/manager.impl.h
```diff
@@ -3160,6 +3160,100 @@ void CollectionManager::checkpointToFile(
...
+      auto file_name = elm.file_name_;
+
+      if (proxy(idx).tryGetLocalPtr() == nullptr) {
+        auto mapped_node = getMappedNode<ColT, IndexType>(proxy, idx);
+        vtAssertExpr(mapped_node != uninitialized_destination);
+        auto this_node = theContext()->getNode();
+
+        using MsgType = RestoreMigrateMsg<ColT>;
+        auto msg = makeMessage<MsgType>(this_node, idx, proxy);
+        if (mapped_node != this_node) {
```
PhilMiller, line 3228: Is guarding against self send necessary here?

tests/unit/collection/test_checkpoint.extended.cc
nlslatt: Thanks for exercising migration.  Does a test with the following sequence of events work?  It's a critical application use case.

1. Load balance
2. Checkpoint
3. Re-create the collection elements at home
4. Restore from the checkpoint

nlslatt (approved):
Looks good

bathmatt:
@lifflander the auto merge of checkpoint and the rest happens tomorrow so if this is ready can we merge away?

lifflander:
> @lifflander the auto merge of checkpoint and the rest happens tomorrow so if this is ready can we merge away?

This is still failing tests. I'm working on it more to figure out what went wrong.

PR #1494, opened 2021-07-06
1445: active: schedule messages on self-send

### LINKED ISSUE #1445: Deliver the message to broadcast sender using scheduler instead of self-send
**What Needs to be Done?**
Currently for default group and objgroup, when we call broadcast, the sender will execute the handler via self-send. We should schedule the message instead.

JacobDomagala:
Fixes #1445

### DIFF
--- a/src/vt/messaging/active.cc
+++ b/src/vt/messaging/active.cc
@@ -233,7 +233,7 @@ EventType ActiveMessenger::sendMsgBytesWithPut(
 
   vtWarnIf(
     !(dest != theContext()->getNode() || is_bcast),
-    "Destination {} should != this node"
+    fmt::format("Destination {} should != this node", dest)
   );
 
   MsgSizeType new_msg_size = base.size();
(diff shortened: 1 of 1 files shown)


PhilMiller:
This looks good to me, though I'll wait for Jonathan's approval as well.

lifflander (approved):
Looks good

lifflander:
@JacobDomagala Can you rebase this so we can merge?

src/vt/messaging/active.cc
```diff
@@ -522,8 +522,29 @@ EventType ActiveMessenger::doMessageSend(
...
+            theContext()->getTask()->send(dest_elm_id, base.size());
+          }
+        }
+      }
+
+      runnable::makeRunnable(base, true, envelopeGetHandler(msg->env), dest)
+        .withTDEpochFromMsg(is_term)
+        .withLBStats(&bare_handler_stats_, bare_handler_dummy_elm_id_for_lb_stats_)
+        .enqueue();
+    }
```
PhilMiller, line 547: How often is this block of functionality repeated? Could it be abstracted out to a separate function?

PhilMiller (approved):
Refactoring should not block this fix

src/vt/messaging/active.cc
lifflander: @JacobDomagala Can you open an issue for refactoring? I think this is repeated twice.

JacobDomagala: https://github.com/DARMA-tasking/vt/issues/1669 created

PR #1498, opened 2021-07-06
1497 Implement function for collections to get local indices

### LINKED ISSUE #1497: Add function to extract local indices from VT for collections
**What Needs to be Done?**

- [ ] First do this
- [ ] Then do that.
- [ ] Also this other thing.

**Additional context**

lifflander:
Fixes #1497

### DIFF
--- a/src/vt/vrt/collection/manager.h
+++ b/src/vt/vrt/collection/manager.h
@@ -1801,6 +1801,16 @@ public:
   IndexT getRange(VirtualProxyType proxy);
 
   /**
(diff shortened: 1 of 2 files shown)


nlslatt (approved):

nlslatt:
> Could I suggest a method that takes a function to call on each index, rather than returning a concrete collection?

That wouldn't solve the particular application problem we are facing.  Really, I think we just need the number of collection elements that are local, but this can serve that purpose.

JacobDomagala (approved):

PR #1500, opened 2021-07-15
1499 Fix collection reductions

### LINKED ISSUE #1499: Collection reductions do not generate proper stamps leading to race conditions
**Describe the bug**
If you do two un-synchronized collection reductions in a row, submitted in the same order (without explicit user-defined stamps) the reductions get confused and race.

lifflander:
Fixes #1499


lifflander:
Closes #1499 
Fixed via the collection refactor #1501

PR #1510, opened 2021-07-22
1509: build: make json headers public

### LINKED ISSUE #1509: json headers are private but includes are visible to applications
**Describe the bug**
The json headers are not visible to applications that use vt, breaking their builds.

For example:
```
In file included from EMPIRE/TPL/vt/src/vt/vrt/collection/balance/node_stats.h:57:0,
                 from EMPIRE/TPL/vt/src/vt/vrt/collection/manager.impl.h:70,
                 from EMPIRE/TPL/vt/src/vt/vrt/collection/manager.h:2087,
                 from EMPIRE/TPL/vt/src/vt/vrt/collection/collection_headers.h:49,
                 from EMPIRE/TPL/vt/src/vt/transport.h:72,
[snip]
EMPIRE/TPL/vt/src/vt/vrt/collection/balance/stats_data.h:54:10: fatal error: nlohmann/json_fwd.hpp: No such file or directory
 #include <nlohmann/json_fwd.hpp>
          ^~~~~~~~~~~~~~~~~~~~~~~
```

I see the `-I/path/to/EMPIRE/TPL/vt/lib/json/include` when building vt source but not EMPIRE source.

nlslatt:
Fixes application builds.

Fixes #1509

### DIFF
--- a/cmake/link_vt.cmake
+++ b/cmake/link_vt.cmake
@@ -53,7 +53,7 @@ function(link_target_with_vt)
 
   if (NOT DEFINED ARG_LINK_JSON AND ${ARG_DEFAULT_LINK_SET} OR ARG_LINK_JSON)
     target_link_libraries(
-      ${ARG_TARGET} PRIVATE ${ARG_BUILD_TYPE} ${JSON_LIBRARY}
+      ${ARG_TARGET} PUBLIC ${ARG_BUILD_TYPE} ${JSON_LIBRARY}
     )
   endif()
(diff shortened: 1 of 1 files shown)


PhilMiller:
I'm curious how the CI builds that are supposed to check vt as an external library and as an embedded package didn't catch this. Do we not have those running, or does the relevant header not get reached from the default includes?

PhilMiller (approved):

Issue #1524, opened 2021-08-17
Failure to build VT+checkpoint+detector in TPL directory in app


lifflander:
**Describe the bug**
```
20:17:27  -- Configuring done
20:17:28  CMake Error: install(EXPORT "EMPIRE-targets" ...) includes target "input_deck" which requires target "checkpoint" that is not in this export set, but in multiple other export sets: cmake/checkpointTargets.cmake, cmake/vtTargets.cmake.
20:17:28  An exported target cannot depend upon another target which is exported multiple times. Consider consolidating the exports of the "checkpoint" target to a single export.
20:17:28  CMake Error: install(EXPORT "EMPIRE-targets" ...) includes target "RandomNumberGenerator" which requires target "checkpoint" that is not in this export set, but in multiple other export sets: cmake/checkpointTargets.cmake, cmake/vtTargets.cmake.
20:17:28  An exported target cannot depend upon another target which is exported multiple times. Consider consolidating the exports of the "checkpoint" target to a single export.
20:17:28  CMake Error: install(EXPORT "EMPIRE-targets" ...) includes target "utils" which requires target "checkpoint" that is not in this export set, but in multiple other export sets: cmake/checkpointTargets.cmake, cmake/vtTargets.cmake.
20:17:28  An exported target cannot depend upon another target which is exported multiple times. Consider consolidating the exports of the "checkpoint" target to a single export.
20:17:28  -- Generating done
20:17:28  CMake Generate step failed.  Build files cannot be regenerated correctly.
```

PhilMiller:
This alone doesn't fix it

PR #1528, opened 2021-08-18
#1524: Fix CMake issues with embedded library usage

PhilMiller:
Assume that the dependencies on `detector` and `checkpoint` will be handled appropriately by the parent project, and don't try to pass them through.

Fixes: #1524


lifflander (approved):
Looks good!

PR #1580, opened 2021-10-13
1577 Fix ubsan for handlers

### LINKED ISSUE #1577: Fix function pointer casting ubsan bug
**What Needs to be Done?**

Basically the error is:

```
/vt/src/vt/runnable/runnable.cc:86:22: runtime error: call to function (unknown) through pointer to incorrect function type 'void (*)(vt::messaging::BaseMsg *)'
(/build/vt/examples/hello_world/hello_world+0xe40e60): note: (unknown) defined here
```

lifflander:
Fixes #1577


jstrzebonski:
@lifflander I cannot add you as a reviewer (you created this PR), but the code is ready for a review.

src/vt/vrt/collection/balance/lb_common.h
```diff
@@ -58,9 +59,17 @@ namespace balance {
 
 using ElementIDType = uint64_t;
 
-struct ElementIDStruct {
+struct ElementIDStruct : vt::Message {
```
PhilMiller, line 62: Why is this suddenly being made a message? This seems rather out of place.

src/vt/registry/auto/auto_registry_common.h
```diff
@@ -61,13 +61,159 @@
...
+
+struct BaseDispatcher {
+  virtual ~BaseDispatcher() = default;
+  virtual void dispatch(void* msg, void* object) const = 0;
+  virtual NodeType dispatch(
+    index::BaseIndex* cur_idx_ptr,
+    index::BaseIndex* range_ptr,
+    NodeType num_nodes
+  ) const = 0;
+};
```
PhilMiller, line 74: The two templates that provide concrete implementations of this abstract class each only really implement one of the two `dispatch` methods. That argues for having two different based types, one for handlers and the other for maps, and if necessary, templating registry instances over the appropriate base type.

PhilMiller:
This largely seems fine. There's the GCC-8 ASan build error, and a few minor comments I left.

PhilMiller:
I'm going to bet that the build failure we now get on GCC-8 represents a true positive instance in which UB was occurring without us noticing.

PhilMiller:
Could you adjust the CI scripts to require "clang-10, ubuntu, mpich", the one with UBSan enabled?

jstrzebonski:
> Could you adjust the CI scripts to require "clang-10, ubuntu, mpich", the one with UBSan enabled?

Sure, I'll turn it ON as soon as I merge this PR.

src/vt/vrt/collection/balance/lb_common.h
jstrzebonski: Removed.

src/vt/registry/auto/auto_registry_common.h
jstrzebonski: Done.

src/vt/registry/auto/auto_registry_common.h
```diff
@@ -61,13 +61,154 @@
...
-using AutoActiveVCType            = vrt::ActiveVirtualFnPtrType;
-using AutoActiveCollectionType    = vrt::collection::ActiveColFnPtrType;
-using AutoActiveCollectionMemType = vrt::collection::ActiveColMemberFnPtrType;
-using AutoActiveMapType           = mapping::ActiveMapFnPtrType;
-using AutoActiveMapFunctorType    = mapping::ActiveMapFnPtrType;
+struct SentinelObject {};
+
+struct BaseHandlersDispatcher {
+  virtual ~BaseHandlersDispatcher() = default;
+  virtual void dispatch(void* msg, void* object) const = 0;
```
jstrzebonski, line 68: Because of the `scatter` we could either except `void*` here or, if we want to kept `messaging::BaseMsg` parameter type here, another BaseDispatcher and separated function making handlers would have to be implemented.

Which solution do we prefer?

jstrzebonski: @lifflander @PhilMiller

PhilMiller: I'd prefer the approach that gives us the most type safety we can get without overhead. So, I think that favors a separate function, and keep this accepting BaseMsg

jstrzebonski: @PhilMiller I've separated dispatcher for scatter. Also I had to implement distinct handler maker/getter, and because of that, I've add another handler field `BaseMsgDerived` used to tell which getHandler should be chosen in `RunnableNew::setupHandler()`.

PhilMiller (approved):

PhilMiller:
The one failing test on GCC7 is due to disk space. Working around that is irrelevant to this PR.

PR #1585, opened 2021-10-26
#1581 add option to choose VT library compilation type

### LINKED ISSUE #1581: Make VT library compilation type flexible (static vs. shared)
**Describe the bug**
Right now, it is hard-coded as static in the cmake. We need to make this flexible there shouldn't be a requirement of static.

jstrzebonski:
Fixes #1581


lifflander (dismissed):
Looks good after you fix the name of the flag as suggested.

nmm0:
I agree that we should conform to BUILD_SHARED_LIBS since that also interacts well if VT is included as a subproject (subdirectory).

CMakeLists.txt
```diff
@@ -18,6 +18,15 @@ include(cmake/check_system_functions.cmake)
...
 set(FCONTEXT_LIBRARY fcontext)
 
+option(vt_shared_lib_enabled "Build VT as shared library" OFF)
+if (vt_shared_lib_enabled)
+  message(STATUS "Building VT as shared library")
+  set(BUILD_SHARED_LIBS ON)
+else()
+  message(STATUS "Building VT as static library")
+  set(BUILD_SHARED_LIBS OFF)
+endif()
```
PhilMiller, line 28: Given `BUILD_SHARED_LIBS`, do we need `vt_shared_lib_enabled` at all? Can we just set `BUILD_SHARED_LIBS` in our CI, like client code would?

jstrzebonski: Yes, we can. Done.

CMakeLists.txt
```diff
@@ -18,6 +18,8 @@ include(cmake/check_system_functions.cmake)
 set(VIRTUAL_TRANSPORT_LIBRARY vt CACHE INTERNAL "" FORCE )
 set(FCONTEXT_LIBRARY fcontext)
 
+option(BUILD_SHARED_LIBS "Build VT as shared library" OFF)
+
```
PhilMiller, line 22: Do we need to declare the option ourself at all? Or does CMake treat it as a built-in option on all projects?

lifflander: We don't need it. Removed

PhilMiller (approved):

cmake/trace_only_functions.cmake
```diff
@@ -101,10 +101,14 @@ function(create_trace_only_target)
...
     ${VT_TRACE_LIB}
-    STATIC
     ${TRACE_HEADER_FILES} ${TRACE_SOURCE_FILES}
   )
 
+  set_target_properties(
+    ${VT_TRACE_LIB} PROPERTIES
+    POSITION_INDEPENDENT_CODE ${BUILD_SHARED_LIBS}
+  )
+
```
PhilMiller, line 111: This apparently needs the option to actually be set

PhilMiller, line 111: We have this for the tracing library; do we need it for others as well? I thought it was there at some point, but maybe it was part of a commit that got rebased out?

lifflander: It's already there for the main VT library

src/CMakeLists.txt
```diff
@@ -199,10 +199,14 @@ endif()
 
 add_library(
   ${VIRTUAL_TRANSPORT_LIBRARY}
-  STATIC
   ${HEADER_FILES} ${SOURCE_FILES}
 )
 
+set_target_properties(
```
lifflander, line 205: Here

lifflander (approved):

PhilMiller (approved):

cmake/trace_only_functions.cmake
PhilMiller: Looks like this may not actually be necessary:
>  For SHARED and MODULE libraries the POSITION_INDEPENDENT_CODE target property is set to ON automatically.

https://cmake.org/cmake/help/latest/command/add_library.html

jstrzebonski: I must have overlooked that.
Fixed -> https://github.com/DARMA-tasking/vt/pull/1599

PR #1600, opened 2021-11-04
1586 fix vt sample project segfault

### LINKED ISSUE #1586: Segfault when running vt-trace-only-sample from vt-sample-project
When compiled with `build_with_tpl=1`:
```bash
root@a942e6b00517:/build/vt/vt-sample-project/build# ./vt-trace-only-sample 
Segmentation fault (core dumped)
```

and with `build_with_libs=1`:
```bash
root@a942e6b00517:/# ./build/vt/vt-sample-project/build/vt-trace-only-sample 
AddressSanitizer:DEADLYSIGNAL
=================================================================
==5395==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000138 (pc 0x0000004d6efb bp 0x7ffe44286fe0 sp 0x7ffe44286d80 T0)
==5395==The signal is caused by a READ memory access.
==5395==Hint: address points to the zero page.
    #0 0x4d6efb in vt::trace::TraceLite::TraceLite(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) /vt/src/vt/trace/trace_lite.cc:128:42
    #1 0x4d5efa in main /build/vt/vt-sample-project/build/../vt-trace-only/src/sample.cc:24:26
    #2 0x7f034a4f90b2 in __libc_start_main /build/glibc-eX1tMB/glibc-2.31/csu/../csu/libc-start.c:308:16
    #3 0x42b05d in _start (/build/vt/vt-sample-project/build/vt-trace-only-sample+0x42b05d)

AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV /vt/src/vt/trace/trace_lite.cc:128:42 in vt::trace::TraceLite::TraceLite(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)
==5395==ABORTING
```

jstrzebonski:
Fixes #1586 

----

`TraceLite` ctor tried to read from config pointer, which wasn't initialized yet.

### DIFF
--- a/ci/build_cpp.sh
+++ b/ci/build_cpp.sh
@@ -259,4 +259,8 @@ then
       "$inclusion_type" \
       ..
     cmake --build .
+
+    # Try to actually run samples
+    mpiexec -n 2 ./vt-runtime-sample
+    mpiexec -n 2 ./vt-trace-only-sample
 fi
--- a/src/vt/trace/trace_lite.cc
+++ b/src/vt/trace/trace_lite.cc
@@ -125,8 +125,11 @@ TraceLite::TraceLite(std::string const& in_prog_name)
    *  incremental_flush_mode_ to Z_FINISH
    */
 
-  auto const use_z_finish = theConfig()->vt_trace_gzip_finish_flush;
-  setFlushType(use_z_finish ? Z_FINISH : Z_SYNC_FLUSH);
+  auto const* conf_ptr = theConfig();
+  if (conf_ptr) {
+    auto const use_z_finish = conf_ptr->vt_trace_gzip_finish_flush;
+    setFlushType(use_z_finish ? Z_FINISH : Z_SYNC_FLUSH);
+  }
 
   // The first (implied) scheduler always starts with an empty event stack.
   event_holds_.push_back(0);
@@ -134,16 +137,19 @@ TraceLite::TraceLite(std::string const& in_prog_name)
 
 TraceLite::~TraceLite() {}
 
-void TraceLite::setFlushType(int flush_type){
+void TraceLite::setFlushType(int flush_type) {
   vtAssert(
     flush_type <= 6 and flush_type >= 0,
-    fmt::format("flush_type={} has to be in [0;6] range!\n", flush_type));
+    fmt::format("flush_type={} has to be in [0;6] range!\n", flush_type)
+  );
 
   incremental_flush_mode_ = flush_type;
 }
 
 #if vt_check_enabled(trace_only)
 void TraceLite::initializeStandalone(MPI_Comm comm) {
+  setFlushType(Z_FINISH);
+
   trace_ptr = this;
   context_ptr = new ctx::Context(true, comm);
 


src/vt/trace/trace_lite.cc
```diff
@@ -125,25 +125,31 @@ TraceLite::TraceLite(std::string const& in_prog_name)
    *  incremental_flush_mode_ to Z_FINISH
    */
 
-  auto const use_z_finish = theConfig()->vt_trace_gzip_finish_flush;
-  setFlushType(use_z_finish ? Z_FINISH : Z_SYNC_FLUSH);
+  auto const* conf_ptr = theConfig();
+  if (conf_ptr) {
+    auto const use_z_finish = conf_ptr->vt_trace_gzip_finish_flush;
+    setFlushType(use_z_finish ? Z_FINISH : Z_SYNC_FLUSH);
+  }
```
jstrzebonski, line 132: I wonder if that code should stay at all?
Is there a case for which `theConfig()` will return something else than nullptr at this point?

cz4rs: Possibly related: https://github.com/DARMA-tasking/vt/blob/b92912381377d37ef9e37e5a7f511204b37da85c/src/vt/runtime/runtime.cc#L110-L136

ci/build_cpp.sh
```diff
@@ -259,4 +259,8 @@ then
       "$inclusion_type" \
       ..
     cmake --build .
+
+    # Try to actually run samples
```
cz4rs, line 263: Do you want this to stay or that's just a debugging leftover?

src/vt/trace/trace_lite.cc
JacobDomagala: Yes, when vt is used in "normal" mode (not trace-only), then `theConfig()` should be valid.

ci/build_cpp.sh
jstrzebonski: I want it to stay.
I think we'd like to know whether changes in don't break anything - and running vt sample seems like a good first, naive test.

src/vt/trace/trace_lite.cc
jstrzebonski: Sure, but `TraceLite` is only used for `trace_only` isn't it?

JacobDomagala: vt::Trace inherits from TraceLite

jstrzebonski: Oh, okay, I've overlooked that!

PhilMiller (approved):
Generally looks good to me, though Cezary maybe has more to say about it before it gets merged.

cz4rs (approved):
Looks good to me.

ci/build_cpp.sh
cz4rs: I was wondering whether we should move that towards testing step, but in the end I think these are fine here.

PR #1601, opened 2021-11-04
1596 Extend lb stats

### LINKED ISSUE #1596: Extend LB statistics to include bare handlers and objgroups
**What Needs to be Done?**

Needed for complete statistic generation.

lifflander:
Fixes #1596


src/vt/elm/elm_id.h
```diff
@@ -41,16 +41,18 @@
...
-#if !defined INCLUDED_VT_VRT_COLLECTION_BALANCE_ELM_ID_H
-#define INCLUDED_VT_VRT_COLLECTION_BALANCE_ELM_ID_H
+#if !defined INCLUDED_VT_ELM_ELM_ID_H
+#define INCLUDED_VT_ELM_ELM_ID_H
 
-#include "vt/config.h"
+#include "vt/configs/types/types_type.h"
+#include "vt/configs/types/types_sentinels.h"
 
+#include <unordered_map>
```
PhilMiller, line 50: What's this doing here? It's not used by anything in this header

src/vt/elm/elm_id.h
```diff
@@ -77,6 +77,7 @@ struct ElementIDStruct {
   ElementIDType id = no_element_id; /**< id must be unique across nodes */
   NodeType home_node = uninitialized_destination; /**< the home node */
   NodeType curr_node = uninitialized_destination; /**< the current node */
+  bool migratable = false; /**< whether entity is migratable */
```
PhilMiller, line 80: Maybe not necessary right now, but this seems like something in which clever representation seems worthwhile - e.g. if there's a `no_node` sentinel value (`uninitialized_destination`?), then setting `curr_node = no_node` should be suitable for a predicate member `bool isMigratable()`

src/vt/objgroup/holder/holder_base.h
```diff
@@ -46,16 +46,31 @@
...
+  using ElementStatsType = elm::ElementStats;
+
   virtual ~HolderBase() = default;
   virtual bool exists() = 0;
   virtual void* getPtr() = 0;
 
   template <typename Serializer>
-  void serialize(Serializer& s) {}
+  void serialize(Serializer& s) {
+    s | elm_id_;
```
PhilMiller, line 64: Why is `stats_` omitted?

src/vt/objgroup/manager.cc
```diff
@@ -113,6 +133,50 @@ ObjGroupProxyType ObjGroupManager::makeCollectiveImpl(
...
+  return theObjGroup()->getHolderBase(handler);
+}
+} /* end namespace detail */
+
+elm::ElementIDStruct ObjGroupManager::getNextElm() {
+  // Avoid startup races
+  if (theNodeStats()) {
+    return theNodeStats()->getNextElm(false);
+  } else {
+    return elm::ElementIDStruct{};
```
PhilMiller, line 176: You've got a sentinel value `elm::no_element_id`; why not return that here?

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
```diff
@@ -479,8 +479,13 @@ void TemperedLB::doLBStages(TimeType start_imb) {
       if (first_iter) {
         // Copy this node's object assignments to a local, mutable copy
         cur_objs_.clear();
-        for (auto obj : *load_model_)
-          cur_objs_[obj] = load_model_->getWork(obj, {balance::PhaseOffset::NEXT_PHASE, balance::PhaseOffset::WHOLE_PHASE});
+        for (auto obj : *load_model_) {
```
PhilMiller, line 482: My revamp of the load model stuff should really include explicit choice of iteration over all objects or migratable objects, so strategies needn't do this

src/vt/vrt/collection/manager.cc
```diff
@@ -92,9 +92,7 @@ balance::ElementIDStruct CollectionManager::getCurrentContext() const {
     }
   }
 #endif
-  return balance::ElementIDStruct{
-    balance::no_element_id, uninitialized_destination, uninitialized_destination
-  };
+  return elm::ElementIDStruct{};
```
PhilMiller, line 95: Again, there's a defined sentinel value `elm::no_element_id`, let's use that

PhilMiller:
This all looks good to me. ZoltanLB needs to be updated. With that, I'd have no objection to merging this as is, and rolling forward to another PR

src/vt/elm/elm_id.h
lifflander: It seems to be required because I'm including the hash function for `ElementIDStruct`

src/vt/vrt/collection/manager.cc
lifflander: The default constructor sets this. I removed this because it seemed like extra redundant code.

src/vt/objgroup/holder/holder_base.h
lifflander: I missed it, good catch.

lifflander: Fixed

src/vt/elm/elm_id.h
PhilMiller: That's technically in `<functional>` - https://en.cppreference.com/w/cpp/utility/hash

src/vt/elm/elm_id.h
```diff
@@ -75,9 +75,10 @@ struct ElementIDStruct {
   }
 
   ElementIDType id = no_element_id; /**< id must be unique across nodes */
-  NodeType home_node = uninitialized_destination; /**< the home node */
   NodeType curr_node = uninitialized_destination; /**< the current node */
-  bool migratable = false; /**< whether entity is migratable */
+
+  bool isMigratable() const;
+  NodeType getHomeNode() const;
```
PhilMiller, line 81: Could we add a `getCurrNode()` accessor too?

lifflander: Done

src/vt/elm/elm_id.h
lifflander: Removed all together now.

src/vt/messaging/active.h
```diff
@@ -1764,6 +1767,10 @@ struct ActiveMessenger : runtime::component::PollableComponent<ActiveMessenger>
 
   // Diagnostic counters for counting forwarded messages
   diagnostic::CounterGauge amForwardCounterGauge;
+
+private:
+  elm::ElementIDStruct elm_id_ = {};
+  elm::ElementStats elm_stats_;
```
PhilMiller, line 1773: These should be commented, and probably named rather more explicitly, since they don't actually correspond to any tangible element instance. Something like `bare_handler_dummy_element_id_for_lb_stats_`

src/vt/objgroup/manager.impl.h
```diff
@@ -258,11 +259,23 @@ void ObjGroupManager::send(ProxyElmType<ObjT> proxy, MsgSharedPtr<MsgT> msg) {
   auto const dest_node = proxy.getNode();
   auto const ctrl = proxy::ObjGroupProxy::getID(proxy_bits);
   auto const han = auto_registry::makeAutoHandlerObjGroup<ObjT,MsgT,fn>(ctrl);
+
+  // set as internal message so it isn't recorded as it routes through bare
+  // handlers
+  envelopeSetInternalMessage(msg->env, true);
```
PhilMiller, line 265: For hypothetical debugging purposes, I think it would be preferable to have a separate envelop bit for whether it's already been recorded in the stats or not, and the the conditional in `active.cc` could test `not internal and not already_recorded`. That keeps the `internal` bit immutable after it's initially set, and avoids setting it on messages that aren't really internal. If there isn't room for said bit, I'm fine with this implementation.

PhilMiller: If we do keep the current implementation, I'd like the bit to be *unset* before it gets passed to any client handlers, so that it doesn't show up confusingly down the line.

tests/unit/lb/test_lb_stats_comm.cc
```diff
@@ -0,0 +1,751 @@
...
+  ap->addElm(*j);
+  ss = ap->finish();
+  // fmt::print("{}\n", ss.str());
+  auto c = DecompressionInputContainer{
+    DecompressionInputContainer::AnyStreamTag{}, std::move(ss)
+  };
+  return StatsData{json::parse(c)};
+#else
+  return *vt::theNodeStats()->getStatsData();
+#endif
```
PhilMiller, line 90: Commented out code should be dropped before this gets merged.

src/vt/objgroup/manager.impl.h
PhilMiller: Oh, since this `internal` bit is being added here solely for this purpose, we can just rename it and call it a day, unless you're planning on also setting it for other stuff inside the runtime as well.

src/vt/vrt/collection/balance/baselb/load_sampler.cc
```diff
@@ -53,13 +53,15 @@ void LoadSamplerBaseLB::buildHistogram() {
     );
```
PhilMiller, line 53: Are the LBs going to become aware of the distinction between the total load on a Node and the cumulative load of the migratable objects? Does this kind of already accomplish that?

PhilMiller:
OK, I've reviewed everything except the big newly added `test_lb_stats_comm`. It all looks good, modulo the comments I've made, and the need to update ZoltanLB.

nlslatt (commented):
It's looking good. I don't have any comments beyond what Phil already said.

tests/unit/lb/test_lb_stats_comm.cc
lifflander: Removed

src/vt/messaging/active.h
lifflander: Done

src/vt/objgroup/manager.impl.h
lifflander: I've renamed it!

src/vt/vrt/collection/balance/baselb/load_sampler.cc
lifflander: Yes, I think this accomplishes that goal. Because the total load will reflect everything but this will only allow objects to be chosen that are migratable.

src/vt/messaging/envelope/envelope_base.h
```diff
@@ -101,6 +101,10 @@ struct ActiveEnvelope {
   // Used only for broadcast to default group
   // Determines whether message should also be sent to the sender
   bool deliver_bcast_to_sender : 1;
+
+  /// Used to denote that the message's bare handlers shouldn't record
+  /// communication statistics due to redundancy
+  bool no_bare_handler_comm_stats : 1;
```
PhilMiller, line 107: Please rename this in the affirmative, possibly as `bool comm_stats_recorded_above_bare_handler`

PhilMiller (approved):

PR #1603, opened 2021-11-08
#1589 Read in communication data from json files

### LINKED ISSUE #1589: Read in communication data from json files
Presently, the constructor of `NodeStats`, which reads from json files, does not populate any communication data. The subphase communication is neither dumped nor read by `NodeStats`.

**What Needs to be Done?**
[ ] Have the constructor read the full-phase communication and populate the `node_comm_` member.
[ ] Add a test that populates a `NodeStats` object, dumps json from it, constructs a new `NodeStats` object from the json file, and then compares the two.

Should we do anything with the subphase communication at this point?

nlslatt:
Fixes #1589

### DIFF
--- a/tests/unit/collection/test_lb.extended.cc
+++ b/tests/unit/collection/test_lb.extended.cc
@@ -327,8 +327,7 @@ TEST_F(TestRestoreStatsData, test_restore_stats_data_1) {
   using CommVolume = vt::vrt::collection::balance::CommVolume;
   using CommBytesType = vt::vrt::collection::balance::CommBytesType;
 
-  // @todo: should do more than one phase, subphase loads, subphase comm, and
-  //        other types of comm
+  // @todo: should do more than one phase and other types of comm
 
   {
     PhaseType phase = 0;
--- a/tests/unit/collection/test_lb.extended.cc
+++ b/tests/unit/collection/test_lb.extended.cc
@@ -48,7 +48,9 @@
 #include "data_message.h"
 
 #include "vt/vrt/collection/manager.h"
+#include "vt/vrt/collection/balance/stats_data.h"
 #include "vt/utils/json/json_reader.h"
+#include "vt/utils/json/json_appender.h"
 
 #include <nlohmann/json.hpp>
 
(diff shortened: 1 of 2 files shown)


tests/unit/collection/test_lb.extended.cc
```diff
@@ -298,6 +300,254 @@ INSTANTIATE_TEST_SUITE_P(
...
+  auto json_writer = std::make_unique<JSONAppender>(
+    "phases", out_file_name, compress
+  );
+  {
+    auto j = sd.toJson(0);
+    json_writer->addElm(*j);
+  }
+  json_writer = nullptr;
+
+  //vt::runInEpochCollective([=]{
```
lifflander, line 384: Should we remove this line?

tests/unit/collection/test_lb.extended.cc
```diff
@@ -298,6 +300,254 @@ INSTANTIATE_TEST_SUITE_P(
...
+    vt::util::json::Reader r(out_file_name);
+    auto json_ptr = r.readFile();
+    auto sd_read = vt::vrt::collection::balance::StatsData(*json_ptr);
+
+    if (sd_read.node_data_.size() != sd.node_data_.size()) {
+      vtAbort(fmt::format(
```
lifflander, line 390: Normally, I would use `EXPECT_EQ` here, but it is nice to have the a good error message. `EXPECT_EQ` allows the test infrastructure to keep going instead of just failing the whole process out. Can we make these just prints when it's not equal?

lifflander (changes requested):

tests/unit/collection/test_lb.extended.cc
nlslatt: Removed.

tests/unit/collection/test_lb.extended.cc
nlslatt: Fixed.

tests/unit/collection/test_lb.extended.cc
```diff
@@ -298,6 +300,290 @@ INSTANTIATE_TEST_SUITE_P(
...
+      "Wrote {} phases of whole-phase comm data but read in {} phases",
+      sd.node_comm_.size(), sd_read.node_comm_.size()
+    );
+  }
+  // @todo: detailed comparison of whole-phase comm data
+
+  // @todo: compare subphase comm when writing/reading is implemented
+  // @todo: detailed comparison of subphase comm data
+
+  // @todo: clean up files
```
nlslatt, line 584: @lifflander Do you think it's important to clean up the files this test produces before merging?

tests/unit/collection/test_lb.extended.cc
```diff
@@ -298,6 +300,289 @@ INSTANTIATE_TEST_SUITE_P(
...
+
+        std::vector<uint64_t> arr;
+        arr.push_back(idx.x());
+        sd.node_idx_[elm_id] = std::make_tuple(proxy.getProxy(), arr);
+      }
+    }
+  }
+
+  auto const compress = false;
+  using JSONAppender = vt::util::json::Appender<std::ofstream>;
```
lifflander, line 377: I would recommend that we do this in-memory to avoid any file system issues:

```C++
StatsData getStatsDataForPhase(vt::PhaseType phase,  vt::vrt::collection::balance::StatsData in) {
  using JSONAppender = vt::util::json::Appender<std::stringstream>;
  using vt::util::json::DecompressionInputContainer;
  using vt::vrt::collection::balance::StatsData;
  using json = nlohmann::json;
  std::stringstream ss{std::ios_base::out | std::ios_base::in};
  auto ap = std::make_unique<JSONAppender>("phases", std::move(ss), false);
  auto j = in.toJson(phase);
  ap->addElm(*j);
  ss = ap->finish();
  //fmt::print("{}\n", ss.str());
  auto c = DecompressionInputContainer{
    DecompressionInputContainer::AnyStreamTag{}, std::move(ss)
  };
  return StatsData{json::parse(c)};
}
```

lifflander (changes requested):

tests/unit/collection/test_lb.extended.cc
lifflander: I think we should just do this in memory. I have a snippet below to do this..

tests/unit/collection/test_lb.extended.cc
nlslatt: I couldn't get this to compile using decompression, so my incorporation of this does not include compression. The json is fairly small so I don't think it's a big deal.

tests/unit/collection/test_lb.extended.cc
nlslatt: I incorporated a slightly modified version of your snippet that does not include compression/decompression.

lifflander (approved):
Looks good to me now!

lifflander:
@nlslatt Can you rebase this PR so we can merge it?

PR #1610, opened 2021-11-16
#1609 clean up workspace before compiling vt-sample-project with vt as TPL

### LINKED ISSUE #1609: Clean up workspace before compiling vt-sample-project with vt as TPL

jstrzebonski:
Fixes #1609


ci/build_vt_sample.sh
```diff
@@ -0,0 +1,65 @@
...
+if test "$VT_INCLUSION_TYPE" = "TPL"
+then
+  inclusion_type="-Dbuild_with_tpl=1"
+elif test "$VT_INCLUSION_TYPE" = "EXT_LIB"
+then
+  inclusion_type="-Dbuild_with_libs=1"
+fi
+
+# Don't build vt-sample on Alpine Linux
+is_alpine="$(grep ID < /etc/os-release | grep -c alpine || true)"
```
cz4rs, line 18: can we remove this part and just not call this script on Alpine?

ci/build_vt_sample.sh
```diff
@@ -0,0 +1,65 @@
...
+if test "$is_alpine" -eq 0 && test "${VT_CI_BUILD:-0}" -eq 1
+then
+export VT=${source_dir}
```
lifflander, line 21: Can you fix the indentation here?

ci/build_vt_sample.sh
cz4rs: (I can see that you have already omitted the call in `alpine-cpp.dockerfile`, so is there any other reason to keep this check?)

cz4rs: Ok, I can see now that this would require a bit more mess in `docker-compose.yml` as well. Up to you if you want to go this path.

cz4rs (approved):
Looks good to me, I like extracting this to a separate script even just for clarity.

ci/build_vt_sample.sh
jstrzebonski: Absolutely! Done.

ci/build_vt_sample.sh
jstrzebonski: Sure, done!

PhilMiller (approved):

PR #1613, opened 2021-11-17
#1602: stop using CMAKE_CXX_FLAGS for setting compile options

### LINKED ISSUE #1602: stop using `CMAKE_CXX_FLAGS` for setting compile options
**What Needs to be Done?**
Specify compile options per target instead of using global `CMAKE_CXX_FLAGS` variable (see `cmake-modules/SetCXXCompilerFlags.cmake` as the main offender).

Preferred solution is to use `target_compile_options` command (which can specify `INTERFACE|PUBLIC|PRIVATE` keywords).

Using `CMAKE_CXX_FLAGS` can affect applications built on top of `vt`.

cz4rs:
fixes #1602


cmake/turn_on_warnings.cmake
```diff
@@ -1,22 +1,25 @@
...
-  enable_cxx_compiler_flag_if_supported("-Wall")
-  enable_cxx_compiler_flag_if_supported("-pedantic")
-  enable_cxx_compiler_flag_if_supported("-Wshadow")
-  enable_cxx_compiler_flag_if_supported("-Wno-unknown-pragmas")
-  enable_cxx_compiler_flag_if_supported("-Wsign-compare")
-  # Not really a warning, is still diagnostic related..
-  enable_cxx_compiler_flag_if_supported("-ftemplate-backtrace-limit=100")
-endif()
+function(turn_on_warnings vt_target)
+  if(NOT hasParent)
```
PhilMiller, line 12: Now that the setting is `target_compile_options(... PRIVATE ...)` rather than `set(CMAKE_CXX_FLAGS ... PARENT_SCOPE)`, I think we might be able to do without the condition `if(NOT hasParent)`. The flags we set here will only affect compilation of vt itself.

cmake-modules/SetCXXCompilerFlags.cmake
```diff
@@ -1,44 +1,44 @@
...
 
 if (vt_ubsan_enabled)
   add_definitions(-DVT_UBSAN_ENABLED)
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=undefined")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-omit-frame-pointer")
+  list(APPEND TARGET_CXX_FLAGS -fsanitize=undefined -fno-omit-frame-pointer)
 endif()
+
+message(DEBUG "Target ${vt_target} compile options: ${TARGET_CXX_FLAGS}")
+target_compile_options(${vt_target} PRIVATE ${TARGET_CXX_FLAGS})
```
PhilMiller, line 42: I'm not sure these should all be `PRIVATE`. Other than the warnings, pretty much everything we set here has an ABI impact that I think we want to expose to any dependent projects as well. That suggests having two variables, `TARGET_PRIVATE_CXX_FLAGS` and `TARGET_PUBLIC_CXX_FLAGS`.

I'll comment on the individual non-warning flags on which group I think they belong in.

cmake-modules/SetCXXCompilerFlags.cmake
```diff
@@ -1,44 +1,44 @@
...
 
 if ("${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU")
   # 4.9.3 complains about std::min not being constexpr
-  if (NOT (CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 5 OR CMAKE_CXX_COMPILER_VERSION VERSION_EQUAL 5))
+  if (NOT (CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 5))
     message("${PROJECT_NAME} currently requires g++ 5 or greater.  If you need it to work with 4.9, please complain.")
   endif ()
   if (CMAKE_CXX_COMPILER_VERSION VERSION_LESS 6)
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-unused-variable")
+    list(APPEND TARGET_CXX_FLAGS -Wno-unused-variable)
```
PhilMiller, line 13: Warnings can stay private.

cmake-modules/SetCXXCompilerFlags.cmake
```diff
@@ -1,44 +1,44 @@
...
   endif()
 elseif ("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -ftemplate-depth=900")
+  list(APPEND TARGET_CXX_FLAGS -ftemplate-depth=900)
```
PhilMiller, line 16: If this is meant to raise a limit that instantiation of vt constructs otherwise hits, then we need it to be public to benefit downstream users.

cmake-modules/SetCXXCompilerFlags.cmake
```diff
@@ -1,44 +1,44 @@
...
   if (APPLE)
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -stdlib=libc++ -DCLI11_EXPERIMENTAL_OPTIONAL=0")
+    list(APPEND TARGET_CXX_FLAGS -stdlib=libc++ -DCLI11_EXPERIMENTAL_OPTIONAL=0)
```
PhilMiller, line 18: These both affect compiled ABI, so they must be public.

cmake-modules/SetCXXCompilerFlags.cmake
```diff
@@ -1,44 +1,44 @@
...
+    list(APPEND TARGET_CXX_FLAGS -DCLI11_EXPERIMENTAL_OPTIONAL=0)
   endif()
   if (CMAKE_CXX_COMPILER_VERSION VERSION_LESS 6)
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-missing-braces")
+    list(APPEND TARGET_CXX_FLAGS -Wno-missing-braces)
   endif()
 elseif (CMAKE_CXX_COMPILER_ID STREQUAL IntelLLVM AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 2021.3.0)
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fhonor-infinites")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fhonor-nans")
+  list(APPEND TARGET_CXX_FLAGS -fhonor-infinites -fhonor-nans)
```
PhilMiller, line 27: These should stay private, since we're just concerned with the behavior of internal code in this case. Applications may legitimately want the greater optimization that hypothetically comes from assuming all floating point values are finite.

However, that means we need to be careful not to use infinity or NaN in template code, where its compilation will be with the downstream project's flags and not our own. If we don't want to restrict ourselves that way, then these need to be public, and projects need to be aware we're setting them.

cmake-modules/SetCXXCompilerFlags.cmake
```diff
@@ -1,44 +1,44 @@
...
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=address")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-omit-frame-pointer")
+  list(APPEND TARGET_CXX_FLAGS -fsanitize=address -fno-omit-frame-pointer)
 endif()
 
 if (vt_ubsan_enabled)
   add_definitions(-DVT_UBSAN_ENABLED)
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=undefined")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-omit-frame-pointer")
+  list(APPEND TARGET_CXX_FLAGS -fsanitize=undefined -fno-omit-frame-pointer)
```
PhilMiller, line 38: ASan and UBSan are most useful if the entire code base is compiled with them, and it's necessary to link with them if any code is compiled with them. So, I think these should be public.

cmake-modules/SetCXXCompilerFlags.cmake
cz4rs: Great, thanks for this - I was wondering if I should enable some of the flags again in tests and examples, but making the right ones public makes more sense.

cmake-modules/SetCXXCompilerFlags.cmake
PhilMiller: Could you check and affirm that anywhere we use infinity or NaN is in non-template code, or templates that are instantiated internally rather than in the public API?

cz4rs: These are the ones that were failing without the flags - non-templated:
```
vt  gg isfinite src/
src/vt/vrt/collection/balance/model/norm.cc:63:  if (std::isfinite(power_)) {
```
https://github.com/DARMA-tasking/vt/blob/cb984cceec59ae2e80017a097ce8e1573dd1c57d/src/vt/vrt/collection/balance/model/norm.cc#L63

```
vt  gg isnan src/
src/vt/vrt/collection/balance/model/norm.cc:54:  vtAssert(not std::isnan(power), "Power must have a real value");
```
https://github.com/DARMA-tasking/vt/blob/cb984cceec59ae2e80017a097ce8e1573dd1c57d/src/vt/vrt/collection/balance/model/norm.cc#L54

Grepping for `nan` and `finit` doesn't bring any more meaningful hits.

PhilMiller (approved):

PhilMiller:
We should probably explicitly test this with EMPIRE sooner rather than later, so we don't get hung up at the next attempt to merge it in.

scripts/get_mpich.sh
```diff
@@ -1,43 +0,0 @@
-#!/bin/sh
```
lifflander, line 1: Why is this removed?

scripts/get_zoltan.sh
```diff
@@ -1,31 +0,0 @@
-#!/usr/bin/env bash
```
lifflander, line 1: Can you put these back?

lifflander (dismissed):

scripts/get_mpich.sh
cz4rs: this is almost identical as `ci/deps/mpich.sh`

scripts/get_zoltan.sh
cz4rs: same as above, almost identical as `ci/deps/zoltan.sh`

PhilMiller:
> We should probably explicitly test this with EMPIRE sooner rather than later, so we don't get hung up at the next attempt to merge it in.

EMPIRE seems to be just fine with this. No further discussion needed.

PR #1614, opened 2021-11-19
#1592 Update clang-format configuration

### LINKED ISSUE #1592: Update clang-format file
**What Needs to be Done?**

- [ ] Fix clang-format identing of preprocessor directives
- [ ] Make clang-format negatively indent access specifiers

jstrzebonski:
Fixes #1592 

----

About the preprocessor directives indention, what style do we prefer?
(paragraph `IndentPPDirectives` at https://clang.llvm.org/docs/ClangFormatStyleOptions.html)

`None`:

```cpp
#if FOO
#if BAR
#include <foo>
#endif
#endif
```

`AfterHash`:

```cpp
#if FOO
#  if BAR
#    include <foo>
#  endif
#endif
```

`BeforeHash`:

```cpp
#if FOO
  #if BAR
    #include <foo>
  #endif
#endif
```

Currently we have `BeforeHash`.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -2,7 +2,7 @@ Language: Cpp
 ColumnLimit: 80
 UseTab: Never
 
-AccessModifierOffset: -4
+AccessModifierOffset: -2
 AlignAfterOpenBracket: AlwaysBreak
 AlignConsecutiveAssignments: false
 AlignConsecutiveDeclarations: false


lifflander:
Let's go with `BeforeHash`

lifflander (approved):

PR #1617, opened 2021-12-03
#1604 put CLI11 in vt namespace

### LINKED ISSUE #1604: ODR violations for applications using CLI11
**Describe the bug**
An application linking to VT and using CLI11 bundled internally can cause ODR violations and potentially crash with an undecipherable errors.

**To Reproduce**
Write an application using VT and CLI11 for parsing its command line. Initialize VT and the program (may) crash. This is because CLI11 symbols are defined in both libvt.a and the application causing an ODR violation. If the definitions are identical this may not crash. The compiler/linker is not required to report ODR violations.

**Expected behavior**
No ODR violations.

**Current Solution**

The single header generator script in CLI11 lets you specify a namespace. When we bundle CLI11 we should define our own namespace, ensuring the symbols are not duplicated.

jstrzebonski:
Fixes #1604 

----

I didn't upgrade CLI11 to the newest version, API changed a bit. If we want to do so, let's open another issue for that. Also script generating single header file was changed, and I opened issue in CLI11 repo reporting problem with support for custom namespace - before upgrading we want it to be fixed

https://github.com/CLIUtils/CLI11/issues/680

### DIFF
--- a/lib/CLI/CLI/CLI11.hpp
+++ b/lib/CLI/CLI/CLI11.hpp
@@ -61,6 +61,9 @@
 #include <utility>
 #include <vector>
 
+namespace vt {
+
+
 
 // Verbatim copy from CLI/Version.hpp:
 
@@ -8220,4 +8223,6 @@ inline std::string Formatter::make_option_usage(const Option *opt) const {
 }
 
 } // namespace CLI
+
+} // namespace vt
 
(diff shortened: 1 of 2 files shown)


cz4rs (approved):
Looks good to me!

PhilMiller (approved):

PR #1618, opened 2021-12-10
#1605 fmt custom namespace

### LINKED ISSUE #1605: Potential ODR violations with fmt
See as reference: #1604 

The issue may also be a problem with libfmt. A potential solution is that fmt uses a macro for the namespace; we could potentially define this macro before including.

jstrzebonski:
Fixes #1605 

---

`fmt` has macros `FMT_BEGIN_NAMESPACE` and `FMT_END_NAMESPACE`, but unfortunately, internally it explicitly uses full namespace (in core.h):

```cpp
namespace detail {
void to_string_view(...);
using fmt::v7::to_string_view;
...
```

To use it, we would have to hardcode our namespace anyway.

To solve this I extracted this "inner" namespace into an additional macro, so it could be simply defined in cmake.

---

I opened an issue in fmt repo regarding this problem
fmtlib/fmt#2642

---

### Everything above is no longer valid. fmt team remove qualifying by inline namespace from `core.h`.

I copied their fix into our `fmt/core.h` and define `FMT_BEGIN_NAMESPACE` and `FMT_END_NAMESPACE`.

### DIFF
--- a/lib/fmt/include/fmt/core.h
+++ b/lib/fmt/include/fmt/core.h
@@ -187,6 +187,10 @@
 #  endif
 #endif
 
+#ifndef FMT_INLINE_NAMESPACE_NAME
+#  define FMT_INLINE_NAMESPACE_NAME v7
+#endif
+
 #ifndef FMT_BEGIN_NAMESPACE
 #  if FMT_USE_INLINE_NAMESPACES
 #    define FMT_INLINE_NAMESPACE inline namespace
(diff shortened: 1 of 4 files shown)


jstrzebonski:
There are some nvcc errors to be fixed.

jstrzebonski:
It seems that nvcc is unable to handle definitions with spaces. I also observed that definition defined within vt's cmake was "leaked" into my test app, where I wanted to use different fmt.

Currently I just hardcoded vt namespace in fmt code. It's just two places in core.h that needed modification, and it fixes both problems.

lib/fmt/include/fmt/core.h
```diff
@@ -500,7 +500,7 @@ constexpr basic_string_view<typename S::char_type> to_string_view(const S& s) {
 
 namespace detail {
 void to_string_view(...);
-using fmt::v7::to_string_view;
+using fmt::to_string_view;
```
cz4rs, line 503: https://github.com/fmtlib/fmt/issues/2642 was fixed and the problematic `v*` was removed.

jstrzebonski: Yeap, that's why I removed it here as well.

cz4rs (approved):
> Currently I just hardcoded vt namespace in fmt code. It's just two places in core.h that needed modification, and it fixes both problems.

Looks good to me, imho that's acceptable "manual" maintenance burden.

jstrzebonski:
> > Currently I just hardcoded vt namespace in fmt code. It's just two places in core.h that needed modification, and it fixes both problems.
> 
> Looks good to me, imho that's acceptable "manual" maintenance burden.

Yes, in case of fmt update, that's literally two characters that needs to be changed.

lifflander (approved):

PR #1622, opened 2021-12-14
#1620 add move assignment to ModifierToken

### LINKED ISSUE #1620: add move assignment to ModifyToken
**What Needs to be Done?**

Currently move assignment is disabled, so modify tokens cannot be temporarily stored. This functionality is useful if, for example, multiple stages are used to modify the token. Also the move constructor/assignment should be noexcept

- [ ] Add move assignment operators, make sure copy assignment is deleted
- [ ] Make move construct and move assign noexcept

nmm0:
Closes #1620


lifflander (approved):
Looks good

PR #1649, opened 2022-01-05
#1443 Make the scheduler loop exception-safe

### LINKED ISSUE #1443: Handle exceptions thrown inside scheduler loop
https://github.com/DARMA-tasking/vt/issues/1394 added new build option to throw an exception on `vtAbort`. This introduces a new issue, when an exception is thrown inside scheduler loop the traces are not handled correctly. 
```c++
void runSchedulerWhile(...) {
   triggerEvent(SchedulerEventType::BeginSchedulerLoop);
   
   while (cond()) {
      // throw here
      runSchedulerOnceImpl();
   }
   triggerEvent(SchedulerEventType::EndSchedulerLoop);
```

*What needs to be done*
Either create RAII object, which triggers BeginSchedulerLoop and EndSchedulerLoop or wrap the runSchedulerOnceImpl with try-catch block

jstrzebonski:
Fixes #1443 

---

In the end, I decided to take a "RAII" approach. I think it will be easier to (if ever necessary) add new functionalities for loop start and loop end this way, than remembering to also change try-catch block.

### DIFF
--- a/src/vt/scheduler/scheduler.cc
+++ b/src/vt/scheduler/scheduler.cc
@@ -289,6 +289,15 @@ void Scheduler::runSchedulerOnceImpl(bool msg_only) {
   }
 }
 
+Scheduler::SchedulerLoopGuard::SchedulerLoopGuard(Scheduler* scheduler)
+  : scheduler_{scheduler} {
+  scheduler_->triggerEvent(SchedulerEventType::BeginSchedulerLoop);
+}
+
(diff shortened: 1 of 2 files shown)


PhilMiller:
Could you please retitle this PR and amend the commit to describe it as "Make the scheduler loop exception-safe"?

Handling exceptions generally implies `catch`ing them.

PhilMiller (approved):

JacobDomagala (approved):

PR #1651, opened 2022-01-07
#1452 remove 'using vt::...' declarations polluting std namespace

### LINKED ISSUE #1452: Stop polluting std namespace with using declarations
**What Needs to be Done?**
Use fully qualified names when defining hash functions and other stuff within `std` namespace instead of adding `using` declarations that pollute the `std` namespace.

jstrzebonski:
Fixes #1452


PhilMiller:
This looks good so far. I'm surprised we were doing that in the first place. Thanks for cleaning it up.

JacobDomagala (approved):

PR #1653, opened 2022-01-10
1652: docs: do not customize m.css library

### LINKED ISSUE #1652: do not customize `m.css` when building documentation
- https://github.com/DARMA-tasking/vt/commit/e86eda790d0af053d0d15515d9fb12f4c5d3526e fixed failing documentation build by removing an assertion from `m.css` library that we use.
- https://github.com/mosra/m.css/issues/190 is now fixed upstream and we should be able to use `m.css` without any modifications.

Remove following line:
https://github.com/DARMA-tasking/vt/blob/9d2e8abf6e75b5a95eda01adf1d3682e795bd57d/ci/build_cpp.sh#L159

cz4rs:
fixes #1652

### DIFF
--- a/ci/build_cpp.sh
+++ b/ci/build_cpp.sh
@@ -156,8 +156,6 @@ then
     git checkout master
     cd ../
 
-    sed -i 's/assert len(self.children) < 16 and len(self.results) < 2048/\#removed assertion/' ./m.css/documentation/_search.py
-
     "$MCSS/documentation/doxygen.py" Doxyfile-mcss
     cp -R docs "$GHPAGE"
     cd "$GHPAGE"


cz4rs:
Note: it seems that there's another issue with the documentation ongoing ([logs](https://github.com/DARMA-tasking/vt/runs/4746707177?check_suite_focus=true)).
I will take a look at that failing build as well.

jstrzebonski (approved):
As I understand the original problem with m.css was fixed. In that case - approve.

JacobDomagala (approved):

PhilMiller (approved):

PR #1657, opened 2022-01-13
1495: fix Alpine build

### LINKED ISSUE #1495: Replace Alpine CI testing with Red Hat/CentOS
**What Needs to be Done?**

Alpine has some issues that aren't worth working around, so we want to replace it.

cz4rs:
fixes #1495

Fixes the stack dumping issues with Alpine build by replacing `libexecinfo` with `libunwind`.
Does a minor cleanup of our backtrace-generating code as well.

---
~cc52339 is the simplest workaround to the failing test - it just disables stack printing. I will try to come up with an actual fix.~
~Confirmed that it works with:~
```
ctest --output-on-failure --repeat-until-fail 100 -R TestMpiAccessGuardDeathTest.test_mpi_access_prevented_proc_2
```
~(unfixed version crashes every couple of runs).~

### DIFF
--- a/ci/deps/libunwind.sh
+++ b/ci/deps/libunwind.sh
@@ -24,7 +24,8 @@ cd ${libunwind_name}
 ./configure \
     --enable-static \
     --enable-shared \
-    --enable-setjmp=no
+    --enable-setjmp=no \
+    --prefix=`/usr/lib/x86_64-linux-gnu/`
 make
 make install
 cd -
--- a/ci/deps/libunwind.sh
+++ b/ci/deps/libunwind.sh
@@ -1,0 +1,31 @@
+#!/usr/bin/env bash
+
+set -exo pipefail
+
+if test $# -lt 1
+then
+    echo "usage: ./$0 <libunwind-version>"
+    exit 1
+fi
+
+libunwind_version=$1
+libunwind_name="libunwind-${libunwind_version}"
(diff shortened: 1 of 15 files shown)


cz4rs:
Related bug report in Alpine's GitLab:
https://gitlab.alpinelinux.org/alpine/aports/-/issues/10896

cz4rs:
TODO:
- expand 5a33c277fd7a34bd30de64ab6242afa0c9a12527 to handle errors correctly (mostly early returns)
- handle introduced FIXMEs
- investigate the errors in `TestPreconfig.test_vt_assert` and `TestCliArguments.test_vt_assert`

PhilMiller (approved):

src/vt/configs/error/pretty_print_stack.h
```diff
@@ -52,7 +52,7 @@
 
 namespace vt { namespace debug { namespace stack {
 
-inline std::string prettyPrintStack(StackVectorType const& stack) {
```
PhilMiller, line 55: This is going to conflict with #1670 which moves this function out of this file, and deletes the file.

PhilMiller (approved):

PhilMiller:
It looks like this is going to cause trouble for EMPIRE, at least as is - I just imported current `vt` to test #1683, and found that the build failed to find `libunwind.h`. Should we be vendoring it into our `lib/` directory?

PR #1658, opened 2022-01-14
#1636 add API for temporarily enabling/disabling debug prints

### LINKED ISSUE #1636: add API to temporarily enable/disable debug prints
**What Needs to be Done?**
It might be useful for an application to only care about certain prints during a particular operation. In addition to globally enabling a debug print (e.g. `--vt_debug_term`), we could allow these flags to be enabled/disabled during program execution. This could help reduce noise when users are debugging their applications.

jstrzebonski:
Fixes #1636

### DIFF
--- a/src/vt/configs/debug/debug_print.cc
+++ b/src/vt/configs/debug/debug_print.cc
@@ -1,0 +1,107 @@
+/*
+//@HEADER
+// *****************************************************************************
+//
+//                                debug_print.cc
+//                       DARMA/vt => Virtual Transport
+//
+// Copyright 2019-2021 National Technology & Engineering Solutions of Sandia, LLC
+// (NTESS). Under the terms of Contract DE-NA0003525 with NTESS, the U.S.
+// Government retains certain rights in this software.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+// * Redistributions of source code must retain the above copyright notice,
+//   this list of conditions and the following disclaimer.
+//
+// * Redistributions in binary form must reproduce the above copyright notice,
+//   this list of conditions and the following disclaimer in the documentation
+//   and/or other materials provided with the distribution.
+//
+// * Neither the name of the copyright holder nor the names of its
+//   contributors may be used to endorse or promote products derived from this
+//   software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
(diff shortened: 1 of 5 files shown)


jstrzebonski:
@nmm0 Is it something that might come in handy for your use cases?

nmm0:
> @nmm0 Is it something that might come in handy for your use cases (take a look at simplified usage in unit tests)?

Yes, this is perfect, especially the scoped version. In my use cases I often want to get the prints from a specific function.

nmm0 (approved):
Looks good to me! Only thing I'd want to add is some documentation just to make developers aware of the feature.

jstrzebonski:
> Looks good to me! Only thing I'd want to add is some documentation just to make developers aware of the feature.

Documentation added.

JacobDomagala (approved):

src/vt/configs/debug/debug_print.h
```diff
@@ -159,6 +159,103 @@
 
 #define vt_option_check_enabled(mode, bit) ((mode & bit) not_eq 0)
 
+/**
+ * \brief Temporarily enable debug print option
+*/
+#define vt_debug_temp_enable_opt(opt) \
+  ::vt::debug::preConfigRef()->vt_debug_##opt = true
```
cz4rs, line 166: I wonder if it makes sense to use `preConfigRef()` instead of simply `theConfig() here.

cz4rs: If we stick with this, consider following scenario:
```cpp
vt_debug_temp_enable_opt(all);

vt::initialize();

EXPECT_EQ(theConfig()->vt_debug_all, true); // or false? vt::initialize might overwrite the value with default
```

src/vt/runtime/runtime_get.cc
```diff
@@ -168,8 +168,16 @@ static arguments::AppConfig preInitAppConfig{};
  *
  * \return A modifiable configuration
  */
-arguments::AppConfig* preConfigRef(){
-  return &preInitAppConfig;
+arguments::AppConfig* preConfigRef() {
```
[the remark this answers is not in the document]
jstrzebonski: This whole change is now reverted.

src/vt/configs/debug/debug_print.h
jstrzebonski: I'm using `theConfig()` instead of `preConfigRef()`, and I added comment about necessity of vt initialization.

cz4rs (approved):
Looks good to me!

lifflander (changes requested):
This looks great to me. The only change I would suggest is checking to see if VT is live, and then printing an appropriate error/exception/abort if it isn't.

jstrzebonski:
> This looks great to me. The only change I would suggest is checking to see if VT is live, and then printing an appropriate error/exception/abort if it isn't.

Done

src/vt/configs/debug/debug_print.cc
```diff
@@ -0,0 +1,107 @@
...
+      : orig_val{theConfig()->vt_debug_##opt} {                                \
+    theConfig()->vt_debug_##opt = val;                                         \
+  }                                                                            \
+                                                                               \
+  ScopedModifier_##opt##_##val::~ScopedModifier_##opt##_##val() {              \
+    theConfig()->vt_debug_##opt = orig_val;                                    \
+  }                                                                            \
+                                                                               \
+  ScopedModifier_##opt##_##val createScopedModifier_##opt##_##val() {          \
+    if (not curRT) {                                                           \
```
jstrzebonski, line 58: Is there a better way to tell if vt's been initialized already?

lifflander: You could just check `if (vt::theConfig())` which might make a little more sense in this context.

src/vt/configs/debug/debug_print.h
```diff
@@ -159,6 +159,111 @@
...
+/**
+ * \brief Temporarily enable debug print option.
+ * vt needs to be initialized before usage.
+ */
+#define vt_debug_temp_enable_opt(opt)                                          \
+  if (not curRT) {                                                             \
+    vtAbort("Trying to read config when VT is not initialized");               \
+  }                                                                            \
+                                                                               \
+  theConfig()->vt_debug_##opt = true
```
lifflander, line 171: We need to qualify all of these calls to `theConfig()` with `::vt::theConfig()` because they might not be in the VT namespace for application code.

lifflander (changes requested):

src/vt/configs/debug/debug_print.h
jstrzebonski: Done.

src/vt/configs/debug/debug_print.cc
jstrzebonski: I agree, checking that looks better in this context, but internally it uses two pointers before it returns config, so if vt is not initialized, I think the program would (at best) segfault.

```cpp
vt::arguments::AppConfig *theConfig() { return &CUR_RT->theArgConfig->config_; }
```

jstrzebonski:
There are still some funky compilation errors with fcontext and unity build enabled.

lifflander (approved):
Looks good once we can get CI passing!

PR #1664, opened 2022-01-28
1660 Associate proxy with chain set to track automatically

### LINKED ISSUE #1660: Associate collection proxy with CollectionChainSet to automatically track indices
**What Needs to be Done?**

One of the motivations for this is bvh's insertable collections, so we can track the indices that exist without manually hooking into the constructor/destructor/migration hooks. We also want to provide a enum to either track the local indices or the home indices. By default, we want this to be the local indices for performance reasons.

lifflander:
Fixes #1660

### DIFF
--- a/src/vt/messaging/collection_chain_set.impl.h
+++ b/src/vt/messaging/collection_chain_set.impl.h
@@ -86,12 +86,15 @@ CollectionChainSet<Index>::CollectionChainSet(
       }
       break;
     case ElementEventEnum::ElementDestroyed:
-    case ElementEventEnum::ElementMigratedOut:
       if (chains_.find(idx) != chains_.end()) {
         removeIndex(idx);
-      } else if (event != ElementEventEnum::ElementMigratedOut) {
+      } else {
         vtAssert(layout == Home, "Must be a home layout");
         p[home].template send<IdxMsg, &ThisType::removeIndexHan>(idx);
+      }
+    case ElementEventEnum::ElementMigratedOut:
+      if (layout == Local) {
+        removeIndex(idx);
       }
       break;
     }
--- a/src/vt/messaging/collection_chain_set.impl.h
+++ b/src/vt/messaging/collection_chain_set.impl.h
@@ -65,14 +65,22 @@ CollectionChainSet<Index>::CollectionChainSet(
   auto p = theObjGroup()->makeCollective<CollectionChainSet<Index>>(this);
 
   auto const this_node = theContext()->getNode();
+  auto const proxy_bits = proxy.getProxy();
 
   ListenerType l = [=](ElementEventEnum event, IndexT idx, NodeType home) {
     switch (event) {
     case ElementEventEnum::ElementCreated:
     case ElementEventEnum::ElementMigratedIn:
       if (layout == Local or (layout == Home and home == this_node)) {
-        addIndex(idx);
-      } else {
+        if (chains_.find(idx) == chains_.end()) {
+          addIndex(idx);
+        } else {
+          vtAssert(
+            layout == Home and event == ElementEventEnum::ElementMigratedIn,
+            "Must be home layout and migrated in"
+          );
+        }
+      } else if (event != ElementEventEnum::ElementMigratedIn) {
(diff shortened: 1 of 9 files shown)


src/vt/messaging/collection_chain_set.impl.h
```diff
@@ -0,0 +1,104 @@
...
+
+  ListenerType l = [=](ElementEventEnum event, IndexT idx, NodeType home) {
+    switch (event) {
+    case ElementEventEnum::ElementCreated:
+    case ElementEventEnum::ElementMigratedIn:
+      if (layout == Local or (layout == Home and home == this_node)) {
+        addIndex(idx);
+      } else {
+        vtAssert(layout == Home, "Must be a home layout");
+        p[home].template send<IdxMsg, &ThisType::addIndexHan>(idx);
```
PhilMiller, line 77: I think this only makes sense for the `ElementCreated` case? Otherwise, won't the home process be told to add the index every time the object migrates?

src/vt/messaging/collection_chain_set.impl.h
```diff
@@ -0,0 +1,104 @@
...
+      }
+      break;
+    case ElementEventEnum::ElementDestroyed:
+    case ElementEventEnum::ElementMigratedOut:
+      if (chains_.find(idx) != chains_.end()) {
+        removeIndex(idx);
+      } else {
+        vtAssert(layout == Home, "Must be a home layout");
+        p[home].template send<IdxMsg, &ThisType::removeIndexHan>(idx);
```
PhilMiller, line 86: Similarly, I think this only applies to the `ElementDeleted` case

src/vt/messaging/collection_chain_set.impl.h
```diff
@@ -0,0 +1,104 @@
...
+      break;
+    }
+  };
+  auto listener = theCollection()->template registerElementListener<ColT>(
+    proxy.getProxy(), l
+  );
+
+  deallocator_ = [=]{
+    theCollection()->template unregisterElementListener<ColT>(
+      proxy.getProxy(), listener
```
PhilMiller, line 97: As a tiny nitpick, I think it'd be preferable to only capture the result of `proxy.getProxy()` rather than `proxy` itself, since this `deallocator` is going to stick around.

PhilMiller:
I'm impressed with how simple this turned out to be

src/vt/messaging/collection_chain_set.impl.h
lifflander: Good point. I missed that when I started collapsing the cases.

lifflander: Fixed

src/vt/messaging/collection_chain_set.impl.h
lifflander: Fixed

src/vt/messaging/collection_chain_set.impl.h
lifflander: Done

src/vt/messaging/collection_chain_set.impl.h
```diff
@@ -0,0 +1,129 @@
...
+  static_assert(std::is_same<IndexT, Index>::value, "Must match index type");
+
+  // Make this chain set an objgroup instance so we can send updates
+  auto p = theObjGroup()->makeCollective<CollectionChainSet<Index>>(this);
+
+  auto const this_node = theContext()->getNode();
+  auto const proxy_bits = proxy.getProxy();
+
+  ListenerType l = [=](ElementEventEnum event, IndexT idx, NodeType home) {
+    switch (event) {
```
PhilMiller, line 71: Could you see if this code looks better if the `if (layout == Home)` were done outside the switch?

PhilMiller:
This all seems fine now.

src/vt/messaging/collection_chain_set.impl.h
lifflander: I tried this but I think it make the code even uglier.

PhilMiller (approved):

PR #1675, opened 2022-02-15
1669: Move the duplicated setup code for LB stats into separate function

### LINKED ISSUE #1669: Move the duplicated setup code for `LBStats` into separate function
See https://github.com/DARMA-tasking/vt/pull/1494#pullrequestreview-871271999

JacobDomagala:
Fixes #1669

### DIFF
--- a/src/vt/messaging/active.cc
+++ b/src/vt/messaging/active.cc
@@ -448,7 +448,7 @@ EventType ActiveMessenger::sendMsgBytes(
     theTerm()->hangDetectSend();
   }
 
-  recordLbStatsCommForSend(dest, base);
+  recordLbStatsCommForSend(dest, base, msg_size);
 
   return event_id;
 }
@@ -512,7 +512,7 @@ EventType ActiveMessenger::doMessageSend(
     if (dest != this_node) {
       sendMsgBytesWithPut(dest, base, send_tag);
     } else {
-      recordLbStatsCommForSend(dest, base);
+      recordLbStatsCommForSend(dest, base, base.size());
 
       runnable::makeRunnable(base, true, envelopeGetHandler(msg->env), dest)
         .withTDEpochFromMsg(is_term)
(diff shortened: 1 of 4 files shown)


src/vt/messaging/active.cc
```diff
@@ -449,17 +449,7 @@ EventType ActiveMessenger::sendMsgBytes(
...
   }
 
-  if (theContext()->getTask() != nullptr) {
-    auto lb = theContext()->getTask()->get<ctx::LBStats>();
-    if (lb) {
-      auto const already_recorded =
-        envelopeCommStatsRecordedAboveBareHandler(msg->env);
-      if (not already_recorded) {
-        auto dest_elm_id = elm::ElmIDBits::createBareHandler(dest);
-        theContext()->getTask()->send(dest_elm_id, msg_size);
```
PhilMiller, line 459: Does `msg_size == base.size()`? We're getting `msg_size` as an argument, so I don't immediately see.

JacobDomagala: I think it's not always the case, I'll fix that. Thanks

src/vt/messaging/active.h
```diff
@@ -1725,6 +1725,18 @@ struct ActiveMessenger : runtime::component::PollableComponent<ActiveMessenger>
...
+  /**
+   * @brief Record LB's statistics for sending a message
+   *
+   * \param[in] dest the destination of the message
+   * \param[in] base the message base pointer
+   * \param[in] msg_size the size of the message being sent
+   */
+  void recordLbStatsCommForSend(
+    NodeType const& dest, MsgSharedPtr<BaseMsgType> const& base,
+    MsgSizeType const& msg_size
```
jstrzebonski, line 1737: I think that dest and msg_size could be passed by value.

cz4rs (approved):
Looks good to me :+1:

lifflander (approved):
Looks good to me!

PR #1676, opened 2022-02-15
1484: fix memory footprint tests

### LINKED ISSUE #1484: Check why Memory footprint tests are failing on non Unity build
**Describe the bug**
When using vt with fcontext (`VT_FCONTEXT_ENABLED`) and non Unity build, it seems that Memory Footprint tests are failing. 
See https://github.com/DARMA-tasking/vt/pull/1481#issuecomment-864025696

cz4rs:
fixes #1484


cz4rs:
a0f5b9a50ef8cc8cb5ab3ce09bca37ca28a0eede adds seemingly reasonable test verifying that `passthru_args` vector is empty after initialization. This fails on `gcc-6` build (and passes fine everywhere else), where the only meaningful difference is that unity build is disabled.

Note: I was able to reproduce this locally with clang 13, so definitely not a compiler issue.

cz4rs:
In the end, this PR is addressing 2 issues:
- function pointers being dereferenced during serialization (fixed by using `countBytes`)
- header issue described in #1702 (fixed by removing majority of includes)

Other changes are a small cleanup, updating workflow definitions and removing tests performing footprinting of single components (these did not add much value, since the coverage is provided by testing `Runtime::printMemoryFootprint()`).

PhilMiller (approved):

PR #1678, opened 2022-02-15
1673 runInEpochRooted should by default use a DS epoch

### LINKED ISSUE #1673: runInEpochRooted should by default use a DS epoch
**Describe the bug**
`runInEpochRooted` should not be using a wave-based epoch. The PhaseManager has rooted hooks registered that calls this function, which adds a ton of overhead for no reason. `runInEpochRooted` should always use a DS epoch I think.

nlslatt:
Closes #1673

### DIFF
--- a/src/vt/termination/termination.h
+++ b/src/vt/termination/termination.h
@@ -213,7 +213,7 @@ public:
    * \return the new epoch
    */
   EpochType makeEpochRooted(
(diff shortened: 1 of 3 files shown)


nlslatt:
This will need to merged into main as well.

PhilMiller (approved):

PR #1683, opened 2022-02-22
1677: dont rerun cmake for git ref

### LINKED ISSUE #1677: Stop using cmake code to generate git header information about sha, etc.
**What Needs to be Done?**

This causes cmake to run every time something changes at all. We want to just add a custom command that updates a cc file that is linked into the VT library that contains only the SHA (not the description or whether it is dirty or not).

nmm0:
Fixes #1677

This PR uses `add_custom_command` to act as a source file generator for `src/vt/configs/generated/vt_git_revision.cc`. It runs a cmake script (adapted from the earlier script) when any of: `.git/HEAD`, `.git/packed-refs` or `.git/refs/<named-branch-ref>` are modified. If they are modified, only vt_git_revision.cc will be rebuilt, and VT will need to be relinked. This will happen automatically. This avoids re-running cmake multiple times for no reason.

This additionally works correctly if a new branch was made and other changes to git occur.


cmake/build_git_info.cmake
```diff
@@ -1,34 +1,45 @@
...
+        set(GIT_PREVIOUS_PARENT "${GIT_PARENT_DIR}")
+        get_filename_component(GIT_PARENT_DIR ${GIT_PARENT_DIR} PATH)
+        if(GIT_PARENT_DIR STREQUAL GIT_PREVIOUS_PARENT)
+            # We have reached the root directory, we are not in git
+            set(${_refspecvar} "GITDIR-NOTFOUND" PARENT_SCOPE)
+            set(${_hashvar} "GITDIR-NOTFOUND" PARENT_SCOPE)
+            return()
+        endif()
+        set(GIT_DIR "${GIT_PARENT_DIR}/.git")
+    endwhile()
```
PhilMiller, line 14: This potentially runs into trouble with `git worktree` or the like.

I know this would be a bigger change, but could we just let Git itself do this searching as it normally would? I think the command we  want is `git rev-parse --verify HEAD`, which outputs the full SHA value.

cmake/run-git.cmake
```diff
@@ -0,0 +1,97 @@
...
+if(res EQUAL 0)
+    set(GIT_CLEAN_STATUS "CLEAN")
+else()
+    set(GIT_CLEAN_STATUS "DIRTY")
+endif()
+
+message(STATUS "Git Clean Status: \"${GIT_CLEAN_STATUS}\"")
+
+message(STATUS "Configuring ${IN_FILE} to generate ${OUT_FILE}.")
+configure_file(${IN_FILE} ${OUT_FILE} @ONLY)
```
PhilMiller, line 97: It's not clear to me that this meets the spec of only modifying the output file if its contents have actually changed. If `HEAD` is unchanged, then there should be no recompilation at all.

cmake/run-git.cmake
```diff
@@ -0,0 +1,97 @@
...
+        OUTPUT_STRIP_TRAILING_WHITESPACE)
+if(NOT res EQUAL 0)
+    message(FATAL_ERROR "could not get the description")
+endif()
+
+message(STATUS "GIT_DESCRIPTION: \"${GIT_DESCRIPTION}\"")
+
+execute_process(COMMAND
+        "${GIT_EXECUTABLE}"
+        diff-index --quiet HEAD --
```
PhilMiller, line 82: There were requests as part of the spur for this overall issue to omit checking for clean/dirty, since recursive filesystem traversal on (e.g.) NFS or GPFS, or worse Lustre, can be painfully slow.

Maybe we can make it optional, so EMPIRE and other production-system builds, CI builds, and so forth can do without?

cmake/run-git.cmake
nmm0: The script will only be invoked through `add_custom_command` so it will only be invoked if the dependencies change. In the current version, that means if the .git folder is modified in any way.

cmake/build_git_info.cmake
nmm0: That sounds a lot less clunky, I just used our previous code for detecting the folder. Would this work in detached head/other corner cases?

cmake/run-git.cmake
nmm0: That can be done via a cmake flag. I'll add that

cmake/build_git_info.cmake
PhilMiller: Detached HEAD is barely even an edge case, compared to worktree. Yeah, it will work fine.

If you're doing the file dependency thing, you may need the code to find the relevant `.git/` anyway. Let me see if there's a git plumbing command for that.

nmm0: I found `rev-parse --git-dir` and that seems to work well. I believe it should also work if it is a submodule. Anyway added that and rev-parse --verify HEAD, and I believe the other issues we will postpone for another PR

cmake/build_git_info.cmake
```diff
@@ -1,34 +1,34 @@
...
-#   DESTINATION      include
-#   CONFIGURATIONS   ${cur_build_type}
-# )
-
-# configure_file(
-#   "${PROJECT_SOURCE_DIR}/vt_git_revision.cc.in"
-#   "${CMAKE_CURRENT_BINARY_DIR}/vt_git_revision.cc"
-#   @ONLY
-# )
+target_sources(${VIRTUAL_TRANSPORT_LIBRARY} PRIVATE ${VT_GIT_CONFIG_FILE})
```
PhilMiller, line 34: Missing EOF newline

cmake/run-git.cmake
```diff
@@ -0,0 +1,100 @@
...
+if(res EQUAL 0)
+    set(GIT_CLEAN_STATUS "CLEAN")
+else()
+    set(GIT_CLEAN_STATUS "DIRTY")
+endif()
+
+message(STATUS "Git Clean Status: \"${GIT_CLEAN_STATUS}\"")
+
+message(STATUS "Configuring ${IN_FILE} to generate ${OUT_FILE}.")
+configure_file(${IN_FILE} ${OUT_FILE} @ONLY)
```
PhilMiller, line 100: Missing EOF newline here, too

cmake/build_git_info.cmake
nmm0: fixed

cz4rs (approved):
This looks good code-wise, still requires confirmation from Phil about the intended behavior.

cmake/run-git.cmake
nmm0: For the future: we might consider a more targeted approach, but right now this seems to be the only workable approach.

cmake/run-git.cmake
nmm0: The flag will be added once we sort out how we want to handle the absence of git information, since in some cases it might not even be desirable for the application to obtain any git info at all.

PR #1694, opened 2022-03-07
#1687 encode RegistryTypeEnum in HandlerType bits

### LINKED ISSUE #1687: Encode the auto_registry::RegistryTypeEnum directly in HandlerType bits
**What Needs to be Done?**

We can reduce a lot of code that makes assumptions and has to know based on context which handler type is being called by encoding the enum type bits directly in the handler. This is related to the bug #1685 because they get out of alignment.

jstrzebonski:
Fixes #1687

### DIFF
--- a/src/vt/handler/handler.cc
+++ b/src/vt/handler/handler.cc
@@ -47,7 +47,8 @@
 namespace vt {
 
 /*static*/ HandlerType HandlerManager::makeHandler(
-  bool is_auto, bool is_functor, HandlerIdentifierType id, bool is_objgroup,
+  bool is_auto, bool is_functor, HandlerIdentifierType id,
+  RegistryTypeEnum const registry_type, bool is_objgroup,
   HandlerControlType control, bool is_trace, bool is_member,
   bool is_base_msg_derived
 ) {
@@ -57,6 +58,7 @@ namespace vt {
   HandlerManager::setHandlerFunctor(new_han, is_functor);
   HandlerManager::setHandlerMember(new_han, is_member);
   HandlerManager::setHandlerBaseMsgDerived(new_han, is_base_msg_derived);
+  HandlerManager::setHandlerRegistryType(new_han, registry_type);
   HandlerManager::setHandlerIdentifier(new_han, id);
 
 #if vt_check_enabled(trace_enabled)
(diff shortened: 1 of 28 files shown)


src/vt/registry/auto/auto_registry_type_enum.h
```diff
@@ -0,0 +1,64 @@
+/*
+//@HEADER
+// *****************************************************************************
+//
+//                            auto_registry_common.h
```
jstrzebonski, line 5: Wrong file name.

jstrzebonski: Done.

src/vt/handler/handler.cc
```diff
@@ -47,7 +47,8 @@
 namespace vt {
 
 /*static*/ HandlerType HandlerManager::makeHandler(
-  bool is_auto, bool is_functor, HandlerIdentifierType id, bool is_objgroup,
+  bool is_auto, bool is_functor, HandlerIdentifierType id,
+  RegistryTypeEnum const registry_type, bool is_objgroup,
```
lifflander, line 51: Can we remove the two flags `is_objgroup` and `is_member` with the handler enum type encoded in the bits?

jstrzebonski: Yes, I think so.

I think I could even remove bits that encodes that values (`eHandlerBits::ObjGroup` and `Member`) as currently they're strictly bind with reg type. And for `isHandlerMember()` and `isHandlerObjGroup()`, I could just check reg type and return based on that, something like that:

```cpp
/*static*/ bool HandlerManager::isHandlerObjGroup(HandlerType han) {
  auto const reg_type = getHandlerRegistryType(han);
  auto const isObjGroup = reg_type == RegistryTypeEnum::RegObjGroup;
  return isObjGroup;
}

/*static*/ bool HandlerManager::isHandlerMember(HandlerType han) {
  auto const reg_type = getHandlerRegistryType(han);
  auto const isMember = reg_type ==  RegistryTypeEnum::RegVrtCollectionMember;
  return isMember;
}
```

lifflander: I think this would reduce the complexity even more then!

jstrzebonski: Done.

src/CMakeLists.txt
```diff
@@ -289,7 +289,8 @@ if(CODE_COVERAGE AND CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
   message(STATUS "VT: enabling code coverage generation")
   # Add required flags (GCC & LLVM/Clang)
   target_compile_options(${VIRTUAL_TRANSPORT_LIBRARY} PUBLIC
-    -Og
+    -O0        # no optimization
```
lifflander, line 292: Why does this change things back?

jstrzebonski: Done.

lifflander (approved):
Looks good!

PhilMiller (approved):

PR #1700, opened 2022-03-14
1699: Update `.clang-format` syntax

### LINKED ISSUE #1699: migrate `.clang-format` to Clang 13 syntax
**What Needs to be Done?**
`IndentPPDirectives: PPDIS_None` is no longer recognized as valid configuration in Clang 13. Use updated value for that option (`None`).

cz4rs:
- Update .clang-format to match Clang 13 syntax.
- Add post-commit hook to format modified code
  - requires `git clang-format` plugin
  - post-commit works better for me, but this could be easily adapted to pre-commit as well

fixes #1699

### DIFF
--- a/scripts/post-commit
+++ b/scripts/post-commit
@@ -1,6 +1,8 @@
 #!/bin/sh
 #
-# Apply formatting style defined in .clang-format on committed changes.
+# An example hook script to apply formatting style defined in .clang-format
+# on committed changes. Copy this to `.git/hooks/` to have it run after
+# every commit.
 # Prints the resulting changes if there are any.
 
 output=$(git clang-format HEAD~1)
--- a/scripts/post-commit
+++ b/scripts/post-commit
@@ -1,0 +1,11 @@
+#!/bin/sh
+#
+# Apply formatting style defined in .clang-format on committed changes.
+# Prints the resulting changes if there are any.
+
+output=$(git clang-format HEAD~1)
+
+if [ "$output" != "" ]
+then
+    git diff
+fi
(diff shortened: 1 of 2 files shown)


scripts/post-commit
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+#
+# Apply formatting style defined in .clang-format on committed changes.
+# Prints the resulting changes if tere are any.
```
jstrzebonski, line 4:
```suggestion
# Prints the resulting changes if there are any.
```

jstrzebonski (approved):
👍🏻 Checked with my clang-format 12.
As for the script, some folks might find it handy, thanks.

nmm0 (commented):
I had to also change the following values:

```
AlignConsecutiveAssignments: None
AlignConsecutiveDeclarations: None
AlignConsecutiveMacros: None
SortIncludes: Never
```

cz4rs:
> I had to also change the following values:
> 
> ```
> AlignConsecutiveAssignments: None
> AlignConsecutiveDeclarations: None
> AlignConsecutiveMacros: None
> SortIncludes: Never
> ```

I will update those and check with Clang 13 👍

cz4rs:
@nmm0 Update the remaining values. Works fine for me, but older version (clang-format-10) didn't like it.
If anyone with an older setup complains we can discuss the solution. For now, I think targeting newer version is the right choice.

jstrzebonski:
My clang-format (12.0.0) doesn't really like `SortIncludes: Never` ;- ) but I can live with that.

cz4rs:
> Does this post-commit hook run automatically or how does it work?

You need to copy/link it to `.git/hooks/`.

JacobDomagala (approved):

PR #1712, opened 2022-03-21
1423 remove maximimum ref-count check

### LINKED ISSUE #1423: Message ref-count limit triggers in simple cases

stmcgovern:
Fixes #1423 by simply removing the upper bound check in the assertion.
Alternative to PR #1432


stmcgovern:
Line 65 comment should be amended to "Message ref-count must never be negative."

lifflander (approved):
Looks good to me! Just need to get the warnings worked out.

PhilMiller:
Code looks fine to me.

stmcgovern:
> Looks good to me! Just need to get the warnings worked out.

- They seem to be related to type conversions (mostly in pipe manager) only when using cuda.
- Here in /vt/src/vt/configs/types/types_sentinels.h(78): warning: integer conversion resulted in a change of sign
`static constexpr RefType const not_shared_message                  = -1000;`
Since that RefType is now uint16  initializing to -1000 no longer makes sense, right?

PR #1718, opened 2022-03-24
#1717 LB: add .br extension for compressed data files

### LINKED ISSUE #1717: Use proper brotli extension for compressed json files
**What Needs to be Done?**

Brotli will not decompress unless one has the proper extension for the file names. Change the output to reflect the proper naming convention.

lifflander:
Fixes #1717

### DIFF
--- a/src/vt/configs/arguments/args.cc
+++ b/src/vt/configs/arguments/args.cc
@@ -776,9 +776,7 @@ std::tuple<int, std::string> ArgConfig::parseToConfig(
 }
 
 namespace {
-static std::string buildFile(
-  std::string const& file, std::string const& dir, bool add_br_extension
-) {
+static std::string buildRankFile(std::string const& file) {
   std::string name = file;
   std::size_t rank = name.find("%p");
   auto str_rank = std::to_string(theContext()->getNode());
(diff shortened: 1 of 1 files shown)


src/vt/configs/arguments/args.cc
```diff
@@ -785,17 +787,22 @@ static std::string buildFile(std::string const& file, std::string const& dir) {
   } else {
     name.replace(rank, 2, str_rank);
   }
+  if (add_br_extension) {
+    if (name.substr(name.length()-3, 3) != ".br") {
```
PhilMiller, line 791: If only we could use C++20, this is just `name.ends_with(".br")`

src/vt/configs/arguments/args.cc
```diff
@@ -785,17 +787,22 @@ static std::string buildFile(std::string const& file, std::string const& dir) {
...
 } /* end anon namespace */
 
 std::string AppConfig::getLBStatsFileOut() const {
-  return buildFile(vt_lb_stats_file, vt_lb_stats_dir);
+  return buildFile(vt_lb_stats_file, vt_lb_stats_dir, vt_lb_stats_compress);
 }
 
 std::string AppConfig::getLBStatsFileIn() const {
-  return buildFile(vt_lb_stats_file_in, vt_lb_stats_dir_in);
+  return buildFile(vt_lb_stats_file_in, vt_lb_stats_dir_in, false);
```
PhilMiller, line 805: Are we never going to read compressed input?

PhilMiller (approved):
Clear, necessary fix

JacobDomagala (approved):

src/vt/configs/arguments/args.cc
lifflander: Yes, the input file that is specified is user defined. Adding `.br` doesn't make sense in this case I think because the user specified it. However, maybe the default file name should change if the user's data is compressed? That does seem a bit complex to implement given that we don't know if it's compressed or not until we try to read it. I could look for both file names I suppose? What do you think?

PhilMiller: I think looking for both compressed and not would be appropriate.

I think that should happen here in this method?

marcinwrobel1986: Hello,
just FYI, this is an implementation in LBAF:
```python
with open(file_name, "rb") as compr_json_file:
    compr_bytes = compr_json_file.read()
    try:
        decompr_bytes = brotli.decompress(compr_bytes)
        decompressed_dict = json.loads(decompr_bytes.decode("utf-8"))
    except brotli.error:
        decompressed_dict = json.loads(compr_bytes.decode("utf-8"))
```

jstrzebonski, line 805: I would introduce some local const var `add_br_extension` to more explicitly show what that third parameter does. In the PR you can see clearly what is going on, but that's not always the case when reading the whole source code.

src/vt/configs/arguments/args.cc
```diff
@@ -776,7 +776,9 @@ std::tuple<int, std::string> ArgConfig::parseToConfig(
 }
 
 namespace {
-static std::string buildFile(std::string const& file, std::string const& dir) {
+static std::string buildFile(
+  std::string const& file, std::string const& dir, bool add_br_extension
```
jstrzebonski, line 780: Personally, I don't like controlling functions' behavior by bool value. For me, it seems that there are two separate functions `buildFileName` and `buildBrFileName`. The latter could simply call `buildFileName()` to get the "base" name, and then do with it exactly what is done within the newly added if block.

lifflander: Done!

nlslatt (approved):

PR #1722, opened 2022-03-28
1721 Loosen assertions on unbounded collections

### LINKED ISSUE #1721: Loosen assertions/restrictions on unbounded collections with list insertion
**Describe the bug**
Currently, the assertion in the collection `makeCollection` construct params is too tight. List insertion should be allowed without bounds as long a the unbounded map is chosen by default in this case.

lifflander:
Fixes #1721


jstrzebonski (approved):
Looks good to me.

PR #1728, opened 2022-04-01
1723 Include user-defined fields in json LB data files

### LINKED ISSUE #1723: Include user-defined fields in json LB data files
Dump flagged entries from `Storable` LB data files in JSON format.

**What Needs to be Done?**

- [x] Add flag to `StoreElm` to determine which user-defined data to dump to json.
- [x] Add virtual `StoreElmBase::toJson()` that only works when type is jsonable.
- [x] Add `Storable::toJson()` that builds json for flagged data.
- [x] Pass a `Storable*` to `NodeStats::addNodeStats()` and have it add json to `StatsData`.
- [x] Include dumping of user-defined data json in `StatsData::toJson()`.
- [x] Add tests.

nlslatt:
This PR adds the ability to dump items stored in collection elements as user-defined data in LB data json files. I also added a commit that removes some stale lines of code from an existing unit test.

Closes #1723

### DIFF
--- a/src/vt/messaging/active.cc
+++ b/src/vt/messaging/active.cc
@@ -161,7 +161,7 @@ void ActiveMessenger::startup() {
   // Hook to collect LB data about objgroups
   thePhase()->registerHookCollective(phase::PhaseHook::End, [this]{
     theNodeLBData()->addNodeLBData(
-      bare_handler_dummy_elm_id_for_lb_data_, &bare_handler_lb_data_
+      bare_handler_dummy_elm_id_for_lb_data_, &bare_handler_lb_data_, nullptr
     );
   });
 #endif
--- a/src/vt/objgroup/manager.cc
+++ b/src/vt/objgroup/manager.cc
@@ -68,7 +68,7 @@ void ObjGroupManager::startup() {
         auto proxy = elm::ElmIDBits::getObjGroupProxy(elm_id.id, false);
         vtAssertExpr(proxy == obj.first);
         theNodeLBData()->registerObjGroupInfo(elm_id, obj.first);
-        theNodeLBData()->addNodeLBData(elm_id, &holder->getLBData());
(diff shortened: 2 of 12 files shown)


src/vt/vrt/collection/balance/stats_data.h
```diff
@@ -51,7 +51,7 @@
 #include <unordered_map>
 #include <memory>
 
-#include <nlohmann/json_fwd.hpp>
+#include <nlohmann/json.hpp>
```
lifflander, line 54: Can we avoid this include here? I see you are using `nlohmann::json` later in the file, could we make that a unique pointer or something? I've so far avoided all inclusions of `json.hpp` in any VT header files due to the burden.

lifflander (changes requested):

lifflander:
Overall, looks great!

src/vt/vrt/collection/balance/stats_data.h
nlslatt: I'm not sure how a unique pointer is supposed to help. I tried changing back the include and changing the data structure to be
```
  std::unordered_map<PhaseType, std::unordered_map<
    ElementIDStruct, std::unique_ptr<nlohmann::json>
  >> user_defined_json_;
```
but it fails to compile even if I comment out all code that actually uses `user_defined_json_`.

lifflander: Oh, maybe we need to move the serializer?

nlslatt: I even had the line in the serializer commented out.

nlslatt: I got it to compile with `shared_ptr`.

lifflander (approved):
Looks good to me!

lifflander (approved):
Looks good to me!

PR #1754, opened 2022-05-02
#1715 allow labels on object groups and collections

### LINKED ISSUE #1715: Allow labels on object collections and groups
**What Needs to be Done?**

Kokkos, Teuchos timers, and our epochs all take labels. Collections and object groups should too

We want to be able to print these labels in things that want to refer to particular entities in output

**Additional context**

jstrzebonski:
Fixes #1715

### DIFF
--- a/src/vt/messaging/collection_chain_set.impl.h
+++ b/src/vt/messaging/collection_chain_set.impl.h
@@ -62,7 +62,9 @@ CollectionChainSet<Index>::CollectionChainSet(
   static_assert(std::is_same<IndexT, Index>::value, "Must match index type");
 
   // Make this chain set an objgroup instance so we can send updates
-  auto p = theObjGroup()->makeCollective<CollectionChainSet<Index>>(this);
+  auto p = theObjGroup()->makeCollective<CollectionChainSet<Index>>(
+    "CollectionChainSet::CollectionChainSet()", this
+  );
 
   auto const this_node = theContext()->getNode();
   auto const proxy_bits = proxy.getProxy();
--- a/src/vt/messaging/collection_chain_set.impl.h
+++ b/src/vt/messaging/collection_chain_set.impl.h
@@ -63,7 +63,7 @@ CollectionChainSet<Index>::CollectionChainSet(
 
   // Make this chain set an objgroup instance so we can send updates
   auto p = theObjGroup()->makeCollective<CollectionChainSet<Index>>(
-    this, "CollectionChainSet::CollectionChainSet()"
+    this, "CollectionChainSet"
   );
 
   auto const this_node = theContext()->getNode();
--- a/src/vt/messaging/collection_chain_set.impl.h
+++ b/src/vt/messaging/collection_chain_set.impl.h
@@ -63,7 +63,7 @@ CollectionChainSet<Index>::CollectionChainSet(
 
   // Make this chain set an objgroup instance so we can send updates
   auto p = theObjGroup()->makeCollective<CollectionChainSet<Index>>(
-    "CollectionChainSet::CollectionChainSet()", this
+    this, "CollectionChainSet::CollectionChainSet()"
   );
 
   auto const this_node = theContext()->getNode();
--- a/src/vt/objgroup/manager.cc
+++ b/src/vt/objgroup/manager.cc
@@ -126,11 +126,7 @@ ObjGroupProxyType ObjGroupManager::makeCollectiveImpl(
     std::forward_as_tuple(proxy),
     std::forward_as_tuple(std::move(base))
   );
-  labels_.emplace(
-    std::piecewise_construct,
-    std::forward_as_tuple(proxy),
-    std::forward_as_tuple(label)
-  );
+  labels_.emplace(proxy, label);
 
   return proxy;
 }
--- a/src/vt/objgroup/manager.cc
+++ b/src/vt/objgroup/manager.cc
@@ -110,7 +110,7 @@ void ObjGroupManager::dispatch(MsgSharedPtr<ShortMessage> msg, HandlerType han) 
 }
 
 ObjGroupProxyType ObjGroupManager::makeCollectiveImpl(
-  HolderBasePtrType base, void* obj_ptr
+  std::string const& label, HolderBasePtrType base, void* obj_ptr
 ) {
   auto const id = cur_obj_id_++;
   auto const node = theContext()->getNode();
@@ -126,6 +126,12 @@ ObjGroupProxyType ObjGroupManager::makeCollectiveImpl(
     std::forward_as_tuple(proxy),
     std::forward_as_tuple(std::move(base))
   );
+  labels_.emplace(
+    std::piecewise_construct,
+    std::forward_as_tuple(proxy),
+    std::forward_as_tuple(label)
+  );
+
   return proxy;
 }
 
--- a/src/vt/vrt/collection/balance/lb_invoke/lb_manager.cc
+++ b/src/vt/vrt/collection/balance/lb_invoke/lb_manager.cc
@@ -77,7 +77,7 @@ namespace vt { namespace vrt { namespace collection { namespace balance {
 /*static*/ std::unique_ptr<LBManager> LBManager::construct() {
   auto ptr = std::make_unique<LBManager>();
   auto proxy = theObjGroup()->makeCollective<LBManager>(
-    ptr.get(), "LBManager::construct()"
+    ptr.get(), "LBManager"
   );
   proxy.get()->setProxy(proxy);
 
@@ -165,8 +165,8 @@ void LBManager::setLoadModel(std::shared_ptr<LoadModel> model) {
 
 template <typename LB>
 LBManager::LBProxyType
-LBManager::makeLB() {
-  auto proxy = theObjGroup()->makeCollective<LB>("LBManager::makeLB()");
+LBManager::makeLB(std::string const& lb_name) {
+  auto proxy = theObjGroup()->makeCollective<LB>(lb_name);
   auto strat = proxy.get();
   strat->init(proxy);
   auto base_proxy = proxy.template castToBase<lb::BaseLB>();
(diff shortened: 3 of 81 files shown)


src/vt/vrt/collection/param/construct_params.h
```diff
@@ -441,28 +449,32 @@ namespace vt {
...
  *
- * \param[in] bounds the bounds for the collection (optional)
+ * \param[in] label collection label
  *
  * \return the parameter configuration object
  */
 template <typename ColT>
-vrt::collection::param::ConstructParams<ColT> makeCollection() {
+vrt::collection::param::ConstructParams<ColT> makeCollection(
+  const std::string& label
```
lifflander, line 458: I think we should make this optional

jstrzebonski: Done.

src/vt/vrt/collection/manager.h
```diff
@@ -1625,14 +1643,16 @@ struct CollectionManager
...
    * \param[in] range the range of the collection to restart
    * \param[in] file_base the base file name for the files to read
    *
    * \return proxy to the new collection
    */
   template <typename ColT>
   CollectionProxyWrapType<ColT> restoreFromFile(
-    typename ColT::IndexType range, std::string const& file_base
+    std::string const& label, typename ColT::IndexType range,
+    std::string const& file_base
```
jstrzebonski, line 1655: I think here label should also be optional.

src/vt/vrt/collection/manager.h
```diff
@@ -278,14 +292,16 @@ struct CollectionManager
...
    * \param[in] range index range for the collection
    * \param[in] cons_fn construct function to create an element on each node
    *
    * \return proxy to the new collection
    */
   template <typename ColT>
   CollectionProxyWrapType<ColT> constructCollective(
-    typename ColT::IndexType range, DistribConstructFn<ColT> cons_fn
+    std::string const& label, typename ColT::IndexType range,
+    DistribConstructFn<ColT> cons_fn
```
jstrzebonski, line 304: I think here label should also be optional.

src/vt/objgroup/manager.cc
```diff
@@ -126,6 +126,8 @@ ObjGroupProxyType ObjGroupManager::makeCollectiveImpl(
     std::forward_as_tuple(proxy),
     std::forward_as_tuple(std::move(base))
   );
+  labels_.emplace(proxy, label);
```
lifflander, line 129: I think we are missing the deletion of the label.

src/vt/vrt/collection/manager.h
```diff
@@ -193,7 +198,9 @@ struct CollectionManager
     typename ColT, mapping::ActiveMapTypedFnType<typename ColT::IndexType> fn
   >
   CollectionProxyWrapType<ColT, typename ColT::IndexType>
-  construct(typename ColT::IndexType range);
+  construct(
```
lifflander, line 201: Since we are going to be deprecating these calls, we probably don't really need to add them here. But.. it's fine as long as we don't also break the interface.

src/vt/vrt/collection/manager.h
```diff
@@ -202,13 +209,16 @@ struct CollectionManager
...
    *
+   * \param[in] label collection label
    * \param[in] range index range for the collection
    *
    * \return proxy to the new collection
    */
   template <typename ColT>
   CollectionProxyWrapType<ColT, typename ColT::IndexType>
-  construct(typename ColT::IndexType range);
+  construct(
```
lifflander, line 219: This looks like a breaking change to me.

src/vt/vrt/collection/manager.h
```diff
@@ -1627,12 +1645,14 @@ struct CollectionManager
...
    * \param[in] range the range of the collection to restart
    * \param[in] file_base the base file name for the files to read
+   * \param[in] label collection label
    *
    * \return proxy to the new collection
    */
   template <typename ColT>
   CollectionProxyWrapType<ColT> restoreFromFile(
-    typename ColT::IndexType range, std::string const& file_base
+    typename ColT::IndexType range, std::string const& file_base,
```
lifflander, line 1654: Instead of providing the label here, shouldn't we just write/read it from the file? I think it should be included in the checkpoint and then restored.

src/vt/vrt/collection/manager.h
```diff
@@ -1700,6 +1720,15 @@ struct CollectionManager
     std::string const& file_base
   );
 
+  /**
+   * \brief Get collection label
+   *
+   * \param[in] proxy Collection proxy
+   * \return collection lable
+   */
+  template <typename ColT>
```
lifflander, line 1729: We should probably make this call typeless `VirtualProxyType` since it doesn't rely on the type at all.

src/vt/vrt/collection/manager.h
```diff
@@ -1808,6 +1838,7 @@ struct CollectionManager
   VirtualIDType next_rooted_id_ = 0;
   TypelessHolder typeless_holder_;
   std::unordered_map<VirtualProxyType, SequentialIDType> reduce_stamp_;
+  std::unordered_map<VirtualProxyType, std::string> labels_;
```
lifflander, line 1841: This should be stored in the `TypelessHolder` or `BaseHolder`

src/vt/vrt/collection/manager.impl.h
```diff
@@ -1334,6 +1340,7 @@ void CollectionManager::insertMetaCollection(
   typeless_holder_.insertCollectionInfo(proxy, holder, [proxy]{
     theCollection()->constructGroup<ColT>(proxy);
```
lifflander, line 1341: Insertion can happen here into the typeless holder.

lifflander (changes requested):

src/vt/objgroup/manager.cc
jstrzebonski: Done.

src/vt/vrt/collection/manager.h
jstrzebonski: Done.

src/vt/vrt/collection/manager.h
jstrzebonski: Done.

src/vt/vrt/collection/manager.h
jstrzebonski: Done.

src/vt/vrt/collection/manager.h
jstrzebonski: Done.

src/vt/vrt/collection/manager.h
jstrzebonski: Done.

src/vt/vrt/collection/manager.impl.h
jstrzebonski: Done.

nlslatt (commented):
Most of it looks great, but there might be one problem.

tests/perf/common/test_harness.cc
```diff
@@ -339,7 +339,9 @@ void PerfTestHarness::StopTimer(std::string const& name) {
 }
 
 void PerfTestHarness::SyncResults() {
-  auto proxy = theObjGroup()->makeCollective<TestNodeObj>(this);
+  auto proxy = theObjGroup()->makeCollective<TestNodeObj>(
+    "PerfTestHarness::SyncResults()", this
```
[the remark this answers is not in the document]
jstrzebonski: This calls that overload:
```cpp
template <typename ObjT, typename... Args>
ProxyType<ObjT> makeCollective(std::string const& label, Args&&... args);
```

I think that maybe, with some funny usage of SFINAE, it could be possible to convert it to `ProxyType<ObjT> makeCollective(Args&&... args);`, and try to either extract the last parameter if it's a `std::string label`, or assume that default label (`""`) should be used. The problem is if `args` contain `std::string` for `ObjT` creation, how would one know if that's the case?

That's why this overload doesn't have a label as the last and optional parameter.

We could maybe add a tag to have a clear difference between `std::string` used as an ObjT's constructor parameter vs objgroup label, but IMHO it would burden interfaces too much.

Maybe others have a different opinion on this?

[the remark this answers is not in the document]
nlslatt: I think it's probably okay. I hadn't noticed that the template parameter type and pointer type were different, so I had been looking at the overload that accepts a raw pointer.

nlslatt (approved):
Looks good to me.

tests/unit/collection/test_checkpoint.extended.cc
```diff
@@ -172,10 +172,13 @@ TEST_F(TestCheckpoint, test_checkpoint_1) {
   auto num_nodes = static_cast<int32_t>(theContext()->getNumNodes());
 
   auto range = vt::Index3D(num_nodes, num_elms, 4);
-  auto checkpoint_name = "test_checkpoint_dir";
+  std::string const checkpoint_name{"test_checkpoint_dir"};
+  std::string const expected_label{"test_checkpoint_1"};
```
PhilMiller, line 176: I just noticed, you're a fan of "east `const`". I'm ambivalent about which way to go, but it definitely reads weird to me in this form.

That said, please don't make another push to change this.

src/vt/messaging/collection_chain_set.impl.h
```diff
@@ -62,7 +62,9 @@ CollectionChainSet<Index>::CollectionChainSet(
   static_assert(std::is_same<IndexT, Index>::value, "Must match index type");
 
   // Make this chain set an objgroup instance so we can send updates
-  auto p = theObjGroup()->makeCollective<CollectionChainSet<Index>>(this);
+  auto p = theObjGroup()->makeCollective<CollectionChainSet<Index>>(
+    this, "CollectionChainSet::CollectionChainSet()"
```
PhilMiller, line 66: Minor nitpick, I don'think the label should describe the object, not the particular constructor, where it's not material to the resulting instance. This appears in a few cases.

PhilMiller: Various managers, maps, etc.

PhilMiller: In the tests, I largely don't care. Just in the main runtime source code.

PhilMiller:
I'm fine with merging as is, but if we're open to a small revision, see my remark about the chosen labels for runtime class instances.

tests/unit/collection/test_mapping.cc
```diff
@@ -216,9 +218,11 @@ TYPED_TEST_P(TestMapping, test_custom_mapping_1) {
 template <typename IndexT>
 struct MyDistMapper : vt::mapping::BaseMapper<IndexT> {
   static vt::ObjGroupProxyType construct() {
-    auto prox =  vt::theObjGroup()->makeCollective<MyDistMapper<IndexT>>();
-    prox.get()->proxy = prox;
-    return prox.getProxy();
+    auto madeProxy = vt::theObjGroup()->makeCollective<MyDistMapper<IndexT>>(
+      "MyDistMapper::construct()"
+    );
+    madeProxy.get()->proxy = proxy;
```
nlslatt, line 224:
```suggestion
    madeProxy.get()->proxy = madeProxy;
```

src/vt/messaging/collection_chain_set.impl.h
nlslatt: I agree since we are labeling the object and now how it came into existence.

jstrzebonski: Done.

src/vt/vrt/collection/balance/lb_invoke/lb_manager.cc
```diff
@@ -161,8 +163,8 @@ void LBManager::setLoadModel(std::shared_ptr<LoadModel> model) {
 
 template <typename LB>
 LBManager::LBProxyType
-LBManager::makeLB() {
-  auto proxy = theObjGroup()->makeCollective<LB>();
+LBManager::makeLB(std::string const& lb_name) {
+  auto proxy = theObjGroup()->makeCollective<LB>(lb_name);
```
jstrzebonski, line 167: Or maybe let's add a function `std::string getName() const` to load balancers, and call it here to get a name instead of passing it as an argument?

nlslatt: There's a data structure in `src/vt/vrt/collection/balance/lb_type.cc` that has the names of all the LBs in it.

jstrzebonski: Done.

nlslatt (approved):
Looks good!

nlslatt:
Unfortunately, there are now merge conflicts. This can be merged once the conflicts are resolved.

PR #1776, opened 2022-05-05
1765: use vt compilation flags for bundled libraries as well

### LINKED ISSUE #1765: use `vt` compilation flags for bundled libraries as well
**Describe the bug**
`icpx` build  is showing following warnings:
```
/vt/lib/EngFormat-Cpp/src/eng_format.cpp:144:15: warning: comparison with NaN always evaluates to false in fast floating point modes [-Wtautological-constant-compare]
    if      ( std::isnan( value ) ) return "NaN";
              ^~~~~~~~~~~~~~~~~~~
/vt/lib/EngFormat-Cpp/src/eng_format.cpp:145:15: warning: comparison with infinity always evaluates to false in fast floating point modes [-Wtautological-constant-compare]
    else if ( std::isinf( value ) ) return "INFINITE";
              ^~~~~~~~~~~~~~~~~~~
2 warnings generated.
```
This is because we only set the project's common compiler flags to `vt` target:
```
$ git grep set_darma_compiler_flags
cmake-modules/SetCXXCompilerFlags.cmake:3:macro(set_darma_compiler_flags vt_target)
src/CMakeLists.txt:243:set_darma_compiler_flags(${VIRTUAL_TRANSPORT_LIBRARY})
```

**Expected behavior**
Use the same (or as similar as possible) set of flags for bundled libraries as for `vt`.

cz4rs:
fixes #1765


cmake-modules/SetCXXCompilerFlags.cmake
```diff
@@ -1,40 +1,34 @@
-# Call this from all CMakeLists.txt files that can be built independently.
-
-macro(set_darma_compiler_flags vt_target)
-
-set(CMAKE_CXX_EXTENSIONS OFF)
```
cz4rs, line 5: this is already done in the main `CMakeLists.txt`

cmake-modules/SetCXXCompilerFlags.cmake
```diff
@@ -1,40 +1,34 @@
...
+    if (APPLE)
+      list(APPEND TARGET_PUBLIC_CXX_FLAGS -stdlib=libc++)
+    endif ()
+  elseif (CMAKE_CXX_COMPILER_ID STREQUAL IntelLLVM AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 2021.3.0)
+    list(APPEND TARGET_PRIVATE_CXX_FLAGS -fhonor-infinites -fhonor-nans)
+  elseif (NOT ${CMAKE_CXX_COMPILER_ID} MATCHES "Intel")
+    message(FATAL_ERROR "Your C++ compiler may not support C++14.")
   endif ()
-  if (CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 7)
-    list(APPEND TARGET_PUBLIC_CXX_FLAGS -DCLI11_EXPERIMENTAL_OPTIONAL=0)
```
cz4rs, line 17: removed: this is no longer present in CLI11

PhilMiller:
I think this is good, but I want to take another look, or have someone else look at it too.

cz4rs:
Rebased on top of `develop`.

PhilMiller (approved):

Issue #1691, opened 2022-03-01
Build fails in the absence of libunwind


PhilMiller:
**Describe the bug**

On a system that doesn't have `libunwind` installed, building a project that incorporates vt leads to a compilation failure for missing `libunwind.h`

This will block integration of updated vt into EMPIRE.

https://github.com/DARMA-tasking/vt/pull/1657#issuecomment-1055754513

**To Reproduce**
1. Example/test/snippet of code that fails
2. Compiler, platform, libraries
3. Run command: number of processors, threading options, etc.

**Expected behavior**

**Screenshots**

**Platform (please complete the following information):**
 - Supercomputer or desktop [e.g., Trinity, Sierra)
 - Compiler environment [e.g., clang-3.9, gcc8]
 - Libraries [e.g., mpich-2, openmpi)
 - Commit ID or release of VT [e.g., vt-1.0.0]

**Additional context**

cz4rs:
DARMA weekly conclusion:
Preferred solution is to check for `libunwind` at CMake level, fallback to `libexecinfo` when not available and just print minimal information (only addresses?) when nothing is available.

PR #1787, opened 2022-05-09
#1786: condition link flag on library's presence

### LINKED ISSUE #1786: linking against libunwind when not present
**Describe the bug**
As a follow-up to #1691, linking flag is still passed for libunwind even when file is not present.

stmcgovern:
Fixes #1786. Follow-up to issue #1691.

### DIFF
--- a/ci/docker/ubuntu-20.04-gnu-openmpi-cpp.dockerfile
+++ b/ci/docker/ubuntu-20.04-gnu-openmpi-cpp.dockerfile
@@ -28,7 +28,6 @@ RUN apt-get update -y -q && \
     valgrind \
     make-guile \
     libomp5 \
-    libunwind-dev \
     ccache \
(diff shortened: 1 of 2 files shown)


cz4rs (changes requested):
To make sure that this is working as intended, we could have one of the builds running without `libunwind`. Maybe Alpine? Removing this line should be enough:
https://github.com/DARMA-tasking/vt/blob/00b1297c4aeacc36bd86b5ff80d0ee1d6f1f5307/ci/docker/alpine-cpp.dockerfile#L23

Looks good otherwise :+1:

stmcgovern:
Thanks @cz4rs. Done.

lifflander (approved):
Looks good to me!

cz4rs (approved):

PR #1794, opened 2022-05-10
1460 fix bug in temperedlb tolerance

### LINKED ISSUE #1460: TemperedLB threshold is inappropriate
**Describe the bug**
The threshold used for determining whether to skip running `TemperedLB` is inappropriately low.  The current code essentially boils down to:

```
static constexpr double const tolerance = 0.05f;
auto const avg  = stats.at(lb::Statistic::P_l).at(lb::StatisticQuantity::avg);
auto const max  = stats.at(lb::Statistic::P_l).at(lb::StatisticQuantity::max);
should_lb = max > tolerance * avg;
```

Because the tolerance is less than 1, `should_lb` will always be true.  The intended value was likely `1.05f`.

nlslatt:
As previously coded, `TemperedLB` would never skip LB due to low imbalance. Instead of changing the value, which is consistent with other LBs, I fixed the formula used for evaluating.

Fixes #1460

### DIFF
--- a/src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
+++ b/src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
@@ -442,7 +442,7 @@ void TemperedLB::runLB(TimeType total_load) {
   }
 
   if (avg > 0.0000000001) {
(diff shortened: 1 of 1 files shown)


lifflander (approved):
Looks good

PR #1798, opened 2022-05-12
1513: fix "`mallinfo` is deprecated" warning

### LINKED ISSUE #1513: warning: 'mallinfo' is deprecated (`memory_usage.cc`)
**Describe the bug**
After upgrading to Ubuntu 21.04 and clang 12, following warning appears during compilation:
**`/vt/src/vt/utils/memory/memory_usage.cc:149:26: warning: 'mallinfo' is deprecated [-Wdeprecated-declarations]`**

Compilation line:
```
 vt  build  /bin/clang++ -DBROTLI_SHARED_COMPILATION -DJSON_USE_IMPLICIT_CONVERSIONS=1 -I../lib/CLI -Idebug -I../src -I../lib/json/include -I../lib/brotli/c/include -I../lib/libfort/lib -I../lib/context/src -Ilib/context -Ilib/checkpoint/src -I../lib/checkpoint/src -I../lib/detector/src -isystem ../lib/fmt/include  -Wall -pedantic -Wshadow -Wno-unknown-pragmas -Wsign-compare -ftemplate-backtrace-limit=100 -fsanitize=undefined  -ftemplate-depth=900 -DCLI11_EXPERIMENTAL_OPTIONAL=0 -g -fcolor-diagnostics -fPIC -std=c++14 -o src/CMakeFiles/vt.dir/vt/utils/memory/memory_usage.cc.o -c /vt/src/vt/utils/memory/memory_usage.cc
/vt/src/vt/utils/memory/memory_usage.cc:149:26: warning: 'mallinfo' is deprecated [-Wdeprecated-declarations]
    struct mallinfo mi = mallinfo();
                         ^
/usr/include/malloc.h:118:48: note: 'mallinfo' has been explicitly marked deprecated here
extern struct mallinfo mallinfo (void) __THROW __MALLOC_DEPRECATED;
                                               ^
```
`glibc` version:
```
 ldd --version 
ldd (Ubuntu GLIBC 2.33-0ubuntu5) 2.33
```

**Expected behavior**
Use `mallinfo2` instead. See [the fix in llvm](https://reviews.llvm.org/D96359) for reference.

cz4rs:
fixes #1513


cmake/check_system_functions.cmake
```diff
@@ -22,6 +22,7 @@ check_function_exists(sysconf vt_has_sysconf)
 
 set(CMAKE_REQUIRED_INCLUDES "malloc.h")
 check_function_exists(mallinfo vt_has_mallinfo)
+check_function_exists(mallinfo2 vt_has_mallinfo2)
```
cz4rs, line 25: This could be replaced with `check_symbol_exists`. [CMake docs](https://cmake.org/cmake/help/latest/module/CheckFunctionExists.html) say:
```
Note

Prefer using [CheckSymbolExists](https://cmake.org/cmake/help/latest/module/CheckSymbolExists.html#module:CheckSymbolExists) instead of this module, for the following reasons:
(...)
    check_function_exists() only verifies linking, it does not verify that the function is declared in system headers.
```

jstrzebonski: If that's the preferred way I think we should use it.

jstrzebonski:
Besides the check_function_exists vs check_symbol_exists usage, that looks good.

src/vt/utils/memory/memory_usage.cc
```diff
@@ -145,13 +145,17 @@ std::string PS::getName() {
...
 std::size_t Mallinfo::getUsage() {
-# if defined(vt_has_mallinfo) && defined(vt_has_malloc_h) && !vt_check_enabled(mimalloc)
-    struct mallinfo mi = mallinfo();
-    unsigned int blocks = mi.uordblks;
-    return static_cast<std::size_t>(blocks);
-# else
-    return 0;
-# endif
+#if vt_check_enabled(mimalloc) || \
+  !defined(vt_has_mallinfo) && !defined(vt_has_mallinfo2)
```
PhilMiller, line 149: Because I don't like to think about operator precedence, I'd prefer if this were parenthesized. I won't call it necessary, though.

cmake/check_system_functions.cmake
PhilMiller: Let's make that a more comprehensive change in a later PR.

PhilMiller (approved):

cmake/check_system_functions.cmake
cz4rs: :ok_hand:

src/vt/utils/memory/memory_usage.cc
cz4rs: Fixed that, it does make it easier to read :+1:

PR #1801, opened 2022-05-16
#1795: store message size in allocated memory block

### LINKED ISSUE #1795: store message size in allocated memory block
**What Needs to be Done?**
- [ ] Store message size in memory block
- [ ] Transmit entire block when sending
- [ ] Ensure that receiving memory block allocates correctly

nmm0:
Closes #1795

### DIFF
--- a/src/vt/messaging/active.cc
+++ b/src/vt/messaging/active.cc
@@ -199,7 +199,7 @@ trace::TraceEventIDType ActiveMessenger::makeTraceCreationSend(
   #endif
 }
 
-void ActiveMessenger::packMsg(
+MsgSizeType ActiveMessenger::packMsg(
   MessageType* msg, MsgSizeType size, void* ptr, MsgSizeType ptr_bytes
 ) {
   vt_debug_print(
@@ -208,8 +208,17 @@ void ActiveMessenger::packMsg(
     size, ptr_bytes, print_ptr(ptr)
   );
 
+  auto const can_grow = thePool()->tryGrowAllocation(msg, ptr_bytes);
+  // Typically this should be checked by the caller in advance
+  vtAssert(can_grow, "not enough space to pack message" );
+
   char* const msg_buffer = reinterpret_cast<char*>(msg) + size;
   std::memcpy(msg_buffer, ptr, ptr_bytes);
+
+  envelopeSetPutTag(msg->env, PutPackedTag);
+  setPackedPutType(msg->env);
+
+  return size + ptr_bytes;
 }
 
 EventType ActiveMessenger::sendMsgBytesWithPut(
@@ -267,10 +276,7 @@ EventType ActiveMessenger::sendMsgBytesWithPut(
       );
     }
     if (direct_buf_pack) {
-      packMsg(msg, base.size(), put_ptr, put_size);
-      new_msg_size += put_size;
-      envelopeSetPutTag(msg->env, PutPackedTag);
-      setPackedPutType(msg->env);
+      new_msg_size = packMsg(msg, base.size(), put_ptr, put_size);
     } else {
       auto const& env_tag = envelopeGetPutTag(msg->env);
       auto const& ret = sendData(
@@ -281,10 +287,6 @@ EventType ActiveMessenger::sendMsgBytesWithPut(
         envelopeSetPutTag(msg->env, ret_tag);
       }
     }
-  } else if (is_put && is_put_packed) {
-    // Adjust size of the send for packed data
-    auto const& put_size = envelopeGetPutSize(msg->env);
-    new_msg_size += put_size;
   }
 
   sendMsgBytes(dest, base, new_msg_size, send_tag);
--- a/src/vt/messaging/active.cc
+++ b/src/vt/messaging/active.cc
@@ -316,12 +316,7 @@ private:
 }
 
 void ActiveMessenger::handleChunkedMultiMsg(MultiMsg* msg) {
-  auto buf =
-#if vt_check_enabled(memory_pool)
-    static_cast<char*>(thePool()->alloc(msg->getSize()));
-#else
-    static_cast<char*>(std::malloc(msg->getSize()));
-#endif
+  auto buf = static_cast<char*>(thePool()->alloc(msg->getSize()));
 
   auto const size = msg->getSize();
   auto const info = msg->getInfo();
@@ -715,16 +710,7 @@ bool ActiveMessenger::recvDataMsgBuffer(
     if (flag == 1) {
       MPI_Get_count(&stat, MPI_BYTE, &num_probe_bytes);
 
-      char* buf =
-        user_buf == nullptr ?
-
-    #if vt_check_enabled(memory_pool)
-        static_cast<char*>(thePool()->alloc(num_probe_bytes)) :
-    #else
-        static_cast<char*>(std::malloc(num_probe_bytes))      :
-    #endif
-
-        static_cast<char*>(user_buf);
+      char* buf = user_buf == nullptr ? static_cast<char*>(thePool()->alloc(num_probe_bytes)) : static_cast<char*>(user_buf);
 
       NodeType const sender = stat.MPI_SOURCE;
 
@@ -763,12 +749,7 @@ void ActiveMessenger::recvDataDirect(
   int nchunks, TagType const tag, NodeType const from, MsgSizeType len,
   ContinuationDeleterType next
 ) {
-  char* buf =
-    #if vt_check_enabled(memory_pool)
-      static_cast<char*>(thePool()->alloc(len));
-    #else
-      static_cast<char*>(std::malloc(len));
-    #endif
+  char* buf = static_cast<char*>(thePool()->alloc(len));
 
   recvDataDirect(
     nchunks, buf, tag, from, len, default_priority, nullptr, next, false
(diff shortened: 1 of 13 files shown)


src/vt/pool/pool.cc
```diff
@@ -67,7 +67,10 @@ Pool::Pool()
 Pool::ePoolSize Pool::getPoolType(
   size_t const& num_bytes, size_t const& oversize
 ) {
-  auto const& total_bytes = num_bytes + oversize + sizeof(HeaderType);
+  // Header size is accounted for internall in MemoryPoolEqual
+  // Blocks will include header size, but getNumBytes won't wo we shouldn't count
```
stmcgovern, line 71: comment typos: internall (l. 70) and wo (l. 71)

src/vt/messaging/active.cc
```diff
@@ -1169,7 +1142,7 @@ void ActiveMessenger::finishPendingActiveMsgAsyncRecv(InProgressIRecv* irecv) {
   }
 
   if (!is_put || put_finished) {
-    processActiveMsg(MsgPtr<MessageType>(base, msg_bytes), sender, true); // Note: use updated msg_bytes for message size
+    processActiveMsg(MsgPtr<MessageType>(base), sender, true); // Note: use updated msg_bytes for message size
```
nlslatt, line 1145: Is this a stale comment?

src/vt/messaging/active.impl.h
```diff
@@ -225,7 +225,7 @@ ActiveMessenger::PendingSendType ActiveMessenger::broadcastMsgSz(
   TagType tag
 ) {
   auto const han = auto_registry::makeAutoHandler<MsgT,f>();
-  MsgSharedPtr<MsgT> msgptr(msg.msg_, msg_size); // Note: use explicitly provided message size
+  MsgSharedPtr<MsgT> msgptr(msg.msg_); // Note: use explicitly provided message size
```
nlslatt, line 228: Stale comment?

src/vt/pool/pool.h
```diff
@@ -132,6 +132,10 @@ struct Pool : runtime::component::Component<Pool> {
    */
   SizeType remainingSize(void* const buf);
 
+  SizeType allocatedSize(void* const buf);
```
nlslatt, line 135: Should `remainingSize()` and `allocatedSize()` be marked const?

tests/unit/collection/test_send.h
```diff
@@ -97,6 +101,28 @@ struct ColMsg : CollectionMessage<TestCol<Args...>> {
...
+  //vt_msg_serialize_if_needed_by_parent();
+  vt_msg_serialize_prohibited();
+
+//  template <typename SerializerT>
+//  void serialize(SerializerT& s) {
+//    s | buff_size;
+//    MessageParentType::serialize(s);
+//  }
+};
+
```
nlslatt, line 125: What's the reason for so much commented code?

tests/unit/collection/test_send.h
```diff
@@ -132,12 +158,35 @@ struct SendHandlers {
...
+  using BaseT = SendHandlers<CollectionT, MessageT, TupleT>;
+  static void handler(MessageT* msg, CollectionT* col) {
+    // Verify payload size is correct
+    EXPECT_EQ(msg->buff_size, sizeof(PayloadT));
+    auto smart_ptr = vt::MsgSharedPtr<MessageT>(msg);
+    EXPECT_EQ(smart_ptr.size(), sizeof(MessageT) + sizeof(PayloadT));
+
+    // Verify payload
+    BaseT::execute(col, *msg->payload());
+    //BaseT::execute(col,msg->tup);
```
nlslatt, line 177: Should this be cleaned up?

tests/unit/collection/test_send.h
```diff
@@ -153,6 +202,7 @@ void test_collection_send_1() {
     auto proxy = theCollection()->construct<ColType>(range);
     for (int i = 0; i < col_size; i++) {
       auto msg = makeMessage<MsgType>(args);
+      EXPECT_EQ(msg.size(), sizeof(MsgType));
       //proxy[i].template send<MsgType,SendHandlers<ColType>::handler>(msg);
```
nlslatt, line 206: Please clean this up while you're here.

tests/unit/collection/test_send.h
```diff
@@ -165,8 +215,30 @@ void test_collection_send_1() {
...
+  using TestParamType = typename ColType::ParamType;
+
+  auto const& this_node = theContext()->getNode();
+  if (this_node == 0) {
+    auto const& col_size = 32;
+    auto range = TestIndex(col_size);
+    TestParamType args = ConstructTuple<TestParamType>::construct();
+    auto proxy = theCollection()->construct<ColType>(range);
+    for (int i = 0; i < col_size; i++) {
+      auto msg = makeMessageSz<MsgType>(sizeof(PayloadType));//, args);
```
nlslatt, line 231: Please clean this up.

nlslatt (commented):
I've provided some feedback but would definitely like more reviews of this PR. Do we have good enough PR testing for both the memory pool on and off?

src/vt/messaging/active.impl.h
nmm0: yep, removed

src/vt/pool/pool.h
nmm0: Yeah they should be, fixed

tests/unit/collection/test_send.h
nmm0: This was some cruft when getting this code to work originally, forgot to remove it. Did so now

tests/unit/collection/test_send.h
nmm0: Removed

tests/unit/collection/test_send.h
nmm0: Removed

src/vt/pool/pool.cc
```diff
@@ -67,6 +67,9 @@ Pool::Pool()
 Pool::ePoolSize Pool::getPoolType(
   size_t const& num_bytes, size_t const& oversize
 ) {
+  // Header size is accounted for internal in MemoryPoolEqual
```
PhilMiller, line 70:
```suggestion
  // Header size is accounted for internally in MemoryPoolEqual
```

src/vt/pool/pool.cc
```diff
@@ -252,6 +247,23 @@ Pool::SizeType Pool::remainingSize(void* const buf) {
...
+
+bool
+Pool::tryGrowAllocation(void* buf, size_t grow_amount) {
+  // For non-pooled alloc, this condition will always be true
+  // since remainingSize(buf) would be 0
+  if ( remainingSize(buf) < grow_amount )
+    return false;
+
+  auto *header = reinterpret_cast<Header*>(HeaderManagerType::getHeaderPtr(reinterpret_cast<char*>(buf)));
+  header->alloc_size += grow_amount;
```
PhilMiller, line 263: I haven't looked at the unmodified bits of the implementation - will `HeaderManagerType::getHeaderBytes` and `HeaderManagerType::getHeaderOversizeBytes` return updated values based on this?

PhilMiller:
Looks pretty good to me.

src/vt/pool/pool.cc
nmm0: `getHeaderBytes` yes -- this actually just modifies alloc_size. I guess the question is should this be modifying overside bytes instead of alloc_size?

src/vt/pool/pool.cc
nmm0: fixed

nlslatt (commented):
I don't have as deep an understanding as I'd like of the code that ties in tightly with this, but I think it looks reasonable.

src/vt/messaging/active.cc
```diff
@@ -711,16 +708,7 @@ bool ActiveMessenger::recvDataMsgBuffer(
...
-        user_buf == nullptr ?
-
-    #if vt_check_enabled(memory_pool)
-        static_cast<char*>(thePool()->alloc(num_probe_bytes)) :
-    #else
-        static_cast<char*>(std::malloc(num_probe_bytes))      :
-    #endif
-
-        static_cast<char*>(user_buf);
+      char* buf = user_buf == nullptr ? static_cast<char*>(thePool()->alloc(num_probe_bytes)) : static_cast<char*>(user_buf);
```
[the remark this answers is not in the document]
nmm0: fixed

src/vt/pool/pool.cc
PhilMiller: I think you're better positioned to answer that than I am. I don't know what the relationship is between the various stored and implied buffer size values.

nlslatt: If I'm understanding right, the logic is probably simpler (or the changes less extensive) the way it is currently implemented. Do you agree with that, @nmm0?

nmm0: I think logic-wise it's the same. The current implementation more closely mimics the old behavior however -- what would happen before is the general "message size" would be increased when sending a put message. Then when the message was received, it would just set alloc_size according to the received size, which would technically include the put size.

However, it does seem kind of weird to not include the put size in the oversize bytes. So I think that the right thing would be to include it in oversize, but maybe we can do that in the future in case it would break things.

nlslatt (approved):
Looks good to me

PR #1823, opened 2022-05-24
1822: Upgrade CI docker image to Alpine 3.16

cz4rs:
fixes #1822

PR #1829, opened 2022-05-25
#1823 rename alpine pipeline

jstrzebonski:
Fixes #1823


cz4rs (approved):

cz4rs:
> While we are on consistency, do we want to rename `azure-clang-alpine-mpich.yml` to `azure-clang-13-alpine-mpich.yml`?

I think filename is fine, since it's the only Alpine build we have (and AFAIK we don't plan to add others). Also note that clang version is not pinned in the dockerfile - it's just using the default (clang-13) that this version of Alpine provides.

JacobDomagala (approved):

PR #1840, opened 2022-06-01
#1839: use unsigned types for left shifts in BitPacker

### LINKED ISSUE #1839: ub left shifting negative signed values
**Describe the bug**
Some bitfields trigger undefined behavior by left-shifting a negative value.

**To Reproduce**
Adding this assert to setFieldDynamic in `bits_packer.impl.h` will reliably fail in our tests:

```cpp
  auto tmp = ((static_cast<BitField>(segment) << nbits) >> nbits);
  if ( std::is_signed< decltype(tmp) >::value && ( tmp < 0 ) )
    vtAbort(fmt::format(
      "hit a negative value of {}, segment: {}, nbits: {}", tmp,
      static_cast<BitField>(segment), nbits ) );
```

The next line is 

```cpp
field = field | (((static_cast<BitField>(segment) << nbits) >> nbits) << start);
```

So this issue being that the `tmp` value here is negative (BitField is signed in some cases) and being left shifted by `start`, which is undefined behavior.

**Expected behavior**
we need to make sure the shifted value is unsigned.

nmm0:
fixes #1839

### DIFF
--- a/src/vt/utils/bits/bits_packer.h
+++ b/src/vt/utils/bits/bits_packer.h
@@ -46,8 +46,26 @@
 
 #include <cassert>
 #include <cstdlib>
+#include <cstring>
 
 namespace vt { namespace utils {
+
+// Non-constexpr C++14 version of bit_cast
+// See: https://en.cppreference.com/w/cpp/numeric/bit_cast
+template<typename Dst, typename Src>
+std::enable_if_t<
+  sizeof(Dst) == sizeof(Src)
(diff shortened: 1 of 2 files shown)


lifflander (approved):

nmm0:
I switch to `std::memcpy` instead of `static_cast` since that's the correct way of bit casting

PhilMiller (approved):

PR #1842, opened 2022-06-06
1835: fix remaining issues in `RawData::getModeledComm`

### LINKED ISSUE #1835: fix subphases handling in `getModeledComm`
https://github.com/DARMA-tasking/vt/blob/3e362c4bc3545c0332d1a9208eb2cef49264d7b3/src/vt/vrt/collection/balance/model/comm_overhead.cc#L78-L84

(notice the old `todo` here as well)

_Originally posted by @cz4rs in https://github.com/DARMA-tasking/vt/pull/1695#discussion_r855481510_

**Describe the bug**
Currently `RawData::getModeledComm` does not consider subphases (note: this method might get moved out of `RawData` model). Fix that by handling them the same way `CommOverhead` does.

### LINKED ISSUE #1836: add per message and per byte weights in `getModeledComm`
**What Needs to be Done?**
Currently messages and bytes hold the same weight in `RawData::getModeledComm`. Add parameters introducing custom weights to them.
See `CommOverhead` for a possible implementation.

### LINKED ISSUE #1837: move the implementation of `getModeledComm` into a separate load model
I'm not sure that the `RawData` model is the right place to compute and return modeled communication. Any thoughts, @PhilMiller?

_Originally posted by @nlslatt in https://github.com/DARMA-tasking/vt/pull/1695#discussion_r883805787_

cz4rs:
fixes #1835 
fixes #1836 
fixes #1837 

- [x] fix subphases handling
- [x] add per message and per byte weights
- [x] move the implementation into a separate load model

Note: these were tiny issues (created per comment), so it's convenient to fix them with a single PR.

TODO:
- [x] add unit test for the new model (see `test_model_comm_overhead.nompi.cc`)
- [x] discussion: should we avoid adding a new model and simply replace the logic in `CommOverhead`? (**no**)

### DIFF
--- a/src/vt/vrt/collection/balance/model/comm_model.cc
+++ b/src/vt/vrt/collection/balance/model/comm_model.cc
@@ -2,7 +2,7 @@
 //@HEADER
 // *****************************************************************************
 //
-//                                 raw_data.cc
+//                                comm_model.cc
 //                       DARMA/vt => Virtual Transport
 //
 // Copyright 2019-2021 National Technology & Engineering Solutions of Sandia, LLC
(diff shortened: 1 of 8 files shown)


src/vt/vrt/collection/balance/model/comm_overhead.h
```diff
@@ -68,7 +68,7 @@ struct CommOverhead : public ComposedModel {
   void setLoads(std::unordered_map<PhaseType, LoadMapType> const* proc_load,
                 std::unordered_map<PhaseType, CommMapType> const* proc_comm) override;
 
-  TimeType getModeledLoad(ElementIDStruct object, PhaseOffset when) override;
+  TimeType getModeledComm(ElementIDStruct object, PhaseOffset when) override;
```
cz4rs, line 71: Optional: should we rename it now, since we have `getModeledComm` concept?

src/vt/vrt/collection/balance/model/comm_model.h
```diff
@@ -0,0 +1,87 @@
...
+    ComposedModel::setLoads(proc_load, proc_comm);
+  }
+
+  TimeType getModeledComm(ElementIDStruct object, PhaseOffset when) override;
+
+private:
+  // observer pointer to the underlying comm data
+  std::unordered_map<PhaseType, CommMapType> const* proc_comm_;
+
+  TimeType per_msg_weight_  = 0.001;
```
cz4rs, line 81: Up for discussion what the default values should be.

src/vt/vrt/collection/balance/model/comm_model.cc
```diff
@@ -0,0 +1,82 @@
...
+
+  if (when.subphase == PhaseOffset::WHOLE_PHASE) {
+    return modeled_comm;
+  } else {
+    // we don't record comm costs for each subphase - split it proportionally
+    auto load = ComposedModel::getModeledLoad(object, when);
+    auto whole_phase_load = ComposedModel::getModeledLoad(
+      object, PhaseOffset{when.phases, PhaseOffset::WHOLE_PHASE});
+
+    return modeled_comm * load / whole_phase_load;
```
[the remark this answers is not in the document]
PhilMiller: We definitely should not use the number of subphases as a denominator. Some subphases may be quite substantial, while others might be negligible. I'm open to other ideas for weighting subphase communications while we don't instrument to that level of specificity, but I think making it proportional to load is reasonable to maintain for now

src/vt/vrt/collection/balance/model/comm_model.h
nlslatt: @ppebay Does LBAF use a per-message weight of zero (meaning it only considers bytes, not the number of messages)?

src/vt/vrt/collection/balance/model/comm_model.h
```diff
@@ -0,0 +1,87 @@
...
+
+#if !defined INCLUDED_VT_VRT_COLLECTION_BALANCE_MODEL_COMM_MODEL_H
+#define INCLUDED_VT_VRT_COLLECTION_BALANCE_MODEL_COMM_MODEL_H
+
+#include "vt/vrt/collection/balance/model/composed_model.h"
+#include <unordered_map>
+
+namespace vt { namespace vrt { namespace collection { namespace balance {
+
+struct CommModel : public ComposedModel {
```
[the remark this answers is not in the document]
nlslatt: Maybe something reflecting that the model is an affine combination of the number of messages and their size?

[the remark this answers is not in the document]
cz4rs: I went with `WeightedMessages`.

src/vt/vrt/collection/balance/model/comm_model.h
cz4rs: Going with following defaults:
```cpp
 per_msg_weight_  = 0.0;
 per_byte_weight_ = 1.0;
```
to make the behavior consistent with current LBAF implementation. This means `TemperedWMin`'s `beta` parameter is to be used as per byte coefficient.

nlslatt (approved):
This looks pretty good. It could be made a little more rigorous by making the different subphases not all have the same load but I'm okay with it as is. I think there are some long lines that should be wrapped.

PR #1847, opened 2022-06-08
#1845: bundled fmt library can conflict with fmt installed in the same prefix

### LINKED ISSUE #1845: bundled fmt library can conflict with fmt installed in the same prefix
**Describe the bug**
The bundled fmt library can overwrite or be overwritten by a different fmt version that is not compatible if installed to the same prefix. This kind of situation is a bit rare but shows up in spack environments and some docker installs.

The fmt library should be prefixed to make it unique.

nmm0:
This PR renames `libfmt{d}.a` to `libfmt-vt{d}.a` in order to prevent conflicts with fmt installed in the same root. Additionally, it renames the include directory from `fmt` to `fmt-vt`.

Closes #1845


nmm0:
We don't have a test for this but I have verified that this works in a spack install with `fmt` installed additionally as a separate package.

lib/fmt/CMakeLists.txt
```diff
@@ -14,6 +14,7 @@ set(FMT_SOURCES src/format.cc)
 
 add_library(fmt ${FMT_SOURCES} ${FMT_HEADERS})
```
PhilMiller, line 15: Should we change the CMake name as well? If a downstream code ends up picking this up, it'll be a bit confusing

nmm0: hmm yeah, there might be a problem with duplicate targets if they happen to also be including fmt as a subdirectory

nmm0: Overall I'd be more in favor of a more permanent solution which would be to optionally force vt to use external versions of these libraries. That would be ideal for the spack case. I'd like to get this merged now however because it is causing some breakages in some internal CI processes

cz4rs (approved):
Looks good for me as is.
Possibly another issue can be opened for the "more permanent solution" as discussed?

PR #1851, opened 2022-06-14
#1850: allow setting dir where lb statistics are dumped

### LINKED ISSUE #1850: Add command-line option for setting directory for dumping LB statistics files

nlslatt:
Closes #1850


src/vt/configs/arguments/args.cc
```diff
@@ -836,6 +840,10 @@ std::string AppConfig::getLBDataFileIn() const {
 
 std::string AppConfig::getLBStatisticsFile() const {
   std::string name = vt_lb_statistics_file;
+  std::string dir = vt_lb_statistics_dir;
+  if (dir.size() > 0) {
```
cz4rs, line 844: Nitpicking:
```suggestion
  if (not dir.empty()) {
```

src/vt/vrt/collection/balance/lb_invoke/lb_manager.cc
```diff
@@ -717,6 +717,13 @@ void LBManager::createStatisticsFile() {
       "LBManager::createStatsFile: file={}\n", file_name
     );
 
+    auto const dir = theConfig()->vt_lb_statistics_dir;
+    // Node 0 creates the directory
+    if (not created_lbstats_dir_ and theContext()->getNode() == 0) {
+      mkdir(dir.c_str(), S_IRWXU);
```
cz4rs, line 723: I've had to double check this, but it seems to be fine when the directory exists as well. `mkdir` will simply return `EEXIST`. Note: I agree that the `created_lbstats_dir_` flag is still necessary to avoid needless calls.

We should probably at least log the error code in case `mkdir` fails for some other reason - permission denied, not enough space.

cz4rs (approved):
Looks good to me, minor comments posted.

src/vt/configs/arguments/args.cc
nlslatt: Thanks for bringing this up. It made me realize that I wasn't checking for an unspecified directory when actually creating the directory.

src/vt/vrt/collection/balance/lb_invoke/lb_manager.cc
nlslatt: Good call. I had copied another instance of directory creation that didn't check the result. I've updated both instances.

cz4rs (approved):

PR #1855, opened 2022-06-16
1849: Add simple unit test for AsyncOpCUDA

### LINKED ISSUE #1849: Add unit test for CUDA AsyncOp

JacobDomagala:
Fixes #1849


tests/unit/active/test_async_op_cuda.cc
```diff
@@ -0,0 +1,209 @@
...
+  int driverVer;
+  if (
+    (cudaDriverGetVersion(&driverVer) == cudaErrorInvalidValue) or
+    (driverVer == 0)) {
+    vtWarn(
+      "Trying to run test_async_op_cuda but CUDA driver is not present!\n"
+    );
+
+    return;
+  }
```
JacobDomagala, line 192: Since Azure runners don't have CUDA driver installed, we don't want to run the test (but we still want to compile the file, so we don't exclude it with vt_ci_build or something similar)

tests/unit/active/test_async_op_cuda.cc
```diff
@@ -0,0 +1,209 @@
...
+
+TEST_F(TestAsyncOp, test_async_op_cuda) {
+  int driverVer;
+  if (
+    (cudaDriverGetVersion(&driverVer) == cudaErrorInvalidValue) or
+    (driverVer == 0)) {
+    vtWarn(
+      "Trying to run test_async_op_cuda but CUDA driver is not present!\n"
+    );
+
```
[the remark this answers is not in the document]
JacobDomagala: Good point :+1:

lifflander (approved):

nlslatt (commented):
I tested this manually on Vortex. If I build with g++, the test is not compiled. If I do a cuda build on the login node and then run on the compute node, the test passes.

If I accidentally run it on the login node instead, it aborts with `cudaMalloc(dataDevicePointer1_) failed with error -> initialization error`. Is that the desired behavior, or was this use case meant to be handled with `GTEST_SKIP`?

tests/unit/active/test_async_op_cuda.cc
```diff
@@ -0,0 +1,206 @@
...
+  auto const this_node = theContext()->getNode();
+  auto p = theObjGroup()->makeCollective<CUDAGroup>("test_async_op_cuda");
+  auto ep = theTerm()->makeEpochRooted(term::UseDS{true});
+
+  // When this returns all the CUDA streams should be done
+  runInEpoch(ep, [p, this_node] {
+    p[this_node].send<MyMsg, &CUDAGroup::cudaHandler>();
+  });
+
+  cudaDeviceSynchronize();
```
PhilMiller, line 200: Might this call hide potential failures? I think we want to run with checks without having synchronized, so we can have a chance of seeing incomplete data in the assertions

tests/unit/active/test_async_op_cuda.cc
```diff
@@ -0,0 +1,204 @@
...
+
+#include <vt/messaging/async_op_cuda.h>
+#include <vt/objgroup/manager.h>
+
+#include <gtest/gtest.h>
+
+
+namespace vt { namespace tests { namespace unit {
+
+#if __CUDACC__
```
PhilMiller, line 54: Tiny detail - we want to check whether the compiler is providing/targeting CUDA support, not that we're using a specific CUDA compiler (nvcc, nvhpc/nvc++, clang)

JacobDomagala: We also only check for `__CUDACC__` define in source files (async_op_cuda.h) so we should also update it there

PhilMiller: Looking over the docs again, `__CUDACC__` may actually be what we want.
https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#nvcc-identification-macro
https://llvm.org/docs/CompileCudaWithLLVM.html#detecting-clang-vs-nvcc-from-code

That's what we already have in `async_op_cuda.h`, so let's just go with this for now, and investigate whether there's adjustment needed later.

PhilMiller (approved):

JacobDomagala:
Ok, rebased the branch

PR #1866, opened 2022-06-29
#1563: cmake: enable MPI guards by default regardless of build

### LINKED ISSUE #1563: Abort when MPI guards are violated in Release mode
**What Needs to be Done?**

We want to abort regardless of the type of build if the MPI guards are violated.

lifflander:
Fixes #1563


nmm0 (approved):
lgtm!

JacobDomagala (approved):

PR #1877, opened 2022-07-20
1876 Fix number of processors for perf tests

### LINKED ISSUE #1876: Fix number of processors for perf tests
**Describe the bug**
The `ping_pong` test is being run with only one processor on our Gitlab CI. It looks like the wrong variable is being passed to `set_tests_properties` (the flag that MPI uses to specify the number of processors rather than the number of processors itself). Fix this.

nlslatt:
I corrected a use of the wrong variable and also made 2 the minimum number of ranks to use for perf tests. `ping_pong` had been trying to run on only 1 rank, which was breaking Gitlab CI.

Fixes #1876


prwolfe (approved):
Makes sense to me.

Issue #1659, opened 2022-01-18
Improve communication statistics in VT


lifflander:
**What Needs to be Done?**

We want to change how `O_c` and `P_c` are computed to make them match up with LBAF.

PR #1888, opened 2022-08-02
1838: add `WeightedCommunicationVolume` load model

### LINKED ISSUE #1838: move `getModeledWork` out of `TemeperedWMin`
Having `getModeledWork` baked into load balancer makes it impossible to compute post-LB statistics. This could possibly be moved into a new load model, but we have to find a way to pass parameters into it (alpha, beta, gamma).
A more general solutions for load models might be required.

---

> Should this be implemented as a new load model instead of integrated into the load balancer? Load models can be composed, so you'd still be able to plug in whatever model you wanted for the loads.

> Having the formula integrated into the LB makes it impossible for vt to compute post-LB statistics following the formula. What we dump to the LB statistics json file post-LB would be incorrect but I'm not sure if it would have any other immediate consequences. I think it might break the (not yet implemented) capability to run multiple load balancers and use the most effective result.

> What do you think, @PhilMiller and @ppebay?

_Originally posted by @nlslatt in https://github.com/DARMA-tasking/vt/pull/1695#discussion_r883861684_

cz4rs:
fixes #1838

### DIFF
--- a/src/vt/vrt/collection/balance/lb_invoke/lb_manager.cc
+++ b/src/vt/vrt/collection/balance/lb_invoke/lb_manager.cc
@@ -629,27 +629,15 @@ void LBManager::computeStatistics(
   using ReduceOp = collective::PlusOp<std::vector<balance::LoadData>>;
 
   total_load_from_model = 0.;
-  TimeType total_work_from_model = 0.;
   std::vector<balance::LoadData> obj_load_model;
-  std::vector<balance::LoadData> obj_work_model;
   for (auto elm : *model) {
-    auto load = model->getModeledLoad(
-      elm, {balance::PhaseOffset::NEXT_PHASE, balance::PhaseOffset::WHOLE_PHASE}
-    );
-    obj_load_model.emplace_back(
-      LoadData{lb::Statistic::Object_load_modeled, load}
-    );
-    total_load_from_model += load;
-
-    // FIXME: this is only TemperedWMin specific and requires composed
-    // WeightedCommunicationVolume model
     auto work = model->getModeledLoad(
       elm, {balance::PhaseOffset::NEXT_PHASE, balance::PhaseOffset::WHOLE_PHASE}
     );
-    obj_work_model.emplace_back(
-      LoadData{lb::Statistic::Object_work_modeled, work}
+    obj_load_model.emplace_back(
+      LoadData{lb::Statistic::Object_load_modeled, work}
     );
-    total_work_from_model += work;
+    total_load_from_model += work;
   }
 
   TimeType total_load_raw = 0.;
@@ -677,14 +665,8 @@ void LBManager::computeStatistics(
   lstats.emplace_back(
     LoadData{lb::Statistic::Rank_load_modeled, total_load_from_model}
   );
-  lstats.emplace_back(
-    LoadData{lb::Statistic::Rank_work_modeled, total_work_from_model}
-  );
   lstats.emplace_back(reduceVec(
     lb::Statistic::Object_load_modeled, std::move(obj_load_model)
-  ));
-  lstats.emplace_back(reduceVec(
-    lb::Statistic::Object_work_modeled, std::move(obj_work_model)
   ));
 
   if (model->hasRawLoad()) {
--- a/src/vt/vrt/collection/balance/lb_invoke/lb_manager.cc
+++ b/src/vt/vrt/collection/balance/lb_invoke/lb_manager.cc
@@ -641,7 +641,9 @@ void LBManager::computeStatistics(
     );
     total_load_from_model += load;
 
-    auto work = model->getModeledWork(
+    // FIXME: this is only TemperedWMin specific and requires composed
+    // WeightedCommunicationVolume model
+    auto work = model->getModeledLoad(
       elm, {balance::PhaseOffset::NEXT_PHASE, balance::PhaseOffset::WHOLE_PHASE}
     );
     obj_work_model.emplace_back(
--- a/src/vt/vrt/collection/balance/lb_invoke/lb_manager.cc
+++ b/src/vt/vrt/collection/balance/lb_invoke/lb_manager.cc
@@ -629,15 +629,25 @@ void LBManager::computeStatistics(
   using ReduceOp = collective::PlusOp<std::vector<balance::LoadData>>;
 
   total_load_from_model = 0.;
+  TimeType total_work_from_model = 0.;
   std::vector<balance::LoadData> obj_load_model;
+  std::vector<balance::LoadData> obj_work_model;
   for (auto elm : *model) {
-    auto work = model->getModeledLoad(
+    auto load = model->getModeledLoad(
       elm, {balance::PhaseOffset::NEXT_PHASE, balance::PhaseOffset::WHOLE_PHASE}
     );
     obj_load_model.emplace_back(
-      LoadData{lb::Statistic::Object_load_modeled, work}
+      LoadData{lb::Statistic::Object_load_modeled, load}
     );
-    total_load_from_model += work;
+    total_load_from_model += load;
+
+    auto work = model->getModeledWork(
+      elm, {balance::PhaseOffset::NEXT_PHASE, balance::PhaseOffset::WHOLE_PHASE}
+    );
+    obj_work_model.emplace_back(
+      LoadData{lb::Statistic::Object_work_modeled, work}
+    );
+    total_work_from_model += work;
   }
 
   TimeType total_load_raw = 0.;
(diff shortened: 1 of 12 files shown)


src/vt/vrt/collection/balance/temperedwmin/temperedwmin.h
```diff
@@ -56,17 +56,9 @@ struct TemperedWMin : TemperedLB {
 
 public:
   void init(objgroup::proxy::Proxy<TemperedWMin> in_proxy);
-  static std::unordered_map<std::string, std::string> getInputKeysWithHelp();
-
-  void inputParams(balance::SpecEntry* spec) override;
 
 protected:
   TimeType getModeledWork(const elm::ElementIDStruct& obj) override;
```
cz4rs, line 61: FIXME: this should have a different name now

nlslatt:
When `TemperedWMin` is used, how will this load model get set up? I can see three potential options: (a) make the app change the load model based on what LB was selected for that phase, which seems very awkward; (b) have `TemperedWMin` compose this load model on top of whatever the app already set up; or (c) make it possible to choose load models from the command-line like we do with load balancers already. I would lean toward (b), which might motivate leaving the `TemperedWMin` LB args as they are instead of removing them.

PhilMiller:
I was kind of expecting option B, if the lb strategy is designed with an assumption that it's working over this model

src/vt/vrt/collection/balance/model/load_model.h
```diff
@@ -263,6 +263,22 @@ struct LoadModel
...
+   *
+   * \param[in] object The object whose work is requested
+   * \param[in] when The interval in which the estimated work is requested
+   *
+   * \return How much work the object is estimated to require
+   *
+   * The `updateLoads` method must have been called before any call to
+   * this.
+   */
+  virtual TimeType getModeledWork(ElementIDStruct object, PhaseOffset when) {
```
cz4rs, line 278: random question: should we attempt to make this, `getModeledComm` and `getModeledLoad` methods `const`?

src/vt/vrt/collection/balance/temperedwmin/temperedwmin.cc
```diff
@@ -94,14 +95,17 @@ void TemperedWMin::inputParams(balance::SpecEntry* spec) {
     "TemperedWMin::inputParams: alpha={}, beta={}, gamma={}\n",
     alpha_, beta_, gamma_
   );
+
+  total_work_model_ = std::make_unique<balance::WeightedCommunicationVolume>(
```
cz4rs, line 99: The model is created here to avoid instantiating it redundantly in a single phase (for example in `getModeledValue`). Should I worry about the underlying model (the one owned by LBManager) changing?

PhilMiller: In theory it could, but right now, the LB strategies get instantiated and destroyed every time they're going to be called, so it doesn't matter anyway.

If you want to code this defensively, save the pointer that it passed here to compose on, and add an assertion at the point of use that the LBManager still returns the same pointer later.

src/vt/vrt/collection/balance/model/load_model.h
PhilMiller: From an interface perspective, that would be a good thing (in a separate PR). Implementation-wise, I think there are a bunch of members we would need to mark `mutable` in some of the classes, because they cache and update things as needed.

That would potentially be a good issue for on-boarding someone new, since it really only requires C++ knowledge, and help them familiarize with the work flow

cz4rs: Added #1903 and labeled it as `intro`.

src/vt/vrt/collection/balance/model/load_model.h
```diff
@@ -263,6 +263,22 @@ struct LoadModel
...
+    return {};
+  }
+
```
PhilMiller, line 281: This might be me being difficult, but can we just do without this, and have the new model override `getModeledLoad()` like everything else does? On object's 'load' is really just its contribution to the abstract cost function that whatever LB strategy is trying to minimize. There's no reason to distinguish some kind of 'normal' from 'special' modeled load. It means whatever the model wants it to mean, and whatever the strategy interprets it as.

PhilMiller (dismissed):

src/vt/vrt/collection/balance/model/load_model.h
PhilMiller: This is probably a point that @ppebay would need to stand up for, since my understanding is that @cz4rs you're just implementing it as requested.

cz4rs: I can see your point here and it doesn't seem like there's much use for `getModeledWork` outside of `TemperedWMin` anyways.

cz4rs: On the other hand, here's the last 6 commits to `load_model.h`:
```
f6ebc31a0 #1672: lb: rename `getComm` to `getModeledComm`
543f320a2 #1672: lb: rename `getLoadMetric` to `getModeledLoad`
1a2ee2e75 #1672: rename `getLoad` to `getLoadMetric`
35b554444 #1672: lb: add TemperedWMin load balancer
26e478bba #1672: lb: add getTotalWork and getComm methods
63c7ee10d #1672: lb: rename getWork to getLoad
```
There's been plenty of renaming already (note that `getModeledLoad()` has started as `getWork()`), so I'd like to make sure that we have a consensus about this.

nlslatt: Do any of the load balancers assume that load summed over ranks is an invariant across LB migrations? If so, then it could be important to allow a load balancer to choose pure load even if the load model in use knows about communication.

PhilMiller: We'd have to think pretty hard to figure out whether any particular LB strategy has such an assumption. 

I would honestly lean toward leaving things pretty loose for now, and if an application is configured with a combination of load model and balancing strategy that fails or produces nonsense, that's on them.

cz4rs: One more thing to consider is having total work reported along with other statistics. See 7f49544 - I have planned to do that in a separate PR for #1659, but it makes sense to include it here.

nlslatt: If we would have to think hard to figure that out, how can we expect applications people to do the right thing? It seems like a bad idea to make a change that makes things more risky for them.

LBs like `GreedyLB` and `HierarchicalLB` assume that loads are the same independent of where they assign the work to be done, right? Using work instead of load is probably going to break them.

Is it really a problem to distinguish between load and work so that apps can switch between one of those and `TemperedWMin` without having to change out the load model? We have an easy way to switch between different load balancers on different phases but switching out the load model is completely left to the application (unless we have `TemperedWMin` continue to read in the `alpha`, `beta`, `gamma`, construct and compose this load model on top of what was already there, and restore the original load model afterward).

ppebay: So `getModeledLoad()` refers to a notion of a "modeled load" that is itself a generic understanding of "load" (i.e., one that might include communication)? In that sense, would `modeledLoad` be a replacement for `Work`?

src/vt/vrt/collection/balance/lb_invoke/lb_manager.cc
```diff
@@ -629,15 +629,25 @@ void LBManager::computeStatistics(
...
       elm, {balance::PhaseOffset::NEXT_PHASE, balance::PhaseOffset::WHOLE_PHASE}
     );
     obj_load_model.emplace_back(
-      LoadData{lb::Statistic::Object_load_modeled, work}
+      LoadData{lb::Statistic::Object_load_modeled, load}
     );
-    total_load_from_model += work;
+    total_load_from_model += load;
+
+    auto work = model->getModeledWork(
```
cz4rs, line 644: FIXME: this is still not enough - `model` here is not the composed total work model that `TemperedWMin` uses.

src/vt/vrt/collection/balance/model/load_model.h
ppebay: In LBAF we have the following concepts:
* "Time" that is the compute time required by an object -- and it's assumed to be constant irrespective of the rank to which it is assigned.
* "Load" that is simply the sum-total of all object times--this is a per-rank quantity;
* "Weight" that represents the amount of data (in bytes?) associated with object-to-object communication (and which can be aggregated at the rank level);
* "Work" that is an (affine) combination of the above -- this is a per-rank variable;
* "Total Work" that is the sum total of "Work" across the entire collection of ranks -- note that this remains constant when only object times are considered

cz4rs: In both `WeightedMessages` and `CommOverhead` models we have:
- `per_msg_weight_`
- `per_byte_weight_`

src/vt/vrt/collection/balance/lb_invoke/lb_manager.cc
```diff
@@ -629,15 +629,27 @@ void LBManager::computeStatistics(
...
+    auto load = model->getModeledLoad(
       elm, {balance::PhaseOffset::NEXT_PHASE, balance::PhaseOffset::WHOLE_PHASE}
     );
     obj_load_model.emplace_back(
-      LoadData{lb::Statistic::Object_load_modeled, work}
+      LoadData{lb::Statistic::Object_load_modeled, load}
+    );
+    total_load_from_model += load;
+
+    // FIXME: this is only TemperedWMin specific and requires composed
```
cz4rs, line 644: `model` here is not the `WeightedCommunicationVolume` model that `TemperedWMin` composes and uses internally.

cz4rs:
`gcc-7` failure is unrelated.

nlslatt (approved):
Looks good

PR #1897, opened 2022-08-03
1896 Use only unique filenames in unit tests

### LINKED ISSUE #1896: Use only unique filenames in unit tests
**What Needs to be Done?**

Change filenames used for I/O in unit tests to be unique across tests, parameterizations, and MPI rank counts.

This may fix #1894 and #1895.

nlslatt:
This PR attempts to stop using the same disk files from multiple unit tests, including runs with different numbers of MPI ranks, so that multiple tests can be run concurrently (as happens in Gitlab CI testing) without file system conflicts. ~~This PR does not yet handle parameterizations or filenames/directories used for LB data files in `addAdditionalArgs()` in `collection/test_lb.extended.cc`. Does anyone have ideas on how to tackle those?~~

Closes #1896

### DIFF
--- a/tests/unit/collection/test_checkpoint.extended.cc
+++ b/tests/unit/collection/test_checkpoint.extended.cc
@@ -172,7 +172,7 @@ TEST_F(TestCheckpoint, test_checkpoint_1) {
   auto num_nodes = static_cast<int32_t>(theContext()->getNumNodes());
 
   auto range = vt::Index3D(num_nodes, num_elms, 4);
-  std::string const checkpoint_name{"test_checkpoint_dir"};
+  std::string const checkpoint_name(getUniqueFilenameWithRanks());
   std::string const expected_label{"test_checkpoint_1"};
 
   {
@@ -251,7 +251,7 @@ TEST_F(TestCheckpoint, test_checkpoint_in_place_2) {
   auto num_nodes = static_cast<int32_t>(theContext()->getNumNodes());
 
   auto range = vt::Index3D(num_nodes, num_elms, 4);
-  auto checkpoint_name = "test_checkpoint_dir";
+  std::string const checkpoint_name(getUniqueFilenameWithRanks());
   auto proxy = vt::theCollection()->constructCollective<TestCol>(
     range, "test_checkpoint_in_place_2"
   );
@@ -323,7 +323,7 @@ TEST_F(TestCheckpoint, test_checkpoint_in_place_3) {
   auto num_nodes = static_cast<int32_t>(theContext()->getNumNodes());
 
   auto range = vt::Index3D(num_nodes, num_elms, 4);
-  auto checkpoint_name = "test_checkpoint_dir_2";
+  std::string const checkpoint_name(getUniqueFilenameWithRanks());
   auto proxy = vt::theCollection()->constructCollective<TestCol>(
     range, "test_checkpoint_in_place_3"
   );
@@ -400,7 +400,7 @@ TEST_F(TestCheckpoint, test_checkpoint_no_elements_on_root_rank) {
   auto num_nodes = static_cast<int32_t>(theContext()->getNumNodes());
 
   auto range = vt::Index3D(num_nodes, num_elms, 4);
-  auto checkpoint_name = "test_null_elm_checkpoint_dir";
+  std::string const checkpoint_name(getUniqueFilenameWithRanks());
 
   {
     auto proxy = vt::makeCollection<TestCol>("test_checkpoint_no_elements_on_root_rank")
--- a/tests/unit/runtime/test_initialization.cc
+++ b/tests/unit/runtime/test_initialization.cc
@@ -44,6 +44,7 @@
 #include <gtest/gtest.h>
 
 #include "test_parallel_harness.h"
+#include "test_helpers.h"
 
 #include <vt/collective/startup.h>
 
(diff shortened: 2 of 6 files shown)


tests/unit/collection/test_checkpoint.extended.cc
```diff
@@ -172,7 +172,9 @@ TEST_F(TestCheckpoint, test_checkpoint_1) {
   auto num_nodes = static_cast<int32_t>(theContext()->getNumNodes());
 
   auto range = vt::Index3D(num_nodes, num_elms, 4);
-  std::string const checkpoint_name{"test_checkpoint_dir"};
+  std::string const checkpoint_name(addNumNodesToFilename(
+    "test_checkpoint_1_dir_"
```
cz4rs, line 176: gtest has a way to get current test's name:
```cpp
testing::UnitTest::GetInstance()->current_test_info()->name(); // "test_checkpoint_1"
testing::UnitTest::GetInstance()->current_test_info()->test_suite_name(); // "TestCheckpoint"
```
Maybe another helper could be added to `test_helpers.h` to generate fully unique names?

nlslatt: Thanks, @cz4rs!

nlslatt: These filenames will get quite long. Are there modern systems on which that will be an issue?

PhilMiller: Unless you're talking kilobytes long, no, I don't think so

nlslatt: I've implemented something along these lines. I think all the files should be unique at this point.

tests/unit/runtime/test_initialization.cc
```diff
@@ -170,7 +171,13 @@ TEST_F(TestInitialization, test_initialize_with_file_and_args) {
...
   static char vt_no_terminate[]{"--vt_no_terminate"};
   static char vt_lb_name[]{"--vt_lb_name=RotateLB"};
-  static char vt_input_config[]{"--vt_input_config=test_cfg.toml"};
+
+  std::string config_file(getUniqueFilenameWithRanks(".toml"));
+  std::string config_flag("--vt_input_config=");
+  std::string config_arg = config_flag + config_file;
+  static char vt_input_config[1024];
+  memset(vt_input_config, 0, 1024);
+  strncpy(vt_input_config, config_arg.c_str(), config_arg.size());
```
PhilMiller, line 180: I think this whole dance could be replaced by just calling `custom_args.emplace_back(strdup(config_arg.c_str()));` below

Given the lifetimes involved, I think it might even be fine without the `strdup`, since `config_arg` should outlive anything that tries to dereference the elements of `custom_args`

tests/unit/test_helpers.h
```diff
@@ -75,6 +76,39 @@ inline bool isOversubscribed() {
   return num_ranks > CMAKE_DETECTED_MAX_NUM_NODES;
 }
 
+/**
+ * Get a unique filename based on the unit test name.
+ */
+inline std::string getUniqueFilename(
+  const std::string &ext = ""
```
cz4rs, line 83: Formatting:
```suggestion
  const std::string& ext = ""
```

tests/unit/test_helpers.h
```diff
@@ -75,6 +76,39 @@ inline bool isOversubscribed() {
...
+  return str;
+}
+
+/**
+ * Construct a filename containing the number of ranks so that
+ * concurrently-running tests will not cause file system race conditions.
+ * Do not call this from .nompi.cc tests or from addAdditionalArgs().
+ */
+inline std::string getUniqueFilenameWithRanks(
+  const std::string &ext = ""
```
cz4rs, line 104:
```suggestion
  const std::string& ext = ""
```
(this could probably fit in one line as well)

tests/unit/test_helpers.h
```diff
@@ -75,6 +76,39 @@ inline bool isOversubscribed() {
...
+inline std::string getUniqueFilename(
+  const std::string &ext = ""
+) {
+  std::stringstream ss;
+  ss << testing::UnitTest::GetInstance()->current_test_info()->test_suite_name()
+     << "_" << testing::UnitTest::GetInstance()->current_test_info()->name()
+     << ext;
+  std::string str(ss.str());
+  std::size_t pos = 0;
+  while ((pos = str.find("/", pos)) != std::string::npos) {
```
cz4rs, line 91: Nitpicking:
```cpp
std::replace(str.begin(), str.end(), '/', '_');
```

tests/unit/collection/test_checkpoint.extended.cc
cz4rs: The filenames generation looks good, minor comments posted.

tests/unit/test_helpers.h
nlslatt: Fixed.

tests/unit/test_helpers.h
nlslatt: Fixed.

tests/unit/test_helpers.h
nlslatt: Fixed.

tests/unit/runtime/test_initialization.cc
nlslatt: Fixed. I left in the `strdup` because the `c_str` is a `const*` and in case CLI11 is doing anything to modify arguments.

PhilMiller: Looks like it's still doing `memset/strncpy` instead of just `strdup`. Is there a missing commit? Or a reason to not do this with `strdup` that I'm misunderstanding?

cz4rs (approved):
Looks good to me apart from `strdup` issue that Phil has mentioned.

tests/unit/runtime/test_initialization.cc
nlslatt: I changed the specific instance you mentioned because it seemed safe. I'm not sure if the others are safe to change. When I hadn't used the `static` keyword on the declarations, bad things happened, which I think is because the variables go out of scope before they are consumed. Are you suggesting `static char *vt_input_config = strdup(config_arg);`? Is the strdup-allocated memory getting leaked?

PhilMiller: Yes, we would leak the memory allocated by `strdup`. I consider that totally reasonable in this sort of unit test.

The declaration could then just be `char *vt_input_config` without the static

nlslatt: Fixed.

nlslatt: Apparently I had overlooked one earlier. That's now fixed too.

PhilMiller (approved):

PR #1900, opened 2022-08-05
1899 improve makerunnable allocation performance

### LINKED ISSUE #1899: Improve makeRunnable allocation performance
**What Needs to be Done?**

- [ ] First do this
- [ ] Then do that.
- [ ] Also this other thing.

**Additional context**

lifflander:
Fixes #1899


src/vt/pool/static_sized/memory_pool_equal.cc
```diff
@@ -146,3 +147,9 @@ template struct MemoryPoolEqual<memory_size_small>;
 template struct MemoryPoolEqual<memory_size_medium>;
 
 }} //end namespace vt::pool
+
+namespace vt { namespace pool {
```
nlslatt, line 151: Why close the namespace and reopen immediately?

src/vt/runnable/runnable.impl.h
```diff
@@ -52,15 +52,25 @@ namespace vt { namespace runnable {
 
 template <typename T>
 T* RunnableNew::get() {
-  for (auto&& ctx : contexts_) {
-    auto t = dynamic_cast<T*>(ctx.get());
+  for (int i = 0; i < ci_; i++) {
+    auto t = dynamic_cast<T*>(contexts_[i].get());
```
nlslatt, line 56: Given the limited number of context types, would it make sense to designate a specific slot to each context type, to minimize the number of `dynamic_cast` calls?

nlslatt (approved):
Do you want to add the `node_` to the debug prints in `set_context.cc`?

PhilMiller:
Code looks reasonable enough. How much performance impact does this have?

nlslatt:
> The gcc-7 CI build is repeatedly hanging the Azure agent at `Linking CXX executable tests/atomic_basic`

I saw many such hangs in this vicinity from my PR to change the number of ranks used for running the ping pong test, so the hang here may not be related to this PR. Any thoughts on how to figure out what's actually going wrong? Do you think we're exceeding memory in our parallel build (trace is enabled so it might be one of the tougher builds)?

src/vt/pool/static_sized/memory_pool_equal.cc
thearusable: Ok I fexed that.

src/vt/runnable/runnable.impl.h
thearusable: I refactored the code to have a separate member for each context type instead of an array. Is that something that you had in mind?

PhilMiller:
Superseded by later smaller PRs

PR #1916, opened 2022-08-16
1908 1909 1910 performance optimizations 1

stmcgovern:
Closes #1908.
Closes #1909.
Closes #1910.
First PR of multiple for fine-grained performance enhancements.

PR #1924, opened 2022-08-23
#1908: Disable preConfig

### LINKED ISSUE #1908: Disable debug print preConfig evaulation when in production mode
**What Needs to be Done?**

- [ ] First do this
- [ ] Then do that.
- [ ] Also this other thing.

**Additional context**

stmcgovern:
Closes #1908.

### DIFF
--- a/src/vt/configs/debug/debug_print.h
+++ b/src/vt/configs/debug/debug_print.h
@@ -85,7 +85,7 @@
 
 #if vt_check_enabled(production_build)
 #define vt_debug_print_impl(force, inconfig, inmode, cat, ctx, ...)     \
-  vt_force_use(__VA_ARGS__)
+  do { if (false) vt_force_use(__VA_ARGS__); } while (false);
 #else
 #define vt_debug_print_impl(force, inconfig, inmode, cat, ctx, ...)     \
   vt::config::ApplyOp<                                                  \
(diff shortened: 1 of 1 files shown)


stmcgovern:
From grouped PR, see:
https://github.com/DARMA-tasking/vt/pull/1916#discussion_r952919356

PhilMiller (changes requested):
Could you pick up my suggestion from the earlier combined PR?

src/vt/configs/debug/debug_print.h
```diff
@@ -83,6 +83,10 @@
 #define vt_debug_argument_option(opt)                                   \
   ::vt::debug::preConfig()->vt_debug_ ## opt
 
+#if vt_check_enabled(production_build)
+#define vt_debug_print_impl(force, inconfig, inmode, cat, ctx, ...)     \
+  do { if (false) vt_force_use(__VA_ARGS__); } while (false);
```
PhilMiller, line 88: The point of using do-while in a macro is that it forces the caller to treat it like a full statement, with a following semicolon. That only works if the macro definition leaves it off

stmcgovern: Yes - fixed.

PhilMiller (approved):

PR #1949, opened 2022-09-06
1948 improve gitlab ci output to cdash

### LINKED ISSUE #1948: Improve gitlab ci output to cdash
**What Needs to be Done?**

- [ ] Increase the output limit for passing tests.
- [ ] Increase the output limit for failing tests.
- [ ] Add a couple additional config options (set at defaults) for each tweaking down the road.
- [ ] Make ctest submission settings more robust.
- [ ] Remove the node number from the hostname and use as the site so that all submissions from the same HPC system show up as a single site. This helps with plotting build and test time histories.
- [ ] Tell ctest not to run perf tests in parallel with other perf tests.

nlslatt:
Closes #1948


ci/ctest_job_script.cmake
```diff
@@ -1,6 +1,17 @@
 include(${CTEST_SCRIPT_DIRECTORY}/CTestConfig.cmake)
 
-site_name(CTEST_SITE)
+set(CTEST_CUSTOM_MAXIMUM_PASSED_TEST_OUTPUT_SIZE "102400" )
+set(CTEST_CUSTOM_MAXIMUM_FAILED_TEST_OUTPUT_SIZE "512000" )
+set(CTEST_CUSTOM_MAXIMUM_NUMBER_OF_ERRORS "50" )
+set(CTEST_CUSTOM_MAXIMUM_NUMBER_OF_WARNINGS "50")
+
+if(COMMAND cmake_host_system_information)
```
lifflander, line 8: Should we fail if this command doesn't exist?

lifflander (approved):

PhilMiller (approved):

ci/ctest_job_script.cmake
nlslatt: Actually, it should revert back to the old behavior in that case. Thanks for catching that.

nlslatt: It may already do that, but I'll make it more explicit.

nlslatt: Fixed.

lifflander:
If this is ready to merge, we can skip all the github CI since it won't be affected by this.

PR #1951, opened 2022-09-07
1950 use circular buffer to increase efficiency of scheduling

### LINKED ISSUE #1950: Use circular buffer to increase efficiency of scheduling
**What Needs to be Done?**

- [ ] First do this
- [ ] Then do that.
- [ ] Also this other thing.

**Additional context**

lifflander:
Fixes #1950

### DIFF
--- a/src/vt/scheduler/queue.h
+++ b/src/vt/scheduler/queue.h
@@ -74,6 +74,7 @@ struct Queue {
   }
 
   T pop() {
+    vtAssert(not buf_.empty(), "Must have at least one element");
     auto t = buf_.pop();
     if (not impl_.empty()) {
       buf_.push(std::move(impl_.front()));
--- a/src/vt/scheduler/queue.h
+++ b/src/vt/scheduler/queue.h
@@ -75,8 +75,9 @@ struct Queue {
 
   T pop() {
     vtAssert(not buf_.empty(), "Must have at least one element");
+    bool const full_buf = buf_.full();
     auto t = buf_.pop();
-    if (not impl_.empty()) {
+    if (full_buf and not impl_.empty()) {
       buf_.push(std::move(impl_.front()));
       impl_.pop();
     }
--- a/src/vt/scheduler/queue.h
+++ b/src/vt/scheduler/queue.h
@@ -45,6 +45,7 @@
 #define INCLUDED_VT_SCHEDULER_QUEUE_H
 
 #include "vt/config.h"
+#include "vt/utils/container/circular_buffer.h"
 
 #include <queue>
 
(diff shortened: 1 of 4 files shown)


src/vt/scheduler/queue.h
```diff
@@ -56,24 +57,48 @@ struct Queue {
...
+    if (buf_.full()) {
+      impl_.emplace(std::move(elm));
+    } else {
+      buf_.push(std::move(elm));
+    }
+  }
 
-  T const& top() { return impl_.front(); }
+  T pop() {
+    auto t = buf_.pop();
```
PhilMiller, line 77: Can `buf` be empty here? If not, please assert so

src/vt/utils/container/circular_buffer.h
```diff
@@ -0,0 +1,123 @@
...
+*/
+
+#if !defined INCLUDED_VT_UTILS_CONTAINER_CIRCULAR_BUFFER_H
+#define INCLUDED_VT_UTILS_CONTAINER_CIRCULAR_BUFFER_H
+
+#include "vt/config.h"
+
+namespace vt { namespace util { namespace container {
+
+template <typename T, int size>
```
PhilMiller, line 51: Could you rename size to capacity, and then rename len() to size(), to be more similar to standard containers

src/vt/utils/container/circular_buffer.h
```diff
@@ -0,0 +1,123 @@
...
+
+  bool empty() const { return head_ == tail_; }
+  bool full() const { return numFree() == 0; }
+  int len() const { return size - 1 - numFree(); }
+
+  template <typename SerializerT>
+  void serialize(SerializerT& s) {
+    s | head_;
+    s | tail_;
+    s | elms_; // this is inefficient, but it's use is footprinting
```
PhilMiller, line 112: Its

src/vt/scheduler/queue.h
```diff
@@ -56,24 +57,48 @@ struct Queue {
...
+    if (not impl_.empty()) {
```
PhilMiller, line 78: I was thinking of a slightly stronger phrasing of the invariant this maintains, that can make this condition cheaper. If buf is less than full, then impl should be empty. Thus, we should be able to check buf, which is already hot in cache, before having to (much less often) look at impl at all

PhilMiller: Note that testing the condition will of course have at least one slot free in buf at this point

lifflander: Done

src/vt/utils/container/circular_buffer.h
lifflander: Done

src/vt/utils/container/circular_buffer.h
lifflander: Fixed

src/vt/utils/container/circular_buffer.h
```diff
@@ -0,0 +1,123 @@
...
+    } else if (head_ < tail_) {
+      return tail_ - head_ - 1;
+    } else {
+      return capacity + tail_ - head_ - 1;
+    }
+  }
+
+  bool empty() const { return head_ == tail_; }
+  bool full() const { return numFree() == 0; }
+  int size() const { return capacity - 1 - numFree(); }
```
stmcgovern, line 106: size should also be changed to len, as in the tests

lifflander: Fixed

PhilMiller (approved):
Ideally, the unit testing would cover behavior near full as well as near empty. I don't see that as strictly necessary to move this along, though

PR #1957, opened 2022-09-13
#1953: sequencer: completely remove component

### LINKED ISSUE #1953: Remove the dead sequencer code
**What Needs to be Done?**

- [ ] First do this
- [ ] Then do that.
- [ ] Also this other thing.

**Additional context**

lifflander:
Fixes #1953


src/vt/context/runnable_context/set_context.cc
```diff
@@ -41,6 +41,7 @@
 //@HEADER
 */
 
+#include "vt/context/context.h"
```
PhilMiller, line 44: What's this doing here?

PhilMiller (approved):

src/vt/context/runnable_context/set_context.cc
lifflander: When I removed the component some transitive dependency was causing this code to no longer compile.

lifflander: And it does depend on the context so that include should be there.

nlslatt (changes requested):
Some files were missed. At the very least, in `sketches/`.

nlslatt:
This merge appears to have broken the Gitlab CI builds. It looks like it might be a transitive includes problem.

PR #1961, opened 2022-09-14
1941 move EpochStack to TerminationDetector

### LINKED ISSUE #1941: Move epoch stack to TerminationDetector

stmcgovern:
Closes #1941. Add EpockStack to TerminationDetector.

### DIFF
--- a/src/vt/messaging/active.impl.h
+++ b/src/vt/messaging/active.impl.h
@@ -445,54 +445,16 @@ ActiveMessenger::PendingSendType ActiveMessenger::broadcastMsgAuto(
   );
 }
 
-inline EpochType ActiveMessenger::getGlobalEpoch() const {
-  vtAssertInfo(
-    epoch_stack_.size() > 0, "Epoch stack size must be greater than zero",
-    epoch_stack_.size()
-  );
-  return epoch_stack_.size() ? epoch_stack_.top() : term::any_epoch_sentinel;
-}
-
 inline void ActiveMessenger::pushEpoch(EpochType const& epoch) {
-  /*
-   * pushEpoch(epoch) pushes any epoch onto the local stack iff epoch !=
-   * no_epoch; the epoch stack includes all locally pushed epochs and the
-   * current contexts pushed, transitively causally related active message
-   * handlers.
-   */
-  vtAssertInfo(
-    epoch != no_epoch, "Do not push no_epoch onto the epoch stack",
-    epoch, no_epoch, epoch_stack_.size(),
-    epoch_stack_.size() > 0 ? epoch_stack_.top() : no_epoch
-  );
-  if (epoch != no_epoch) {
-    epoch_stack_.push(epoch);
-  }
+  return theTerm()->pushEpoch(epoch);
 }
 
 inline EpochType ActiveMessenger::popEpoch(EpochType const& epoch) {
-  /*
-   * popEpoch(epoch) shall remove the top entry from epoch_size_, iif the size
-   * is non-zero and the `epoch' passed, if `epoch != no_epoch', is equal to the
-   * top of the `epoch_stack_.top()'; else, it shall remove any entry from the
-   * top of the stack.
-   */
-  auto const& non_zero = epoch_stack_.size() > 0;
-  vtAssertExprInfo(
-    non_zero and (epoch_stack_.top() == epoch or epoch == no_epoch),
-    epoch, non_zero, epoch_stack_.top()
-  );
-  if (epoch == no_epoch) {
-    return non_zero ? epoch_stack_.pop(),epoch_stack_.top() : no_epoch;
-  } else {
-    return non_zero && epoch == epoch_stack_.top() ?
-      epoch_stack_.pop(),epoch :
-      no_epoch;
-  }
+  return theTerm()->popEpoch(epoch);
 }
 
 inline EpochType ActiveMessenger::getEpoch() const {
-  return getGlobalEpoch();
+  return theTerm()->getEpoch();
 }
 
 template <typename MsgT>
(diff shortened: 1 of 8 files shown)


src/vt/termination/termination.impl.h
```diff
@@ -83,6 +83,104 @@ inline bool TerminationDetector::isDS(EpochType epoch) {
...
+  /*
+   * popEpoch(epoch) shall remove the top entry from epoch_size_, iif the size
+   * is non-zero and the `epoch' passed, if `epoch != no_epoch', is equal to the
+   * top of the `epoch_stack_.top()'; else, it shall remove any entry from the
+   * top of the stack.
+   */
+  auto const& non_zero = epoch_stack_.size() > 0;
+  vtAssertExprInfo(
+    non_zero and (epoch_stack_.top() == epoch.get() or epoch == no_epoch),
+    epoch, non_zero, epoch_stack_.top()
```
PhilMiller, line 172:
```suggestion
    epoch, non_zero, non_zero ? epoch_stack_.top() : no_epoch
```

src/vt/termination/termination.h
```diff
@@ -71,6 +71,18 @@ namespace vt { namespace term {
...
+struct EpochStack {
+  using DataType = epoch::detail::EpochImplType;
+
+  void push(DataType in) { stack_[cur_++] = in; }
+  DataType top() const { return stack_[cur_-1]; }
+  void pop() { cur_--; }
+  unsigned int size() const { return cur_; }
+
+  int cur_ = 0;
+  std::array<DataType, 64> stack_;
```
PhilMiller, line 83: We'd discussed making this a potentially dynamically sized structure.

As in the scheduler queue, where we use a circular buffer, we may want to do something clever where we keep the `std::array` for the shallow case, and do a bulk hand-off to a growable structure for the *bottom* of the stack that's not changing as rapidly

PhilMiller: I'd suggest opening an issue and doing that in a follow-up commit

lifflander (approved):
Looks good to me!

src/vt/termination/termination.h
```diff
@@ -102,13 +114,27 @@ struct TerminationDetector :
...
    */
   TerminationDetector();
 
-  virtual ~TerminationDetector() {}
+  virtual ~TerminationDetector() {
+    //Pop all extraneous epochs off the stack greater than 1
+      auto stack_size = epoch_stack_.size();
+      while (stack_size > 1) {
+        stack_size = (epoch_stack_.pop(), epoch_stack_.size());
+       }
```
PhilMiller, line 129: The way this is written seems overly clever. The variable `stack_size` isn't used outside the loop control, so the whole thing could be written

```suggestion
      // Pop all extraneous epochs off the stack greater than 1
      while (epoch_stack_.size() > 1) {
        epoch_stack_.pop();
      }
```

Also, the indentation was messed up.

That said, this is not a big concern.

src/vt/termination/termination.h
```diff
@@ -772,11 +796,25 @@ struct TerminationDetector :
    */
   static void epochContinueHandler(TermMsg* msg);
 
-private:
+public:
+  inline EpochType getEpoch() const;
+  inline void pushEpoch(EpochType const& epoch);
+  inline EpochType popEpoch(EpochType const& epoch = no_epoch);
```
PhilMiller, line 802: These should take `epoch` by value

src/vt/termination/termination.h
```diff
@@ -772,11 +796,25 @@ struct TerminationDetector :
...
+
+  inline void pushEpochFast(EpochType epoch) {
+    epoch_stack_.push(epoch.get());
+  }
+  inline void popEpochFast(EpochType epoch) {
```
PhilMiller, line 807: Why is this version taking the `epoch` argument, if it's not going to check it?

src/vt/termination/termination.h
lifflander: You can blame the "cleverness" on me (it was copied from AM).

src/vt/messaging/active.impl.h
```diff
@@ -445,54 +445,16 @@ ActiveMessenger::PendingSendType ActiveMessenger::broadcastMsgAuto(
...
-      epoch_stack_.pop(),epoch :
-      no_epoch;
-  }
+  return theTerm()->popEpoch(epoch);
 }
 
 inline EpochType ActiveMessenger::getEpoch() const {
-  return getGlobalEpoch();
+  return theTerm()->getEpoch();
 }
```
PhilMiller, line 459: I don't really like keeping these wrappers here, but I guess it spares us rewriting all the call sites at the same time. It does also allow for caching `theTerm()` in `ActiveMessenger` for internal usage

PhilMiller:
Merging vs fixing stuff is up to you, @lifflander

src/vt/termination/termination.h
stmcgovern: done

src/vt/termination/termination.h
stmcgovern: done

src/vt/termination/termination.h
stmcgovern: done

src/vt/messaging/active.impl.h
stmcgovern: I've not changed this. Do you think it's worth an issue to remove them or to go with the caching, @PhilMiller @lifflander ?

PhilMiller: Let's leave it be for now, and think about caching if that lookup ever turns up as a hot spot in profiles

PhilMiller (approved):

lifflander:
Since these problems already existed, I would prefer to merge and then fix them in a separate issue.

PR #1962, opened 2022-09-14
1943 scheduler check optimizations

### LINKED ISSUE #1943: Scheduler check optimizations

stmcgovern:
Closes #1943. Add vt_check_enabled statements to reduce unneeded code in the scheduler.

### DIFF
--- a/src/vt/scheduler/suspended_units.cc
+++ b/src/vt/scheduler/suspended_units.cc
@@ -50,6 +50,7 @@ namespace vt { namespace sched {
 void SuspendedUnits::addSuspended(
   ThreadIDType tid, RunnablePtrType runnable, PriorityType p
 ) {
+#if vt_check_enabled(fcontext)
   vtAssert(runnable->isSuspended(), "Runnable must be suspended to add");
 
   units_.emplace(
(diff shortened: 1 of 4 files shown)


stmcgovern:
Deleting the pointer relies on the upcoming changes to runnable. Simply remove for now and revisit after #1899.

PhilMiller:
Is that class's destructor supposed to be virtual?

Separately, is that a memory leak in the current code base?

PhilMiller:
Why does only nvcc think it's not a complete type?

stmcgovern:
The current code base uses a unique pointer- In work for #1899, changes were made to RunnablePtrType. We should consider them there when the PR is ready. 
The checks here are independent of that. 
I don't know about nvcc.

src/vt/scheduler/suspended_units.cc
```diff
@@ -59,15 +60,18 @@ void SuspendedUnits::addSuspended(
       detail::SuspendedRunnable{std::move(runnable), p}
     )
   );
+#endif
```
PhilMiller, line 63: Could you add `#else` `vtAbort` here and below, to make sure we don't have accidental inappropriate calls?

PhilMiller (approved):

src/vt/scheduler/suspended_units.cc
PhilMiller: Or, if we want to be sparing about code size in release builds, it could be `vtAssert(false, ...)`

stmcgovern: Done

PhilMiller (approved):

cz4rs (approved):

PR #1966, opened 2022-09-19
#1958: lib: update googletest to 1.12.1

### LINKED ISSUE #1958: Update GoogleTest (GTest) to at least 1.12.0 to get newer CMake dependency
We've started getting warnings about Google Test's CMakeLists.txt specifying a minimum dependency on CMake 2.8, which is old enough that newer CMake doesn't want to maintain compatibility with it much longer.

Google Test 1.12.0 bumps that to CMake 3.5. Let's try upgrading to 1.12.1, the last release before their 'live at HEAD' transition.

Per #1807, 

* Update GTest to a more recent version
* Verify that our tests work correctly with the new GTest

lifflander:
Fixes #1958


tests/extern/googletest/CMakeLists.txt
```diff
@@ -1,23 +1,21 @@
 # Note: CMake support is community-based. The maintainers do not use CMake
 # internally.
 
-cmake_minimum_required(VERSION 3.17)
+cmake_minimum_required(VERSION 3.5)
```
PhilMiller, line 4: I think we'd perhaps tweaked this?

nmm0: Yeah I think we updated it to avoid the 2.x compat warnings... this change in googletest would address those warnings, but 3.5 is a very old version, so we might end up with problems in the future again

nlslatt: We did: `d09d18510e (Sean McGovern        2022-05-24 15:27:36 +0000  4) cmake_minimum_required(VERSION 3.17)`

nlslatt: The recent warnings were probably coming from `tests/extern/googletest/googletest/CMakeLists.txt`, which didn't get tweaked before.

I'm running this branch through Gitlab CI right now to make sure there won't be any surprises.

lifflander: Shall I apply his CMake version update here? Also, there is a commit that adds C++11 explicitly. @stmcgovern Do you know why [this change was made](https://github.com/DARMA-tasking/vt/commit/3cc2dbc3e2d8d2bd56872143d6f69213a52032dc) and if it's still needed?

nlslatt: This doesn't make sense, but the Intel 18 CI job on this branch still had this in the configure log:
```
CMake Deprecation Warning at tests/extern/googletest/CMakeLists.txt:4 (cmake_minimum_required):
  Compatibility with CMake < 2.8.12 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.


CMake Deprecation Warning at tests/extern/googletest/googletest/CMakeLists.txt:56 (cmake_minimum_required):
  Compatibility with CMake < 2.8.12 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.

```

Also, all four Gitlab CI builds on this branch indicate there was a warning in the configure step, but ONLY the Intel 18 build shows any warnings in the captured output.

stmcgovern: @lifflander I made the change you refer to in order to keep the CMakeLists as it was in PR1776 (and before). That was related to using vt compilation flags for bundled libraries.

stmcgovern: Those specifications are from over 2 years ago, so maybe are outdated now...

cz4rs: > This doesn't make sense, but the Intel 18 CI job on this branch still had this in the configure log:
This looks like the wrong version being run in CI 🤷 Hard to explain otherwise.

cz4rs: Regarding `cmake_minimum_required(VERSION 3.5)`, if the CI is happy with it then I think we cal leave it and worry about any later warning when they appear.

tests/extern/googletest/googletest/src/gtest.cc
```diff
@@ -6222,8 +6244,8 @@ void UnitTestImpl::UnshuffleTests() {
 // For example, if Foo() calls Bar(), which in turn calls
 // GetCurrentOsStackTraceExceptTop(..., 1), Foo() will be included in
 // the trace but Bar() and GetCurrentOsStackTraceExceptTop() won't.
-std::string GetCurrentOsStackTraceExceptTop(UnitTest* /*unit_test*/,
-                                            int skip_count) {
+GTEST_NO_INLINE_ GTEST_NO_TAIL_CALL_ std::string
```
cz4rs, line 6247: Intel icpc:
```
/vt/tests/extern/googletest/googletest/src/gtest.cc(6247): warning #3175: unrecognized gcc optimization level
  GTEST_NO_INLINE_ GTEST_NO_TAIL_CALL_ std::string
```

tests/extern/googletest/googletest/src/gtest-internal-inl.h
```diff
@@ -640,7 +632,8 @@ class GTEST_API_ UnitTestImpl {
   // For example, if Foo() calls Bar(), which in turn calls
   // CurrentOsStackTraceExceptTop(1), Foo() will be included in the
   // trace but Bar() and CurrentOsStackTraceExceptTop() won't.
-  std::string CurrentOsStackTraceExceptTop(int skip_count) GTEST_NO_INLINE_;
+  std::string CurrentOsStackTraceExceptTop(int skip_count)
+      GTEST_NO_INLINE_ GTEST_NO_TAIL_CALL_;
```
cz4rs, line 636: Intel icpc:
```
warning #3175: unrecognized gcc optimization level
        GTEST_NO_INLINE_ GTEST_NO_TAIL_CALL_;
```

tests/extern/googletest/CMakeLists.txt
nlslatt: > > This doesn't make sense, but the Intel 18 CI job on this branch still had this in the configure log:
> 
> This looks like the wrong version being run in CI 🤷 Hard to explain otherwise.

Now that we have clean/dirty repo indicators on CDash, I remember that Intel 18 downloads an old version of google test because the newer ones don't build with it. That's why the warning is there, so it has nothing to do with this PR. The CMake warnings listed by the others all seem to be related to variables not used by the project.

tests/extern/googletest/googletest/src/gtest.cc
nlslatt: Based on a similar problem noted by Christian Trott (that they refused to fix for him), I think the problem here is that some non-GCC compilers define `__GNUC__` but don't fully support everything that GCC does. The part of the `googletest/include/gtest/internal/gtest-port.h` file that defines `GTEST_NO_TAIL_CALL_` has changed since the 1.12.1 release (**very** recently, actually). https://github.com/google/googletest/pull/4025 What do you all think about trying out gtest `main`? I have no idea how stable it is or if that PR will even help us. We could potentially hack in something ourselves that excludes `icpc` from trying to use the requested optimization.

nlslatt:
My attempt to bump this to googletest `main` did not resolve the warning, so I propose we rebase out that commit and just hack the definition of `GTEST_NO_TAIL_CALL_`.

nlslatt:
Looks like my hack eliminated the warnings. It would be great to get reviews.

cz4rs (approved):
Looks good to me!

PhilMiller (approved):

PR #1975, opened 2022-09-22
#1974: fix various c++17 deprecations and compile errors

### LINKED ISSUE #1974: address C++17 deprecations
**What Needs to be Done?**

- [ ] Replace `std::is_pod` with `std::is_standard_layout && std::is_trivial`
- [ ] Remove `std::iterator`

nmm0:
Fixes #1974

### DIFF
--- a/src/vt/termination/interval/integral_set.h
+++ b/src/vt/termination/interval/integral_set.h
@@ -378,9 +378,7 @@ public:
   // using reference         = DomainT&;
 
   template <typename Impl>
-  struct IntervalSetIter :
-    std::iterator<std::bidirectional_iterator_tag, DomainT, DomainT>
-  {
+  struct IntervalSetIter {
     using Iter              = IntervalSetIter;
     using iterator_category = std::bidirectional_iterator_tag;
     using difference_type   = DomainT;
--- a/src/vt/messaging/envelope/envelope_base.h
+++ b/src/vt/messaging/envelope/envelope_base.h
@@ -113,8 +113,7 @@ namespace vt {
 
 using Envelope = messaging::ActiveEnvelope;
 
-static_assert(std::is_pod<Envelope>(), "Envelope must be POD");
-static_assert(std::is_trivially_destructible<Envelope>(), "Envelope must be trivially destructible");
+static_assert(std::is_standard_layout<Envelope>::value && std::is_trivial<Envelope>::value, "Envelope must be standard layout and trivial");
 
 } /* end namespace vt */
 
--- a/src/vt/messaging/envelope/envelope_extended.h
+++ b/src/vt/messaging/envelope/envelope_extended.h
@@ -104,19 +104,19 @@ using EpochEnvelope    = messaging::EpochActiveEnvelope;
 using TagEnvelope      = messaging::TagActiveEnvelope;
 using EpochTagEnvelope = messaging::EpochTagActiveEnvelope;
 
-static_assert(std::is_pod<EpochEnvelope>(),    "EpochEnvelope must be POD");
-static_assert(std::is_pod<TagEnvelope>(),      "TagEnvelope must be POD");
-static_assert(std::is_pod<EpochTagEnvelope>(), "EpochTagEnvelope must be POD");
+static_assert(std::is_standard_layout<EpochEnvelope>::value,    "EpochEnvelope must be standard layout");
+static_assert(std::is_standard_layout<TagEnvelope>::value,      "TagEnvelope must be standard layout");
+static_assert(std::is_standard_layout<EpochTagEnvelope>::value, "EpochTagEnvelope must be standard layout");
 static_assert(
-  std::is_trivially_destructible<EpochEnvelope>(),
+  std::is_trivial<EpochEnvelope>::value,
   "EpochEnvelope must be trivial"
 );
 static_assert(
-  std::is_trivially_destructible<TagEnvelope>(),
+  std::is_trivial<TagEnvelope>::value,
   "TagEnvelope must be trivial"
 );
 static_assert(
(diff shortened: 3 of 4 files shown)


PhilMiller:
Lol at those compiler errors

src/vt/termination/interval/integral_set.h
```diff
@@ -378,8 +378,7 @@ struct IntegralSetBase {
   // using reference         = DomainT&;
 
   template <typename Impl>
-  struct IntervalSetIter :
-    std::iterator<std::bidirectional_iterator_tag, DomainT, DomainT>
+  struct IntervalSetIter
```
PhilMiller, line 381: I'm not familiar with this bit of C++ evolution - what previously depended on having that inheritance?

src/vt/vrt/collection/balance/lb_common.h
```diff
@@ -76,6 +76,10 @@ static constexpr ElementIDType const no_element_id = elm::no_element_id;
 struct PhaseOffset {
   PhaseOffset() = delete;
 
+  PhaseOffset(int ph, unsigned int sub)
+    : phases(ph), subphase(sub)
+  {}
+
```
PhilMiller, line 82: What changed with C++17 that we need an explict constructor rather than just an implicit initializer list construction?

nmm0:
> Lol at those compiler errors

:facepalm: maybe I should have tested with C++14.... well I guess I'm going to have to find another way. Maybe the least gross way will be to just do explicit capture for everything, no default capture

src/vt/termination/interval/integral_set.h
nmm0: Yeah I was curious about that too -- I found this explanation: https://www.fluentcpp.com/2018/05/08/std-iterator-deprecated/

tl;dr basically `std::iterator` was just a helper and would take as template parameters the trait types (e.g. value_type). We never needed it since we specify the types directly.

PhilMiller:
> > Lol at those compiler errors
> 
> 🤦 maybe I should have tested with C++14.... well I guess I'm going to have to find another way. Maybe the least gross way will be to just do explicit capture for everything, no default capture

Or wait a bit, and we can fold in "require C++17" with the changes to adapt to it. EMPIRE will be able to accept that in the next few weeks, BVH is already there as I understand it, and I can check on our other app.

src/vt/termination/interval/integral_set.h
PhilMiller: Thanks, I see now.

It seems that the types as specified maybe aren't right, though? They even disagree with the comment just above.

@lifflander ?

nmm0: Hmm yeah it looks like pointer should be `DomainT*` and reference `DomainT&`

src/vt/vrt/collection/balance/lb_common.h
nmm0: Ok funny story, I apparently I had a C++20 flag from a dependency. That was causing the issue, though it _shouldn't_ have. Having a deleted constructor doesn't count as a user-defined constructor in C++17 or C++20. Anyway, I'll revert that change since it isn't necessary for this PR though it does fail in C++20 mode

nmm0:
> > > Lol at those compiler errors
> > 
> > 
> > facepalm maybe I should have tested with C++14.... well I guess I'm going to have to find another way. Maybe the least gross way will be to just do explicit capture for everything, no default capture
> 
> Or wait a bit, and we can fold in "require C++17" with the changes to adapt to it. EMPIRE will be able to accept that in the next few weeks, BVH is already there as I understand it, and I can check on our other app.

Yeah that is my bad, I was getting the C++20 flag from Kokkos. I forgot implicit this is deprecated in 20, not 17. I'll remove it from the PR

PhilMiller:
Incidentally, you're "I'll revert that" are a good argument for committing distinct changes separately - it would be really easy to just rebase to put the `iterator` change we're keeping first, push that on this branch, and push the rest on a C++20 branch.

We'll want to figure out how to write code that's happy compiled as C++17 or C++20 starting once we require 17.

src/vt/termination/interval/integral_set.h
```diff
@@ -378,9 +378,7 @@ struct IntegralSetBase {
   // using reference         = DomainT&;
 
   template <typename Impl>
-  struct IntervalSetIter :
-    std::iterator<std::bidirectional_iterator_tag, DomainT, DomainT>
-  {
+  struct IntervalSetIter {
```
PhilMiller, line 381: Could you also open an issue about the type aliases here being suspicious?

PhilMiller (approved):

nmm0:
> Incidentally, you're "I'll revert that" are a good argument for committing distinct changes separately - it would be really easy to just rebase to put the `iterator` change we're keeping first, push that on this branch, and push the rest on a C++20 branch.
> 
> We'll want to figure out how to write code that's happy compiled as C++17 or C++20 starting once we require 17.

Split into two commits, made a third one for implicit `this` on the C++20 deprecations branch: https://github.com/DARMA-tasking/vt/tree/1976-address-cpp20-deprecations

src/vt/termination/interval/integral_set.h
nmm0: https://github.com/DARMA-tasking/vt/issues/1977

src/vt/termination/interval/integral_set.h
nmm0: https://github.com/DARMA-tasking/vt/issues/1977

nmm0:
This PR is broken because our checking scripts don't allow '+' in the name of the branch. New PR: https://github.com/DARMA-tasking/vt/pull/1978

PR #1984, opened 2022-09-28
1983 Runtime optimizations 2 -- objgroup dispatch/locman/contexts

### LINKED ISSUE #1983: Make location manager more efficient
**What Needs to be Done?**

- [ ] First do this
- [ ] Then do that.
- [ ] Also this other thing.

**Additional context**

lifflander:
Fixes #1983


src/vt/topos/location/location.impl.h
```diff
@@ -277,14 +277,36 @@ void EntityLocationCoord<EntityID>::clearCache() {
...
+
+template <typename MsgT, typename=void>
+struct IsSerializable {
+  static constexpr bool const is_ser = false;
+};
+
+template <typename MsgT>
+struct IsSerializable<
+  MsgT,
+  typename std::enable_if_t<true
```
PhilMiller, line 290: What's up with the `true and ...` thing here?

src/vt/topos/location/location.impl.h
```diff
@@ -533,8 +536,9 @@ void EntityLocationCoord<EntityID>::routeMsgNode(
     // set the instance on the message to deliver to the correct manager
     msg->setLocInst(this_inst);
 
+    auto m = msg;
```
PhilMiller, line 538: Huh?

src/vt/topos/location/location.impl.h
```diff
@@ -668,7 +668,22 @@ void EntityLocationCoord<EntityID>::routeMsgHandler(
...
+}
+
+template <typename EntityID>
+template <typename MessageT>
+void EntityLocationCoord<EntityID>::routeMsgHandlerLocal(
+  MsgSharedPtr<MessageT> const& msg
+) {
+  runnable::makeRunnable(msg, true, msg->getHandler(), theContext()->getNode())
+    .withTDEpochFromMsg()
+    .run();
```
PhilMiller, line 713: Should this be `.run()` or `.enqueue()`?

src/vt/topos/location/location.impl.h
```diff
@@ -668,7 +668,22 @@ void EntityLocationCoord<EntityID>::routeMsgHandler(
 # endif
 
   msg->setHandler(handler);
-  return routeMsg<MessageT>(id,home_node,msg);
+
+  if (local_registered_.find(id) == local_registered_.end()) {
+    return routeMsg<MessageT>(id,home_node,msg);
+  } else {
+    return routeMsgHandlerLocal(msg);
```
PhilMiller: Can we somehow take advantage of having already done the lookup in `local_registered_` in this case?

src/vt/runnable/runnable.cc
```diff
@@ -185,7 +157,11 @@ void RunnableNew::run() {
     vt_force_use(is_threaded_, tid_)
 #endif
 
-    task_();
+    if (is_scatter_) {
+      f_.func_scat_->dispatch(msg_ == nullptr ? nullptr : msg_.get(), obj_);
+    } else {
+      f_.func_->dispatch(msg_ == nullptr ? nullptr : msg_.get(), obj_);
```
PhilMiller, line 166: If these have the same calling signature, why do we need the union of the two different type aliases?

src/vt/objgroup/dispatch/dispatch.impl.h
```diff
@@ -47,25 +47,28 @@
...
-  (obj_->*type_func)(msg);
-  if (cur_epoch != no_epoch) {
-    theMsg()->popEpoch(cur_epoch);
-  }
+  auto m = promoteMsg(tmsg);
+  auto holder = detail::getHolderBase(han);
+  auto const& elm_id = holder->getElmID();
+  auto lb_data = &holder->getLBData();
+
+  runnable::makeRunnable(m, true, han, from_node)
```
PhilMiller, line 66: Can we do `std::move(m)` here and take it at `MsgSharedPtr<T> &&`? Are there cases in the code where this isn't a transfer of ownership?

src/vt/runnable/make_runnable.h
```diff
@@ -261,28 +276,28 @@ struct RunnableMaker {
   }
 
   /**
-   * \brief Enqueue the runnable in the scheduler for execution later
+   * \brief Run the runnable immediately with a lambda
    */
-  void enqueue();
+  void runLambda(ActionType action) {
```
PhilMiller, line 281:
```suggestion
  void runLambda(ActionType&& action) {
```

src/vt/topos/location/location.impl.h
lifflander: This is just a stylistic thing. Look at message_serialize.h. It's prevalent.

src/vt/topos/location/location.impl.h
lifflander: We know it's local so we create the runnable directly for it to execute. `local_registered_` does not hold the pointer to the actual object (it does not manage objects), just an entry that corresponds to its existence.

src/vt/runnable/runnable.cc
lifflander: `func_` and `func_scat_` are different types. I don't think there is a safe way around this.

src/vt/topos/location/location.impl.h
lifflander: This is a difficult one. `.run()` is correct for all the use cases that currently exist. It runs a system handler on the collection manager that generates the next runnable that is enqueued.

src/vt/objgroup/dispatch/dispatch.impl.h
lifflander: `makeRunnable` is actually taking a `const&` here. It ends up populating the actual Runnable (which is not const&). In theory, a move might be possible. I'm not sure if there is always a valid transfer of ownership at these callsites. I will have to think about this more.

src/vt/runnable/runnable.cc
lifflander: And they actually do have different signatures. It's just not obvious from these callsites. One treats `msg_` as a vt::Message; the other treats it as non-message bytes.

src/vt/runnable/make_runnable.h
lifflander: Fixed

src/vt/topos/location/location.impl.h
PhilMiller: That seems kinda wasteful. Could we add a method in the collection manager to do that more directly?

src/vt/objgroup/dispatch/dispatch.impl.h
```diff
@@ -52,20 +52,15 @@ namespace vt { namespace objgroup { namespace dispatch {
 
 template <typename ObjT>
 void Dispatch<ObjT>::run(HandlerType han, BaseMessage* msg) {
-  using ActiveFnType = void(ObjT::*)(vt::BaseMessage*);
+  //using ActiveFnType = void(ObjT::*)(vt::BaseMessage*);
```
PhilMiller: Delete commented-out code?

src/vt/objgroup/dispatch/dispatch.impl.h
```diff
@@ -52,20 +52,15 @@ namespace vt { namespace objgroup { namespace dispatch {
...
-  auto type_func = reinterpret_cast<ActiveFnType>(base_func);
-  (obj_->*type_func)(msg);
-  if (cur_epoch != no_epoch) {
-    theMsg()->popEpoch(cur_epoch);
-  }
+  auto m = promoteMsg(tmsg);
+  runnable::makeRunnable(m, true, han, theContext()->getNode())
+    .withObjGroup(obj_)
+    .withTDEpochFromMsg()
+    .enqueue();
```
PhilMiller, line 71: Why the switch from running inline to enqueueing here? This is a `run()` method, so if it really should be enqueueing, then we should rename it

src/vt/messaging/active.cc
```diff
@@ -498,10 +498,16 @@ EventType ActiveMessenger::doMessageSend(
...
       recordLBDataCommForSend(dest, base, base.size());
 
-      runnable::makeRunnable(base, true, envelopeGetHandler(msg->env), dest)
-        .withTDEpochFromMsg(is_term)
-        .withLBData(&bare_handler_lb_data_, bare_handler_dummy_elm_id_for_lb_data_)
-        .enqueue();
+      auto han = envelopeGetHandler(msg->env);
+      bool const is_obj = HandlerManagerType::isHandlerObjGroup(han);
+      if (is_obj) {
+        objgroup::dispatchObjGroup(base, han, dest, nullptr);
```
PhilMiller, line 504: There's an apparent inconsistency here - the third argument here is called `dest`, but the corresponding parameter is called `from_node`, which matches the argument at all the other call sites. Which is it?

src/vt/objgroup/manager.static.h
```diff
@@ -99,6 +103,75 @@ messaging::PendingSend broadcast(MsgSharedPtr<MsgT> msg, HandlerType han) {
...
+  auto& objs = getObjs();
+  auto obj_iter = objs.find(proxy);
+  vt_debug_print(
+    normal, objgroup,
+    "dispatch: try ctrl={:x}, han={:x}, has dispatch={}\n",
+    ctrl, han, obj_iter != objs.end()
+  );
+  if (obj_iter == objs.end()) {
+    auto const epoch = envelopeGetEpoch(msg->env);
+    if (epoch != no_epoch and epoch != term::any_epoch_sentinel) {
```
PhilMiller, line 146: I forget - is `any_epoch_sentinel` distinct from the global epoch? Having a message sitting enqueued like this should inhibit detection of global termination

src/vt/objgroup/manager.static.h
```diff
@@ -99,6 +103,75 @@ messaging::PendingSend broadcast(MsgSharedPtr<MsgT> msg, HandlerType han) {
...
+      theTerm()->produce(epoch);
+    }
+    auto& pending = getPending();
+    pending[proxy].emplace_back([=]{
+      auto& objs2 = getObjs();
+      auto obj_iter2 = objs2.find(proxy);
+      vtAssert(obj_iter2 != objs2.end(), "Obj must exist");
+      detail::dispatchImpl(msg, han, from_node, cont, obj_iter2->second->getPtr());
```
PhilMiller, line 154: Is this lambda the only thing of its sort that gets pushed into `pending[proxy]`? If so, we should probably make that a container of structs that hold the relevant captured values, and move this code out of a lambda, so that we avoid additionally allocating behind the scenes for the captured variables.

src/vt/objgroup/manager.static.h
lifflander: No, they are the same.

src/vt/objgroup/manager.static.h
lifflander: I realized this capture was expensive, but it's completely off the critical path. Thus, I decided not to optimize it needlessly. It's only the case where a message arrives and races with the creation of the objgroup.

src/vt/messaging/active.cc
lifflander: This is the case where a self-send is happening; thus, the destination is the from node.

src/vt/objgroup/manager.static.h
lifflander: Fixed this

src/vt/objgroup/dispatch/dispatch.impl.h
lifflander: I've managed to do this

PhilMiller:
Could you squash down the small patch-up commits that correct mistakes in earlier commits in the PR into those earlier commits?

PhilMiller (approved):
Would be good to get at least one other review on this, since it's pretty big

nmm0:
Any chance this could be split up into separate optimizations/PRs with maybe a short description of what's going on in it? It's a bit difficult to follow what all of the changes are doing or how they are more efficient.

For example there are a few changes (the `std:function` optimization) that I could easily approve already but the changes to objGroup dispatches I'm still trying to understand.

lifflander:
> Any chance this could be split up into separate optimizations/PRs with maybe a short description of what's going on in it? It's a bit difficult to follow what all of the changes are doing or how they are more efficient.
> 
> For example there are a few changes (the `std:function` optimization) that I could easily approve already but the changes to objGroup dispatches I'm still trying to understand.

Basically, the objgroup dispatch wasn't going through RunnableNew the same way a collection goes through it causing extra overhead.

src/vt/runnable/runnable.cc
```diff
@@ -53,83 +53,58 @@
...
-void RunnableNew::setupHandler(HandlerType handler, bool is_void) {
+void RunnableNew::setupHandler(HandlerType handler) {
   using HandlerManagerType = HandlerManager;
+
   bool const is_obj = HandlerManagerType::isHandlerObjGroup(handler);
+  vtAssert(not is_obj, "Must not be object");
 
-  if (not is_void) {
-    if (is_obj) {
-      task_ = [=] { objgroup::dispatchObjGroup(msg_, handler); };
```
lifflander, line 62: This is the old code, where we had to create a std::function that calls back into the objgroup manager code (because the object isn't known here).

src/vt/topos/location/location.impl.h
```diff
@@ -656,6 +656,35 @@ template <typename MessageT, ActiveTypedFnType<MessageT> *f>
...
+) {
+  setupMessageForRouting<MessageT, f>(id, home_node, msg);
+
+  routePreparedMsgHandler(msg);
+
+  if (local_registered_.find(id) == local_registered_.end()) {
+    return routeMsg<MessageT>(id,home_node,msg);
+  } else {
+    return routeMsgHandlerLocal(msg);
+  }
```
nlslatt, line 668: This looks redundant. Are you sure it's correct?

src/vt/topos/location/location.impl.h
```diff
@@ -688,50 +715,58 @@ void EntityLocationCoord<EntityID>::routeMsgHandlerLocal(
...
   bool const use_eager = useEagerProtocol(msg);
   auto const epoch = theMsg()->getEpochContextMsg(msg);
 
   vt_debug_print(
     verbose, location,
     "routeMsg: inst={}, home={}, msg_size={}, is_large_msg={}, eager={}, "
-    "in_from={}, from={}, msg{}, msg from={}, epoch={:x}\n",
-    this_inst, home_node, msg_size, msg_size > small_msg_max_size, use_eager,
-    from_node, from, print_ptr(msg.get()), msg->getLocFromNode(),
+    "msg{}, msg from={}, epoch={:x}\n",
```
nlslatt, line 728: Is this a typo?

src/vt/topos/location/location.impl.h
lifflander: Good catch. I think this is a bug.

src/vt/topos/location/location.impl.h
lifflander: Fixed

src/vt/runnable/make_runnable.h
nmm0: Did you forget to push this, I still see the copy?

src/vt/runnable/runnable.h
```diff
@@ -334,7 +339,12 @@ struct RunnableNew {
 private:
   detail::Contexts contexts_;               /**< The contexts  */
   MsgSharedPtr<BaseMsgType> msg_ = nullptr; /**< The associated message */
-  ActionType task_ = nullptr;               /**< The runnable's task  */
+  void* obj_ = nullptr;                     /**< Object pointer */
+  union {
```
nmm0, line 343: I don't think we should hold this up because of it, since C++17 is still WIP, but this would be much more elegant with C++17 `variant` and `std::visit`

nmm0 (approved):
Just some minor suggestions -- about the `auto` thing it's more up to personal preference. Just one thing to flag is Phil's comment, you mentioned it changed but it didn't look like the diff changed.

src/vt/objgroup/manager.cc
```diff
@@ -163,12 +137,12 @@ elm::ElementIDStruct ObjGroupManager::getNextElm(ObjGroupProxyType proxy) {
   }
 }
 
-void dispatchObjGroup(MsgSharedPtr<ShortMessage> msg, HandlerType han) {
-  vt_debug_print(
-    verbose, objgroup,
-    "dispatchObjGroup: han={:x}\n", han
-  );
-  return theObjGroup()->dispatch(msg,han);
+std::unordered_map<ObjGroupProxyType, std::unique_ptr<holder::HolderBase>>& getObjs() {
```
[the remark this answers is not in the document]
lifflander: Done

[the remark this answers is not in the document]
lifflander: I actually tried to do this and it causes a compile-time failure because the friend declaration is ambiguous once the function is `auto&`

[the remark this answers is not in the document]
nmm0: ooohh got it

PR #1990, opened 2022-10-10
#1940: add commit date in CI results comment

### LINKED ISSUE #1940: Add commit time beside commit hash in CI results comment
When looking at the PR comment showing the results of the various CI build/test jobs, it would be helpful to see the commit time (not the author time) of the referenced commit, so that one could more easily tell which jobs were run on the most recently rebased/pushed commits

jstrzebonski:
Fixes #1940

### DIFF
--- a/scripts/report_logs_in_comment.sh
+++ b/scripts/report_logs_in_comment.sh
@@ -105,7 +105,7 @@ fi
 commit_sha="$(git log --skip=1 -1  --pretty=format:%H)"
 commit_date="$(TZ=UTC0 git show -s --format=%cd --date=format-local:'%Y-%m-%d %H:%M:%S' "$commit_sha")"
 build_link='[Build log](https://dev.azure.com/DARMA-tasking/DARMA/_build/results?buildId='"$build_id"'&view=logs&j='"$job_id"'&t='"$task_id)"
-comment_body="Build for $commit_sha ($commit_date)\n\n"'```'"\n$val\n"'```'"\n\n$build_link"
+comment_body="Build for $commit_sha ($commit_date UTC)\n\n"'```'"\n$val\n"'```'"\n\n$build_link"
 
 # Fix new lines
 new_line="\n"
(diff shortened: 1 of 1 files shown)


scripts/report_logs_in_comment.sh
```diff
@@ -103,8 +103,9 @@ fi
 
 # Build comment
 commit_sha="$(git log --skip=1 -1  --pretty=format:%H)"
+commit_date="$(git show -s --format=%cd --date=format:'%Y-%m-%d %H:%M:%S' "$commit_sha")"
```
cz4rs, line 106: If I understand correctly, this will use committer's timezone. I guess this could work, but please put a comment here to make it explicit.

jstrzebonski: Yes.
As I understood, @PhilMiller intention was to make build logs even more readable for PR's owner in the first place.

But if we want to standardize this, I can print the date using the UTC timezone.

cz4rs (approved):

scripts/report_logs_in_comment.sh
PhilMiller: I think UTC would be preferable. It would avoid confusion between commits made by (for example) me in California, and you in Poland.

scripts/report_logs_in_comment.sh
```diff
@@ -103,8 +103,9 @@ fi
 
 # Build comment
 commit_sha="$(git log --skip=1 -1  --pretty=format:%H)"
+commit_date="$(TZ=UTC0 git show -s --format=%cd --date=format-local:'%Y-%m-%d %H:%M:%S' "$commit_sha")"
 build_link='[Build log](https://dev.azure.com/DARMA-tasking/DARMA/_build/results?buildId='"$build_id"'&view=logs&j='"$job_id"'&t='"$task_id)"
-comment_body="Build for $commit_sha\n\n"'```'"\n$val\n"'```'"\n\n$build_link"
+comment_body="Build for $commit_sha ($commit_date)\n\n"'```'"\n$val\n"'```'"\n\n$build_link"
```
nlslatt, line 108: Can you add `UTC` to this comment line so that we don't forget what timezone we're using?

jstrzebonski: Done.

PhilMiller (approved):

PhilMiller:
The messages here make a great demonstration of exactly how this is working :-)

PR #1996, opened 2022-10-17
#1934: Add parameter to control minimal retention of historical LB data

### LINKED ISSUE #1934: Add configuration for minimum phase of historical LB data to retain
**What Needs to be Done?**

Right now, the number of phases worth of historical LB data that the system retains is based on how many the installed load model says it wants to use. We should add a configuration parameter to set a minimum retention, independent of the load model. This will support later programmatic use of historical data, like outputting based on some trigger event/observation, with historical data before activation for context.

**Additional context**

thearusable:
Closes #1934

### DIFF
--- a/tests/unit/collection/test_lb_data_retention.cc
+++ b/tests/unit/collection/test_lb_data_retention.cc
@@ -139,10 +139,17 @@ struct TestCol : vt::Collection<TestCol,vt::Index1D> {
     auto sp_load_phase_count = lb_data.getSubphaseLoadPhaseCount();
     auto sp_comm_phase_count = lb_data.getSubphaseCommPhaseCount();
 
-    EXPECT_EQ(load_phase_count,    std::size_t{1});
-    EXPECT_EQ(sp_load_phase_count, std::size_t{1});
-    EXPECT_EQ(comm_phase_count,    std::size_t{1});
-    EXPECT_EQ(sp_comm_phase_count, std::size_t{1});
+    #if vt_check_enabled(lblite)
+      EXPECT_EQ(load_phase_count,    std::size_t{1});
+      EXPECT_EQ(sp_load_phase_count, std::size_t{1});
+      EXPECT_EQ(comm_phase_count,    std::size_t{1});
+      EXPECT_EQ(sp_comm_phase_count, std::size_t{1});
+    #else
+      EXPECT_EQ(load_phase_count,    std::size_t{0});
+      EXPECT_EQ(sp_load_phase_count, std::size_t{0});
+      EXPECT_EQ(comm_phase_count,    std::size_t{0});
+      EXPECT_EQ(sp_comm_phase_count, std::size_t{0});
+    #endif
   }
 };
 


tests/unit/collection/test_lb_data_retention.cc
```diff
@@ -226,4 +230,44 @@ TEST_F(TestLBDataRetention, test_lbdata_retention_last4) {
...
+  theLBManager()->setLoadModel(persist);
+
+  for (int i=0; i<num_phases; ++i) {
+    runInEpochCollective([&]{
+      // Do some work.
+      proxy.broadcastCollective<MyMsg<TestCol>, TestCol::colHandler>();
+    });
+    // Go to the next phase.
+    vt::thePhase()->nextPhaseCollective();
+  }
```
PhilMiller, line 270: Should this test go on to access data older than the model's requirement?

src/vt/vrt/collection/balance/node_lb_data.cc
```diff
@@ -331,7 +331,11 @@ void NodeLBData::addNodeLBData(
   in->updatePhase(1);
 
   auto model = theLBManager()->getLoadModel();
-  in->releaseLBDataFromUnneededPhases(phase, model->getNumPastPhasesNeeded());
+  if(min_hist_lb_data_ > 0){
+    in->releaseLBDataFromUnneededPhases(phase, min_hist_lb_data_);
```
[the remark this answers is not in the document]
thearusable: Yes you are right! Fixed.

src/vt/vrt/collection/balance/node_lb_data.h
```diff
@@ -249,6 +249,13 @@ struct NodeLBData : runtime::component::Component<NodeLBData> {
    */
   LBDataHolder* getLBData() { return lb_data_.get(); }
 
+  /**
+   * \internal \brief Set the minimal amount of historical LB data which should be hold
```
cz4rs, line 253:
```suggestion
   * \internal \brief Set the minimal amount of historical LB data which should be held
```

cz4rs: Or maybe even better:
```suggestion
   * \internal \brief Set the minimal amount of historical LB data which should be retained
```
?

cz4rs: Whichever option you choose, the same applies to line 255 below.

tests/unit/collection/test_lb_data_retention.cc
thearusable: I extended the amount of the for loop iterations to the configured minimum data retention. Is that something you had in mind?

src/vt/vrt/collection/balance/lb_invoke/lb_manager.cc
```diff
@@ -465,7 +469,7 @@ void LBManager::finishedLB(PhaseType phase) {
     "finishedLB\n"
   );
 
-  theNodeLBData()->startIterCleanup(phase, model_->getNumPastPhasesNeeded());
+  theNodeLBData()->startIterCleanup(phase, min_hist_lb_data_);
```
PhilMiller, line 472: `NodeLBData` now keeps that as a member variable, so it shouldn't be necessary to pass it as an argument any more.

tests/unit/collection/test_lb_data_retention.cc
PhilMiller: Ah, I see that `TestCol::colHandler` does actually check that the data for older phases is present. The tests look good.

PhilMiller:
Other than the minor interface thing I noted, I'd approve this.

src/vt/vrt/collection/balance/lb_invoke/lb_manager.cc
thearusable: Ok, it's removed now.

cz4rs (dismissed):
Looks good to me.

PhilMiller (approved):
This looks ready to go.

Could you think through whether there are any weird concerns if a new model gets set and that changes the number of phases retained? In particular, I can think of the possibility that we keep some stale data forever if a new model demands fewer phases than its predecessor.

thearusable:
@PhilMiller Sorry for the delay, I missed your message. 

From what I understand the scenario that you described is the only one which can have some weird behaviors. LB data is being cleaned after each phase so that will occur rather frequently. So if there is more data than model needs then it will not be retained for long. 
The only way to retain that data even if it is not needed is to use the configuration  `LBType::NoLB`, in this case LB will not be run so also the data will not be removed.

src/vt/vrt/collection/balance/lb_invoke/lb_manager.cc
```diff
@@ -159,6 +161,8 @@ LBType LBManager::decideLBToRun(PhaseType phase, bool try_file) {
 void LBManager::setLoadModel(std::shared_ptr<LoadModel> model) {
   model_ = model;
   auto nlb_data = theNodeLBData();
+  min_hist_lb_data_ = std::max(model->getNumPastPhasesNeeded(), theConfig()->vt_lb_data_retention);
+  nlb_data->setMinLBDataHistory(min_hist_lb_data_);
```
PhilMiller, line 165: To address the edge case I brought up, could you add a bit of code here to loop from `max(0, current_phase - previous_min_hist_lb_data_)` to `max(0, current_phase - min_hist_lb_data_)` and clear out those phases?

An appropriate unit test for that would be to install a model that needs say 10 phases, advance 11 phases, install a different model requiring only 1 phase, and confirm that only 1 phase worth of data is present.

Unit test coverage of the case where fewer phases have run would also be good

PhilMiller (changes requested):
See my latest comment about covering the edge case

PhilMiller:
The reason I see the potential for retention is that the cleaning removes data from a particular past phase at a fixed offset from the current phase. if the model's requested retention falls, then there will be a range of phases whose data gets skipped over.

PhilMiller (approved):

PhilMiller (changes requested):
Looks like the unit tests fail. so something is missing in the cleanup code.

thearusable:
@PhilMiller I added the test cases which you mentioned but I wanted to ask you about the checks which should be in those cases. 
Currently I'm checking the amount of phases LB data in `theNodeLBData()` and also amount of data in `TestCol`. 
The first one is working fine, the data is removed as expected. 
The second check is failing, the amount of LB data in the `TestCol` is not changed. Should that be the case? Or I should look for some way to remove that data also, maybe by adding additional code in `theCollection()` for example?

What are the steps to change the Load Model? Currently I'm only doing the `theLBManager()->setLoadModel(model_1_phase);`. Is that an only required step or I missing something here?

PhilMiller:
Calling `setLoadModel` should be all that's necessary.

Is cleanup not happening at all for `TestCol`, or is it just a phase behind or something like that? If the cleanup would happen with some finite delay or some finite overage relative to `theNodeLBData()`, I'd be fine with loosening the test to reflect that.

It may be acceptable (for now, at least) if `TestCol` were to always retain at most (one more than) the maximum number of retained steps ever requested. The case we really want to avoid is unbounded data growth. As long as you can show that with the logic and tests, I'll support moving this functionality forward.

There's maybe an edge case of "extra data retained per call to `setLoadModel()`" if an application were to do something pathological, like alternately set models that wanted 10 and 20 steps of data retained repeatedly. Maybe we can accept something like that for now, with just an open issue to address it later.

That all said, if the data is/were retained in an ordered `map`, then perhaps instead of looping over keys we think need to be deleted, we could take a somewhat different approach to cleanup of (roughly)
```c++
for (auto i = data.begin(); *i < current_phase - retention; /* nothing */)
  i = data.erase(i);
```
That is, just delete anything the map may have that's older than what we're interested in.

thearusable:
> Calling `setLoadModel` should be all that's necessary.
> 
> Is cleanup not happening at all for `TestCol`, or is it just a phase behind or something like that? If the cleanup would happen with some finite delay or some finite overage relative to `theNodeLBData()`, I'd be fine with loosening the test to reflect that.
> 
> It may be acceptable (for now, at least) if `TestCol` were to always retain at most (one more than) the maximum number of retained steps ever requested. The case we really want to avoid is unbounded data growth. As long as you can show that with the logic and tests, I'll support moving this functionality forward.
> 
> There's maybe an edge case of "extra data retained per call to `setLoadModel()`" if an application were to do something pathological, like alternately set models that wanted 10 and 20 steps of data retained repeatedly. Maybe we can accept something like that for now, with just an open issue to address it later.
> 
> That all said, if the data is/were retained in an ordered `map`, then perhaps instead of looping over keys we think need to be deleted, we could take a somewhat different approach to cleanup of (roughly)
> 
> ```c++
> for (auto i = data.begin(); *i < current_phase - retention; /* nothing */)
>   i = data.erase(i);
> ```
> 
> That is, just delete anything the map may have that's older than what we're interested in.

Cleanup is not happening for `TestCol`, it will retain the maximum number of phases ever requested by the model.
Cache data in node is stored in `unordered_map` so I kept the previous implementation, I cleaned it a little bit and move it to `node_lb_data`.

tests/unit/collection/test_lb_data_retention.cc
```diff
@@ -104,6 +117,30 @@ struct TestCol : vt::Collection<TestCol,vt::Index1D> {
...
+
+  static void expect7PhasesOfData(MyMsg<TestCol>* msg, TestCol* col) {
+    auto& lb_data = col->lb_data_;
+    auto load_phase_count = lb_data.getLoadPhaseCount();
+    auto comm_phase_count = lb_data.getCommPhaseCount();
+    auto sp_load_phase_count = lb_data.getSubphaseLoadPhaseCount();
+    auto sp_comm_phase_count = lb_data.getSubphaseCommPhaseCount();
+
+    #if vt_check_enabled(lblite)
+      auto model = theLBManager()->getLoadModel();
```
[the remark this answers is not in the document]
thearusable: It's not needed. Removed.

src/vt/vrt/collection/balance/node_lb_data.cc
```diff
@@ -129,6 +129,21 @@ void NodeLBData::startIterCleanup(PhaseType phase, unsigned int look_back) {
...
 }
 
+void NodeLBData::trimLBDataHistory(PhaseType phase) {
+  if (phase != no_lb_phase && phase > min_hist_lb_data_) {
+    // Clean phases which are no longer needed
+    for(auto i = 0ull; i < phase - min_hist_lb_data_; i++) {
+      lb_data_->node_data_.erase(i);
+      lb_data_->node_comm_.erase(i);
+      lb_data_->node_subphase_comm_.erase(i);
+    }
```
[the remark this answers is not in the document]
thearusable: Good point! This function is now changed to use the range based `erase` variant.

src/vt/vrt/collection/balance/node_lb_data.cc
```diff
@@ -129,6 +129,26 @@ void NodeLBData::startIterCleanup(PhaseType phase, unsigned int look_back) {
...
+
+    // Trim all maps to have 'min_hist_lb_data_ + 1' phases of data
+    auto fromData = lb_data_->node_data_.find(phaseToLookFor);
+    lb_data_->node_data_.erase(fromData, lb_data_->node_data_.end());
+
+    auto fromComm = lb_data_->node_comm_.find(phaseToLookFor);
+    lb_data_->node_comm_.erase(fromComm, lb_data_->node_comm_.end());
+
+    auto fromSub = lb_data_->node_subphase_comm_.find(phaseToLookFor);
+    lb_data_->node_subphase_comm_.erase(fromSub, lb_data_->node_subphase_comm_.end());
```
nlslatt, line 144: Is there some guarantee that erasing from unordered containers using these iterators will erase phases numbered `<= phase - min_hist_lb_data_ - 1`?

PhilMiller: No, there isn't. That's why I suggested a switch to an ordered `map` 
https://github.com/DARMA-tasking/vt/pull/1996#issuecomment-1361716885

thearusable: Ok I missed that. Now it's changed to ordered `map`

src/vt/vrt/collection/balance/lb_data_holder.h
```diff
@@ -105,11 +105,11 @@ struct LBDataHolder {
...
-  std::unordered_map<PhaseType, LoadMapType> node_data_;
+  std::map<PhaseType, LoadMapType> node_data_;
   /// Node communication graph for each local object
-  std::unordered_map<PhaseType, CommMapType> node_comm_;
+  std::map<PhaseType, CommMapType> node_comm_;
   /// Node communication graph for each subphase
-  std::unordered_map<PhaseType, std::unordered_map<SubphaseType, CommMapType>> node_subphase_comm_;
+  std::map<PhaseType, std::unordered_map<SubphaseType, CommMapType>> node_subphase_comm_;
   /// User-defined data from each phase
   std::unordered_map<PhaseType, std::unordered_map<
```
PhilMiller, line 114: Do we need to be clearing this one as well? @nlslatt @ppebay @lifflander

src/vt/vrt/collection/balance/node_lb_data.cc
```diff
@@ -129,6 +129,26 @@ void NodeLBData::startIterCleanup(PhaseType phase, unsigned int look_back) {
   node_objgroup_lookup_.clear();
 }
 
+void NodeLBData::trimLBDataHistory(PhaseType phase) {
+  if (phase != no_lb_phase && phase > min_hist_lb_data_) {
+    auto phaseToLookFor = phase - min_hist_lb_data_;
+
+    lb_data_->node_data_.erase(
+      lb_data_->node_data_.begin(), lb_data_->node_data_.find(phaseToLookFor));
```
PhilMiller, line 137: I think this may be incorrect, since if `phaseToLookFor` specifically is not present, the `find` call will return the `end()` iterator, and the whole map will be erased. I think simply replacing `find` with `lower_bound` or `upper_bound` would be correct.

Could you write up a test plan of the different cases and how they're covered?

PhilMiller (changes requested):
See comment about `find` vs `lower_bound`

src/vt/vrt/collection/balance/node_lb_data.cc
thearusable: After the refactor there are really only two cases. 
1. When the maps to clean are empty or they have lower amount of elements than `min_hist_lb_data_ + 1` -> Then we will only clean the lookup tables with: 
`NodeLBData::node_migrate_.clear();`
`node_collection_lookup_.clear();`
`node_objgroup_lookup_.clear();`

2. When the maps have more elements than we need to retain -> We will remove the unwanted phases and also clean the lookup tables. 

Previous implementation had problems with missing phases in map. Let's say we have phases `1 2 3 4 5 6 7 8 10`(missing 9 for some reason) and our new model needs only 1 phase of data so we want to retain 2 phases.
With `find` we will clear the entire map because we will not find phase `9` and with `lower_bound` we will find `10` and remove everything before it so we will end up with only one phase `10`.
With current implementation we will retain last two phases so it will be `8` and `10`

src/vt/vrt/collection/balance/lb_data_holder.h
nlslatt: Yes.

src/vt/vrt/collection/balance/node_lb_data.cc
thearusable: @PhilMiller

src/vt/vrt/collection/balance/lb_data_holder.h
thearusable: @PhilMiller / @nlslatt I added clearing logic for `user_defined_json_`

thearusable:
@PhilMiller I updated the implementation to use `lower_bound`. This PR can be reviewed again.

src/vt/vrt/collection/balance/node_lb_data.cc
```diff
@@ -346,7 +366,7 @@ void NodeLBData::addNodeLBData(
   in->updatePhase(1);
 
   auto model = theLBManager()->getLoadModel();
```
nlslatt, line 368: Looks like we don't need the model here anymore.

tests/unit/collection/test_lb_data_retention.cc
```diff
@@ -218,6 +266,166 @@ TEST_F(TestLBDataRetention, test_lbdata_retention_last4) {
...
+
+  // Check amount of phase data in the node
+  checkNodeLBData(1 + 1);
+
+  // Check that amount of the retained data in TestCol is not changed and still contains 7 phases of data
+  for (uint32_t i=0; i<10; ++i) {
+    runInEpochCollective([&]{
+      // Do some work.
+      proxy.broadcastCollective<TestCol::expect7PhasesOfData>();
+    });
```
nlslatt, line 422: Why would we want the collection to retain obsolete data like the node used to? Shouldn't we make the same type of change there as well? @lifflander @PhilMiller

nlslatt (commented):
Do you have any tests of setting `vt_lb_data_retention` lower than what the model needs? Yes, it will be ignored as your comments say, but I don't see where you've tested that it actually does get ignored.

Do you have any tests with `vt_lb_data_retention` is higher than what the model needs that run more phases than the value of `vt_lb_data_retention`, to make sure that the earliest phases still get deleted?

Is there a reason why the same type of changes here are not being applied to the history stored with each element?

src/vt/vrt/collection/balance/node_lb_data.cc
thearusable: Oh yes, I forgot to remove it.

thearusable:
@nlslatt I updated the tests to better reflect how many and which phases are being kept. 

Test `test_lbdata_retention_last2` - has `vt_lb_data_retention` lower than what model needs. 
Test `test_lbdata_config_retention_higher` - has `vt_lb_data_retention` higher than what model needs.

thearusable:
@nlslatt, @lifflander - I rebased this PR and fixed all conflicts. PR is ready for review.

src/vt/vrt/collection/balance/lb_data_holder.h
```diff
@@ -125,17 +125,17 @@ struct LBDataHolder {
...
 public:
   /// Node timings for each local object
-  std::unordered_map<PhaseType, LoadMapType> node_data_;
+  std::map<PhaseType, LoadMapType> node_data_;
   /// Node communication graph for each local object
-  std::unordered_map<PhaseType, CommMapType> node_comm_;
+  std::map<PhaseType, CommMapType> node_comm_;
   /// Node communication graph for each subphase
-  std::unordered_map<PhaseType, std::unordered_map<SubphaseType, CommMapType>> node_subphase_comm_;
-  /// User-defined data from each phase for JSON output
```
nlslatt, line 133: Please restore this full comment that was truncated.

tests/unit/collection/test_lb_data_retention.cc
nlslatt: @lifflander I'm not clear on whether Phil's comment means that we should correct the inconsistency or not.

src/vt/vrt/collection/balance/lb_data_holder.h
thearusable: Fixed

tests/unit/collection/test_lb_data_retention.cc
```diff
@@ -58,6 +58,29 @@
 
 namespace vt { namespace tests { namespace unit {
 
+void validatePersistedPhases(std::vector<PhaseType> expected_phases) {
+  #if vt_check_enabled(lblite)
+    // Check maps size
+    EXPECT_EQ(expected_phases.size(), theNodeLBData()->getLBData()->node_comm_.size());
+    EXPECT_EQ(expected_phases.size(), theNodeLBData()->getLBData()->node_data_.size());
+    EXPECT_EQ(expected_phases.size(), theNodeLBData()->getLBData()->node_subphase_comm_.size());
+    EXPECT_EQ(expected_phases.size(), theNodeLBData()->getLBData()->user_defined_json_.size());
```
nlslatt, line 67: `user_defined_lb_info_` should be added to the checks in several places.

nlslatt (dismissed):
Almost there.

src/vt/vrt/collection/balance/node_lb_data.cc
```diff
@@ -135,6 +137,25 @@ void NodeLBData::startIterCleanup(PhaseType phase, unsigned int look_back) {
   node_objgroup_lookup_.clear();
 }
 
+void NodeLBData::trimLBDataHistory() {
+  auto trim_data = [this](auto& map){
+    if(map.size() > min_hist_lb_data_ + 1) {
```
nlslatt, line 142: What's the reason for the +1 here? I think either the +1 should be removed or the comparison should be changed to `>=`.

nlslatt: It looks like you made both changes instead of just one, which I think will make the cleanup code sometimes trigger unnecessarily (performance issue rather than correctness).

thearusable: Ah yes, I applied both instead of one of the solutions. Now it's changed to `>`.

src/vt/elm/elm_lb_data.h
```diff
@@ -115,24 +116,23 @@ struct ElementLBData {
...
-  friend struct vrt::collection::balance::NodeLBData;
+  void setHistoryCapacity(unsigned int hist_lb_data_count);
 
 protected:
   bool cur_time_started_ = false;
   TimeType cur_time_ = TimeType{0.0};
   PhaseType cur_phase_ = fst_lb_phase;
-  std::unordered_map<PhaseType, LoadType> phase_timings_ = {};
-  std::unordered_map<PhaseType, CommMapType> phase_comm_ = {};
+  util::container::CircularPhasesBuffer<LoadType> phase_timings_ = {1};
```
thearusable, line 130: I've set a default size of the buffer to be 1. There is a call to `stop()` and `addTime()` which will save the timings in `phase_timings_`. Before adding new data in  all buffers will be resized.

As an alternative to that I could add a check to see if the buffers are initialized so then that first timings data would be ignored. 
Second option is to add another type of the buffer which will be able to grow it's size up to a first call to resize which would freeze it's size.

src/vt/utils/container/circular_phases_buffer.h
```diff
@@ -0,0 +1,411 @@
...
+    if (new_size == vector_.size()) {
+      return;
+    }
+
+    auto new_vec =
+      std::vector<StoredPair>(new_size, StoredPair{invalid_, StoredType{}});
+
+    if (new_size > 0) {
+      if (new_size < size()) {
+        auto tmp_tail = head_ - new_size + 1;
```
JacobDomagala, line 198: nvcc is right, we should explicitly make it signed int.

thearusable: Thanks for noticing it, I missed the warning from nvcc.

src/vt/elm/elm_lb_data.h
thearusable: This is no longer the case. Now it will grow dynamically until we set a specific buffer size

src/vt/utils/container/circular_phases_buffer.h
```diff
@@ -0,0 +1,280 @@
...
+  void reset() {
+    m_requested_capacity = 0;
+    clear();
+  }
+
+  template <typename SerializeT>
+  void serialize(SerializeT& s) {
+    s | m_requested_capacity;
+    // TODO: Checkpoint is not supporting the queue
+    // s | m_indexes;
```
thearusable, line 232: I'll check in `checkpoint` why this line results in compilation errors.

tests/unit/collection/test_lb_data_retention.cc
```diff
@@ -58,35 +58,64 @@
...
+        EXPECT_EQ(sp_load_phase_count, buffers_size);
+        EXPECT_EQ(comm_phase_count,    buffers_size);
+        EXPECT_EQ(sp_comm_phase_count, buffers_size);
       } else if (phase == 0) {
-        EXPECT_EQ(load_phase_count,    phase);
-        EXPECT_EQ(sp_load_phase_count, phase);
-        EXPECT_EQ(comm_phase_count,    phase);
-        EXPECT_EQ(sp_comm_phase_count, phase);
+        EXPECT_EQ(load_phase_count,    1);
+        EXPECT_EQ(sp_load_phase_count, 1);
```
thearusable, line 116: We will have initial timings added to `ElementLBData` because in the tests I used `proxy.broadcastCollective<TestCol::insertValue>()` to properly setup `TestCol` to populate members like `node_user_attributes` in `LBDataHolder`.

src/vt/utils/container/circular_phases_buffer.h
```diff
@@ -0,0 +1,300 @@
...
+  void removeOldest() {
+    m_buffer.erase(m_indexes.front());
+    m_indexes.pop();
+  }
+
+private:
+  std::size_t m_requested_capacity;
+  std::queue<PhaseType> m_indexes;
+  std::unordered_map<PhaseType, StoredType> m_buffer;
+};
```
PhilMiller, line 296: A stylistic thing I noticed - the rest of the vt codebase uses a trailing underscore to indicate member variables, rather than an `m_` prefix. This should probably be consistent with that?

src/vt/utils/container/circular_phases_buffer.h
```diff
@@ -0,0 +1,300 @@
...
+   * @brief Add new phase to the cache and remove oldest one if buffer exceeds requested size
+   *
+   * @param pair the pair<PhaseType, StoredType> to be stored
+   */
+  void addToCache(StoredPair&& pair) {
+    if (m_requested_capacity > 0 && m_buffer.size() >= m_requested_capacity) {
+      removeOldest();
+    }
+    m_indexes.push(pair.first);
+    m_buffer[pair.first] = std::move(pair.second);
```
PhilMiller, line 281: I think it's slightly preferable to use `insert` or `emplace` in this case, since the key `pair.first` should not exist, and it shouldn't create the node empty and then move into it.

src/vt/utils/container/circular_phases_buffer.h
```diff
@@ -0,0 +1,300 @@
...
+
+  template <typename SerializeT>
+  void serialize(SerializeT& s) {
+    s | m_requested_capacity;
+    s | m_indexes;
+    s | m_buffer;
+  }
+
+  /**
+   * @brief Get iterator to the begging of the buffer
```
PhilMiller, line 229: Typo, repeated a few times: `begging` -> `beginning`

src/vt/utils/container/circular_phases_buffer.h
```diff
@@ -0,0 +1,300 @@
...
+   */
+  auto cend() const { return m_buffer.cend(); }
+
+private:
+  /**
+   * @brief Add new phase to the cache and remove oldest one if buffer exceeds requested size
+   *
+   * @param pair the pair<PhaseType, StoredType> to be stored
+   */
+  void addToCache(StoredPair&& pair) {
```
PhilMiller, line 276: It's a little weird that this method takes a `pair`, when all of the all sites call it with a `make_pair` argument, and the implementation pulls the pair apart. Why not just take the two elements as separate parameters?

src/vt/utils/container/circular_phases_buffer.h
thearusable: Ah, yes, fixed.

src/vt/utils/container/circular_phases_buffer.h
thearusable: Good point! I fixed that.

src/vt/utils/container/circular_phases_buffer.h
thearusable: Fixed.

src/vt/utils/container/circular_phases_buffer.h
thearusable: Ok, I fixed that, now the function takes phase and data as separate parameters.

src/vt/utils/container/circular_phases_buffer.h
```diff
@@ -0,0 +1,301 @@
...
+   * \param[in] phase the phase for which data will be stored
+   * \param[in] data the data to store
+   */
+  void addToCache(const PhaseType phase, StoredType const& data) {
+    if (requested_capacity_ > 0 && buffer_.size() >= requested_capacity_) {
+      removeOldest();
+    }
+    indexes_.push(phase);
+    buffer_.emplace(phase, data);
+  }
```
PhilMiller, line 283: Based on the overloads of `store`, it seems like we want to avoid copying `data` when possible. That seems smart if it could be a container or something else expensive to copy. However, by taking a `const&` and passing it to `emplace` as such, there's still guaranteed to be a copy into the `unordered_map` entry.

To resolve that, `addToCache` would need to take either `StoredType data` or `StoredType&& data` and `std::move` it into the `emplace` call. Similar reasoning then needs to be worked up the call stack, to make sure that avoidable copies aren't forced at any level.

src/vt/utils/container/circular_phases_buffer.h
thearusable: Done.

src/vt/utils/container/circular_phases_buffer.h
thearusable: Thanks for catching that! It's fixed now.

src/vt/utils/container/circular_phases_buffer.h
```diff
@@ -0,0 +1,301 @@
...
+  auto cend() const { return buffer_.cend(); }
+
+private:
+  /**
+   * @brief Add new phase to the cache and remove oldest one if buffer exceeds requested size
+   *
+   * \param[in] phase the phase for which data will be stored
+   * \param[in] data the data to store
+   */
+  void addToCache(const PhaseType& phase, const StoredType&& data) {
```
PhilMiller, line 277: `const StoredData&&` is no more helpful in avoiding copies than a const L-value reference. The callee needs to be able to take ownership of any contents of `data` away from the passed argument, and that requires it to be non-`const`

thearusable: Right, thanks for catching this! I removed it. I think there are no more unnecessary copies being done.

src/vt/utils/container/circular_phases_buffer.h
```diff
@@ -0,0 +1,301 @@
...
+   */
+  void removeOldest() {
+    buffer_.erase(indexes_.front());
+    indexes_.pop();
+  }
+
+private:
+  std::size_t requested_capacity_;
+  std::queue<PhaseType> indexes_;
+  std::unordered_map<PhaseType, StoredType> buffer_;
```
thearusable, line 296: @lifflander, @nlslatt - Should I change the implementation to use a `vector` or can it stay as a `unordered_map` + `queue`?

With this implementation, we have a quick `resize()` method which only needs to remove excess elements during the resizing. After changing to a `vector`, there will be cases when the entire cache would need to be copied over(or elements would need to be rotated). Currently, the resizing is only being used in `NodeLBData` when changing the model and in `ElementLBData` before running the first phase.

I'm not sure how frequently the method will be used in the future; for example, would there be a use case that would require a frequent model change?

thearusable:
Things to do:
- [ ] Prepare benchmark for existing implementation
- [ ] Prepare different implementation based on `std::vector` and prepare benchmark for it.

src/vt/utils/container/circular_phases_buffer.h
```diff
@@ -0,0 +1,391 @@
...
+   */
+  auto end() const { return ForwardIterator(); }
+
+  /**
+   * \brief Finds the newest phase in the buffer.
+   *
+   * \return the newest phase in the buffer.
+   */
+  PhaseType frontPhase() const {
+    return inserted_ == 0 ? 0 : head_phase_ - 1;
```
nlslatt, line 259: This returns the same value whether phase 0 exists in the buffer or not, right? I believe this will cause the loops you converted (in replay, for example) to malfunction when the buffer is empty.

src/vt/vrt/collection/balance/workload_replay.cc
```diff
@@ -249,19 +249,19 @@ readInWorkloads(const std::string &filename) {
   auto json = r.readFile();
   auto sd = std::make_shared<LBDataHolder>(*json);
 
-  for (auto &phase_data : sd->node_data_) {
+  for (auto phase = sd->node_data_.backPhase(); phase <= sd->node_data_.frontPhase(); phase++) {
```
nlslatt, line 252: This loop, and ones like it, will now have one iteration whether the buffer is empty or has one phase, right? That is a problem.

src/vt/utils/container/circular_phases_buffer.h
```diff
@@ -0,0 +1,391 @@
...
+
+  /**
+   * \brief Finds the oldest phase in the buffer.
+   *
+   * \return the oldest phase in the buffer.
+   */
+  PhaseType backPhase() const {
+    if (head_ == inserted_) {
+      return 0;
+    }
```
nlslatt, line 279: Are you assuming here that the first phase must always be zero? That might not be a safe assumption.

nlslatt (commented):
I have not reviewed the whole PR, but I have concerns about the CircularPhasesBuffer implementation. In addition to addressing the specific concerns flagged, please add code comments to indicate assumptions being made within the member functions so that we can figure out a performant way to do sufficient sanity checks.

src/vt/utils/container/circular_phases_buffer.h
```diff
@@ -0,0 +1,391 @@
...
+  using StoredPair = std::pair<PhaseType, StoredType>;
+
+  static const auto no_pos = std::numeric_limits<std::size_t>::max();
+
+public:
+
+  /**
+   * \brief Construct a CircularPhasesBuffer
+   *
+   * \param[in] in_capacity the requested size of the buffer
```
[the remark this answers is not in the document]
thearusable: Fixed.

src/vt/utils/container/circular_phases_buffer.h
```diff
@@ -0,0 +1,391 @@
...
+  std::size_t findTail() const {
+    if (inserted_ == 0) {
+      return no_pos;
+    } else if (head_ == inserted_) {
+      return 0;
+    }
+    return head_;
+  }
+
+  std::size_t phaseToIndex(const PhaseType phase) const {
```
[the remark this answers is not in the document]
thearusable: I added a check to validate if the phase we want to add is bigger than `head_phase_ - buffer_.size()`.

src/vt/utils/container/circular_phases_buffer.h
```diff
@@ -0,0 +1,391 @@
...
+    return nullptr;
+  }
+
+  /**
+   * \brief Store data in the buffer.
+   *
+   * \param[in] phase the phase for which data will be stored
+   * \param[in] obj the data to store
+   */
+  void store(const PhaseType& phase, StoredType data) {
```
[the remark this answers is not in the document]
thearusable: After the update, this method will check if the phase is within the valid range, meaning it is bigger than `head_phase_ - buffer_.size()`.

src/vt/utils/container/circular_phases_buffer.h
```diff
@@ -0,0 +1,391 @@
...
+    inserted_ = inserted_ + 1;
+  }
+
+  /**
+   * \brief Resize buffer to the requested size.
+   *
+   * \param[in] new_size the requested new size of the buffer
+   */
+  void resize(const std::size_t new_size) {
+    if (new_size == buffer_.size() || new_size == 0) {
```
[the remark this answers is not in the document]
thearusable: Good point, it's fixed now. There is no option now of setting the size of the buffer to 0.

tests/unit/utils/test_circular_phases_buffer.nompi.cc
```diff
@@ -0,0 +1,246 @@
...
+    buffer[i] = {i};
+  }
+  validatePresentPhases(buffer, {28, 29, 30, 31, 32, 33, 34, 35});
+
+  buffer.resize(1);
+  validatePresentPhases(buffer, {35});
+  validateMissingPhases(buffer, {28, 29, 30, 31, 32, 33, 34});
+
+  buffer.resize(0);
+  validatePresentPhases(buffer, {35});
```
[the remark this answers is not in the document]
thearusable: Fixed.

src/vt/utils/container/circular_phases_buffer.h
thearusable: Good point, now the `head_phase_ ` starts from the maximum value for `PhaseType`. We will get that value when `frontPhase()` is called on an empty buffer.

src/vt/vrt/collection/balance/workload_replay.cc
thearusable: I updated the behavior, although when creating the iterator to the beginning of the buffer, I need to call advance at least once to find the first element.

src/vt/utils/container/circular_phases_buffer.h
thearusable: I changed the implementation to use a modulo operation, so now empty pairs will have a phase value equal to the maximum value for `PhaseType`. It is checked when accessing the element if the phase value is equal to the phase value stored in the buffer - it will not allow access to some invalid element when used with the maximum value of `PhaseType`.

tests/unit/collection/test_lb_data_retention.cc
```diff
@@ -218,6 +467,47 @@ TEST_F(TestLBDataRetention, test_lbdata_retention_last4) {
...
+  // Check the phases persisted in the node
+  validatePersistedPhases({4,5,6,7});
+
+  for (int i=0; i<num_phases; ++i) {
+    runInEpochCollective([&]{
+      // Do some work.
+      proxy_new.broadcastCollective<TestCol::colHandler>();
+    });
+    // Go to the next phase.
+    vt::thePhase()->nextPhaseCollective();
```
thearusable, line 506: When we reset the current phase to 0 in `restoreFromFileInPlace`, then this line will cause vt to abort due to having different phases in LBData and in the stats message: https://github.com/DARMA-tasking/vt/blob/5538883df647dc7ed93f19f751d425a5d03b24a0/src/vt/vrt/collection/balance/col_lb_data.impl.h#L73

@lifflander / @nlslatt - Do we need to reset the current phase to 0 in `restoreFromFileInPlace`?

thearusable: Based on my recent testing, it appears that the `resetPhase` method is essential for the proper functioning of VT.
My previous tests were successful because I did not fully recreate VT, which resulted in the preservation of `PhaseManager`, `NodeLBData`, and other VT modules.
When we create a new fresh instance and restore the collection from the checkpoint, every module will commence execution from phase 0.

Upon restoring the collection from the checkpoint, we will retain only one phase of history, even if multiple phases are serialized. Upon restoration, the current phase is reset to 0, rendering other phases inaccessible.

PR #1998, opened 2022-10-18
1938 remove unneeded calls to mpi wtime in trace and lb

### LINKED ISSUE #1938: Remove unneeded calls to MPI_Wtime in Trace and LB
**What Needs to be Done?**
As a follow-on to #1901, pass the current time from the scheduler down to these components.

stmcgovern:
Closes #1938.
~~Pass the current time from the scheduler to LB and Trace contexts.~~ 
This has been modified to allow the scheduler to hold a `current_sched_time_` and then the appropriate runnable contexts (here LB and Trace) can query their start time from the scheduler. This change was primarily the result of not wanting to force all runnable contexts to take a time (even in contexts where that is inappropriate). Still targets the issue by reducing calls to MPI_Wtime (although increased dereferencing of the `theSched()`)

### DIFF
--- a/src/vt/parameterization/parameterization.h
+++ b/src/vt/parameterization/parameterization.h
@@ -107,7 +107,7 @@ static void dataMessageHandler(DataMsg<Tuple>* msg) {
     NodeType const& from_node = theContext()->getFromNodeCurrentTask();
 
     processing_tag =
-      theTrace()->beginProcessing(ep, msg_size, event, from_node);
+      theTrace()->beginProcessing(ep, msg_size, event, from_node, timing::getCurrentTime());
   }
 #endif
 
--- a/src/vt/parameterization/parameterization.h
+++ b/src/vt/parameterization/parameterization.h
@@ -121,7 +121,7 @@ static void dataMessageHandler(DataMsg<Tuple>* msg) {
   }
 
 #if vt_check_enabled(trace_enabled)
-  theTrace()->endProcessing(processing_tag);
+  theTrace()->endProcessing(processing_tag, timing::getCurrentTime());
 #endif
 }
 
--- a/src/vt/runnable/invoke.h
+++ b/src/vt/runnable/invoke.h
@@ -140,7 +140,7 @@ static trace::TraceProcessingTag BeginProcessingInvokeEvent() {
   const auto trace_event = theTrace()->messageCreation(trace_id, 0);
   const auto from_node = theContext()->getNode();
 
-  return theTrace()->beginProcessing(trace_id, 0, trace_event, from_node);
+  return theTrace()->beginProcessing(trace_id, 0, trace_event, from_node, timing::getCurrentTime());
 }
 
 template <typename Callable, Callable f, typename... Args>
--- a/src/vt/runnable/invoke.h
+++ b/src/vt/runnable/invoke.h
@@ -145,7 +145,7 @@ static trace::TraceProcessingTag BeginProcessingInvokeEvent() {
 
 template <typename Callable, Callable f, typename... Args>
 static void EndProcessingInvokeEvent(trace::TraceProcessingTag processing_tag) {
-  theTrace()->endProcessing(processing_tag);
+  theTrace()->endProcessing(processing_tag, timing::getCurrentTime());
 
   const auto trace_id = CallableWrapper<Callable, f>::GetTraceID();
   theTrace()->messageCreation(trace_id, 0);
--- a/src/vt/runnable/runnable.cc
+++ b/src/vt/runnable/runnable.cc
@@ -185,49 +185,49 @@ void RunnableNew::run() {
 #endif
 }
 
-void RunnableNew::start() {
+void RunnableNew::start(TimeType time) {
   contexts_.setcontext.start();
   if (contexts_.has_td) contexts_.td.start();
   if (contexts_.has_col) contexts_.col.start();
-  if (contexts_.has_lb) contexts_.lb.start();
+  if (contexts_.has_lb) contexts_.lb.start(time);
 #if vt_check_enabled(trace_enabled)
-  if (contexts_.has_trace) contexts_.trace.start();
+  if (contexts_.has_trace) contexts_.trace.start(time);
 #endif
 }
 
-void RunnableNew::finish() {
+void RunnableNew::finish(TimeType time) {
   contexts_.setcontext.finish();
   if (contexts_.has_td) contexts_.td.finish();
   if (contexts_.has_col) contexts_.col.finish();
   if (contexts_.has_cont) contexts_.cont.finish();
-  if (contexts_.has_lb) contexts_.lb.finish();
+  if (contexts_.has_lb) contexts_.lb.finish(time);
 #if vt_check_enabled(trace_enabled)
-  if (contexts_.has_trace) contexts_.trace.finish();
+  if (contexts_.has_trace) contexts_.trace.finish(time);
 #endif
 }
 
-void RunnableNew::suspend() {
+void RunnableNew::suspend(TimeType time) {
 #if vt_check_enabled(fcontext)
   contexts_.setcontext.suspend();
   if (contexts_.has_td) contexts_.td.suspend();
   if (contexts_.has_col) contexts_.col.suspend();
-  if (contexts_.has_lb) contexts_.lb.suspend();
+  if (contexts_.has_lb) contexts_.lb.suspend(time);
 
 # if vt_check_enabled(trace_enabled)
-    if (contexts_.has_trace) contexts_.trace.suspend();
+    if (contexts_.has_trace) contexts_.trace.suspend(time);
 # endif
 #endif
 }
 
-void RunnableNew::resume() {
+void RunnableNew::resume(TimeType time) {
 #if vt_check_enabled(fcontext)
   contexts_.setcontext.resume();
   if (contexts_.has_td) contexts_.td.resume();
   if (contexts_.has_col) contexts_.col.resume();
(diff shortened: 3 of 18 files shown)


src/vt/event/event.h
```diff
@@ -177,7 +177,7 @@ struct AsyncEvent : runtime::component::PollableComponent<AsyncEvent> {
   void removeEventID(EventType const& event);
   EventStateType testEventComplete(EventType const& event);
   EventType attachAction(EventType const& event, ActionType callable);
-  void testEventsTrigger(int const& num_events = num_check_actions);
+  void testEventsTrigger(int const& num_events = num_check_actions, TimeType current_time=0.0);
```
PhilMiller, line 180: I'm a bit uncomfortable with an internal interface like this having a default argument. Are there call sites that are being left unmodified, for which the default value would be is sensible to pass?

src/vt/runnable/runnable.cc
```diff
@@ -212,46 +212,46 @@ void RunnableNew::run() {
 #endif
 }
 
-void RunnableNew::begin() {
+void RunnableNew::begin(TimeType current_time) {
```
PhilMiller, line 215: Not relevant to this PR, but I realize that we should probably not have methods named `begin/end` that are unrelated to getting the relevant iterators of a container-like class. Perhaps we should consider renaming them to something like `start/finish`

src/vt/runnable/runnable.h
```diff
@@ -236,25 +236,25 @@ struct RunnableNew {
    * \internal \brief Loop through all the contexts associated with this
    * runnable and invoke \c begin() on them.
    */
-  void begin();
+  void begin(TimeType current_time = 0.0);
```
PhilMiller, line 239: Same comment about default values

src/vt/vrt/collection/balance/baselb/baselb.cc
```diff
@@ -66,19 +66,20 @@ std::shared_ptr<const balance::Reassignment> BaseLB::startLB(
   balance::LoadModel* model,
   StatisticMapType const& in_stats,
   ElementCommType const& in_comm_lb_data,
-  TimeType total_load
+  TimeType total_load,
+  TimeType current_time
 ) {
-  start_time_ = timing::getCurrentTime();
+  start_time_ = current_time;
```
PhilMiller, line 72: This assignment makes sense, but I don't see why it's being passed down the lb stack any further. There are no uses of it, and the strategies can all access the member assigned here anyway.

src/vt/vrt/collection/manager.cc
```diff
@@ -69,7 +69,8 @@ void CollectionManager::startup() {
       elm.second();
     }
     auto const cur_phase = thePhase()->getCurrentPhase();
-    theLBManager()->selectStartLB(cur_phase);
+    auto  current_time = thePhase()->getStartTime();
```
PhilMiller, line 72: Is the beginning of the phase the right time to be passing in as the start of the LB? Is this getting called just after the phase manager ticks forward? Since it's in an End hook, isn't that not the case?

PhilMiller:
I would approve the first 3 commits of this PR as resolving 1938 (modulo my comments about default parameters). The latter two commits seem to be less motivated, and less clearly correct.

stmcgovern:
Thanks for your comments @PhilMiller. I'll drop the last 2 commits and see about the default arguments.

lifflander (dismissed):
Looks good!

PhilMiller (changes requested):
Frankly, I was much happier with the approach of just passing a time down the call stack.

This PR now adds a bunch of calls to `timing::getCurrentTime`, though I haven't checked whether they're in specifically cold paths.

This change in approach may have been discussed, but that discussion is supposed to be memorialized in comments here. I'll note that the PR description no longer quite matches the code.

Let's talk about this in the meeting this week.

src/vt/scheduler/scheduler.cc
```diff
@@ -258,7 +258,8 @@ void Scheduler::runSchedulerOnceImpl(bool msg_only) {
     }
   } else if (work_queue_.empty()) {
     if (curRT->needsCurrentTime()) {
-      runProgress(msg_only, timing::getCurrentTime());
+      current_sched_time_ = timing::getCurrentTime();
+      runProgress(msg_only, current_sched_time_);
```
PhilMiller, line 262: This may not be worth changing, but since `runProgress` is a sibling member, it doesn't need to take `current_sched_time_` as an argument, since it can access it directly.

stmcgovern:
> Frankly, I was much happier with the approach of just passing a time down the call stack.
> 
> This PR now adds a bunch of calls to `timing::getCurrentTime`, though I haven't checked whether they're in specifically cold paths.
> 
> This change in approach may have been discussed, but that discussion is supposed to be memorialized in comments here. I'll note that the PR description no longer quite matches the code.
> 
> Let's talk about this in the meeting this week.

Thanks @PhilMiller. Yes, the idea was to just pass time down the call stack. After removing the default arguments (per your first review), then a time argument needed to be passed through to any runnable (`void RunnableNew::run()`. Aside from complicating the `RunnableNew` interface, @lifflander remarked that not all runnables need a time. Then the change of tactic to have the scheduler store a time and allow the appropriate runnable contexts (in particular LB and trace) to ask for a start time was adopted. 

`theTrace()->beginProcessing` called `timing::getCurrentTime` by default. Removing this implicit use required getting the current time explicitly at certain call sites. 

I'll update the PR description to reflect these changes.

src/vt/runnable/invoke.h
```diff
@@ -140,7 +140,7 @@ static trace::TraceProcessingTag BeginProcessingInvokeEvent() {
   const auto trace_event = theTrace()->messageCreation(trace_id, 0);
   const auto from_node = theContext()->getNode();
 
-  return theTrace()->beginProcessing(trace_id, 0, trace_event, from_node);
+  return theTrace()->beginProcessing(trace_id, 0, trace_event, from_node, timing::getCurrentTime());
```
PhilMiller, line 143: This may actually illustrate a case for why I like the 'pass the time as an argument to contexts' approach better - if we want to add LB attribution for methods called through `invoke`, we would want to share this timer call for the start time, rather than have it reference the scheduler's recorded time, or need another new interface for the divergent case.

stmcgovern:
Status Update: PR uses a recent time held by the scheduler which is passed down to runnables that need a time. Currently only LB context uses this (Trace is TODO). Consider how to test for basic correctness confirmation.

src/vt/parameterization/parameterization.h
```diff
@@ -107,7 +107,7 @@ static void dataMessageHandler(DataMsg<Tuple>* msg) {
     NodeType const& from_node = theContext()->getFromNodeCurrentTask();
 
     processing_tag =
-      theTrace()->beginProcessing(ep, msg_size, event, from_node);
+      theTrace()->beginProcessing(ep, msg_size, event, from_node, timing::getCurrentTime());
```
PhilMiller, line 110: @lifflander Should we be create a `Runnable` here?

src/vt/scheduler/scheduler.cc
```diff
@@ -239,29 +239,30 @@ void Scheduler::runProgress(bool msg_only, TimeType current_time) {
...
   } else if (work_queue_.empty()) {
     if (curRT->needsCurrentTime()) {
-      runProgress(msg_only, timing::getCurrentTime());
+      runProgress(msg_only, getRecentTime());
     } else {
       runProgress(msg_only, TimeType{0});
     }
   }
 
+
```
PhilMiller, line 265: Stray new line.
```suggestion
```

src/vt/scheduler/base_unit.cc
```diff
@@ -62,6 +62,7 @@ void BaseUnit::execute() {
     #endif
   } else if (work_) {
     work_();
+    theSched()->setRecentTimeToStale();
```
PhilMiller, line 65: Should this be outside the `if / else if`, or duplicated in both arms? Why shouldn't the time be updated after we've run stuff in a thread?

src/vt/parameterization/parameterization.h
PhilMiller: Code's going away, so we don't care.

PhilMiller:
I think I'm generally ok with this, and just need to see the couple comments addressed.

This does need to be run through the same fine-grain overhead benchmarks we used to identify time calls as a concern to begin with. I'd like to be confident that added calls to `theSched()` aren't losing us whatever ground we're gaining, or even a large fraction of it. If they are, then I'd push for accepting that every context is going to take a time argument, useful or not, and shifting the logic for when to check the clock and what time to pass down the stack more thoroughly into the scheduler.

src/vt/scheduler/scheduler.cc
stmcgovern: fixed

src/vt/scheduler/base_unit.cc
stmcgovern: Yeah, it's in the run function (runnable.cc line 183). Should we just pull it out to here (outside the if/else)?

PhilMiller:
Some runs of the microbenchmark were getting ~30% reduction in time, directly attributable to reduced time spent in `PMPI_Wtime`. So, I think we're good on achieving the goal here.

lifflander (approved):

src/vt/scheduler/base_unit.cc
stmcgovern: Left as is, since the runnable branch will also need to get a finish time.

stmcgovern:
Improvements to the tests will be pursued in VT #2017

PR #2014, opened 2022-11-15
#1906: Add parameter to control which version of C++ standard is used

### LINKED ISSUE #1906: Enable C++17 in some CI builds
**What Needs to be Done?**
- [x] make it possible to select C++ standard version when building `vt`
- [x] make sure that C++14 remains the default
- [x] enable C++17 in at least one `clang` and one `gcc` build

thearusable:
Closes #1906


CMakeLists.txt
```diff
@@ -36,6 +36,14 @@ if (ccache_binary)
 endif()
 
 set(CMAKE_CXX_EXTENSIONS OFF)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+
+if(vt_cxx_standard)
```
PhilMiller, line 41: Please wrap this whole condition in a check whether CMAKE_CXX_STANDARD is already set. We should not override an external general setting, such as from an application embedding vt.

thearusable: Ok, fixed

CMakeLists.txt
```diff
@@ -36,6 +36,12 @@ if (ccache_binary)
 endif()
 
 set(CMAKE_CXX_EXTENSIONS OFF)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+
+if(NOT CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 14 CACHE STRING "The C++ standard to use")
+endif()
+message(STATUS "CXX_STANDARD: ${CMAKE_CXX_STANDARD}")
```
PhilMiller, line 44: When printing this, please include the `CMAKE_` prefix. CMake also understands a variable banked without it, but it is the target specific setting for which the prefixed variable gives the default

thearusable: Fixed.

cz4rs:
@thearusable Notice the error in `icpx` build:
```
/vt/src/vt/vrt/collection/balance/lb_common.cc:154:15: error: variable 'i' set but not used [-Werror,-Wunused-but-set-variable]
  std::size_t i = 0;
```
This seems overzealous, as `i` is later used inside a loop (I guess what compiler complains about is the case when loop is never executed). I'd suggest running this build with docker locally with `-Werror` disabled to see how many warnings show up. Last resort solution would be to disable `-Werror` for `icpx` build in CI.


Overall this seems to be caused by an `icpx` update. Compare last build on `develop`:
```
-- The CXX compiler identification is IntelLLVM 2022.1.0
```
and this PR:
```
-- The CXX compiler identification is IntelLLVM 2022.2.1
```
Some additional explanation for this: in CI, we usually just pull our docker images and build vt inside them. Only upon detecting changes in Dockerfiles (as happens in this PR) the containers will be rebuilt (instead of pulled). Since we don't use fixed version of icpx, this rebuild resulted in a newer version being used.

PhilMiller (approved):

cz4rs (approved):
Looks good apart from the `icpx` issue.

JacobDomagala (approved):

cz4rs (approved):
👍

lifflander (approved):

PR #2020, opened 2022-11-21
#2019: add move constructor/assign to TraceScopedEvent and TraceScopedNote

### LINKED ISSUE #2019: TraceScopedEvent and TraceScopedNote should have deleted copy constructors and defined move constructors/assign
Currently, they only have a deleted copy assignment operator.

nmm0:
Fixes #2019

### DIFF
--- a/src/vt/trace/trace_user.h
+++ b/src/vt/trace/trace_user.h
@@ -218,7 +218,20 @@ struct TraceScopedEvent final {
       event_(event)
   { }
 
+  TraceScopedEvent(TraceScopedEvent const&) = delete;
+  TraceScopedEvent(TraceScopedEvent &&other) noexcept
+  {
+    std::swap(begin_, other.begin_);
+    std::swap(event_, other.event_);
+  }
+
   TraceScopedEvent& operator=(TraceScopedEvent const&) = delete;
+  TraceScopedEvent& operator=(TraceScopedEvent &&other) noexcept
+  {
+    std::swap(begin_, other.begin_);
+    std::swap(event_, other.event_);
+    return *this;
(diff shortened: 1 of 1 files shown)


PhilMiller (commented):
I know calling swap is idiomatic, but I'd be more comfortable if the moved from objects had their event fields assigned the no event value explicitly, moreso in the move assignment case.

nmm0:
> I know calling swap is idiomatic, but I'd be more comfortable if the moved from objects had their event fields assigned the no event value explicitly, moreso in the move assignment case.

For the move constructor it is set explicitly then swapped (from the initializer on event_), but for move assignment it has to use a swap to be correct if not the old value (if it is not `no_trace_event`) will never be released

PhilMiller:
I feel like if that case ever arises, it represents a logic bug, which we should maybe assert against.

What's the actual use case for allowing any sort of move construction or assignment of these types? I figured it was for something like moving into a closure, where there's no existing event to be dealt with.

nmm0:
> I feel like if that case ever arises, it represents a logic bug, which we should maybe assert against.
> 
> What's the actual use case for allowing any sort of move construction or assignment of these types? I figured it was for something like moving into a closure, where there's no existing event to be dealt with.

One I can think of off the top of my head would be an object that needs to extend the lifetime of the TraceScopedEvent/Note until it is passed a new event. When passed a new event (lets say at the end of an iteration or something), it would need to move the new event into its storage, releasing the previous scope.

PhilMiller:
Ok, I'll buy it.

cz4rs (approved):
Looks good to me.

PR #2029, opened 2022-12-01
2028 setcontext doesnt suspend correct task with threads

### LINKED ISSUE #2028: SetContext doesn't suspend correct task with threads
**Describe the bug**
When there are nested tasks, the suspended task isn't tracked properly causing an assert to fail.

lifflander:
Fixes #2028

### DIFF
--- a/src/vt/context/runnable_context/set_context.cc
+++ b/src/vt/context/runnable_context/set_context.cc
@@ -80,7 +80,13 @@ void SetContext::suspend() {
     print_ptr(this), print_ptr(prev_task_.get()), print_ptr(cur_task_.get())
   );
 
-  finish();
+  // Get the innermost task, that actually called for this thread to block
+  suspended_task_ = theContext()->getTask();
+
+  if (suspended_task_ != nullptr)
+    vtAssert(prev_task_ == nullptr, "There should be no prev_task");
+
+  ContextAttorney::setTask(prev_task_);
 }
 
 void SetContext::resume() {
@@ -90,7 +96,9 @@ void SetContext::resume() {
     print_ptr(this), print_ptr(prev_task_.get()), print_ptr(cur_task_.get())
   );
 
-  start();
+  prev_task_ = theContext()->getTask();
+  ContextAttorney::setTask(suspended_task_);
+  suspended_task_ = nullptr;
 }
 
 }} /* end namespace vt::ctx */
--- a/src/vt/context/runnable_context/set_context.cc
+++ b/src/vt/context/runnable_context/set_context.cc
@@ -83,8 +83,7 @@ void SetContext::suspend() {
   // Get the innermost task, that actually called for this thread to block
   suspended_task_ = theContext()->getTask();
 
-  if (suspended_task_ != nullptr)
-    vtAssert(prev_task_ == nullptr, "There should be no prev_task");
+  vtAssert(prev_task_ == nullptr, "There should be no prev_task");
 
   ContextAttorney::setTask(prev_task_);
 }
--- a/src/vt/scheduler/thread_action.cc
+++ b/src/vt/scheduler/thread_action.cc
@@ -73,11 +73,16 @@ ThreadAction::~ThreadAction() {
 void ThreadAction::run() {
   ctx_ = make_fcontext_stack(stack_, runFnImpl);
 
+  // `jump_fcontext` transfers control to the fcontext thread indicated by `ctx_`
+  // that we just set up.  Control will 'return' from `jump_fcontext` to here when
+  // that thread eventually suspends or completes
   auto prev_running = cur_running_;
   cur_running_ = this;
 
   transfer_in_ = jump_fcontext(ctx_, static_cast<void*>(this));
 
+  // reset the current running thread ID after we finish or suspend the current
+  // thread
   cur_running_ = prev_running;
 }
 
@@ -91,6 +96,7 @@ void ThreadAction::resume() {
     return;
   }
 
+  // the same logic in run() applies here
   auto prev_running = cur_running_;
   cur_running_ = this;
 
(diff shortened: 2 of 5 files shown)


PhilMiller (approved):

PhilMiller:
Someone else should review this besides me, because Jonathan and I pair-programmed it. In particular, it would be valuable to get some assurance that someone else can understand what's going on, or get us to write the comments that enable that understanding.

PhilMiller:
The threading test looks good. I'm glad it turned out to be so simple to set up and catch the issue

tests/unit/active/test_async_op_threads.cc
```diff
@@ -71,7 +73,6 @@ struct MyObjGroup {
 
     // get the epoch stack and store the original size
```
nlslatt, line 74: Stale comment.

tests/unit/active/test_async_op_threads.cc
```diff
@@ -104,7 +105,7 @@ struct MyObjGroup {
     theMsg()->pushEpoch(cur_ep);
     theMsg()->pushEpoch(cur_ep);
 
-    EXPECT_EQ(epoch_stack.size(), original_epoch_size + 2);
+    auto const previous_stack_size = epoch_stack.size();
```
nlslatt, line 108: Could you make the epoch stack size names less ambiguously self-documenting? The relationship between `previous_stack_size` and `stack_before` is not at all clear.

tests/unit/active/test_async_op_threads.cc
```diff
@@ -91,11 +92,11 @@ struct MyObjGroup {
     }
     auto op2 = std::make_unique<messaging::AsyncOpMPI>(
       req2,
-      [this,original_epoch_size]{
+      [this]{
         done_ = true;
         // stack should be the size before running this method since we haven't
         // resumed the thread yet!
-        EXPECT_EQ(theTerm()->getEpochStack().size(), original_epoch_size - 2);
+        EXPECT_EQ(theTerm()->getEpochStack().size(), stack_before);
```
nlslatt, line 99: The comment above doesn't adequately explain this to me.

nlslatt (commented):
It would be helpful for me to see more comments related to the transfer in/out stuff in `ThreadAction` (I have no experience with fcontext).

PhilMiller:
Testing as a hot patch in EMPIRE also passed

src/vt/context/runnable_context/set_context.cc
```diff
@@ -80,7 +80,13 @@ void SetContext::suspend() {
     print_ptr(this), print_ptr(prev_task_.get()), print_ptr(cur_task_.get())
   );
 
-  finish();
+  // Get the innermost task, that actually called for this thread to block
+  suspended_task_ = theContext()->getTask();
+
+  if (suspended_task_ != nullptr)
```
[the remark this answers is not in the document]
PhilMiller: I don't think there is any. If there's no nested task, then `suspended_task_ == this`

We can probably remove the `if` and just assert unconditionally.

PhilMiller:
> It would be helpful for me to see more comments related to the transfer in/out stuff in `ThreadAction` (I have no experience with fcontext).

I agree. I don't entirely understand it either, and Jonathan had to work at explaining it to me.

Do you think that's essential to moving this forward?

src/vt/context/runnable_context/set_context.cc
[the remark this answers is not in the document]
PhilMiller: What's going on here is essentially that a threaded task is always supposed to be the outermost, which corresponds to the assertion.

(writing the following mostly for my own edification / future source of documentation, rather than in answer to this particular question)

When `ThreadAction::suspend()` gets called, the `jump_fcontext` switches control to whichever of `ThreadAction::run` or `ThreadAction::resume` called into it, and the return from those goes back to the bits of runnable that call the `SetContext` methods, as well as all of the other runnable contexts bits (LB, tracing, TD, etc). At that point, it will call `SetContext::suspend` on the runnable owning the thread that it ran, rather than any nested runnable task that the thread had called into (e.g. via `invoke` or some other handler executed in-line). That runnable should always have been called directly from the vt scheduler.

Which makes me realize that we still have a hypothetical nesting problem if some code inside a threaded handler calls `runInEpoch*` or `runThroughEpoch` or anything that turns the scheduler, if there might then be suspensions of threads run within that. I think we can address that by evolving those to run the nested scheduler in a fresh thread, rather than recursively, but that's the next level of stuff to figure out and fix.

@lifflander want to write a test case for the (... in a different issue, please?)?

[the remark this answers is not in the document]
PhilMiller: Testing of the current codebase passed with that assertion being made unconditionally, at least.

src/vt/scheduler/thread_action.cc
```diff
@@ -72,7 +72,13 @@ ThreadAction::~ThreadAction() {
 
 void ThreadAction::run() {
   ctx_ = make_fcontext_stack(stack_, runFnImpl);
+
+  auto prev_running = cur_running_;
+  cur_running_ = this;
+
```
PhilMiller, line 78:
```suggestion

// `jump_fcontext` transfers control to the fcontext thread indicated by `ctx_` that we just set up.
// Control will 'return' from `jump_fcontext` to here when that thread eventually suspends or completes
```

nlslatt:
@lifflander @PhilMiller Is someone going to address my other review comments like deleting the stale comment?

PhilMiller:
Should be Jonathan, as author of those tests

lifflander:
Ok, I've finally finished with my comments and fixes and pushed.

nlslatt (approved):
Looks good to me

Issue #2025, opened 2022-11-30
Default value in build script


stmcgovern:
Following #1906, the build_cpp.sh does not complete as `CMAKE_CXX_STANDARD` is not set by default. Provide a default value (i.e. 14).

stmcgovern:
Closing as this is superseded by #2024 and PR #2032.

PR #2032, opened 2022-12-02
#2024: Fix CI issue with empty CXX standard variable

### LINKED ISSUE #2024: Documentation pipeline is broken
**Describe the bug**
```
-- Found Git: /usr/bin/git (found version "2.17.1") 
fatal: detected dubious ownership in repository at '/vt'
To add an exception for this directory, call:

	git config --global --add safe.directory /vt
```

thearusable:
Closes #2024


cz4rs (approved):

PhilMiller:
Also, what's the relationship between this change and the error shown in the linked issue?

PhilMiller:
I believe Cezary understands, given his quick approval, but it's helpful to write things down so the rest of us can follow later

thearusable:
@PhilMiller Force option is for overriding the already existing value in `CMAKE_CXX_STANDARD`, without that CMake's set function will not modify the value. In our case it was incorrectly set to an empty string. 
So now even if we set that variable to `""` in the script then it will still work properly because it would override it with the default value of `14`.
Previously there was also a small issue in the scripts which will pass `""` when no value is set. Now they will set a proper default value.

PhilMiller:
It seems that if the variable is set with nonsense, we shouldn't carry on and try to do something that looks sensible. If it's a likely user mistake, we should detect it and report the error. Otherwise, I don't think we should try to compensate for a script bug that we know about and are fixing.

Is there some subtlety here I'm missing?

thearusable:
No, you are right. Ok so I'll remove the FORCE option. Scripts will not create that issue again and if user set some gibberish then it would fail as on the pipeline right now.

PhilMiller (approved):

stmcgovern (approved):
I'll close #2025 since this PR includes that and an additional fix.

lifflander (approved):

PR #2034, opened 2022-12-06
#2033: lb: rewrite and fix bugs in OfflineLB/LBDataRestartReader

### LINKED ISSUE #2033: Simplify and fix OfflineLB
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

lifflander:
Fixes #2033


lifflander:
Tests are coming.

src/vt/vrt/collection/balance/lb_data_restart_reader.cc
```diff
@@ -78,45 +78,15 @@ void LBDataRestartReader::startup() {
...
-
-std::deque<std::set<ElementIDType>> LBDataRestartReader::readIntoElementHistory(
-  LBDataHolder const& lbdh
-) {
-  std::deque<std::set<ElementIDType>> element_history;
-  for (PhaseType phase = 0; phase < lbdh.node_data_.size(); phase++) {
-    std::set<ElementIDType> buffer;
+void LBDataRestartReader::readHistory(LBDataHolder const& lbdh) {
+  num_phases_ = lbdh.node_data_.size();
+  for (PhaseType phase = 0; phase < num_phases_; phase++) {
```
nlslatt, line 83: This seems to assume that a file with `N` phases will contain specifically phases `0` through `N-1`. Now that we have the ability to selectively output phases to the LB data files, I think this is no longer a safe assumption.

Do we want to handle cases where the file contains gaps in phase ranges that may not even be reached by this run? We should at least give a better error message than just an out of range exception when we fail to find an expected phase.

Do we want to allow cases where `OfflineLB` is only used for a subset of particularly challenging phases, eliminating the requirement to have a continuous range of phases of history available?

Do we want to allow reading in only a subset of phases from potentially very large files?

tests/unit/lb/test_offlinelb.cc
```diff
@@ -0,0 +1,144 @@
...
+// POSSIBILITY OF SUCH DAMAGE.
+//
+// Questions? Contact darma@sandia.gov
+//
+// *****************************************************************************
+//@HEADER
+*/
+
+#include <vt/transport.h>
+
```
nlslatt, line 45: Looks like you need an additional include.

nlslatt (commented):
I wonder if we need to make this more general. Please see the comment below.

src/vt/vrt/collection/balance/lb_data_restart_reader.cc
lifflander: Good point. I think you are right that we need to consider cases where not all phases are output.

src/vt/vrt/collection/balance/lb_data_restart_reader.cc
```diff
@@ -78,45 +78,22 @@ void LBDataRestartReader::startup() {
...
-    for (auto const& obj : lbdh.node_data_.at(phase)) {
-      if (obj.first.isMigratable()) {
-        buffer.insert(obj.first.id);
+void LBDataRestartReader::readHistory(LBDataHolder const& lbdh) {
+  num_phases_ = lbdh.node_data_.size();
+  PhaseType prev_known_phase = 0;
+  for (PhaseType phase = 0; phase < num_phases_; phase++) {
+    auto iter = lbdh.node_data_.find(phase);
+    if (iter == lbdh.node_data_.end()) {
+      history_[phase] = history_[prev_known_phase];
```
nlslatt, line 87: This is an interesting solution to the problem. Could we end up using too much memory this way? My original thought was that the LB interval would likely not be 1, so only a subset of the phases would be needed anyway. I think it's also possible to mix `OfflineLB` with another load balancer (on different phases) by using an LB config file.

lifflander: I agree that it uses more memory than necessary. It was just the easiest way to write the code. I can do a search each time instead that will find the last previous known phase.

lifflander: Ok based on our conversation, for the sake of this PR, we are going to assume that the phase specification is dense is fully specified for every phase up to num_phases.

src/vt/vrt/collection/balance/lb_data_restart_reader.h
```diff
@@ -83,67 +83,96 @@ struct LBDataRestartReader : runtime::component::Component<LBDataRestartReader>
...
   template <typename Serializer>
   void serialize(Serializer& s) {
-    s | msgsReceived
-      | totalMove
-      | proxy_
-      | proc_move_list_
-      | proc_phase_runs_LB_;
+    s | proxy_
+      | changed_distro_
+      | num_phases_;
```
nlslatt, line 122: Should the serializer include `history_`?

lifflander: Fixes

nlslatt (approved):
Looks good to me

PR #2039, opened 2022-12-07
2027: Create Runnable inside Objgroup's invoke

### LINKED ISSUE #2027: Objgroup invoke does not create a runnable
**Describe the bug**
When proxy[x].invoke is called for an objgroup, a runnable is not created. This contrasts to a collection invoke which does create a runnable.

JacobDomagala:
Fixes #2027


src/vt/runnable/make_runnable.h
```diff
@@ -272,6 +272,8 @@ struct RunnableMaker {
    * \param[in] task_ the task to execute
    */
   RunnableMaker&& withExplicitTask(ActionType task_) {
+    // Set this to 'true' as we're not going through normal handler
+    set_handler_ = true;
```
JacobDomagala, line 276: Not doing this causes issues inside `RunnableMaker::run()` (it's trying to setup handler and fails). For Collection this is handled inside `RunnableMaker::withCollection()`. 
Not sure if that's the best way to do it though. Should we add additional field `bool has_explicit_task` or something like it?

lifflander: Sorry, I meant to post this on the issue. I added `withObjGroup` in this PR https://github.com/DARMA-tasking/vt/pull/1984, which has not been merged yet.

PhilMiller: Can that aspect be picked out of there?

JacobDomagala: With #1984 merged, I'll get back to this PR

lifflander (approved):

src/vt/objgroup/manager.impl.h
```diff
@@ -211,20 +211,45 @@ void ObjGroupManager::invoke(
...
-      dest_node
-    )
-  );
-
-  return runnable::invoke<Type, f>(get(proxy), std::forward<Args>(args)...);
+      "Attempting to invoke handler on node:{} instead of node:{}!\n",
+      this_node, dest_node));
+
+  using Ret = typename util::FunctionWrapper<Type>::ReturnType;
+  constexpr bool is_void = std::is_same<Ret, void>::value;
```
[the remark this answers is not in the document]
PhilMiller: Or just inline that expression in the condition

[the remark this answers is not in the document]
JacobDomagala: Done

src/vt/objgroup/manager.impl.h
```diff
@@ -211,20 +211,45 @@ void ObjGroupManager::invoke(
...
+  constexpr bool copyable = std::is_copy_constructible<Ret>::value;
+
+  if constexpr (not is_void) {
+    Ret result;
+
+    runnable::makeRunnableVoid(false, uninitialized_handler, this_node)
+      .withObjGroup(get(proxy))
+      .runLambda([&] {
+        if constexpr (copyable) {
```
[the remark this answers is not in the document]
JacobDomagala: Done

src/vt/objgroup/manager.impl.h
```diff
@@ -211,20 +211,43 @@ void ObjGroupManager::invoke(
...
+      .runLambda([&] {
+        if constexpr (std::is_copy_constructible_v<Ret>) {
+          result =
+            runnable::invoke<Type, f>(get(proxy), std::forward<Args>(args)...);
+        } else {
+          auto&& ret =
+            runnable::invoke<Type, f>(get(proxy), std::forward<Args>(args)...);
+          result = std::move(ret);
+        }
+      });
```
PhilMiller, line 241: Does this all do the correct thing if `f` returns an l-value reference to a moveable but non-copyable type? Without looking at `FunctionWrapper`, if `Ret` preserves the reference, then `Ret result` just won't compile, I think.

My concern is that we may get back a reference to something that needs to be left alone, and then move from it.

PhilMiller: More deeply, is there a reason that `runLambda` doesn't itself pass through the return value from the thing it runs, such that we could do 
```c++
return runnable::makeRunnableVoid(...).runLambda([] { 
  return runnable::invoke<Type, f>(get(proxy), forward(args));
});
```

If it needs to manage making `finish` calls on the various context bits, those could be move to a scope guard structure that will call them in their destructors, rather than in straight-line code, so that we can return without declaring temporaries that would need to be conscientious about reference types.

lifflander: Good point. Now that runLambda does it immediately we can push the return value through

src/vt/runnable/make_runnable.h
```diff
@@ -273,12 +273,21 @@ struct RunnableMaker {
...
+      std::invoke(std::forward<Callable>(c), std::forward<Args>(args)...);
+      delete impl_;
+      impl_ = nullptr;
+      is_done_ = true;
+    } else {
+      decltype(auto) r{impl_->runLambda(std::forward<Callable>(c), std::forward<Args>(args)...)};
+      delete impl_;
+      impl_ = nullptr;
+      is_done_ = true;
+      return r;
```
PhilMiller, line 289: The `decltyp(auto)` form preserves references, right?

To not have to think about it, I'd slightly prefer something like

```c++
auto local_impl = unique_ptr<Blah>(impl_);
impl_ = nullptr;
is_done_ = true;
return local_impl->runLambda(...);
```
With that formulation, I believe we don't even need to distinguish the `void` case, because C++ allows returning void but not creating a local variable.

src/vt/objgroup/manager.h
```diff
@@ -234,7 +234,7 @@ struct ObjGroupManager : runtime::component::Component<ObjGroupManager> {
    * \param[in] proxy proxy to the object group
    * \param[in] args function arguments
    */
-  template <typename ObjT, typename Type, Type f, typename... Args>
+  template <typename ObjT, auto f, typename... Args>
```
PhilMiller, line 237: The client interface for `invoke` changes from this, right? Are we concerned about maintaining compatibility?

lifflander: Yes, it changes. I think breaking this is fine. We can update any app code.

src/vt/runnable/make_runnable.h
lifflander: Yes, it preserves references.

src/vt/objgroup/manager.h
PhilMiller: Makes it a bit harder for EMPIRE to `git bisect` on vt issues, but I don't expect that to matter.

src/vt/vrt/collection/manager.h
```diff
@@ -665,37 +665,13 @@ struct CollectionManager
...
-   * through scheduler
-   *
-   * \param[in] proxy the collection proxy
-   * \param[in] args function params
-   */
-  template <typename ColT, typename Type, Type f, typename... Args>
-  util::IsVoidReturn<Type>
-  invoke(VirtualElmProxyType<ColT> const& proxy, Args... args);
+  template <typename ColT, auto f, typename... Args>
+  auto invoke(VirtualElmProxyType<ColT> const& proxy, Args&&... args);
```
PhilMiller, line 674: Potentially a more involved interface change, but if you put `auto f` before `typename ColT`, then `ColT` can be deduced, right?

lifflander: This is an internal interface. The actual interface doesn't require ColT because it's called on the proxy which has it already

lifflander: proxy[idx].invoke<&MyCol::handler>(args...)

PhilMiller (approved):

PR #2047, opened 2022-12-20
2045 Switch default builds to C++17 and remove old CUDA build

### LINKED ISSUE #2045: Switch default builds to C++17, remove old CUDA build
**What Needs to be Done?**

lifflander:
Fixes #2045


cz4rs:
A bunch of warnings promoted to errors in `clang` builds:
```
/vt/src/vt/utils/adt/union.h:727:5: error: expression result unused [-Werror,-Wunused-value]
```

cz4rs:
Regarding the `PR checks (PR description format)` failure - possibly a regex didn't expect `+` character in the branch name?
Also note that the comment aggregating build results / errors is missing.

thearusable:
> Regarding the `PR checks (PR description format)` failure - possibly a regex didn't expect `+` character in the branch name?
> Also note that the comment aggregating build results / errors is missing.

Yes, regex is not expecting the '+' character: https://github.com/DARMA-tasking/check-pr-fixes-issue/blob/master/src/helpers.js#L3

cz4rs:
`icpc` fails because of:
```
The following tests FAILED:
	399 - vt:TestPreconfig.test_vt_assert_no_mpi (SEGFAULT)
```
This might be worth investigating, but on the other hand the log is also littered with:
```
icpc: remark #10441: The Intel(R) C++ Compiler Classic (ICC) is deprecated and will be removed from product release in the second half of 2023. The Intel(R) oneAPI DPC++/C++ Compiler (ICX) is the recommended compiler moving forward. Please transition to use this compiler. Use '\''-diag-disable=10441'\'' to disable this message.
```

Can we drop `icpc` support right now?

cz4rs:
On the other hand, `clang-5.0` failure seems to be directly related to this PR:
```
In file included from /vt/src/vt/worker/worker_group_comm.cc:45:
/vt/src/vt/context/context.h:225:3: error: non-type template argument refers to subobject '&tls_static_str_thisWorker__'
  DeclareClassInsideInitTLS(Context, WorkerIDType, thisWorker_, no_worker_id)
  ^
```

PhilMiller:
For `icpc`, we'll probably want to pin whatever version of Intel's oneAPI is being supported by Trilinos and used by the Sandia apps.

The clang 5 failure in worker will go away when the removal PR is merged

PR #2064, opened 2023-01-13
276 Automatically synthesize and deduce message types for sends to arbitrary handlers

### LINKED ISSUE #276: Automatically synthesize message types for sends to arbitrary handlers

lifflander:
Fixes #276


cmake/load_bundled_libraries.cmake
```diff
@@ -3,7 +3,7 @@
 include(SetCXXCompilerFlags)
 
 # Export a minimum version flag for any bundled libraries that don't set their own
-set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD 17)
```
PhilMiller, line 6: #2067 clears this out.

examples/hello_world/hello_world.cc
```diff
@@ -49,24 +49,21 @@ struct HelloMsg : vt::Message {
...
   vt::NodeType num_nodes = vt::theContext()->getNumNodes();
-
   if (num_nodes == 1) {
     return vt::rerror("requires at least 2 nodes");
   }
 
   if (this_node == 0) {
-    auto msg = vt::makeMessage<HelloMsg>(this_node);
-    vt::theMsg()->broadcastMsg<HelloMsg, hello_world>(msg);
+    vt::theMsg()->send<helloWorld>(1, 10, 20, 11.3f);
```
PhilMiller, line 66: Oof, that first argument being undistinguished is ugly. Won't come up with proxies, but it might be good to use a wrapped up Node type that's not just an integer alias

lifflander: Yes, I was thinking the same thing

src/vt/registry/auto/auto_registry_general.h
```diff
@@ -154,10 +155,11 @@ struct FunctorAdapterArgs<ObjTypeT, MsgT> {
 
 // Need to provide a non-pointer overload for parameterization auto-registered
 // functions for GCC
```
PhilMiller, line 157: Was this here to address a limitation in older versions of GCC, that may no longer be applicable?

examples/hello_world/hello_world.cc
PhilMiller: If the destination argument became a strong type, I'm ready to approve the rest.

lifflander: Done

PhilMiller (approved):

PhilMiller:
Separately, I think it would be worthwhile to do a refactor pushing that strong `Node` type all the way down the stack, to the point that we don't call `get()` on it except in MPI call arguments.

PhilMiller (approved):

lifflander:
@thearusable Can you focus on the objgroup and collection send/broadcast APIs?

thearusable:
> @thearusable Can you focus on the objgroup and collection send/broadcast APIs?

Sure, will do

src/vt/group/group_manager.h
```diff
@@ -193,6 +193,24 @@ struct GroupManager : runtime::component::Component<GroupManager> {
...
   void sendMsg(GroupType const group, MsgT* msg);
 
+  template <typename ReturnT, typename... Args>
+  struct FunctionTraits;
+
+  template <typename ReturnT, typename T>
+  struct FunctionTraits<ReturnT(*)(T*)> {
+    using MsgT = T;
+    using ReturnType = ReturnT;
+  };
```
PhilMiller, line 203: These `FunctionTraits` helpers are proliferating. They should probably just getting written in a common place once.

PhilMiller:
I'd be a bit more comfortable if the changes for `reduce` and `scatter` were split out to follow-up PRs, just because their interface is a bit more involved.

PhilMiller:
OK, `scatter` is fine to keep in.. That's small and similar to `send/broadcast`

src/vt/registry/auto/auto_registry_common.h
```diff
@@ -153,48 +115,27 @@ struct BaseMapsDispatcher {
...
   ) const override {
-    return DispatchImpl<HandlerT>::run(
-      static_cast<IndexT*>(cur_idx_ptr),
-      static_cast<IndexT*>(range_ptr),
-      num_nodes,
-      fn_ptr_
-    );
+    using T = HandlerT;
+
+    if constexpr (std::is_same_v<T, vt::mapping::ActiveMapTypedFnType<IndexT>*>) {
```
PhilMiller, line 128: `icpc` somewhat justifiably complains that if this condition isn't satisfied, then this function is missing a `return`

src/vt/registry/auto/auto_registry_common.h
```diff
@@ -69,77 +69,39 @@ struct BaseHandlersDispatcher {
...
+      if constexpr (std::is_same_v<ObjT, SentinelObject>) {
+        std::apply(fp, msg->getTuple());
+      } else {
+        if constexpr (std::is_same_v<T, ColMemberTypedFnType>) {
+          (elm->*fp)(msg);
+        } else {
+          std::apply(fp, std::tuple_cat(std::make_tuple(elm), msg->getTuple()));
+        }
+      }
+    }
```
PhilMiller, line 100: Technically not part of the issue here, but I'm happy to see the simplification, and it doesn't raise any interesting questions.

src/vt/registry/auto/auto_registry_common.h
PhilMiller: Actually, I think this particular replacement of `enable_if` with `if constexpr` is wrong. If something inappropriately tries calling this now, it will end up with a no-op rather than a compilation failure. The condition at least needs to be a `static_assert`

src/vt/registry/auto/auto_registry_common.h
```diff
@@ -204,32 +145,19 @@ struct BaseScatterDispatcher {
...
 
 public:
-  void dispatch(void* msg, void* object) const override {
-    DispatchImpl<HandlerT>::run(static_cast<MsgT*>(msg), object, fn_ptr_);
+  void dispatch(void* msg, void*) const override {
+    using T = HandlerT;
+
+    if constexpr (std::is_same_v<T, ActiveTypedFnType<MsgT>*>) {
+      fp(static_cast<MsgT*>(msg));
+    }
```
PhilMiller, line 156: Ditto the comment about `MapsDispatcher`

src/vt/utils/fntraits/fntraits.h
```diff
@@ -0,0 +1,98 @@
...
+struct NoMsg {};
+
+template <typename enabled, typename ObjT, typename... Args>
+struct ObjFuncTraits;
+
+template <typename ObjT, typename Return, typename Msg>
+struct ObjFuncTraits<
+  std::enable_if_t<std::is_convertible<Msg, vt::Message>::value>,
+  ObjT,
+  Return(*)(Msg*, ObjT*)
```
PhilMiller, line 58: Jonathan noted that the current ordering in this case impedes parameterization and automatic synthesis. He's reversing the order. Technically a breaking interface change, but we think it will have no external impact.

src/vt/registry/auto/auto_registry_common.h
lifflander: Added a static_assert

src/vt/group/group_manager.h
lifflander: Yes, that is done

src/vt/messaging/active.h
```diff
@@ -86,7 +89,12 @@ using ContinuationDeleterType =
 
 } /* end namespace vt */
 
-namespace vt { namespace messaging {
+namespace vt {
+
+struct StrongNodeType { };
+using Node = Strong<NodeType, uninitialized_destination, StrongNodeType>;
```
PhilMiller, line 95: This should maybe be in a lower-level shared header, so that we're more able to use it across the codebase.

src/vt/messaging/active.h
```diff
@@ -704,6 +733,60 @@ struct ActiveMessenger : runtime::component::PollableComponent<ActiveMessenger>
...
+   *
+   * \return the \c PendingSend for the sent message
+   */
+  template <auto f, typename... Params>
+  PendingSendType broadcast(Params&&... params) {
+    using Tuple = typename FuncTraits<decltype(f)>::TupleType;
+    using MsgT = ParamMsg<Tuple>;
+    auto msg = vt::makeMessage<MsgT>(std::forward<Params>(params)...);
+    auto han = auto_registry::makeAutoHandlerParam<decltype(f), f, MsgT>();
+    return broadcastMsg<MsgT>(han, msg, true, no_tag);
```
PhilMiller, line 787:
```suggestion
    constexpr bool deliver_to_sender = true;
    return broadcastMsg<MsgT>(han, msg, deliver_to_sender, no_tag);
```

src/vt/objgroup/proxy/proxy_objgroup.impl.h
```diff
@@ -82,6 +84,26 @@ typename Proxy<ObjT>::PendingSendType Proxy<ObjT>::broadcast(Args&&... args) con
   return broadcastMsg<MsgT,fn>(makeMessage<MsgT>(std::forward<Args>(args)...));
 }
 
+template <typename ObjT>
+template <auto f, typename... Params>
+typename Proxy<ObjT>::PendingSendType
+Proxy<ObjT>::broadcast(Params&&... params) const {
+  using MsgT = typename ObjFuncTraits<decltype(f)>::MsgT;
+  if constexpr (std::is_same_v<MsgT, NoMsg>) {
```
PhilMiller, line 92: Could you add a comment on how this case comes up, and how we end up with parameterization to do when `MsgT == NoMsg`?

src/vt/registry/auto/auto_registry_common.h
```diff
@@ -204,32 +144,21 @@ struct BaseScatterDispatcher {
...
 public:
-  void dispatch(void* msg, void* object) const override {
-    DispatchImpl<HandlerT>::run(static_cast<MsgT*>(msg), object, fn_ptr_);
+  void dispatch(void* msg, void*) const override {
+    using T = HandlerT;
+
+    if constexpr (std::is_same_v<T, ActiveTypedFnType<MsgT>*>) {
+      fp(static_cast<MsgT*>(msg));
+    } else {
+      vtAssert(false, "Invalid function type for scatter handler");
```
PhilMiller, line 156:
```suggestion
      vtAbort("Invalid function type for scatter handler");
```
So it would still be there in a debug build that selected the wrong path.

src/vt/registry/auto/auto_registry_common.h
```diff
@@ -153,48 +111,30 @@ struct BaseMapsDispatcher {
...
+    using T = HandlerT;
+
+    if constexpr (std::is_same_v<T, vt::mapping::ActiveMapTypedFnType<IndexT>*>) {
+      return fp(
+        static_cast<IndexT*>(cur_idx_ptr),
+        static_cast<IndexT*>(range_ptr),
+        num_nodes
+      );
+    } else {
+      vtAssert(false, "Invalid function type for map handler");
```
PhilMiller, line 131:
```suggestion
      vtAbort("Invalid function type for map handler");
```

src/vt/messaging/active.h
PhilMiller: That can happen in a follow-up PR

PhilMiller: See #2099

src/vt/utils/fntraits/fntraits.h
PhilMiller: There's two call sites in EMPIRE impacted, but in both cases the improved template `auto` parameter deduction makes them simpler

src/vt/registry/auto/auto_registry_common.h
```diff
@@ -69,77 +69,35 @@ struct BaseHandlersDispatcher {
...
+  explicit HandlersDispatcher(HandlerT in_fn_ptr) : fp(in_fn_ptr) { }
 
 public:
-  void dispatch(messaging::BaseMsg* msg, void* object) const override {
-    DispatchImpl<HandlerT>::run(static_cast<MsgT*>(msg), object, fn_ptr_);
+  void dispatch(messaging::BaseMsg* base_msg, void* object) const override {
+    using T = HandlerT;
+
+    [[maybe_unused]] auto msg = static_cast<MsgT*>(base_msg);
+    [[maybe_unused]] auto elm = static_cast<ObjT*>(object);
```
cz4rs, line 82: Is the `[[maybe_unused]]` necessary here? (a.k.a. can the compilers account for an `else` at the end?)

PhilMiller: I've seen Intel's classic compiler warn on other code with the same pattern. It's a bug that they're never going to fix, because it's end of life.

src/vt/registry/auto/auto_registry_general.h
cz4rs: Should checking this be converted into a separate issue? (to avoid delaying merge for this PR)

PR #2072, opened 2023-01-18
#2071: make pre-worker-removal init call deprecated

### LINKED ISSUE #2071: Make old initialization syntax deprecated instead of removed
The old initialization syntax taking the number of workers should be deprecated, since it can break applications even if they don't use workers.

nmm0:
Closes #2071


PhilMiller (approved):

nmm0:
I think some things are failing because of the unity builds

nmm0:
> I think some things are failing because of the unity builds

nvm it was just an inconsistency between the ways that compilers different compilers handled things

PR #2089, opened 2023-02-10
2086 loosen when lbdatarestartreader is created

### LINKED ISSUE #2086: Loosen when LBDataRestartReader is created
**Describe the bug**
Right now the condition is far too specific.

```C++
bool Runtime::needLBDataRestartReader() {
  #if vt_check_enabled(lblite)
    if (arg_config_->config_.vt_lb_data) {
      auto lbNames = vrt::collection::balance::get_lb_names();
      auto mapLB = vrt::collection::balance::LBType::OfflineLB;
      if (arg_config_->config_.vt_lb_name == lbNames[mapLB]) {
        return true;
      }
    }
  #endif
  return false;
}
```

lifflander:
Fixes #2086

### DIFF
--- a/src/vt/runtime/runtime.cc
+++ b/src/vt/runtime/runtime.cc
@@ -385,7 +385,7 @@ bool Runtime::tryFinalize(bool const disable_sig) {
 bool Runtime::needLBDataRestartReader() {
   #if vt_check_enabled(lblite)
   if (true) {
-    return arg_config_->config_.vt_lb_data_in != "";
+    return arg_config_->config_.vt_lb_data_in;
   } else
   #endif
     return false;
--- a/src/vt/runtime/runtime.cc
+++ b/src/vt/runtime/runtime.cc
@@ -384,15 +384,11 @@ bool Runtime::tryFinalize(bool const disable_sig) {
 
 bool Runtime::needLBDataRestartReader() {
   #if vt_check_enabled(lblite)
-    if (arg_config_->config_.vt_lb_data) {
-      auto lbNames = vrt::collection::balance::get_lb_names();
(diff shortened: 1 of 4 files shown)


src/vt/runtime/runtime.cc
```diff
@@ -385,15 +385,11 @@ bool Runtime::tryFinalize(bool const disable_sig) {
...
   #if vt_check_enabled(lblite)
-    if (arg_config_->config_.vt_lb_data) {
-      auto lbNames = vrt::collection::balance::get_lb_names();
-      auto mapLB = vrt::collection::balance::LBType::OfflineLB;
-      if (arg_config_->config_.vt_lb_name == lbNames[mapLB]) {
-        return true;
-      }
-    }
+  if (true) {
+    return arg_config_->config_.vt_lb_data_in != "";
```
nlslatt, line 389: I'm guessing this was meant to say `vt_lb_data_dir_in` given that you made the default value of that argument empty. If the user wants to read from the working directory, they would have to set that to `.` in order for this logic to enable the reader. Is there a reason we can't add boolean `vt_lb_data_in` like we already have a `vt_lb_data` (for outputting)?

nlslatt (changes requested):

nlslatt (changes requested):
The printing of the banner items for inputting LB data is conditional on the output flag, `vt_lb_data`.

nlslatt:
> The printing of the banner items for inputting LB data is conditional on the output flag, `vt_lb_data`.

Also, `vt_lb_data_in` is missing from the serializer in `app_config.h`.

nlslatt (approved):
Looks good

PR #2093, opened 2023-02-14
#2092: location: call function pointer directly

### LINKED ISSUE #2092: Runtime optimizations stage 2
**What Needs to be Done?**

- Move messages instead of passing MsgSharedPtr around
- Remove runnable from location manager (~200ns speedup!)
-

lifflander:
Fixes #2092


PhilMiller:
Are the first two commits clearing up a first/second handler kind of situation? Why shouldn't a runnable be made and called there?

PhilMiller:
More deeply, do we need to maintain some notion of 'system mode' runnables that do all of the relevant instrumentation for occasions when we want to trace and account for low-level activity?

lifflander:
> More deeply, do we need to maintain some notion of 'system mode' runnables that do all of the relevant instrumentation for occasions when we want to trace and account for low-level activity?

This is for the particular case where the collection message send is completely local. It goes through the location manager and then needs to call back into the collection manager. Creating a runnable in this case isn't particularly useful because the context it preserved on the stack (the runnable should not be enqueued). Once the collection's runnable is created, it preserves the context.

If the send is off node, the location manager will create runnables for its handlers to forward, etc.

Do you think this case is fine to optimize out?

src/vt/messaging/active.cc
```diff
@@ -963,7 +963,8 @@ void ActiveMessenger::prepareActiveMsgToRun(
   if (is_obj) {
     objgroup::dispatchObjGroup(base, handler, from_node, cont);
   } else {
-    runnable::makeRunnable(base, not is_term, handler, from_node)
+    auto m = base;
```
PhilMiller, line 966: Why is `base` declared as `const&` here, necessitating the additional reference count bump and then move of the copy? Can ownership transfer flow further up the stack?

src/vt/runnable/make_runnable.h
```diff
@@ -331,19 +329,19 @@ struct RunnableMaker {
...
 ) {
-  auto r = new RunnableNew(msg, is_threaded);
+  auto r = new RunnableNew(std::move(msg), is_threaded);
 #if vt_check_enabled(trace_enabled)
   auto const han_type = HandlerManager::getHandlerRegistryType(handler);
   if (han_type == auto_registry::RegistryTypeEnum::RegVrt or
       han_type == auto_registry::RegistryTypeEnum::RegGeneral or
       han_type == auto_registry::RegistryTypeEnum::RegObjGroup) {
-    r->addContextTrace(msg, handler, from);
+    r->addContextTrace(r->getMsg(), handler, from);
```
PhilMiller, line 340: If `r` has the message pointer as a member now, then this method can just refer to it internally, can't it? I feel like we may have discussed that at some point in the past, and there was a reason not to back then.

src/vt/topos/location/location.impl.h
```diff
@@ -549,7 +549,7 @@ void EntityLocationCoord<EntityID>::routeMsgNode(
 
     theTerm()->produce(epoch);
 
-    auto trigger_msg_handler_action = [=](EntityID const& hid) {
+    auto trigger_msg_handler_action = [=](EntityID const& hid, MsgSharedPtr<MessageT>&& m) mutable {
```
PhilMiller, line 552: If this lambda is getting `m` as an argument now, does it need to capture `msg` at all? Should all of the `msg->foo` inside become `m->foo`?

PhilMiller:
Please squash the 'fix' commits into the commit they're fixing, for atomicity of review.

PhilMiller:
This generally looks good to me, even if only part of the code base is transformed to avoid copies, and there's more to be done later.

PhilMiller:
> > More deeply, do we need to maintain some notion of 'system mode' runnables that do all of the relevant instrumentation for occasions when we want to trace and account for low-level activity?
> 
> This is for the particular case where the collection message send is completely local. It goes through the location manager and then needs to call back into the collection manager. Creating a runnable in this case isn't particularly useful because the context it preserved on the stack (the runnable should not be enqueued). Once the collection's runnable is created, it preserves the context.
> 
> If the send is off node, the location manager will create runnables for its handlers to forward, etc.
> 
> Do you think this case is fine to optimize out?

That sounds good then. We don't need to separately account those two bits different bits of overhead, since we already have a clear context in which it's occurring.

stmcgovern:
Not ready yet, but took draft off to get more testing.

src/vt/runnable/make_runnable.h
nlslatt: @lifflander @stmcgovern Can one of you please respond to this?

src/vt/serialization/messaging/serialized_messenger.impl.h
```diff
@@ -428,14 +428,14 @@ template <typename MsgT, typename BaseT>
...
           if (is_obj) {
             objgroup::dispatchObjGroup(
               user_msg.template to<BaseMsgType>(), typed_handler, node, nullptr
             );
           } else {
-            runnable::makeRunnable(user_msg, true, typed_handler, node)
+            runnable::makeRunnable(std::move(user_msg), true, typed_handler, node)
               .withTDEpochFromMsg()
               .enqueue();
           }
```
nlslatt, line 441: Why does this lambda keep using `user_msg` (including to cast it to `BaseMsgType` again) instead of using `in`?

src/vt/topos/location/location.impl.h
```diff
@@ -727,15 +730,17 @@ void EntityLocationCoord<EntityID>::routePreparedMsg(
 
   if (use_eager) {
     theMsg()->pushEpoch(epoch);
-    routeMsgEager<MessageT>(msg->getEntity(), msg->getHomeNode(), msg);
+    routeMsgEager<MessageT>(msg->getEntity(), msg->getHomeNode(), std::move(msg));
```
nlslatt, line 733: Do you need to extract the entity and home node before this call in case the move is evaluated first?

src/vt/topos/location/location.impl.h
```diff
@@ -727,15 +730,17 @@ void EntityLocationCoord<EntityID>::routePreparedMsg(
...
     theTerm()->produce(epoch);
     // non-eager protocol: get location first then send message after resolution
-    getLocation(msg->getEntity(), msg->getHomeNode(), [=](NodeType node) {
+    auto entity = msg->getEntity();
+    auto home_node = msg->getHomeNode();
+    getLocation(entity, home_node, [this, epoch, m = std::move(msg)](NodeType node) mutable {
       theMsg()->pushEpoch(epoch);
       routeMsgNode<MessageT>(
-        msg->getEntity(), msg->getHomeNode(), node, msg
+        m->getEntity(), m->getHomeNode(), node, std::move(m)
```
nlslatt, line 743: Do you need to extract the entity and home node before this call in case the move is evaluated first?

nlslatt (commented):
Looks pretty good but there are some open questions.

stmcgovern:
This demonstrates performance improvement in the following performance test

From Develop @ [6109deb](https://github.com/DARMA-tasking/vt/commit/6109deb32f635df3b29665fd84e10fd99dd818e0)
```
Test results for test_collection_local_send running on 1 nodes:
[0] Results for test_collection_local_send (avg: 661.343ms stdev: 4.613ms min: 655.371ms max: 668.093ms)

----- TIMERS -----
[0] Results for colSend 1000000 (avg: 661.300ms stdev: 4.615ms min: 655.294ms max: 668.051ms)
```
From this branch 2092
```
Test results for test_collection_local_send running on 1 nodes:
[0] Results for test_collection_local_send (avg: 436.370ms stdev: 3.647ms min: 431.374ms max: 446.114ms)

----- TIMERS -----
[0] Results for colSend 1000000 (avg: 436.331ms stdev: 3.648ms min: 431.345ms max: 446.079ms)

```

PR #2097, opened 2023-03-04
#2036 Update PerfTestHarness to allow for multiple tests in a single file

### LINKED ISSUE #2036: Modify the perf test harness for multiple test configurations
In #1930, the performance test `collection_local_send` was split into two files to allow for two different test configurations (i.e., preallocate or not messages for sending). It would be nice to have the ability to configure multiple test variants (for this test and others).

JacobDomagala:
Fixes #2036


tests/perf/collection_local_send.cc
```diff
@@ -81,6 +82,20 @@ struct NodeObj {
...
+
+    theTerm()->disableTD();
+
+    test_obj_->StartTimer(fmt::format("colSend {}", num_iters));
+    perfRunBenchmark();
+    test_obj_->StopTimer(fmt::format("colSend {}", num_iters));
+
+    theTerm()->enableTD();
+  }
+
```
nlslatt, line 98: Can you pull the common code out of `perfMakeRunnablePreAllocate` and `perfMakeRunnable` and put it into a common function or would that adversely impact performance?

nlslatt (commented):
This looks reasonable to me

tests/CMakeLists.txt
```diff
@@ -264,7 +264,7 @@ foreach(PERF_TEST ${PROJECT_PERF_TESTS})
     PROPERTIES
     PROCESSORS ${PERF_TEST_NUMPROC}
     RUN_SERIAL TRUE  # do not run in parallel with any other tests
-    TIMEOUT 300
+    TIMEOUT 500
```
JacobDomagala, line 267: Increased the timeout, as we now run multiple tests inside single test file

tests/perf/collection_local_send.cc
JacobDomagala: Done

tests/perf/collection_local_send.cc
nlslatt, line 1: @JacobDomagala @lifflander I'm trying to understand what the `collection_local_send.cc` perf tests actually do. Is it true that only rank `0` runs the timed perf test, and only does a send to collection element `0`, which is assumed to be local to rank `0` but not verified? It seems like some stuff could be simplified, like (a) consistently using `my_node_` instead of `theContext()->getNode()`, (b) also only pre-allocating on rank `0`, and (c) using `preallocate_` to conditionally pre-allocate based on what happened in `initialiaze` instead of having the separate `perfMakeRunnable` and `perfMakeRunnablePreAllocate` calls.

JacobDomagala: https://github.com/DARMA-tasking/vt/issues/2155 should address those issues

stmcgovern (approved):
Looks good to me

lifflander (approved):

Issue #2100, opened 2023-03-10
Disable unity build for gcc 8 CI that hits OOM


PhilMiller:
**What Needs to be Done?**

Change flags to do separate compilation as usual.

**Is your feature request related to a problem? Please describe.**

**Describe potential solution outcome**

**Describe alternatives you've considered**

**Additional context**

PR #2101, opened 2023-03-10
#2100: Disable unity builds for gcc 7/8

PhilMiller:
Closes: #2100


cz4rs (changes requested):
You need to update the .yml files using `generate-workflow.pl` script:
```
./scripts/generate-workflow.pl ./scripts/azure-workflow-template.yml ./scripts/workflows-azure.ini
```

cz4rs:
Note: we could also drop `gcc-7` build and move its configuration (trace + code coverage) to some other build (for example `clang-12` and `clang-13` are using pure default settings).

cz4rs (approved):
LGTM

PR #2106, opened 2023-03-14
2091: fix `missing return statement` warning

### LINKED ISSUE #2091: runLambda needs to explicitly return void
**Describe the bug**
The ARM compiler emits this warning:

`/.../vt/src/vt/runnable/runnable.h(245): warning #1011: missing return statement at end of non-void function "vt::runnable::RunnableNew::runLambda(Callable &&, Args &&...) [with Callable=bool (NodeObj::*)(), Args=<NodeObj *>]"`

about this code from `runnable/runnable.h`:

```
  /**
   * \brief Run the task as a lambda!
   */
  template <typename Callable, typename... Args>
  decltype(auto) runLambda(Callable&& c, Args&&... args) {
    start();
    if constexpr(std::is_void_v<std::invoke_result_t<Callable, Args...>>) {
      std::invoke(std::forward<Callable>(c), std::forward<Args>(args)...);
      finish();
    } else {
      decltype(auto) r{std::invoke(std::forward<Callable>(c), std::forward<Args>(args)...)};
      finish();
      return r;
    }
  }
```

We probably just need to explicitly `return;` for the case where the lambda returns void.

cz4rs:
fixes #2091

### DIFF
--- a/src/vt/runnable/runnable.h
+++ b/src/vt/runnable/runnable.h
@@ -239,6 +239,7 @@ public:
       std::invoke(std::forward<Callable>(c), std::forward<Args>(args)...);
       auto finish_time = timing::getCurrentTime();
       finish(finish_time);
+      return;
     } else {
       decltype(auto) r{std::invoke(std::forward<Callable>(c), std::forward<Args>(args)...)};
       auto finish_time = timing::getCurrentTime();
--- a/src/vt/runnable/runnable.h
+++ b/src/vt/runnable/runnable.h
@@ -235,17 +235,19 @@ public:
   decltype(auto) runLambda(Callable&& c, Args&&... args) {
     auto start_time = timing::getCurrentTime();
     start(start_time);
-    if constexpr(std::is_void_v<std::invoke_result_t<Callable, Args...>>) {
-      std::invoke(std::forward<Callable>(c), std::forward<Args>(args)...);
-      auto finish_time = timing::getCurrentTime();
-      finish(finish_time);
-      return;
-    } else {
-      decltype(auto) r{std::invoke(std::forward<Callable>(c), std::forward<Args>(args)...)};
-      auto finish_time = timing::getCurrentTime();
-      finish(finish_time);
-      return r;
-    }
+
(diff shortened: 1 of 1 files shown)


PhilMiller:
icpx build was warning-free

PhilMiller:
nvcc 11 still has the warning

PhilMiller:
nvcc 11.2 as well as 11.0 :-(

PhilMiller:
icpc was clean. This might be the best we can do

PhilMiller (commented):
We may have to convert `runLambda` back into a pair of specializations based on `enable_if` if we actually want to silence the warning across all compilers.

src/vt/runnable/runnable.h
```diff
@@ -235,16 +235,19 @@ struct RunnableNew {
...
-      finish(finish_time);
-      return r;
-    }
+
+    // Arrange a scope guard to call finish() without any sort of dynamic allocation
+    struct finisher {
+      RunnableNew* r;
+      finisher(RunnableNew* in_r) : r(in_r){};
+      ~finisher() {
+        auto finish_time = timing::getCurrentTime();
```
cz4rs, line 244: `finish_time` doesn't seem correct here (before `std::invoke` is called), but let's see if this version silences the warning at least.
Based on [Phil's comment](https://github.com/DARMA-tasking/vt/issues/2091#issuecomment-1430193837) in the original issue.

cz4rs: Never mind, `finish_time` value should be correct since this is a destructor :see_no_evil:

lifflander (approved):

cz4rs:
Converting to draft as there are more warnings showing up:
```
/vt/src/vt/objgroup/proxy/proxy_objgroup_elm.impl.h(98)
/vt/src/vt/vrt/collection/send/sendable.impl.h(156)
/vt/src/vt/vrt/collection/broadcast/broadcastable.impl.h(200)
```

PhilMiller (approved):

PR #2108, opened 2023-03-15
2102 collection broadcast loses message data

### LINKED ISSUE #2102: Collection broadcast loses message data due to moves when promoted to CollectionMessage
**Describe the bug**
When a broadcast to a collection occurs using a normal `vt::Message` (not a collection message) and it is serialized, the message is de-allocated preemptively before all the collection elements on a rank receive it (after the first one does). Due to a template bug in `proxy.broadcast` it won't actually accept a `vt::Message` (it looks for a type in `vt::Message` that doesn't exist). So the only way to get it to fail is by using a callback that targets a collection as a broadcast.

**To Reproduce**
1. Run examples/callback/callback.cc

**Expected behavior**

**Screenshots**

 - Compiler environment [e.g., clang-4.0, gcc8]

**Additional context**

lifflander:
Fixes #2102

### DIFF
--- a/src/vt/registry/auto/auto_registry_common.h
+++ b/src/vt/registry/auto/auto_registry_common.h
@@ -58,6 +58,13 @@
 #include <cstdlib>
 #include <functional>
 
+namespace vt::vrt::collection {
+
+template <typename ColT, typename UserMsgT, typename BaseMsgT>
+struct ColMsgWrap;
+
+} /* end namespace vt::vrt::collection */
+
 namespace vt { namespace auto_registry {
 
 struct SentinelObject {};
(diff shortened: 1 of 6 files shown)


nlslatt (approved):
I don't see any obvious problems, but I'd feel better with others reviewing this as well.

thearusable (approved):

src/vt/registry/auto/auto_registry_common.h
```diff
@@ -86,9 +105,23 @@ struct HandlersDispatcher final : BaseHandlersDispatcher {
...
+        fp(elm, msg);
+      } else {
+        auto wrap_msg = static_cast<
+          vrt::collection::ColMsgWrap<ObjT, MsgT, vt::Message>*
+        >(base_msg);
+        fp(elm, &wrap_msg->getMsg());
+      }
     } else if constexpr (std::is_same_v<T, ColMemberTypedFnType>) {
-      (elm->*fp)(msg);
+      if constexpr (IsColMsgTrait<MsgT>::value or not IsColTrait<ObjT>::value) {
```
PhilMiller, line 117: I have a slight preference for reversing the sense of this test, 
`if (collection and not collection message) { wrapped case } else { simple case }`

I think it would be somewhat more comprehensible. It says when we need to do something special, as opposed to when we can do without.

PhilMiller (approved):
Looks great to see a bug fixed by simplifying the logic.

PhilMiller:
If you're going to edit the condition, maybe squash the commits, too

PhilMiller (approved):

PR #2110, opened 2023-03-21
#2107: CI: remove builds for unsupported compilers

### LINKED ISSUE #2107: remove `gcc-7` build and reproduce its configurations on different build
**What Needs to be Done?**
We have raised the minimum version requirements for both `gcc` and `clang`.

- [x] Remove obsolete builds
- [x] Remove obsolete badges from README
- [ ] Make sure that CI and [build requirements on wiki](https://github.com/DARMA-tasking/vt/wiki/Build-Requirements#supported-compilers) are matching
- [x] Move the configurations used in the removed builds to some other builds (tracing + coverage, address sanitizer)

cz4rs:
- remove build for `gcc-7` (no longer supported)
- move its configuration to `gcc-11` build
- move LB schema test to `gcc-9` build (to avoid having too many options on `gcc-11`)

fixes #2107


lifflander (approved):

.github/workflows/macosx-clang-mpich.yml
```diff
@@ -1,4 +1,4 @@
-name: PR tests (clang-8, macosx, mpich)
+name: PR tests (clang-14, macosx, mpich)
```
PhilMiller, line 1: The cache name and maybe other stuff still refer to clang 8.  I don't see what in this file actually set the compiler version, though.

cz4rs: Updated the version number everywhere.

Compiler version is not set explicitly, we use the default for that image: https://github.com/actions/runner-images/blob/macOS-12/20230318.1/images/macos/macos-12-Readme.md.

.github/workflows/build-docs.yml
```diff
@@ -16,7 +16,7 @@ jobs:
       ARCH: amd64
       UBUNTU: 18.04
       COMPILER_TYPE: gnu
-      COMPILER: gcc-7
+      COMPILER: gcc-8
```
cz4rs, line 19: Switching to the newest version (Ubuntu 22.04 + gcc-12) causes error in the docs generation. I have not investigated it and opted for a minimal bump instead.

This seems to be working fine:
https://github.com/DARMA-tasking/vt/actions/runs/4563477715/jobs/8052013152

.github/workflows/macosx-clang-mpich.yml
```diff
@@ -15,7 +15,7 @@ concurrency:
 
 jobs:
   build:
-    runs-on: macos-latest
+    runs-on: macos-12
```
cz4rs, line 18: Changing `latest` to a pinned version (this was considered preferable for Intel compilers, we can probably use the same approach here).

cz4rs: AFAIK this should be enough to make the compiler version pinned as well, so we don't have to configure it explicitly elsewhere.

PhilMiller (approved):

PR #2116, opened 2023-03-27
#2115: Update CMake min required version to 3.23

### LINKED ISSUE #2115: Update CMake min required version
Trilinos has recently raised CMake minimum required version to `3.23` (see https://github.com/trilinos/Trilinos/commit/35ca59d5cf91b60ac0f110e0928203bb9754281d). This breaks our Docker image for `gcc-9` build ([when building Zoltan](https://dev.azure.com/DARMA-tasking/DARMA/_build/results?buildId=48886&view=logs&j=3dc8fd7e-4368-5a92-293e-d53cefc8c4b3&t=1364e02f-4e1e-51cf-2821-fd252a3c0e34)).

**What Needs to be Done?**
- [ ] use CMake `3.23` (or newer) in our CI
- [ ] consider raising the requirement in vt's CMake code (we currently require `3.17`)

cz4rs:
fixes #2115

### DIFF
--- a/scripts/check_containers.sh
+++ b/scripts/check_containers.sh
@@ -1,11 +1,14 @@
 #!/usr/bin/env bash
-# Check if there were any dockerfiles modified in the latest pull request.
-# Note: this is specific to the way Azure checks out git repo.
+# Check if there were any dockerfiles modified in the PR.
+# Note: this is specific to Azure - in the pipeline's local branch git history
+# contains one merge commit for all of the changes in the PR.
 
 diff_latest() {
-  git diff --name-only "$(git log --skip=1 -1 --merges --pretty=format:%H)"..
+  printf "Files modified in the pull request:\n" >&2
+  git diff --name-only HEAD^1 HEAD | tee >(cat >&2)
 }
 
+# Decide which docker-compose command to use:
 if diff_latest | grep -i dockerfile > /dev/null
 then
     echo "build --pull"
--- a/scripts/workflow-azure-template.yml
+++ b/scripts/workflow-azure-template.yml
@@ -62,6 +62,8 @@ stages:
       [% workflow_runs_on %]
     timeoutInMinutes: 180
     steps:
+      - checkout: self
+        fetchDepth: 0
       - task: Bash@3
         displayName: Build timestamp for caching
         continueOnError: true
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.17 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.23 FATAL_ERROR)
 
 file(READ "VERSION" _vt_version_str)
 string(STRIP "${_vt_version_str}" _vt_version_str)
--- a/ci/azure/azure-clang-10-ubuntu-mpich.yml
+++ b/ci/azure/azure-clang-10-ubuntu-mpich.yml
@@ -74,6 +74,8 @@ stages:
       vmImage: 'ubuntu-22.04'
     timeoutInMinutes: 180
     steps:
+      - checkout: self
+        fetchDepth: 0
       - task: Bash@3
         displayName: Build timestamp for caching
         continueOnError: true
--- a/ci/azure/azure-clang-11-ubuntu-mpich.yml
+++ b/ci/azure/azure-clang-11-ubuntu-mpich.yml
@@ -74,6 +74,8 @@ stages:
       vmImage: 'ubuntu-22.04'
     timeoutInMinutes: 180
     steps:
+      - checkout: self
+        fetchDepth: 0
       - task: Bash@3
         displayName: Build timestamp for caching
         continueOnError: true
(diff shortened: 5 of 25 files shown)


PhilMiller (approved):

thearusable (approved):

JacobDomagala (approved):

PhilMiller:
Should this PR and the issue be re-titled to reflect that it's not raising the minimum requirement, just what's provided in CI to match an optional library?

lifflander (approved):

cz4rs:
> Somehow the nvcc 11.2 build failed with the old CMake version still being used.

I have improved reliability of `check_containers` script, which triggers Docker images rebuilds. Now with `detector` being integrated into `magistrate`, this has to wait for https://github.com/DARMA-tasking/vt/pull/2120 to be merged:
```
In file included from /build/detector/install/include/detector_headers.h:49:
/build/detector/install/include/none_such.h:57:8: error: redefinition of '\''NoneSuch'\''
struct NoneSuch final {
       ^
/build/checkpoint/install/include/checkpoint/detector.h:55:8: note: previous definition is here
struct NoneSuch final {
       ^
```

PhilMiller:
(rebased on the wrong computer - I'll re-apply the signatures later if no one else gets to it first.)

cz4rs:
> (rebased on the wrong computer - I'll re-apply the signatures later if no one else gets to it first.)

Done.

PhilMiller:
Looks like the container updating script still has some sort of issue that's breaking the nvcc build

scripts/check_containers.sh
```diff
@@ -1,11 +1,14 @@
...
-# Check if there were any dockerfiles modified in the latest pull request.
-# Note: this is specific to the way Azure checks out git repo.
+# Check if there were any dockerfiles modified in the PR.
+# Note: this is specific to Azure - in the pipeline's local branch git history
+# contains one merge commit for all of the changes in the PR.
 
 diff_latest() {
-  git diff --name-only "$(git log --skip=1 -1 --merges --pretty=format:%H)"..
+  printf "Files modified in the pull request:\n" >&2
+  git diff --name-only HEAD~ HEAD | tee >(cat >&2)
```
PhilMiller, line 8:
```suggestion
  git diff --name-only HEAD^1 HEAD | tee >(cat >&2)
```
Maybe that would work better, if HEAD is a merge commit. It will give the first parent (i.e. the tip of the target branch), rather than a unique parent.

cz4rs: Yeah, I was scratching my head thinking what could be the issue. I will give it a try :+1:

PhilMiller:
I wish git's error were a bit more explicit, so I didn't have to guess
among the possibilities it lists, but ambiguity is definitely more likely
than HEAD having no parent at all.

PhilMiller:
The `fetch` and `checkout` command in the 11.2 case looks really wrong. A commit ID shouldn't be getting stuck in the ref space. Is the script expanding some variable with that appended?

cz4rs:
> The `fetch` and `checkout` command in the 11.2 case looks really wrong. A commit ID shouldn't be getting stuck in the ref space. Is the script expanding some variable with that appended?

I think the key difference is `--depth=1` here.
After checking those commands locally it results in a single commit checkout:
```
~  azure-test  git log --oneline 
0bb2e65 (grafted, HEAD, origin/0bb2e65940c5dd3c782aefa625376ac8a95de134) Merge e550caa377616309e4ea1270c6e75f62bc04024e into 6822a099b2bfc52153cc30a2f119acd2816f27ca
```

cz4rs:
No idea why Azure uses 2 different ways for checkout, but for now I think we'll just have to support them both in `check_containers` script.

PhilMiller:
Is there some checkout action we're using with an un-/under-specified version?

cz4rs:
> Oh, indeed, with `--depth=1` it's impossible to get anything about the parent commit. I still think the appearance of commit IDs in expressions like `refs/remotes/origin/0bb2e65940c5dd3c782aefa625376ac8a95de134` is wrong too, though.

Ugh, I've just realized that this single commit contains all of repo's code:
```
git show --stat HEAD
(...)
 1612 files changed, 349090 insertions(+)
```
so the information on PR's changes cannot be obtained from it.

> Is there some checkout action we're using with an un-/under-specified version?
Azure checks out parent repo by default, so we don't specify anything in our `.yml`s.
We could use `checkout: self` and [other parameters](https://learn.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/steps-checkout?view=azure-pipelines#properties) explicitly.

cz4rs:
Alright, I think I found the culprit:
image

scripts/workflow-azure-template.yml
```diff
@@ -62,6 +62,9 @@ stages:
       [% workflow_runs_on %]
     timeoutInMinutes: 180
     steps:
+      steps:
+      - checkout: self
+        fetchDepth: 0
```
cz4rs, line 67: This disables shallow fetch in Azure workflows. Required for `check_containers` script to work correctly.

cz4rs: Note: maybe we could also make it work by setting depth to 2 :shrug:

PhilMiller: Even `fetchDepth: 2` should work fine for our use case. The issue is that a fetch depth of 1 means that the commit we're building won't have even its parents available to compare against.

PhilMiller: That said, if this works, let's leave it for now, and revisit if we ever care to optimize that point. I don't think we will.

PR #2118, opened 2023-03-27
#2117: location: add new assertions to detect duplicate insertions

### LINKED ISSUE #2117: Check for duplication elements during collection insertion
**What Needs to be Done?**

Add a layer of debugging for this case

lifflander:
Fixes #2117


src/vt/vrt/collection/collection_builder.impl.h
```diff
@@ -131,7 +131,8 @@ void CollectionManager::makeCollectionImpl(param::ConstructParams<ColT>& po) {
 
   // Invoke getCollectionLM() to create a new location manager instance for
   // this collection
-  theLocMan()->getCollectionLM<IndexType>(proxy);
```
PhilMiller, line 134: Kinda ugly that a method named `getFoo` has a side effect of creating said Foo

PR #2120, opened 2023-03-28
#2046: build: remove detector as dependency

### LINKED ISSUE #2046: Replace `detector` library with C++17
**What Needs to be Done?**
Replace, as much as possible, use of [detector](https://github.com/DARMA-tasking/detector) with C++17 constructs.

If it is possible to get rid of it completely, remember to remove it from build scripts, CMake and documentation, and archive the repository as well.

lifflander:
Fixes #2046


PhilMiller (approved):

PhilMiller:
The branch format check that failed doesn't like the `++` in the branch name, I don't think. If that's all, we can just override it.

cz4rs (approved):
Looks good to me.

cz4rs:
`gcc-10` build now fails because of `vt-sample-project`:
```
CMake Error at build_with_tpl.cmake:13 (add_subdirectory):
  add_subdirectory given source "TPL/detector" which is not an existing
  directory.
  CMakeLists.txt:11 (include)


CMake Error at build_with_tpl.cmake:14 (add_subdirectory):
  The source directory

    /build/vt/vt-sample-project/TPL/checkpoint

  does not contain a CMakeLists.txt file.
  CMakeLists.txt:11 (include)
```

lifflander:
Yes, this is going to fail until we update vt-sample-project.

ci/build_vt_sample.sh
```diff
@@ -20,9 +20,8 @@ if test "$is_alpine" -eq 0 && test "${VT_CI_BUILD:-0}" -eq 1
 then
     export VT=${source_dir}
     export VT_BUILD=${build_dir}/vt
-    export VT_INSTALL=${VT_BUILD}/install
-    export DETECTOR=${build_dir}/detector
-    export CHECKPOINT=${DETECTOR}/build/checkpoint
+    export VT_INSTALL=${VT_BUILD}/install/cmake
+    export CHECKPOINT=${build_dir}/checkpoint
```
PhilMiller, line 24: As a separate matter, we should probably move away from this script seemingly getting checkpoint from the build directory. Either it should be installed, or it should be nested in vt.

PR #2123, opened 2023-03-29
#2119: cmake: install target for checkpoint when included in lib

### LINKED ISSUE #2119: vt doesn't install checkpoint and detector if they are added as subdirectories to vt
**Describe the bug**

Reproduced by @lifflander . If VT includes checkpoint and detector as subdirectories, it doesn't install them properly and downstream code will fail to find them.

lifflander:
Fixes #2119

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -340,3 +340,10 @@ export(
   FILE                      "vtTargets.cmake"
   NAMESPACE                 vt::runtime::
 )
+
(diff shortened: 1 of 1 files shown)


PhilMiller:
This looks sensible, but my knowledge of CMake is too limited to be confident that it's right.

stmcgovern:
This works for me. Thanks @lifflander

stmcgovern (approved):

PR #2129, opened 2023-04-08
#2122: Silence `icpc` deprecation warning

### LINKED ISSUE #2122: Add `-diag-disable=10441` to `CMAKE_CXX_FLAGS` for icpc build
**What Needs to be Done?**

On the `icpc` CI pipeline builds, we should add `-diag-disable=10441` to `CMAKE_CXX_FLAGS` so that we're not bombarded with warnings about the known EOL status of that compiler.

We currently get this message for every file:
> `intel-cc: remark #10441: The Intel(R) C++ Compiler Classic (ICC) is deprecated and will be removed from product release in the second half of 2023. The Intel(R) oneAPI DPC++/C++ Compiler (ICX) is the recommended compiler moving forward. Please transition to use this compiler. Use '-diag-disable=10441' to disable this message.`

JacobDomagala:
Fixes #2122


cmake/load_bundled_libraries.cmake
```diff
@@ -41,9 +41,12 @@ set(BROTLI_DISABLE_TESTS ON)
 # we need to disable bundled mode so it will install properly
 set(BROTLI_BUNDLED_MODE OFF)
 set(BROTLI_BUILD_PORTABLE ON)
-set(BROTLI_LIBRARY brotlicommon-static brotlienc-static brotlidec-static)
+set(BROTLI_BUILD_SHARED_LIBS OFF)
```
JacobDomagala, line 44: `vt` is only using static libs. This shouldn't affect anyone, I think.

cmake-modules/SetCXXCompilerFlags.cmake
```diff
@@ -11,6 +11,12 @@ function(set_darma_compiler_flags vt_target)
     endif ()
   elseif (CMAKE_CXX_COMPILER_ID STREQUAL IntelLLVM AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 2021.3.0)
     list(APPEND TARGET_PRIVATE_CXX_FLAGS -fhonor-infinites -fhonor-nans)
+  elseif (${CMAKE_CXX_COMPILER_ID} MATCHES "Intel")
```
PhilMiller, line 14: Should the be `MATCHES` or `STREQUAL`? Will we end up adding this flag for `icpx`?

lifflander: Can you resolve this @JacobDomagala?

lifflander (approved):
Approved once Phil's comment is addressed.

JacobDomagala:
> Why are the Brotli build changes part of this PR?

Brotli builds static and shared libs and both of them generate `icpc` deprecation warnings. To fix that I disabled generation of shared libs (we don't need them since `vt` is only using static ones).

cmake-modules/SetCXXCompilerFlags.cmake
JacobDomagala: Changed to `STREQUAL`

nlslatt (approved):

PR #2136, opened 2023-04-18
276 Callback parameterization

### LINKED ISSUE #276: Automatically synthesize message types for sends to arbitrary handlers

lifflander:
Fixes #276

This replaces the current proxy.reduce syntax with three calls that are symmetrical across object groups and collections.

```C++
proxy.allreduce<&MyCol::handler, collective::PlusOp>(value1, value2);
proxy.reduce<&MyCol::handler, collective::PlusOp>(proxy[1], value1, value2);
proxy.reduce<collective::PlusOp>(some_callback, value1, value2);
```

### DIFF
--- a/src/vt/collective/reduce/get_reduce_stamp.h
+++ b/src/vt/collective/reduce/get_reduce_stamp.h
@@ -48,6 +48,7 @@
 #include "vt/collective/reduce/reduce_scope.h"
 #include "vt/messaging/message.h"
 
+#include <tuple>
 #include <utility>
 #include <type_traits>
 
--- a/src/vt/collective/reduce/get_reduce_stamp.h
+++ b/src/vt/collective/reduce/get_reduce_stamp.h
@@ -2,7 +2,7 @@
 //@HEADER
 // *****************************************************************************
 //
-//                             epoch_graph_reduce.h
+//                              get_reduce_stamp.h
 //                       DARMA/vt => Virtual Transport
 //
 // Copyright 2019-2021 National Technology & Engineering Solutions of Sandia, LLC
@@ -41,39 +41,59 @@
 //@HEADER
 */
 
-#if !defined INCLUDED_VT_TERMINATION_GRAPH_EPOCH_GRAPH_REDUCE_H
-#define INCLUDED_VT_TERMINATION_GRAPH_EPOCH_GRAPH_REDUCE_H
+#if !defined INCLUDED_VT_COLLECTIVE_REDUCE_GET_REDUCE_STAMP_H
+#define INCLUDED_VT_COLLECTIVE_REDUCE_GET_REDUCE_STAMP_H
 
 #include "vt/config.h"
 
-namespace vt { namespace collective { namespace reduce { namespace operators {
+namespace vt { namespace collective { namespace reduce {
 
-template <typename T>
-struct ReduceDataMsg;
-
-}}}} /* end namespace vt::collective::reduce::operators */
-
-namespace vt { namespace termination { namespace graph {
-
-// Must be templated (where T = graph::EpochGraph) because of the circular
-// dependency between termination.h and reduce.h
-template <typename T>
-struct EpochGraphMsg : collective::reduce::operators::ReduceDataMsg<T> {
-  using MessageParentType = collective::reduce::operators::ReduceDataMsg<T>;
-  vt_msg_serialize_if_needed_by_parent();
-
-  EpochGraphMsg() = default;
-
-  explicit EpochGraphMsg(std::shared_ptr<T> const& graph)
-    : collective::reduce::operators::ReduceDataMsg<T>(*graph)
-  { }
-
-  template <typename SerializerT>
-  void serialize(SerializerT& s) {
-    MessageParentType::serialize(s);
+template <typename enable = void, typename... Args>
+struct GetReduceStamp : std::false_type {
+  template <typename MsgT>
+  static auto getMsg(Args&&... args) {
+    return vt::makeMessage<MsgT>(std::tuple{std::forward<Args>(args)...});
   }
 };
 
-}}} /* end namespace vt::termination::graph */
+template <>
+struct GetReduceStamp<
+  std::enable_if_t<std::is_same_v<void, void>>
+> : std::false_type {
+  template <typename MsgT>
+  static auto getMsg() {
+    return vt::makeMessage<MsgT>(std::tuple<>{});
+  }
+};
 
-#endif /*INCLUDED_VT_TERMINATION_GRAPH_EPOCH_GRAPH_REDUCE_H*/
+template <typename... Args>
+struct GetReduceStamp<
+  std::enable_if_t<
+    std::is_same_v<
+      std::decay_t<std::tuple_element_t<sizeof...(Args) - 1, std::tuple<Args...>>>,
+      collective::reduce::ReduceStamp
+    >
+  >,
+  Args...
+> : std::true_type {
+  template <typename... Params, std::size_t... Is>
+  static constexpr auto getMsgHelper(
+    std::tuple<Params...> tp, std::index_sequence<Is...>
+  ) {
+    return std::tuple{std::get<Is>(tp)...};
+  }
+
+  template <typename MsgT>
+  static auto getMsg(Args&&... args) {
+    return vt::makeMessage<MsgT>(
+      getMsgHelper(
+        std::tie(std::forward<Args>(args)...),
+        std::make_index_sequence<sizeof...(Args) - 1>{}
+      )
+    );
+  }
+};
+
+}}} /* end namespace vt::collective::reduce */
+
+#endif /*INCLUDED_VT_COLLECTIVE_REDUCE_GET_REDUCE_STAMP_H*/
--- a/src/vt/collective/reduce/get_reduce_stamp.h
+++ b/src/vt/collective/reduce/get_reduce_stamp.h
@@ -57,8 +57,12 @@ namespace vt { namespace collective { namespace reduce {
 template <typename enable = void, typename... Args>
 struct GetReduceStamp : std::false_type {
   template <typename MsgT>
-  static auto getMsg(Args&&... args) {
-    return vt::makeMessage<MsgT>(std::tuple{std::forward<Args>(args)...});
+  static auto getStampMsg(Args&&... args) {
+    return
+      std::make_tuple(
+	collective::reduce::ReduceStamp{},
+	vt::makeMessage<MsgT>(std::tuple{std::forward<Args>(args)...})
+      );
   }
 };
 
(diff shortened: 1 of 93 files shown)


tests/unit/pipe/test_callback_send_collection.extended.cc
nlslatt, line 1: I think you can delete `testHandler`, `CallbackDataMsg`, `testHandlerEmpty`, and `CallbackMsg`.

nlslatt (commented):
Here's my feedback so far. I think it's important that others review this as well.

src/vt/collective/reduce/reduce.h
```diff
@@ -168,6 +168,41 @@ struct Reduce : virtual collective::tree::Tree {
...
+    return reduce<OpT, f>(root, std::forward<Params>(params)...);
+  }
+
+  template <typename Op, auto f, typename... Params>
+  PendingSendType reduce(Node root, Params&&... params) {
+    using Tuple = typename FunctionTraitsArgs<decltype(f)>::TupleType;
+    using MsgT = ReduceTMsg<Tuple>;
+
+    auto msg = vt::makeMessage<MsgT>(std::tuple{std::forward<Params>(params)...});
+    auto id = detail::ReduceStamp{};
```
PhilMiller: Does this imply that there won't be a way to stamp a call to the parameter-synthesized version of reduce?

src/vt/vrt/collection/balance/lb_data_restart_reader.h
```diff
@@ -153,9 +153,9 @@ struct LBDataRestartReader : runtime::component::Component<LBDataRestartReader>
    * \brief Reduce distribution changes globally to find where migrations need
    * to occur
    *
-   * \param[in] msg the reduction message
+   * \param[in] vec the vector of booleans
    */
-  void reduceDistroChanges(ReduceMsg* msg);
+  void reduceDistroChanges(std::vector<bool> vec);
```
PhilMiller: `std::vector<bool>` is generally something to be avoided. I'm assuming it was already in use as a member of `ReduceMsg` here?

src/vt/collective/reduce/operators/default_op.impl.h
```diff
@@ -49,28 +49,46 @@
...
-    if (cb.valid()) {
+    if (msg->hasValidCallback()) {
       envelopeUnlockForForwarding(msg->env);
-      cb.template send<MsgT>(msg);
+      if (msg->isParamCallback()) {
+        if constexpr (IsTuple<typename MsgT::DataT>::value) {
+          msg->getParamCallback().sendTuple(std::move(msg->getVal()));
+        }
+      } else {
+        // We need to force the type the the more specific one here
```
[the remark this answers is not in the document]
PhilMiller:
```suggestion
        // We need to force the type to the more specific one here
```

PhilMiller:
I haven't reviewed this in full detail, but it generally looks good. My comments above are marked 'Outdated', but they still apply to the code as it appears now.

tutorial/tutorial_2b.h
```diff
@@ -66,41 +67,25 @@ struct ReduceCol : ::vt::Collection<ReduceCol,Index1D> {
...
+  int val = 100;
 
   // Invoke the reduce!
-  proxy.reduce<ReduceOp,PrintReduceResult>(reduce_msg.get());
+  proxy.allreduce<&ReduceCol::reduceTarget>(val);
+}
+
+void ReduceCol::reduceTarget(int value) {
+  fmt::print("collection reduce value={}\n", value);
+  assert(32 * 100 == value);
```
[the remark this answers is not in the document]
lifflander: Fixed

src/vt/collective/reduce/operators/default_op.impl.h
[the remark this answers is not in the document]
lifflander: Fixed

src/vt/vrt/collection/balance/greedylb/greedylb.h
```diff
@@ -103,7 +103,7 @@ struct GreedyLB : LoadSamplerBaseLB {
   void recvObjs(GreedySendMsg* msg);
   void recvObjsBcast(GreedyBcastMsg* msg);
   void finishedTransferExchange();
-  void collectHandler(GreedyCollectMsg* msg);
+  void collectHandler(GreedyPayload payload);
```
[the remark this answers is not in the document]
lifflander: Removed

src/vt/vrt/collection/balance/lb_data_restart_reader.h
lifflander: Yes, this already existed like this so I don't think there's a reason to change it here. Another PR is refactoring this code to generally improve it.

tests/unit/collection/test_collection_group_recreate.cc
```diff
@@ -49,26 +49,25 @@
...
+  void doReduce() {
+    auto const proxy = getCollectionProxy();
+    int val = getIndex().x();
+    proxy.allreduce<&MyCol::checkVal, collective::PlusOp>(val);
+  }
 
-  void doWork(TestMsg2* msg) {}
+  void checkVal(int val) {
+    cb_counter++;
+    fmt::print("at root: final num={} cb_counter={}\n", val, cb_counter);
```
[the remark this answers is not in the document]
lifflander: Fixed

tests/unit/pipe/test_callback_send_collection.extended.cc
lifflander: Removed

src/vt/collective/reduce/reduce.h
lifflander: I have fixed this limitation and cleaned up the code.

src/vt/utils/fntraits/fntraits.h
```diff
@@ -141,60 +162,227 @@ struct ObjFuncTraitsImpl<
...
+struct ObjFuncTraitsImpl<
   std::enable_if_t<
     not (
       std::is_convertible<Arg, vt::Message*>::value or
       std::is_convertible<Arg, vt::ShortMessage*>::value or
       std::is_convertible<Arg, vt::EpochMessage*>::value or
       std::is_convertible<Arg, vt::PayloadMessage*>::value
-    )
+    ) and
+    not std::is_pointer<Arg>::value
```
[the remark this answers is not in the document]
lifflander: Fixed

src/vt/collective/reduce/get_reduce_stamp.h
```diff
@@ -0,0 +1,116 @@
...
+
+namespace vt { namespace collective { namespace reduce {
+
+template <typename enable = void, typename... Args>
+struct GetReduceStamp : std::false_type {
+  template <typename MsgT>
+  static auto getStampMsg(Args&&... args) {
+    return
+      std::make_tuple(
+	collective::reduce::ReduceStamp{},
```
PhilMiller, line 63: Does the default constructor internally do something that gives back a unique value?

src/vt/collective/reduce/operators/default_op.impl.h
```diff
@@ -49,28 +49,46 @@
...
-    if (cb.valid()) {
+    if (msg->hasValidCallback()) {
       envelopeUnlockForForwarding(msg->env);
-      cb.template send<MsgT>(msg);
+      if (msg->isParamCallback()) {
+        if constexpr (IsTuple<typename MsgT::DataT>::value) {
+          msg->getParamCallback().sendTuple(std::move(msg->getVal()));
+        }
+      } else {
+        // We need to force the type the more specific one here
```
PhilMiller, line 74:
```suggestion
        // We need to force the type to the more specific one here
```

src/vt/vrt/collection/balance/baselb/baselb.h
```diff
@@ -134,7 +134,7 @@ struct BaseLB {
   void migrateObjectTo(ObjIDType const obj_id, NodeType const node);
   void transferSend(NodeType from, TransferVecType const& transfer);
   void transferMigrations(TransferMsg<TransferVecType>* msg);
-  void finalize(CountMsg* msg);
```
PhilMiller, line 137: Ditto Nicole's comment on GreedyLB - do we still need `CountMsg`, or can its definition be deleted, too?

tutorial/tutorial_2b.h
```diff
@@ -66,41 +67,27 @@ struct ReduceCol : ::vt::Collection<ReduceCol,Index1D> {
...
   auto proxy = getCollectionProxy();
-  auto reduce_msg = makeMessage<ReduceMsg>();
-
-  // Get a reference to the value to set it in this reduce msg
-  reduce_msg->getVal() = 100;
+  int val = 100;
 
   // Invoke the reduce!
-  proxy.reduce<ReduceOp,PrintReduceResult>(reduce_msg.get());
+  proxy.allreduce<&ReduceCol::reduceTarget>(val);
```
PhilMiller, line 82: Is addition now an implicit default operation if none is specified? That's pretty conventional, but I'm not really a fan of it being a special case where it's not spelled out.

src/vt/collective/reduce/get_reduce_stamp.h
lifflander: No, the constructor generates a default constructed object which is basically an empty stamp. An empty stamp tells the reduction manager to generate the next in the sequence for a given reduction scope. Scopes are the collection you are running in, the object group, etc.

src/vt/collective/reduce/operators/default_op.impl.h
lifflander: Fixed

tutorial/tutorial_2b.h
lifflander: This is indeed a bug. I have added a static_assert to stop this from compiling.

stmcgovern (approved):

PR #2137, opened 2023-04-25
#2130: Add spec for skipped and identical phases

### LINKED ISSUE #2130: Add specification for LB data file for phases skipped or identical
**What Needs to be Done?**

We want to specify for every phase in the data file whether data is skipped or identical. This is related to issue https://github.com/DARMA-tasking/vt/issues/1935 which will allow phase output to be skipped for some outputs. The identical to previous is for data input where the mapping is the same as a previous phase and thus does not need to be output again.

```json
{
	"metadata": {
		"rank": 0,
		"shared_node": {
			"id": 0,
			"num_nodes": 1,
			"rank": 0,
			"size": 4,
			"available_memory": 32768
		},
		"type": "LBDatafile"
                "phases": {
                  "count": 100,
                  "skipped": {
                    "list": [10, 13, 15],
                    "range": [[20, 25], [30, 35]]
                  }
                  "identical_to_previous": {
                    "list": [50, 53, 58],
                    "range": [[60, 70]]
                  }
                }
	},
	"phases": [

	]
}
```

*Edited to `count`

thearusable:
Closes #2130

### DIFF
--- a/tests/unit/collection/test_lb_data_holder.cc
+++ b/tests/unit/collection/test_lb_data_holder.cc
@@ -1,0 +1,196 @@
+/*
+//@HEADER
+// *****************************************************************************
+//
+//                            test_lb_data_holder.cc
+//                       DARMA/vt => Virtual Transport
+//
+// Copyright 2019-2021 National Technology & Engineering Solutions of Sandia, LLC
+// (NTESS). Under the terms of Contract DE-NA0003525 with NTESS, the U.S.
+// Government retains certain rights in this software.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+// * Redistributions of source code must retain the above copyright notice,
+//   this list of conditions and the following disclaimer.
+//
+// * Redistributions in binary form must reproduce the above copyright notice,
+//   this list of conditions and the following disclaimer in the documentation
+//   and/or other materials provided with the distribution.
+//
+// * Neither the name of the copyright holder nor the names of its
+//   contributors may be used to endorse or promote products derived from this
+//   software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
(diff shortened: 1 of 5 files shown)


lifflander (approved):
I think this looks good. The next step is to use this data to determine for LBDataRestartReader when a phase is identical or skipped and use to to orchestrate the object distribution.

tests/unit/collection/test_lb_data_holder.cc
```diff
@@ -0,0 +1,196 @@
...
+
+struct TestOptionalPhasesMetadata : TestParallelHarnessParam<bool> { };
+
+void addPhasesDataToJson(nlohmann::json& json, PhaseType amountOfPhasesToAdd, std::vector<PhaseType> toSkip) {
+  using ElementIDStruct = vt::vrt::collection::balance::ElementIDStruct;
+  using LBDataHolder = vt::vrt::collection::balance::LBDataHolder;
+  using LoadSummary = vt::vrt::collection::balance::LoadSummary;
+
+  std::unordered_map<PhaseType, std::vector<ElementIDStruct>> ids;
+  for (PhaseType i = 0; i < 2; i++) {
```
nlslatt, line 69: Why are you using `PhaseType` here?

tests/unit/collection/test_lb_data_holder.cc
```diff
@@ -0,0 +1,196 @@
...
+TEST_F(TestOptionalPhasesMetadata, test_no_lb_phases_metadata) {
+  using LBDataHolder = vt::vrt::collection::balance::LBDataHolder;
+
+  nlohmann::json json;
+  json["metadata"]["type"] = "LBDatafile";
+
+  addPhasesDataToJson(json, num_phases, {});
+
+  LBDataHolder testObj(json);
+  ASSERT_EQ(testObj.count_, 0);
```
nlslatt, line 102: Why is this `0` instead of `num_phases`? `addPhasesDataToJson()` should be adding all 10, right?

tests/unit/collection/test_lb_data_holder.cc
```diff
@@ -0,0 +1,196 @@
...
+  std::set<PhaseType> expectedSkipped = {};
+  ASSERT_EQ(testObj.skipped_phases_, expectedSkipped);
+  std::set<PhaseType> expectedIdentical = {};
+  ASSERT_EQ(testObj.identical_phases_, expectedIdentical);
+
+  auto outJsonPtr = testObj.metadataToJson();
+  ASSERT_TRUE(outJsonPtr != nullptr);
+
+  ASSERT_EQ((*outJsonPtr)["count"], 0);
```
nlslatt, line 111: Why is this `0` instead of `num_phases`? If there were 10 before and nothing is flagged as skipped or identical, where did those phases go?

nlslatt (changes requested):
There are test failures that need to be resolved.

tests/unit/collection/test_lb_data_holder.cc
thearusable: Yes. I pushed a change in code but forgot to push an update in tests.

tests/unit/collection/test_lb_data_holder.cc
thearusable: It's not really needed, I've changed to `unsigned`

nlslatt (approved):
Should all the checks in the tests really be `ASSERT_*` or should some be `EXPECT_*`? I recommend only using `ASSERT_*` when the checks that follow are no longer valid (and might segfault or do something else bad). Using `EXPECT_*` gives the opportunity to see what independent checks are failing later on in the test.

PR #2139, opened 2023-04-26
#2099: Make `NodeType` a strong type and use it across the codebase

### LINKED ISSUE #2099: Use strong Node type across code base to prevent confusion
This should maybe be in a lower-level shared header, so that we're more able to use it across the codebase.

_Originally posted by @PhilMiller in https://github.com/DARMA-tasking/vt/pull/2064#discussion_r1128338848_

JacobDomagala:
Fixes #2099


JacobDomagala:
vt-sample has to be updated

lifflander:
> 

Yes, I do think we should support arithmetic operations.

JacobDomagala:
We probably would want to come up with a way to not break `vt` dependent code that uses old `NodeType`

PR #2146, opened 2023-05-09
2134 typo removal

### LINKED ISSUE #2134: Typo Removal
**What Needs to be Done?**
Remove Typos

nlslatt:
amdarr was kind enough to fix some of our typos for us. I've split up some of his commits that had substantive changes, gave them more appropriate commit messages, signed them, and pulled out a small amount of stuff that probably shouldn't get changed. Most of the changes are in comments, but there are some renames of classes and files.

Fixes #2134


tests/unit/collection/test_collection_construct_common.h
```diff
@@ -74,7 +74,7 @@ struct ConstructHandlers {
   static void handler(CollectionT* col, MessageT* msg) {
     // fmt::print(
     //   "{}: constructed TestCol: idx.x()={}\n",
-    //   theContext()->getNode(), col->getInex().x(), print_ptr(col)
+    //   theContext()->getNode(), col->getIndex().x(), print_ptr(col)
```
nmm0, line 77: Dead code should we just remove?

tests/unit/termination/test_term_cleanup.cc
```diff
@@ -58,7 +58,7 @@ struct TestTermCleanup :  TestParallelHarness {
   using TestMsgType = TestStaticBytesNormalMsg<64>;
 
   static void handler(TestMsgType* msg) {
-    //fmt::print("recieve msg\n");
+    //fmt::print("receive msg\n");
```
nmm0, line 61: Should we just delete this line as it is dead code?

tests/unit/collection/test_collection_construct_common.h
nlslatt: Looks like it's probably there for debugging purposes. Remove it anyway?

PhilMiller (approved):

tests/unit/collection/test_collection_construct_common.h
PhilMiller: Let's consider dead code removal separately. Everything here is a clear improvement right now.

nmm0: Hmm... I guess it's in a test, but I still don't like having commented out code like this because there is no guarantee it remains relevant or correct. If we want to add it back in at some point we can look at the git history. But if people think we should keep it I'm pretty flexible on this.

PR #2148, opened 2023-05-15
#2147: fix incorrect order of parameters for call to beginProcessing

### LINKED ISSUE #2147: invalid trace timings
**Describe the bug**
Projections will complain about corrupted or invalid trace files, or that the time of a handler cannot fit into an integer. Occasionally, I see large negative numbers in the latency of a handler in the trace.

This makes me think that there may be some uninitialized values in the trace.

This bug seems to be introduced by #1998 as I have bisected and versions of VT from before that was merged work correctly.

nmm0:
Closes #2147

I'm going to see if it is possible to add some tests, since we don't have any trace tests.

### DIFF
--- a/src/vt/trace/trace.cc
+++ b/src/vt/trace/trace.cc
@@ -359,7 +359,7 @@ TraceProcessingTag Trace::beginProcessing(
   );
 
   if (theConfig()->vt_trace_memory_usage) {
-    addMemoryEvent(theMemUsage()->getFirstUsage());
+    addMemoryEvent(theMemUsage()->getFirstUsage(), time);
   }
 
   return TraceProcessingTag{ep, loggedEvent};
@@ -404,7 +404,7 @@ void Trace::endProcessing(
   );
 
   if (theConfig()->vt_trace_memory_usage) {
-    addMemoryEvent(theMemUsage()->getFirstUsage());
+    addMemoryEvent(theMemUsage()->getFirstUsage(), time);
   }
 
   // Final event is same as original with a few .. tweaks.
--- a/src/vt/trace/trace_lite.h
+++ b/src/vt/trace/trace_lite.h
@@ -252,6 +252,14 @@ struct TraceLite  {
     return static_cast<TimeIntegerType>(time * 1e6);
   }
 
+  std::size_t getNumTraceEvents() const {
+    return traces_.size();
+  }
+
+  const LogType* getLastTraceEvent() const noexcept {
+    return traces_.empty() ? nullptr : &traces_.back();
+  }
+
 protected:
   /**
    * \brief Emit a 'stop' trace for previous open event or a '[re]start' trace
--- a/src/vt/trace/trace_lite.h
+++ b/src/vt/trace/trace_lite.h
@@ -349,7 +349,7 @@ protected:
    *
    * \return computed bytes used for tracing (lower bound)
    */
-  std::size_t getTracesSize() const {
+  std::size_t getTracesSize() const noexcept {
     return traces_.size() * sizeof(Log);
   }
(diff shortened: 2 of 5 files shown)


src/vt/trace/trace.cc
```diff
@@ -359,7 +359,7 @@ TraceProcessingTag Trace::beginProcessing(
   );
 
   if (theConfig()->vt_trace_memory_usage) {
-    addMemoryEvent(theMemUsage()->getFirstUsage());
+    addMemoryEvent(theMemUsage()->getFirstUsage(), time);
```
nmm0, line 362: I added this so that the recorded events would be consistent (otherwise it uses the current time)

src/vt/trace/trace.cc
```diff
@@ -404,7 +404,7 @@ void Trace::endProcessing(
...
```
nmm0, line 407: Same as above

src/vt/trace/trace_lite.h
```diff
@@ -252,6 +252,24 @@ struct TraceLite  {
...
+
+  /**
+   * @brief Get the last recorded trace event
+   *
+   * @return the last recorded trace event
+   */
+  const LogType* getLastTraceEvent() const noexcept {
+    return traces_.empty() ? nullptr : &traces_.back();
+  }
+
```
nmm0, line 272: These functions are mostly used for testing but might be useful in other circumstances

src/vt/trace/trace_lite.h
```diff
@@ -349,7 +367,7 @@ struct TraceLite  {
    *
    * \return computed bytes used for tracing (lower bound)
    */
-  std::size_t getTracesSize() const {
+  std::size_t getTracesSize() const noexcept {
```
nmm0, line 370: I driveby-changed this to noexcept because it should be

tests/unit/trace/test_runnable_context_trace.cc
```diff
@@ -0,0 +1,99 @@
...
+  if (!theTrace()->checkDynamicRuntimeEnabled())
+    GTEST_SKIP() << "trace tests require --vt_trace to be set";
+
+  auto msg = makeMessage< TraceMsg >();
+
+  auto handler = auto_registry::makeAutoHandler< TraceMsg, handler_func >();
+
+  HandlerManager::setHandlerTrace(handler, true);
+
+  // Give some nonesense parameters but Trace shouldn't touch them
```
PhilMiller, line 71:
```suggestion
  // Give some nonsense parameters but Trace shouldn't touch them
```

tests/unit/trace/test_runnable_context_trace.cc
```diff
@@ -0,0 +1,99 @@
...
+  last_trace = theTrace()->getLastTraceEvent();
+  EXPECT_NE(last_trace, nullptr);
+  // Counterintuitive, but we restart the open event as the last action
+  EXPECT_EQ(last_trace->type, trace::eTraceConstants::BeginProcessing);
+  EXPECT_EQ(last_trace->time, TimeType{7}); // Time should still match though
+}
+
+}}} // end namespace vt::tests::unit
+
+#endif // vt_check_enabled(trace_enabled)
```
PhilMiller, line 99: EOF newline, so Git doesn't complain, please

PhilMiller:
Should the PR description be made a bit more comprehensive?

PhilMiller:
Trivial remarks aside, this looks good to me.

lifflander (approved):

JacobDomagala (approved):

nlslatt (approved):

PR #2151, opened 2023-05-26
#2150: Types: Make `Time` a strong type and use it across the codebase

### LINKED ISSUE #2150: Make Time a strong type
**What Needs to be Done?**
To avoid implicit conversions from ints/doubles, time should be a strong type.

See https://github.com/DARMA-tasking/vt/issues/2147 for a situation in which this came up

JacobDomagala:
Fixes #2150

### DIFF
--- a/src/vt/elm/elm_lb_data.cc
+++ b/src/vt/elm/elm_lb_data.cc
@@ -51,8 +51,8 @@
 namespace vt { namespace elm {
 
 void ElementLBData::start(TimeType time) {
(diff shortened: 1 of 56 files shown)


nlslatt:
I confirmed last week that this PR does not change what appears in the json output, which is good.

I would like to hear @lifflander's opinion on whether the changes from `TimeTypeWrapper` to the not strong `LoadType` type are okay in those contexts.

PR #2152, opened 2023-05-26
#2144: Fix unwanted cache export

### LINKED ISSUE #2144: Embedded vt as a subdirectory exports `ccache` usage to enclosing project
**Describe the bug**

An application that embeds vt and refers to it via `add_subdirectory` finds that it gets `ccache` picked up automatically, regardless of whether the person building the application wanted it, or whether that application has its own configuration mechanism to choose to use `ccache`. Vt should definitely not be doing this in the embedded case.

Offending CMake code is here:
https://github.com/DARMA-tasking/vt/blob/develop/CMakeLists.txt#L28

This corresponds to DARMA-tasking/magistrate#302

**To Reproduce**

**Expected behavior**

**Screenshots**

 - Compiler environment [e.g., clang-4.0, gcc8]

**Additional context**

thearusable:
Fixes #2144

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,14 +25,16 @@ set(
   "${CMAKE_CURRENT_SOURCE_DIR}/cmake/"
 )
 
-# Try to find ccache to speed up compilation
-find_program(ccache_binary ccache)
-if (ccache_binary)
-  message(
-    STATUS
-    "VT: Found ccache binary: ${ccache_binary}; adding launch rule"
-  )
-  set_directory_properties(PROPERTIES RULE_LAUNCH_COMPILE "${ccache_binary}")
+if (NOT DEFINED CMAKE_CXX_COMPILER_LAUNCHER)
+  # Try to find ccache to speed up compilation
+  find_program(ccache_binary ccache)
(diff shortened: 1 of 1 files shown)


nlslatt:
@thearusable What is the new behavior introduced by this PR? Does it avoid forcing ccache on the parent project but still use it for vt embedded in that other project's build?

PhilMiller:
> @thearusable What is the new behavior introduced by this PR? Does it avoid forcing ccache on the parent project but still use it for vt embedded in that other project's build?

I realize reading this PR that I should have thought a bit more about what to say in opening this and https://github.com/DARMA-tasking/magistrate/issues/302. It's one thing if vt or Magistrate uses ccache automatically when built on its own. When built as part of a larger project, it should build the same way the larger project does, and not try to customize things like that.

thearusable:
@nlslatt Yes. After the latest changes it will respect the CMAKE_CXX_COMPILER_LAUNCHER parameter. It will search for ccache only when it's not defined. In other cases it will respect whatever was set in CMAKE_CXX_COMPILER_LAUNCHER for this or parent project.

PR #2185, opened 2023-08-15
#1668: initialize collection stats phase

### LINKED ISSUE #1668: collection stats phase not initialized correctly
**Describe the bug**
Collection stats's phase is not initialized correctly (it defaults to 0). If a collection is created during a non-zero phase, this will create a mismatch between the phase in `CollectStatsMsg` and the phase associated with the collection, causing an assert.

**To Reproduce**
1. call `thePhase()->nextPhaseCollective()`
2. call `makeCollection()`
3. send a message from the collection
4. end the phase, `syncNextPhase` will assert because `stats.getPhase()` is 0 while `msg->getPhase()` is 1

**Expected behavior**
No assert, the phase should be updated in the collection.

**What needs to be done**
 `insertCollectionElement` should call `updatePhase`

nmm0:
Fixes #1668


src/vt/elm/elm_lb_data.cc
```diff
@@ -149,10 +149,13 @@ void ElementLBData::updatePhase(PhaseType const& inc) {
...
-  subphase_timings_[cur_phase_];
-  phase_comm_[cur_phase_];
-  subphase_comm_[cur_phase_];
+  if ( cur_phase_ != 0 )
+  {
+    phase_timings_[cur_phase_];
+    subphase_timings_[cur_phase_];
+    phase_comm_[cur_phase_];
+    subphase_comm_[cur_phase_];
+  }
```
nmm0, line 158: It's not clear to me that this is the best solution, so I'd like feedback. Basically this was causing a test to fail because they were looking at the length of these fields (and I call update in the construction of collection elements with this PR).

My reasoning was that the 0th phase shouldn't really have any timings associated with it. However, an alternative solution would be to avoid calling updatePhase if the phase is 0 from within insertElement.

tests/unit/collection/test_lb_lite.extended.cc
```diff
@@ -150,6 +150,41 @@ TEST_F(TestLB, test_lb_1) {
...
+  auto proxy2 = theCollection()->constructCollective<LBTest>(
+    range, "test_lb_phases_stats_2"
+  );
+
+  for (int i = 0; i < num_iter; i++) {
+    vt::runInEpochCollective([=]{
+      proxy2.broadcastCollective<IterMsg,&LBTest::iterWork>(i);
+    });
+
+    vt::thePhase()->nextPhaseCollective();
```
nmm0, line 184: This will assert without the fix (from mismatched phases)

src/vt/elm/elm_lb_data.cc
nmm0: In the meeting we decided to remove this in favor of a setPhase function

src/vt/vrt/collection/manager.impl.h
```diff
@@ -1182,6 +1182,9 @@ bool CollectionManager::insertCollectionElement(
     elm_holder->insert(idx, typename Holder<IndexT>::InnerHolder{
       std::move(vc)
     });
+    auto raw_ptr = elm_holder->lookup(idx).getRawPtr();
+    raw_ptr->getLBData().resetPhase();
+    raw_ptr->getLBData().updatePhase(thePhase()->getCurrentPhase());
```
nmm0, line 1187: This will be changed to the new setPhase

nmm0:
There appear to be some issues with this implementation that are not caught by new or existing tests, but show up in one of my application runs. So I'm investigating what the cause is.

src/vt/elm/elm_lb_data.cc
thearusable: Updated to `setPhase`

src/vt/vrt/collection/manager.impl.h
thearusable: Updated to `setPhase`

nmm0:
> @nmm0 Were you able to identify the issue?

Hey @thearusable sorry, I haven't had a chance, I've been out sick for a little while. I'll be looking into it but the main thing is that on one of the NimbleSM application runs, it eventually fails with a mismatched phase number. I'm not sure what the exact cause of it is, but I think it may be an inserted element not getting its phase number set correctly (I did remove that logic from the insert function but thought it would be covered in the other function it got moved to).

lifflander (approved):

PR #2186, opened 2023-08-16
#2177: Add helper method to remove vt parameters

### LINKED ISSUE #2177: Add utility to grab all arguments that are prefixed with `--vt_`
**What Needs to be Done?**

- [ ] Write method to strip out VT arguments prefixed with `--vt_` or `!--vt_`

thearusable:
Closes #2177


nlslatt (changes requested):
I don't think this meets our needs. Our use case is conditionally using vt from an application. We will not know if we need to use vt until after the command-line arguments for the application have been parsed and the input files have been read in. If the input files specify needing vt capabilities, vt will then be initialized. I believe we need the following:

1. The full original content of `argc` and `argv` is cached somehow for later processing by vt. Saving only the vt arguments is insufficient. At the very least, we also need `argv[0]` (expected by CLI11) and anything MPI-related (as we currently pass the full argument list to `MPI_Init`).
2. vt command-line arguments get removed from the passed `argc` and `argv` as currently happens in the vt initialization process, but *separate* from that process, for the benefit of the application. `argv[0]` and anything not directly vt-related must remain.
3. vt consumes the cached `argc` and `argv` at the time of initialization.

We could store the original `argc` and `argv` internally or pass them back to the application (as something distinct from the argc and argv the application cares about), which later passes them into the call to initialize vt.

Please let me know if any of this isn't clear.

lifflander (commented):
@thearusable I think the approach with `ArgvContainer` makes sense, considering that we don't have a runtime at that point to store arguments in. However, I would like to avoid any manual allocations/deallocations, so if we can try to use smart pointers instead of new/delete that would be preferable.

lifflander (approved):

lifflander:
@nlslatt Can you update your review on this?

nlslatt (commented):
This should meet my application needs.

We currently allow finalizing vt and then reinitializing it, but the preconfigure implementation here doesn't support it.

lifflander (changes requested):
After some discussion with @nlslatt, we've realized that this approach won't work if VT is initialized, finalized, and then re-initialized. Also if there are multiple instances of VT, this presents a problem.

Thus, we want to add a second interface that does not use the static variable, and instead returns the a unique_ptr to the `ArgvContainer`. Then we can add an optional parameter to `initializePreconfigured` that takes that `ArgvContainer`.

thearusable:
@nlslatt, @lifflander - I updated the interface to return the `unique_ptr` to `ArgvContainer`.

nlslatt (commented):
What happens when someone wants to add a third call to `initializePreconfigured` to your test, passing the original `vtConfig` object again (which was moved out in the second call)? We need to support passing the `ArgvContainer` to multiple calls.

lifflander (approved):
I think this looks good now

PR #2191, opened 2023-09-14
#2190: collection: fix group construction with listInsertHere

### LINKED ISSUE #2190: Group is not created in collection building when listInsertHere is used
**Describe the bug**
The group is not constructed so reduces won't work.

lifflander:
Fixes #2190


nlslatt (approved):
Looks great. You might want to add a test that has at least one rank with zero collection elements.

nlslatt (commented):
Actually, you'll need to use or remove `global_constructed_elms`.

PR #2199, opened 2023-10-13
#2171: Implement addUserDefinedData function to PhaseManager

### LINKED ISSUE #2171: Add application-specific information to phase (e.g., timesteps) in JSON output
**What Needs to be Done?**

We want to be able to match up timesteps in NimbleSM with phases IDs.

JacobDomagala:
Fixes #2171

### DIFF
--- a/src/vt/phase/phase_manager.h
+++ b/src/vt/phase/phase_manager.h
@@ -188,10 +188,16 @@ void addUserDefinedData(PhaseType phase, const KeyT& key, const ValueT& value) {
     "PhaseManager::addUserDefinedData: ValueT type is not jsonable"
   );
 
-  auto j = std::make_shared<nlohmann::json>();
-  j->emplace(key, value);
+  auto perPhase = theNodeLBData()->getLBData()->user_per_phase_json_;
 
-  theNodeLBData()->getLBData()->user_per_phase_json_[phase] = j;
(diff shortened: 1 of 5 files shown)


lifflander (approved):
Looks good to me

src/vt/phase/phase_manager.h
```diff
@@ -169,6 +170,30 @@ struct PhaseManager : runtime::component::Component<PhaseManager> {
...
+  );
+  static_assert(
+    is_jsonable<ValueT>::value,
+    "PhaseManager::addUserDefinedData: ValueT type is not jsonable"
+  );
+
+  auto j = std::make_shared<nlohmann::json>();
+  j->emplace(key, value);
+
+  theNodeLBData()->getLBData()->user_per_phase_json_[phase] = j;
```
nlslatt, line 194: Does this approach limit the user-defined data to a single key-value pair? I think we should allow as much as the application wants to provide.

nlslatt (changes requested):
Please see concern below.

src/vt/phase/phase_manager.h
JacobDomagala: Updated

nlslatt (approved):
Looks good

PR #2204, opened 2023-10-18
410 Dependent Epochs rewritten

### LINKED ISSUE #410: Semantics of collection chain set wrt nextStepCollective
Per a discussion with @PhilMiller, we've concluded that a mechanism that actually holds back messages for a given collective step (from one chain to another) could provide a clearer semantic.

First, note that a message marked with a collective epoch could arrive before that epoch is even created on another node. This is because `theTerm()->makeCollectiveEpoch(..)` is not synchronized by design. This definitely complicates any mechanism that we build.

My current proposition:

1. Add a bit to epoch that indicates if the epoch is *ready to execute* immediately or must be made ready. Currently, all are ready immediately.

2. If that bit is set as unready, the virtual context collection manager will not deliver the message until the user indicates it ready.

3. Add a call `proxy(idx).ready(epoch)` that allows delivery

4. Add the proxy to collection chain set so it can call this, for each index at the right time, and set the epoch bit appropriately.

lifflander:
Fixes #410

- After discussions about what is needed to support the resilience work, I've reopened this issue and semi-rebased/rewrote the old PR (that was closed) in a much cleaner way.
- I believe the current implementation adds very little overhead to any of the current code paths.
- This generalizes epochs allowing more complex dependency structures.

### DIFF
--- a/src/vt/messaging/active.cc
+++ b/src/vt/messaging/active.cc
@@ -934,12 +934,13 @@ void ActiveMessenger::prepareActiveMsgToRun(
   using MsgType = ShortMessage;
   auto msg = base.to<MsgType>().get();
 
-  auto const is_term = envelopeIsTerm(msg->env);
-  auto const is_bcast = envelopeIsBcast(msg->env);
-  auto const dest = envelopeGetDest(msg->env);
-  auto const handler = envelopeGetHandler(msg->env);
-  auto const epoch = envelopeIsEpochType(msg->env) ?
+  auto const is_term   = envelopeIsTerm(msg->env);
+  auto const is_bcast  = envelopeIsBcast(msg->env);
+  auto const dest      = envelopeGetDest(msg->env);
+  auto const handler   = envelopeGetHandler(msg->env);
+  auto const epoch     = envelopeIsEpochType(msg->env) ?
     envelopeGetEpoch(msg->env) : term::any_epoch_sentinel;
+
   auto const from_node = is_bcast ? dest : in_from_node;
 
   if (!is_term || vt_check_enabled(print_term_msgs)) {
@@ -963,6 +964,13 @@ void ActiveMessenger::prepareActiveMsgToRun(
   if (is_obj) {
     objgroup::dispatchObjGroup(base, handler, from_node, cont);
   } else {
+    if (epoch != term::any_epoch_sentinel and epoch::EpochManip::isDep(epoch)) {
+      if (not theTerm()->epochReleased(epoch)) {
+        pending_epoch_msgs_[epoch].emplace_back(base, from_node);
+        return;
+      }
+    }
+
     runnable::makeRunnable(base, not is_term, handler, from_node)
       .withContinuation(cont)
       .withTDEpochFromMsg(is_term)
@@ -978,6 +986,17 @@ void ActiveMessenger::prepareActiveMsgToRun(
   if (not is_term) {
     theTerm()->consume(epoch,1,in_from_node);
     theTerm()->hangDetectRecv();
+  }
+}
+
+void ActiveMessenger::releaseEpochMsgs(EpochType epoch) {
+  auto iter = pending_epoch_msgs_.find(epoch);
+  if (iter != pending_epoch_msgs_.end()) {
+    auto msgs = std::move(iter->second);
+    pending_epoch_msgs_.erase(iter);
+    for (auto&& m : msgs) {
+      prepareActiveMsgToRun(m.buffered_msg, m.from_node, true, m.cont);
+    }
   }
 }
 
--- a/src/vt/messaging/active.cc
+++ b/src/vt/messaging/active.cc
@@ -859,7 +859,7 @@ void ActiveMessenger::finishPendingDataMsgAsyncRecv(InProgressDataIRecv* irecv) 
       theTerm()->consume(term::any_epoch_sentinel,1,sender);
       theTerm()->hangDetectRecv();
     };
-    theSched()->enqueue(irecv->priority, run);
+    theSched()->enqueueLambda(irecv->priority, run);
   }
 }
 
@@ -964,13 +964,6 @@ void ActiveMessenger::prepareActiveMsgToRun(
   if (is_obj) {
     objgroup::dispatchObjGroup(base, handler, from_node, cont);
   } else {
-    if (epoch != term::any_epoch_sentinel and epoch::EpochManip::isDep(epoch)) {
-      if (not theTerm()->epochReleased(epoch)) {
-        pending_epoch_msgs_[epoch].emplace_back(base, from_node);
-        return;
-      }
-    }
-
     runnable::makeRunnable(base, not is_term, handler, from_node)
       .withContinuation(cont)
       .withTDEpochFromMsg(is_term)
@@ -986,17 +979,6 @@ void ActiveMessenger::prepareActiveMsgToRun(
   if (not is_term) {
     theTerm()->consume(epoch,1,in_from_node);
     theTerm()->hangDetectRecv();
-  }
-}
-
-void ActiveMessenger::releaseEpochMsgs(EpochType epoch) {
-  auto iter = pending_epoch_msgs_.find(epoch);
-  if (iter != pending_epoch_msgs_.end()) {
-    auto msgs = std::move(iter->second);
-    pending_epoch_msgs_.erase(iter);
-    for (auto&& m : msgs) {
-      prepareActiveMsgToRun(m.buffered_msg, m.from_node, true, m.cont);
-    }
   }
 }
 
--- a/src/vt/messaging/active.h
+++ b/src/vt/messaging/active.h
@@ -327,8 +327,6 @@ struct ActiveMessenger : runtime::component::PollableComponent<ActiveMessenger> 
   using SendFnType           = std::function<SendInfo(PtrLenPairType,NodeType,TagType)>;
   using UserSendFnType       = std::function<void(SendFnType)>;
   using ContainerPendingType = std::unordered_map<TagType,PendingRecvType>;
-  using MsgContType          = std::list<BufferedMsgType>;
-  using EpochWaitType        = std::unordered_map<EpochType, MsgContType>;
   using ReadyHanTagType      = std::tuple<HandlerType, TagType>;
   using HandlerManagerType   = HandlerManager;
   using PendingSendType      = PendingSend;
@@ -1652,13 +1650,6 @@ struct ActiveMessenger : runtime::component::PollableComponent<ActiveMessenger> 
   # endif
   }
 
-  /*
-   * \brief Deliver messages that are now released with a dependent epoch
-   *
-   * \param[in] epoch the epoch to release
-   */
-  void releaseEpochMsgs(EpochType epoch);
-
 private:
   /**
    * \internal \brief Allocate a new, unused tag.
@@ -1773,7 +1764,6 @@ private:
   elm::ElementIDStruct bare_handler_dummy_elm_id_for_lb_data_ = {};
   elm::ElementLBData bare_handler_lb_data_;
   MPI_Comm comm_ = MPI_COMM_NULL;
-  EpochWaitType pending_epoch_msgs_                           = {};
 };
 
 }} // end namespace vt::messaging
--- a/src/vt/messaging/active.h
+++ b/src/vt/messaging/active.h
@@ -268,7 +268,7 @@ struct BufferedActiveMsg {
 
   BufferedActiveMsg(
     MessageType const& in_buffered_msg, NodeType const& in_from_node,
-    ActionType in_cont
+    ActionType in_cont = nullptr
   ) : buffered_msg(in_buffered_msg), from_node(in_from_node), cont(in_cont)
   { }
 
@@ -327,6 +327,8 @@ struct ActiveMessenger : runtime::component::PollableComponent<ActiveMessenger> 
   using SendFnType           = std::function<SendInfo(PtrLenPairType,NodeType,TagType)>;
   using UserSendFnType       = std::function<void(SendFnType)>;
   using ContainerPendingType = std::unordered_map<TagType,PendingRecvType>;
+  using MsgContType          = std::list<BufferedMsgType>;
+  using EpochWaitType        = std::unordered_map<EpochType, MsgContType>;
   using ReadyHanTagType      = std::tuple<HandlerType, TagType>;
   using HandlerManagerType   = HandlerManager;
   using PendingSendType      = PendingSend;
@@ -1650,6 +1652,13 @@ struct ActiveMessenger : runtime::component::PollableComponent<ActiveMessenger> 
   # endif
   }
 
+  /*
+   * \brief Deliver messages that are now released with a dependent epoch
+   *
+   * \param[in] epoch the epoch to release
+   */
+  void releaseEpochMsgs(EpochType epoch);
+
 private:
   /**
    * \internal \brief Allocate a new, unused tag.
@@ -1764,6 +1773,7 @@ private:
   elm::ElementIDStruct bare_handler_dummy_elm_id_for_lb_data_ = {};
   elm::ElementLBData bare_handler_lb_data_;
   MPI_Comm comm_ = MPI_COMM_NULL;
+  EpochWaitType pending_epoch_msgs_                           = {};
 };
 
 }} // end namespace vt::messaging
--- a/src/vt/objgroup/proxy/proxy_objgroup_elm.impl.h
+++ b/src/vt/objgroup/proxy/proxy_objgroup_elm.impl.h
@@ -140,6 +140,30 @@ ObjT* ProxyElm<ObjT>::get() const {
   return theObjGroup()->get<ObjT>(proxy);
 }
 
+template <typename ObjT>
+bool ProxyElm<ObjT>::isReleased(EpochType epoch) const {
+  if (node_ == theContext()->getNode()) {
+    return theSched()->isReleasedEpochObjgroup(epoch, proxy_);
+  } else {
+    vtAbort("isReleased can only be called locally");
+    return false;
+  }
+}
+
+template <typename ObjT>
+inline void releaseRemoteObjGroup(ProxyElm<ObjT> proxy, EpochType ep) {
+  proxy.release(ep);
+}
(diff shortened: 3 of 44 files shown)


src/vt/messaging/active.cc
```diff
@@ -963,6 +964,13 @@ void ActiveMessenger::prepareActiveMsgToRun(
   if (is_obj) {
     objgroup::dispatchObjGroup(base, handler, from_node, cont);
   } else {
+    if (epoch != term::any_epoch_sentinel and epoch::EpochManip::isDep(epoch)) {
```
PhilMiller: Why is this branch not applicable in the `is_obj` case?

src/vt/messaging/active.cc
```diff
@@ -963,6 +964,13 @@ void ActiveMessenger::prepareActiveMsgToRun(
...
+      if (not theTerm()->epochReleased(epoch)) {
+        pending_epoch_msgs_[epoch].emplace_back(base, from_node);
```
PhilMiller: Why is `cont` not being passed along here? The insertion is passing the default argument value of `nullptr` instead

src/vt/messaging/active.h
```diff
@@ -327,6 +327,8 @@ struct ActiveMessenger : runtime::component::PollableComponent<ActiveMessenger>
   using SendFnType           = std::function<SendInfo(PtrLenPairType,NodeType,TagType)>;
   using UserSendFnType       = std::function<void(SendFnType)>;
   using ContainerPendingType = std::unordered_map<TagType,PendingRecvType>;
+  using MsgContType          = std::list<BufferedMsgType>;
```
PhilMiller: Usually preferable to just use `vector` and rely on amortization of insertion, gaining locality of access.

src/vt/termination/termination.h
```diff
@@ -243,23 +243,27 @@ struct TerminationDetector :
...
    * \param[in] parent parent epoch that waits for this new epoch
+   * \param[in] is_dep whether it's a dependent epoch
    *
    * \return the new epoch
    */
   EpochType makeEpochRooted(
     UseDS use_ds = UseDS{true},
-    ParentEpochCapture parent = ParentEpochCapture{}
+    ParentEpochCapture parent = ParentEpochCapture{},
+    bool is_dep = false
```
PhilMiller, line 253: Boolean parameters (especially defaulted ones) are a bit of an anti-pattern. Preferable to use named `enum class` members instead

src/vt/termination/termination.h
```diff
@@ -358,6 +369,54 @@ struct TerminationDetector :
...
+  void onReleaseEpoch(EpochType epoch, ActionType action);
+
+  /**
+   * \brief Test if an epoch is dependent and if it is, if that epoch has been released
+   *
+   * \param[in] epoch the epoch
+   *
+   * \return if it is released
+   */
+  bool epochReleased(EpochType epoch);
```
PhilMiller, line 394: Perhaps better phrased as "isEpochClear", since a non-dependent epoch will never actually get 'released'

src/vt/termination/termination.cc
```diff
@@ -1105,11 +1117,121 @@ void TerminationDetector::initializeCollectiveEpoch(
...
+    makeEpochRooted(label, use_ds, successor, is_dep);
+}
+
+void TerminationDetector::releaseEpoch(EpochType epoch) {
+  bool const is_dep = isDep(epoch);
+
+  if (is_dep) {
+    // Put the epoch in the released set, which is not conclusive due to
+    // dependencies, which effects the status. An epoch is *released* iff the
+    // epoch is in the released set and all succesrros are *released* (or there
```
PhilMiller: Typo `successors`

src/vt/termination/termination.h
```diff
@@ -358,6 +369,54 @@ struct TerminationDetector :
...
+   */
+  void releaseEpoch(EpochType epoch);
+
+  /**
+   * \brief Action to run on release of a dependent epoch
+   *
+   * \param[in] epoch the epoch
+   * \param[in] action the action
+   */
+  void onReleaseEpoch(EpochType epoch, ActionType action);
```
PhilMiller, line 385: This name really doesn't sit well with me, but I'm struggling to offer an improvement.

On approach is that the comments could at least be a lot more explicit:
- run immediately if `epoch` is non-dependent or already released
- enqueue `action` otherwise

Alternately, drop the 'running immediately' case entirely, and make it strictly enqueuing, with the (single) caller checking the `is_dep` condition itself. The function name should still be clarified.

src/vt/termination/termination.h
```diff
@@ -823,16 +886,22 @@ struct TerminationDetector :
...
   // list of remote epochs pending status report of finished
-  std::unordered_set<EpochType> epoch_wait_status_      = {};
+  std::unordered_set<EpochType> epoch_wait_status_ = {};
   // has printed epoch graph during abort
-  bool has_printed_epoch_graph                          = false;
+  bool has_printed_epoch_graph = false;
   NodeType this_node_ = uninitialized_destination;
   EpochStackType epoch_stack_;
+  // released epoch list for dependent epochs
+  std::unordered_set<EpochType> epoch_released_ = {};
```
PhilMiller, line 893: Shouldn't this be a range container as is used for other stuff like this, so that it won't tend to grow without bound?

src/vt/termination/termination.cc
```diff
@@ -1105,11 +1117,121 @@ void TerminationDetector::initializeCollectiveEpoch(
...
+  bool const is_dep = isDep(epoch);
+  if (is_dep) {
+    bool const is_term = theEpoch()->getTerminatedWindow(epoch)->isTerminated(
+      epoch
+    );
+    if (is_term) {
+      auto iter = epoch_released_.find(epoch);
+      if (iter != epoch_released_.end()) {
+        epoch_released_.erase(iter);
+      }
```
PhilMiller, line 1195: This can just be `epoch_released_.erase(epoch);`

src/vt/termination/termination.cc
```diff
@@ -1105,11 +1117,121 @@ void TerminationDetector::initializeCollectiveEpoch(
...
+
+void TerminationDetector::runReleaseEpochActions(EpochType epoch) {
+  auto iter = epoch_release_action_.find(epoch);
+  if (iter != epoch_release_action_.end()) {
+    auto actions = std::move(iter->second);
+    epoch_release_action_.erase(iter);
+    for (auto&& fn : actions) {
+      fn();
+    }
+  }
```
PhilMiller, line 1150: TIL, with C++17, this can be done more cleanly as 
```c++
auto node = epoch_release_action_.extract(epoch);
if (node) {
    for (auto&& fn : node.value()) {
      fn();
    }
}
```

src/vt/termination/termination.cc
```diff
@@ -1105,11 +1117,121 @@ void TerminationDetector::initializeCollectiveEpoch(
...
+    makeEpochCollective(label, successor, is_dep) :
+    makeEpochRooted(label, use_ds, successor, is_dep);
+}
+
+void TerminationDetector::releaseEpoch(EpochType epoch) {
+  bool const is_dep = isDep(epoch);
+
+  if (is_dep) {
+    // Put the epoch in the released set, which is not conclusive due to
+    // dependencies, which effects the status. An epoch is *released* iff the
```
PhilMiller: affects

src/vt/termination/termination.cc
```diff
@@ -1105,11 +1117,121 @@ void TerminationDetector::initializeCollectiveEpoch(
...
+
+  // Terminated epochs are always released
+  bool const is_term = theEpoch()->getTerminatedWindow(epoch)->isTerminated(
+    epoch
+  );
+  if (is_term) {
+    return true;
+  }
+
+  // All successors must be released for an epoch to be released even if its in
```
PhilMiller: it's

tests/unit/termination/test_term_dep_epoch_active.cc
```diff
@@ -0,0 +1,129 @@
...
+    return vt::messaging::PendingSend(msg, [=](MsgVirtualPtr<vt::BaseMsgType>){
+      EXPECT_EQ(TestDep::num_dep, 0);
+      theTerm()->releaseEpoch(epoch);
+    });
+  });
+
+  vt::theTerm()->addAction([=]{
+    EXPECT_EQ(TestDep::num_non_dep, 1);
+    EXPECT_EQ(TestDep::num_dep, (num_nodes - 1)*k);
+  });
```
PhilMiller, line 116: I'm not a fan of the style choice of having the scheduler run and the test assertions get checked as a result of running to quiescence as the process is finishing up. If nothing else, it makes for really funky looking stack traces if/when an assertion fails.

src/vt/messaging/active.cc
PhilMiller: NM, this code went away in a later commit.

src/vt/scheduler/scheduler.cc
```diff
@@ -384,6 +384,17 @@ void Scheduler::resume(ThreadIDType tid) {
   suspended_.resumeRunnable(tid);
 }
 
+void Scheduler::releaseEpoch(EpochType ep) {
+  auto iter = pending_work_.find(ep);
```
PhilMiller: Another possible use of `pending_work_.extract(ep)`

src/vt/scheduler/scheduler.impl.h
```diff
@@ -92,7 +92,15 @@ void Scheduler::enqueue(bool is_term, RunT r) {
   if (is_term) {
     num_term_msgs_++;
   }
-  work_queue_.emplace(UnitType(is_term, r));
+  auto ep = r->getEpoch();
+  if (
+    ep != no_epoch and ep != term::any_epoch_sentinel and
+    not theTerm()->epochReleased(ep))
```
PhilMiller: Doesn't `epochReleased()` already check for the `no_epoch` and `any_epoch_sentinel` values?

src/vt/objgroup/proxy/proxy_objgroup_elm.impl.h
```diff
@@ -140,6 +140,30 @@ ObjT* ProxyElm<ObjT>::get() const {
...
+inline void releaseRemoteObjGroup(ProxyElm<ObjT> proxy, EpochType ep) {
+  proxy.release(ep);
+}
+
+template <typename ObjT>
+void ProxyElm<ObjT>::release(EpochType epoch) const {
+  if (node_ == theContext()->getNode()) {
+    theSched()->releaseEpochObjgroup(epoch, proxy_);
+  } else {
+    theMsg()->send<releaseRemoteObjGroup<ObjT>>(vt::Node(node_), *this, epoch);
```
PhilMiller, line 163: Why is it going to be valid to release an epoch for a single node?

src/vt/scheduler/scheduler.cc
```diff
@@ -412,6 +412,24 @@ void Scheduler::releaseEpochObjgroup(EpochType ep, ObjGroupProxyType proxy) {
...
 
+void Scheduler::releaseEpochCollection(EpochType ep, UntypedCollection* untyped) {
+  if (
+    auto iter = pending_collection_work_.find(ep);
+    iter != pending_collection_work_.end())
+  {
+    if (auto iter2 = iter->second.find(untyped); iter2 != iter->second.end()) {
+      auto& container = iter2->second;
+      while (container.size() > 0) {
+        work_queue_.emplace(container.pop());
```
PhilMiller, line 423: Maybe better to `move` the values from `container`, rather than actually removing them, so that no work needs to be done on the structure of `container` itself, which is going away right after this loop.

I.e.
```
for (auto &unit : container) {
  work_queue_.emplace(std::move(unit));
}
```

src/vt/scheduler/scheduler.impl.h
```diff
@@ -120,22 +120,33 @@ void Scheduler::enqueue(MsgT* msg, RunT r) {
...
+            }
+          } else {
+            auto untyped = static_cast<UntypedCollection*>(obj);
+            if (not untyped->isReleasedEpoch(ep)) {
+              pending_collection_work_[ep][untyped].push(unit);
+              return;
+            }
+          }
+        } else if (not theTerm()->epochReleased(ep)) {
+          pending_work_[ep].push(unit);
```
PhilMiller, line 147: The nesting is getting really deep here. It might be nice to peel off some of the cases as early returns, if/as possible.

PhilMiller:
So, the one big question I have in reviewing this is: why the move to releasing epochs per-collection/objgroup/object?

Everything else I've marked in comments is pretty minor, and can/should be addressed after writing up an answer for the big question.

Also, squashing fixup commits into their progenitors would be good.

Matthew-Whitlock:
It looks like creating an epoch within a dependent epoch circumvents waiting for release on all of the messages within the nested epoch. As is, I don't think we could make any assertions about the behavior of general code like this:
```
auto epoch = vt::theTerm()->makeEpochCollective(term::ParentEpochCapture{}, true);
vt::theMsg()->pushEpoch(epoch);

some::library::function();

vt::theMsg()->popEpoch(epoch);
vt::theTerm()->finishedEpoch(epoch);
```

tests/unit/termination/test_term_dep_epoch_collection.cc
```diff
@@ -0,0 +1,139 @@
...
+  bool const bcast = true;
+  int const k = 10;
+  int const r = 48;
+
+  auto range = vt::Index1D(r);
+  auto proxy = vt::theCollection()->constructCollective<TestDepColl>(range);
+  vt::theCollective()->barrier();
+
+  auto epoch = vt::theTerm()->makeEpochCollective(term::ParentEpochCapture{}, true);
+  vt::theMsg()->pushEpoch(epoch);
```
Matthew-Whitlock, line 98: Since dependent epochs make this pattern necessary, a returning equivalent of `vt::runInEpoch<Collective/Rooted>`  would be a nice helper for users.

src/vt/termination/termination.cc
```diff
@@ -1105,11 +1117,84 @@ void TerminationDetector::initializeCollectiveEpoch(
...
+    auto actions = std::move(iter->second);
+    epoch_release_action_.erase(iter);
+    for (auto&& fn : actions) {
+      fn();
+    }
+  }
+  theSched()->releaseEpoch(epoch);
+}
+
+void TerminationDetector::onReleaseEpoch(EpochType epoch, ActionType action) {
```
Matthew-Whitlock, line 1154: This should use the same approach as term actions after https://github.com/DARMA-tasking/vt/pull/2196

src/vt/termination/termination.cc
```diff
@@ -1105,11 +1117,84 @@ void TerminationDetector::initializeCollectiveEpoch(
...
+    return true;
+  }
+
+  // Terminated epochs are always released
+  bool const is_term = theEpoch()->getTerminatedWindow(epoch)->isTerminated(
+    epoch
+  );
+  if (is_term) {
+    return true;
+  }
```
Matthew-Whitlock, line 1178: I don't think this should be true, or an empty dependent epoch could terminate before being released, which seems counter-intuitive. 
```
auto epoch = vt::theTerm()->makeEpochCollective(term::ParentEpochCapture{}, true);
vt::theTerm()->finishedEpoch(epoch);
vt::theTerm()->addAction(epoch, {...});
```
Your added action could run before releasing in the above, or in cases where a set of functions may or may not actually send messages. We should produce once on dependent epoch creation and consume on release, to align the no-messages behavior with typical behavior.

lifflander:
> So, the one big question I have in reviewing this is: why the move to releasing epochs per-collection/objgroup/object?

The short answer to this question is that it will allow us to express dependencies per-object. An object can release work when it's ready, which is a per-object state.

> Everything else I've marked in comments is pretty minor, and can/should be addressed after writing up an answer for the big question.
> 
> Also, squashing fixup commits into their progenitors would be good. Or, even rewriting the history entirely to provide a clearer to understand narrative sequence of commits.

Yes, I am planning on re-writing the history.

lifflander:
@PhilMiller @Matthew-Whitlock Please take a look at the follow-on PR #2206. I nearly have the new abstraction working on top of this `taskCollective`. Here is the example Jacobi program. This might clarify why the dependencies are per-object.

https://github.com/DARMA-tasking/vt/blob/99a63d99a7d156cafae33d157e56c2aea3145986/examples/collection/jacobi1d_vt.cc#L233-L284

PhilMiller:
Ok, I'll take a look. That maybe answers my follow-up question of what it means to release an epoch for some object/collection, but not for others.

src/vt/termination/termination.h
lifflander: Yes, I was planning this improvement. The bounds are lower than one might think because they are always removed once the epoch terminates.

src/vt/messaging/active.cc
lifflander: This code is old now and has been deleted

src/vt/termination/termination.cc
lifflander: This code is removed

src/vt/termination/termination.h
lifflander: I'm actually removing this function all together. No need for it.

src/vt/termination/termination.cc
lifflander: This code is removed.

src/vt/termination/termination.h
lifflander: Renamed

src/vt/termination/termination.cc
lifflander: Fixed

src/vt/termination/termination.cc
lifflander: Done

src/vt/scheduler/scheduler.cc
lifflander: The problem with this is that `container` is not iterable. The container is a hybrid queue implemented differently depending on whether priorities are enabled. If they are not, it is a 64 entry circular buffer with an overflow `std::queue` container. We could make it iterable, but it would need a custom iterator.

src/vt/termination/termination.cc
nlslatt: @lifflander What are your thoughts on Matthew's concern?

PR #2208, opened 2023-10-27
#2125: Fix sprintf warnings

### LINKED ISSUE #2125: sprintf warning on macOS
**What Needs to be Done?**

When building on mac I am seeing a warning of `'sprintf' is deprecated`
It recommends using `snprintf` .

Also might be useful to enable `-Werror` in macos builds.

thearusable:
Closes #2125

### DIFF
--- a/src/vt/group/collective/group_info_collective.cc
+++ b/src/vt/group/collective/group_info_collective.cc
@@ -701,7 +701,7 @@ void InfoColl::finalize() {
       buf[0] = '\0';
       int cur = 0;
       for (auto&& elm : collective_->span_children_) {
-        cur += snprintf(buf + cur, max_buffer_length, "%d,", elm);
+        cur += snprintf(buf + cur, max_buffer_length - cur, "%d,", elm);
       }
 
       auto const& num_children = collective_->span_children_.size();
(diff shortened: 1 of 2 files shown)


JacobDomagala (approved):

src/vt/group/collective/group_info_collective.cc
```diff
@@ -696,11 +696,12 @@ void InfoColl::finalize() {
...
 
     if (!vt_check_enabled(production_build)) {
-      char buf[256];
+      constexpr int max_buffer_length = 256;
+      char buf[max_buffer_length];
       buf[0] = '\0';
       int cur = 0;
       for (auto&& elm : collective_->span_children_) {
-        cur += sprintf(buf + cur, "%d,", elm);
+        cur += snprintf(buf + cur, max_buffer_length, "%d,", elm);
```
nlslatt, line 704: Should this use `max_buffer_length - cur` instead of `max_buffer_length` because we are potentially writing mid-buffer?

nlslatt (changes requested):
Please see comment below.

src/vt/group/collective/group_info_collective.cc
thearusable: Ah, Yes, You're right.

nlslatt (approved):
Looks good

lifflander (approved):

PR #2213, opened 2023-11-08
2212 add message properties for generated messages

### LINKED ISSUE #2212: Add message properties for generated messages
**What Needs to be Done?**

We need to be able to set properties on the message, like the priority or epoch or whether it's traced when generated by the system.

lifflander:
Fixes #2212

- Adds message properties to generated/synthesized messages that are parameterized.
- I had to move the parameters going into `ParamMsg` from the constructor to a method `setParams` so I could safely modify the message is `MsgProps` are passed. Otherwise, I would have to save the `MsgProps` to apply them after the constructor finishes.

### DIFF
--- a/src/vt/collective/reduce/reduce_manager.cc
+++ b/src/vt/collective/reduce/reduce_manager.cc
@@ -46,12 +46,12 @@
 
 namespace vt { namespace collective { namespace reduce {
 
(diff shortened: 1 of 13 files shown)


nlslatt (approved):
Looks good to me

PR #2219, opened 2023-11-22
#2218: location: fix case when migrated but doesn't update home throu…

### LINKED ISSUE #2218: Fix bug in location manager not updating home node
**Describe the bug**
Occurring in NimbleSM, but the bug can happen anywhere. @nmm0

lifflander:
Fixes #2218


nmm0 (approved):
Looks good to me and fixes the livelocking issue

nlslatt (approved):

PR #2203, opened 2023-10-18
#2201: implement memory aware TemperedLB in vt

ppebay:
Resolves #2201 

This PR in particular:
* adds the necessary apparatus to allow for several transfer strategies, to be chosen by the user;
* refactors the existing transfer stage to make it the `Original` strategy;
* implements the `SwapClusters` transfer strategy; i.e. a simplification of LBAF's `ClusteringStrategy` (without sub-clustering).

PR #2228, opened 2023-12-01
2227 Add user LB info to load models

### LINKED ISSUE #2227: Add user-defined data to load models
**What Needs to be Done?**

Integrate user-defined LB info into the load models, where the data will be accessible by the load balancers even under replay.

nlslatt:
This PR adds access to the user-defined LB info to the load models so that load balancers can access it whether running on a live application or replaying from json files.

I have confirmed that, with the appropriate changes to #2203, TemperedLB can access the user-defined LB info for local objects that were loaded from a json file for a different rank, while still being able to access info provided live by a running application. I will comment on #2203 what changes are necessary.

Closes #2227


lifflander (approved):
This looks good to me. I think eventually we should change `setLoads` to a more general name now that it's setting the loads, communication, and user-defined data.

PR #2230, opened 2023-12-20
#2175: Add YAML as input as alternative to command-line arguments

### LINKED ISSUE #2175: Add YAML as input as alternative to command-line arguments
**What Needs to be Done?**

- [ ] Using the yaml-cpp library

cwschilly:
Fixes #2175

## Usage

Pass a yaml config file with
```sh
--vt_input_config_yaml /path/to/yaml/file
```

The order of precedence is currently
1. command line arguments
2. input `toml` or `ini` file
3. input yaml,

meaning that the command line arguments will overwrite everything (and a toml config will overwrite a yaml config).

## Changes to `yaml-cpp`

I had to make changes to the yaml-cpp library in order to avoid variable shadowing errors. I mostly used the solution from [this PR](https://github.com/jbeder/yaml-cpp/pull/815), which added `enum class` where appropriate.

In one case, I had to change a variable name from `YAML::Null` to `YAML::BaseNull` (to avoid shadowing `YAML::NodeType::Null`). I don't foresee this having consequences for our development (if we want to check if some input is null, we would use `YAML::NodeType::Null`).

I also made `yaml-cpp` a system include directory to silence all warnings (`-Wtautological-constant-compare` was throwing an error on the intel icpx pipeline--[this issue](https://github.com/assimp/assimp/issues/4450) explains why it is not a meaningful warning and can be safely silenced).


src/vt/configs/arguments/args.cc
```diff
@@ -122,13 +123,24 @@ std::tuple<int, std::string> parseArguments(
...
+    if (config_ending == ".yml") | (config_ending == "yaml") {
+      // use yaml-cpp
+      parse_yaml(config_file, appConfig);
+    } else if (config_ending == ".ini") | (config_ending == "toml") {
+      // use CLI parser
+      app.set_config(
+        "--vt_input_CLI_config",
+        config_file,
+        "Read in a config file with CLI library",
+        false // not required
```
cwschilly, line 140: The idea is that the user would not pass `--vt_input_CLI_config`; they would pass their file with `--vt_input_config` (as is the current API), and if it is a `toml` or `ini` file it would be set as the default value for the `app.set_config()` function.

lifflander: This looks good to me!

cwschilly:
I'm not sure what's going on with the `PR checks (git --check)` pipeline--there isn't any trailing white space in the file it's talking about

JacobDomagala:
I think we should strip it down to bare minimum. We don't need `test` and `contributing` dirs. That would also resolve the trailing whitespace issue : ) 

Should we also support using external `yaml-cpp`? Same way we intent to do with `fmt` (https://github.com/DARMA-tasking/vt/pull/2221)

cwschilly:
> Should we support using external `yaml-cpp`? Same way we intent to do with `fmt` (#2221)

I think this makes sense, especially since it seems the yaml-cpp source code doesn't pass the CI

JacobDomagala:
Also, is the use of a yaml-based config file incentivized by something (like Empire already using that file type)? If not, maybe we should consider using JSON for the config file as well? We already have it set up, and its structure is well-known to everyone.

lifflander:
> Also, is the use of a yaml-based config file incentivized by something (like Empire already using that file type)? If not, maybe we should consider using JSON for the config file as well? We already have it set up, and its structure is well-known to everyone.

Yes, Empire is using YAML for their input and we want to be able to pass that directly to VT. LBAF and vt-tv are also using YAML so it seems that that route makes sense.

cwschilly:
I've taken the approach of [this PR](https://github.com/jbeder/yaml-cpp/pull/815) to yaml-cpp in order to resolve the clang warnings.

lib/yaml-cpp/include/yaml-cpp/null.h
```diff
@@ -0,0 +1,26 @@
...
+class Node;
+
+struct YAML_CPP_API _Null {};
+inline bool operator==(const _Null&, const _Null&) { return true; }
+inline bool operator!=(const _Null&, const _Null&) { return false; }
+
+YAML_CPP_API bool IsNull(const Node& node);  // old API only
+YAML_CPP_API bool IsNullString(const std::string& str);
+
+extern YAML_CPP_API _Null BaseNull;
```
cwschilly, line 23: To avoid the shadowing warning caused by this (`YAML::Null`) and `YAML::NodeType::Null`, I've renamed this variable to `YAML::BaseNull`. I don't imagine that this will have any far-reaching consequences, but I want to highlight it just in case.

lib/yaml-cpp/CMakeLists.txt
```diff
@@ -0,0 +1,174 @@
...
+set_property(TARGET yaml-cpp
+  PROPERTY
+    CXX_STANDARD_REQUIRED ON)
+
+target_include_directories(yaml-cpp
+  PUBLIC
+    $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
+    $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
+  PRIVATE
+    $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/src>)
```
JacobDomagala, line 86: If we don't care about the warning itself, we can probably just silence any warnings with: 

```suggestion
target_include_directories(yaml-cpp
  SYSTEM PUBLIC
    $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
    $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
  PRIVATE
    $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/src>)

```

JacobDomagala: There's a duplication of logs now that we've added verbose builds. Most recent icpx is now clean.

cwschilly: Thanks @JacobDomagala , I made this change

lifflander:
Summary of our discussion:
- Let's output the current configuration as a YAML file
- Use this output for testing input is working properly
- For the debug print configuration, take a list of components to output debug print statements
- Create further nesting for inputs, like the "LB Data Output"/"Statistics"/"LB Data Input"

cz4rs:
@cwschilly I have started to look at this, I will finish the review and post the comments tomorrow.

lib/yaml-cpp/.clang-format
```diff
@@ -0,0 +1,46 @@
+---
```
cz4rs, line 1: Let's not add any dotfiles from the library.

cz4rs:
> The order of precedence is currently
> 
>     1. input yaml
> 
>     2. command line arguments
> 
>     3. input `toml` or `ini` file,
> 
> 
> meaning that the YAML will overwrite the command line arguments (which in turn overwrite the input `toml` file).

Intuitively I'd expect the command line arguments to overwrite the configuration files. Whatever the order ends up being, this needs to be mentioned explicitly in the docs and possibly in usage / help message in the library.

cmake/load_bundled_libraries.cmake
```diff
@@ -38,6 +38,10 @@ else()
   set_darma_compiler_flags(${FMT_LIBRARY})
 endif()
 
+# yaml-cpp always included in the build
+set(YAMLCPP_LIBRARY yaml-cpp)
+add_subdirectory(${PROJECT_LIB_DIR}/yaml-cpp)
```
cz4rs, line 43: > Should we support using external `yaml-cpp`? Same way we intent to do with `fmt` (#2221)

Getting back to this question. Current approach is to always use bundled version, can we expect any issues in the applications because of this?

cmake/link_vt.cmake
```diff
@@ -41,7 +42,7 @@ function(link_target_with_vt)
   set(allKeywords ${noValOption} ${singleValArg} ${multiValueArg})
 
   cmake_parse_arguments(
-    ARG "${noValOption}" "${singleValArg}" "${multiValueArgs}" ${ARGN}
+    ARG "${noValOption}" "${singleValArg}" "${multiValueArg}" ${ARGN}
```
cz4rs, line 45: Curious that it didn't cause any bugs before :thinking:
Maybe no one uses `vt` with `CUSTOM_LINK_ARGS`.

src/vt/configs/arguments/args.cc
```diff
@@ -185,6 +193,241 @@ std::tuple<int, std::string> parseArguments(
   return std::make_tuple(-1, std::string{});
 }
 
+void addYamlConfig(CLI::App& app, AppConfig& appConfig) {
+  auto yaml_description = "Read in a yaml config file for VT";
```
cz4rs, line 197:
```suggestion
  auto yaml_description = "Read in a YAML config file for VT";
```

tests/unit/test_helpers.h
```diff
@@ -101,6 +102,48 @@ inline std::string getUniqueFilenameWithRanks(const std::string& ext = "") {
...
+          } catch(...) {
+            try {
+              EXPECT_EQ(yaml_1_val.as<int>(), yaml_2_val.as<int>());
+            } catch(...) {
+              EXPECT_EQ(yaml_1_val.as<std::string>(), yaml_2_val.as<std::string>());
+            }
+          }
+        }
+        else {
+          EXPECT_TRUE(false);
```
cz4rs, line 142: Provide a clear error message for this case.

src/vt/configs/arguments/args.cc
```diff
@@ -185,6 +193,241 @@ std::tuple<int, std::string> parseArguments(
...
+      if (bool_iter == 1) appConfig.vt_user_1 = user_input;
+      else if (bool_iter == 2) appConfig.vt_user_2 = user_input;
+      else if (bool_iter == 3) appConfig.vt_user_3 = user_input;
+      else too_many_user_args = true;
+      appConfig.user_args["vt_user_" + std::to_string(bool_iter)] = key;
+      bool_iter++;
+    } catch(...) {
+      try {
+        int user_input = user_options[key].as<int>();
+        if (int_iter == 1) appConfig.vt_user_int_1 = user_input;
```
cz4rs, line 392: Place the instruction in a separate line after an `if` or `else`:
```suggestion
        if (int_iter == 1)
          appConfig.vt_user_int_1 = user_input;
```

cz4rs: When in doubt you can use `clang-format` (for example `git clang-format HEAD~1`). Note that the `.clang-format` configuration file we have does not reflect `vt`'s formatting style perfectly :shrug:

src/vt/configs/arguments/app_config.h
```diff
@@ -77,6 +82,8 @@ struct AppConfig {
     return vt_debug_print_flush;
   }
 
+  std::string vt_input_config_yaml = "";
```
cz4rs, line 85: This should probably be the only change in this header, all other should be moved elsewhere (preferably .cpp file to avoid pulling in `yaml-cpp` header).

src/vt/configs/arguments/app_config.h
```diff
@@ -411,6 +430,242 @@ struct AppConfig {
 
       | passthru_args;
   }
+
+  using variantArg_t = std::variant<bool, int32_t, int64_t, double, std::size_t, std::string>;
+
+  std::vector<std::string> splitString(std::string input_str) {
```
cz4rs, line 436: Take a look at the existing `DemanglerUtils::splitString`.

src/vt/configs/arguments/args.cc
```diff
@@ -185,6 +193,241 @@ std::tuple<int, std::string> parseArguments(
...
+        else if (str_iter == 3) appConfig.vt_user_str_3 = user_input;
+        else too_many_user_args = true;
+        appConfig.user_args["vt_user_str_" + std::to_string(str_iter)] = key;
+        str_iter++;
+      }
+    }
+  }
+  if (too_many_user_args) {
+    throw std::runtime_error("Only three user-defined arguments of each type (bool, int, or string) are supported.");
+  }
```
cz4rs, line 411: Use `vtAbortIf` here (we almost never use exceptions for error handling in `vt`).

tests/unit/runtime/test_initialization.cc
```diff
@@ -259,65 +261,235 @@ TEST_F(TestInitialization, test_initialize_with_file_args_and_appconfig) {
...
-  arguments::AppConfig appConfig{};
-  appConfig.vt_lb_name = "GreedyLB";
+  EXPECT_EQ(custom_argc, 2);
 
   int this_rank;
   MPI_Comm_rank(comm, &this_rank);
   if (this_rank == 0) {
     std::ofstream cfg_file_{config_file.c_str(), std::ofstream::out | std::ofstream::trunc};
-    cfg_file_ << "vt_lb_name = RandomLB\n";
+    cfg_file_ << "Output Control:\n"
```
cz4rs, line 287: Use raw string literal so you don't have to use `<<` and `\n` everywhere.

tests/unit/runtime/test_initialization.cc
```diff
@@ -259,65 +261,235 @@ TEST_F(TestInitialization, test_initialize_with_file_args_and_appconfig) {
   EXPECT_EQ(custom_argv[2], nullptr);
 }
 
-TEST_F(TestInitialization, test_preconfigure_and_initialization) {
+TEST_F(TestInitialization, test_initialize_with_yaml) {
```
cz4rs, line 264: We need a test that shows order of precedence too. No need to test all combinations, but at least case that has `yaml`, `toml` and command line args.

src/vt/configs/arguments/args.cc
```diff
@@ -185,6 +193,241 @@ std::tuple<int, std::string> parseArguments(
...
+    const std::string& key = it->first.as<std::string>();
+    try {
+      bool user_input = user_options[key].as<bool>();
+      if (bool_iter == 1) appConfig.vt_user_1 = user_input;
+      else if (bool_iter == 2) appConfig.vt_user_2 = user_input;
+      else if (bool_iter == 3) appConfig.vt_user_3 = user_input;
+      else too_many_user_args = true;
+      appConfig.user_args["vt_user_" + std::to_string(bool_iter)] = key;
+      bool_iter++;
+    } catch(...) {
```
cz4rs, line 389: Catch a specific class of exceptions if possible.

src/vt/configs/arguments/args.cc
```diff
@@ -185,6 +193,241 @@ std::tuple<int, std::string> parseArguments(
...
+  appConfig.vt_debug_level = debug_print_configuration["Level"].as<std::string>("terse");
+  appConfig.vt_debug_all = debug_print_configuration["Enable All"].as<bool>(false);
+  appConfig.vt_debug_none = debug_print_configuration["Disable All"].as<bool>(false);
+
+  std::vector<std::string> debug_enables;
+  if (debug_print_configuration["Enable"]) {
+      debug_enables = debug_print_configuration["Enable"].as<std::vector<std::string>>();
+  }
+
+  appConfig.vt_debug_gen = (std::find(debug_enables.begin(), debug_enables.end(), "gen") != debug_enables.end());
```
cz4rs, line 280: You can wrap this `std::find` call in a lambda that captures `debug_enables` and takes the string as a parameter.

lib/yaml-cpp/include/yaml-cpp/null.h
cz4rs: It seems like the author is refusing to fix this upstream, so there's not much else to be done.

src/vt/configs/arguments/args.cc
```diff
@@ -185,6 +193,241 @@ std::tuple<int, std::string> parseArguments(
...
+  app.add_option("--vt_input_config_yaml", appConfig.vt_input_config_yaml, yaml_description);
+}
+
+void parseYaml(AppConfig& appConfig, std::string inputFile) {
+  // Read in the YAML configuration file
+  auto yaml_input = YAML::LoadFile(inputFile);
+
+  // Output control
+  YAML::Node output_control = yaml_input["Output Control"];
+  appConfig.vt_color = output_control["Color"].as<bool>(true);
```
cz4rs, line 207: Consider what happens when user passes `--vt_color=false` in the command line.

Use current value of configuration variable as fallback:
```suggestion
  appConfig.vt_color = output_control["Color"].as<bool>(appConfig.vt_color);
```
This should make sense whatever order of precedence we end up with.

cz4rs (changes requested):
@cwschilly Plenty of smaller changes requested.
I will still try to think of a couple more :wink: 

The "Should we support using external `yaml-cpp`" question should probably be brought up on Slack, because it might get lost among other comments here.

src/vt/configs/arguments/app_config.h
cz4rs: `args.cc` looks like the right place for the conversion code.

src/vt/configs/arguments/args.cc
```diff
@@ -185,6 +193,241 @@ std::tuple<int, std::string> parseArguments(
...
+
+  // Memory Usage Reporting
+  YAML::Node memory_usage_reporting = yaml_input["Memory Usage Reporting"];
+  std::string default_vt_memory_reporters =
+# if (vt_feature_mimalloc != 0)
+  "mimalloc,"
+# endif
+  "mstats,machinfo,selfstat,selfstatm,sbrk,mallinfo,getrusage,ps";
+  appConfig.vt_memory_reporters = memory_usage_reporting["Memory Reporters"].as<std::string>(default_vt_memory_reporters);
+  appConfig.vt_print_memory_each_phase = memory_usage_reporting["Print Memory Each Phase"].as<bool>(false);
```
cz4rs, line 226: Duplication: all the option strings like "Print Memory Each Phase" are repeated `app_config.h`. Use named variables (possibly file scoped?).

lib/yaml-cpp/.clang-format
cwschilly: Fixed

cmake/link_vt.cmake
```diff
@@ -41,7 +42,7 @@ function(link_target_with_vt)
   set(allKeywords ${noValOption} ${singleValArg} ${multiValueArg})
```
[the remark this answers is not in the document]
cwschilly: I could not find it being used anywhere, so I removed it

src/vt/configs/arguments/args.cc
cwschilly: Fixed

tests/unit/test_helpers.h
cwschilly: Fixed

src/vt/configs/arguments/args.cc
cwschilly: Fixed, thanks

src/vt/configs/arguments/app_config.h
cwschilly: Fixed; all conversion code is now in `args.cc`

src/vt/configs/arguments/app_config.h
cwschilly: Fixed, thanks for pointing this out

src/vt/configs/arguments/args.cc
cwschilly: Fixed

tests/unit/runtime/test_initialization.cc
cwschilly: Fixed

tests/unit/runtime/test_initialization.cc
cwschilly: New test added that shows order of precedence of toml->args->yaml

src/vt/configs/arguments/args.cc
cwschilly: Fixed

src/vt/configs/arguments/args.cc
cwschilly: Fixed

src/vt/configs/arguments/args.cc
cwschilly: Fixed

src/vt/configs/arguments/args.cc
cwschilly: Fixed; I've moved everything into `args.cc` and stored all YAML labels as variables at the start of the file.

cmake/load_bundled_libraries.cmake
cwschilly: I don't anticipate any issues with always using bundled version (this is the same logic we use with the CLI library). However, adding support for an external `yaml-cpp` lib will be done in follow-on issue #2322

src/vt/configs/arguments/args.cc
```diff
@@ -118,34 +280,60 @@ std::tuple<int, std::string> parseArguments(
...
+    } catch (CLI::Error &ex) {
+      // Return exit code and message, delaying logic processing of such.
+      // The default exit code for 'help' is 0.
+      std::stringstream yaml_message_stream;
+      int yaml_result = app.exit(ex, yaml_message_stream, yaml_message_stream);
+      return std::make_tuple(yaml_result, yaml_message_stream.str());
+    }
+  }
+
+  // Parse the YAML parameters
```
cz4rs, line 313: We need to document the order of precedence for the users. I'd suggest putting it in the usage string (`VtFormatter::make_usage` in `args.cc`).

src/vt/configs/arguments/args.cc
```diff
@@ -763,6 +1420,10 @@ std::tuple<int, std::string> ArgConfig::parseToConfig(
 
   app.set_help_flag("--vt_help", "Display help");
 
+  // Find a YAML config file, if present
+  addYamlConfig(app, appConfig);
+
+  // Parse remaining args
```
cz4rs, line 1426: Misleading comments: parsing happens later, this is just preparation:
```suggestion
  // setup CLI parsing
  addYamlConfig(app, appConfig);
```

src/vt/configs/arguments/args.cc
```diff
@@ -763,6 +1420,10 @@ std::tuple<int, std::string> ArgConfig::parseToConfig(
 
   app.set_help_flag("--vt_help", "Display help");
```
cz4rs, line 1422: Actually, maybe you could move the `app.set_config(...)` call from line 293 to have the CLI::App setup in one place.

src/vt/configs/arguments/args.cc
```diff
@@ -118,34 +280,60 @@ std::tuple<int, std::string> parseArguments(
...
   // so node 0 can output in the runtime once MPI is init'ed
   if (appConfig.vt_output_config) {
-    appConfig.vt_output_config_str = app.config_to_str(true, true);
+    auto output_file = appConfig.vt_output_config_file;
+    auto config_file_ending = output_file.substr(output_file.size()-4);
+    if (config_file_ending == ".yml" || config_file_ending == "yaml") {
+      appConfig.vt_output_config_str = convertConfigToYamlString(appConfig);
+    } else {
+      appConfig.vt_output_config_str = app.config_to_str(true, true);
+    }
```
cz4rs, line 336: Extract that logic into a helper function:
```suggestion
      appConfig.vt_output_config_str = convertConfigToString(appConfig);
```

src/vt/configs/arguments/args.cc
```diff
@@ -186,6 +374,256 @@ std::tuple<int, std::string> parseArguments(
   return std::make_tuple(-1, std::string{});
 }
 
+void addYamlConfig(CLI::App& app, AppConfig& appConfig) {
```
cz4rs, line 377: Rename this for consistency:
```suggestion
void addYamlArgs(CLI::App& app, AppConfig& appConfig) {
```

src/vt/configs/arguments/args.cc
```diff
@@ -186,6 +374,256 @@ std::tuple<int, std::string> parseArguments(
...
+  auto yaml_description = "Read in a YAML config file for VT";
+  app.add_option("--vt_input_config_yaml", appConfig.vt_input_config_yaml, yaml_description);
+}
+
+void parseYaml(AppConfig& appConfig, std::string inputFile) {
```
cz4rs, line 382: Make the string a const ref.

src/vt/configs/arguments/args.cc
```diff
@@ -186,6 +374,256 @@ std::tuple<int, std::string> parseArguments(
...
+  appConfig.vt_debug_vrt_coll = is_debug_enabled(vt_debug_vrt_coll_label);
+  appConfig.vt_debug_worker = is_debug_enabled(vt_debug_worker_label);
+  appConfig.vt_debug_group = is_debug_enabled(vt_debug_group_label);
+  appConfig.vt_debug_broadcast = is_debug_enabled(vt_debug_broadcast_label);
+  appConfig.vt_debug_objgroup = is_debug_enabled(vt_debug_objgroup_label);
+  appConfig.vt_debug_phase = is_debug_enabled(vt_debug_phase_label);
+  appConfig.vt_debug_context = is_debug_enabled(vt_debug_context_label);
+  appConfig.vt_debug_epoch = is_debug_enabled(vt_debug_epoch_label);
+  appConfig.vt_debug_print_flush = debug_print_configuration[vt_debug_print_flush_label].as<bool>(appConfig.vt_debug_print_flush);
+  // appConfig.vt_debug_replay = debug_print_configuration["Enable replay"].as<bool>(false);
```
cz4rs, line 497: Should this be removed or uncommented?

src/vt/configs/arguments/args.cc
```diff
@@ -186,6 +374,256 @@ std::tuple<int, std::string> parseArguments(
...
+  appConfig.vt_lb_statistics = lb_stats[vt_lb_statistics_label].as<bool>(appConfig.vt_lb_statistics);
+  appConfig.vt_lb_statistics_compress = lb_stats[vt_lb_statistics_compress_label].as<bool>(appConfig.vt_lb_statistics_compress);
+  appConfig.vt_lb_statistics_file = lb_stats[vt_lb_statistics_file_label].as<std::string>(appConfig.vt_lb_statistics_file);
+  appConfig.vt_lb_statistics_dir = lb_stats[vt_lb_statistics_dir_label].as<std::string>(appConfig.vt_lb_statistics_dir);
+
+  appConfig.vt_lb_self_migration = load_balancing[vt_lb_self_migration_label].as<bool>(appConfig.vt_lb_self_migration);
+  appConfig.vt_lb_spec = load_balancing[vt_lb_spec_label].as<bool>(appConfig.vt_lb_spec);
+  appConfig.vt_lb_spec_file = load_balancing[vt_lb_spec_file_label].as<std::string>(appConfig.vt_lb_spec_file);
+
+  // appConfig.vt_lb_run_lb_first_phase = load_balancing["vt_lb_run_lb_first_phase"].as<bool>(false);
```
cz4rs, line 531: Commented out.

src/vt/configs/arguments/args.cc
```diff
@@ -186,6 +374,256 @@ std::tuple<int, std::string> parseArguments(
...
+}
+
+void parseYaml(AppConfig& appConfig, std::string inputFile) {
+  // Read in the YAML configuration file
+  auto yaml_input = YAML::LoadFile(inputFile);
+
+  // Output control
+  YAML::Node output_control = yaml_input["Output Control"];
+  appConfig.vt_color = output_control[vt_color_label].as<bool>(appConfig.vt_color);
+  appConfig.vt_no_color = not output_control[vt_color_label].as<bool>(not appConfig.vt_no_color);
```
cz4rs, line 389: This uses `not` twice.

src/vt/configs/arguments/args.cc
```diff
@@ -52,12 +53,170 @@
...
 namespace vt { namespace arguments {
 
 // Temporary variables used only for parsing artifacts.
 namespace {
 
+// Set the YAML labels for all CLI args
+
+// Output Control
+static const std::string vt_color_label = "Color";
+static const std::string vt_no_color_label = "Color";
```
cz4rs, line 67: This should probably be:
```suggestion
static const std::string vt_no_color_label = "No color";
```

src/vt/configs/arguments/args.cc
cz4rs: Also, instead of negating value for `vt_color_label` we should probably just use `vt_no_color_label`.

cz4rs: Alright, I took a better look at it and we can improve :crossed_fingers: the code here with a following lambda:
```cpp
  // read the config value from YAML using default (current) value as fallback
  auto update_config = [](
    auto& config_entry, std::string const& label, YAML::Node const& node
  ) {
    using config_type = std::remove_reference_t<decltype(config_entry)>;
    config_entry = node[label].as<config_type>(config_entry);
  };
```

This should make the code less repetitive and slightly easier to follow:
```cpp
  update_config(appConfig.vt_color, vt_color_label, output_control);
```

src/vt/configs/arguments/args.cc
```diff
@@ -186,6 +374,256 @@ std::tuple<int, std::string> parseArguments(
...
+  appConfig.vt_debug_level = debug_print_configuration[vt_debug_level_label].as<std::string>(appConfig.vt_debug_level);
+  appConfig.vt_debug_all = debug_print_configuration[vt_debug_all_label].as<bool>(appConfig.vt_debug_all);
+  appConfig.vt_debug_none = debug_print_configuration[vt_debug_none_label].as<bool>(appConfig.vt_debug_none);
+
+  std::vector<std::string> debug_enables;
+  if (debug_print_configuration["Enable"]) {
+      debug_enables = debug_print_configuration["Enable"].as<std::vector<std::string>>();
+  }
+
+  auto is_debug_enabled = [&debug_enables](const std::string& debug_type) {
```
cz4rs, line 461: Maybe add a comment here that the default value for all `vt_debug_` categories is `false`, so we don't need to check current value.

src/vt/configs/arguments/args.cc
```diff
@@ -186,6 +374,256 @@ std::tuple<int, std::string> parseArguments(
...
+        else if (str_iter == 3)
+          appConfig.vt_user_str_3 = user_input;
+        else too_many_user_args = true;
+        user_args_labels["vt_user_str_" + std::to_string(str_iter)] = key;
+        str_iter++;
+      }
+    }
+  }
+
+  vtAbortIf(too_many_user_args, "Only three user-defined arguments of each type (bool, int, or string) are supported.");
```
cz4rs, line 606: Extract that into a separate function.

src/vt/configs/arguments/args.cc
```diff
@@ -696,6 +1134,225 @@ void addThreadingArgs(
...
+      throw std::runtime_error("Argument type not recognized.");
+    }
+}
+
+std::string convertConfigToYamlString(AppConfig& appConfig) {
+  // First, create a converter vector of tuples {Node(s), Key, Value}
+  std::vector<std::tuple<std::string, std::string, variantArg_t>> cli_to_yaml_args = {
+      // Output Control
+      {"Output Control", vt_color_label, static_cast<variantArg_t>(appConfig.vt_color)},
+      {"Output Control", vt_no_color_label, static_cast<variantArg_t>(not appConfig.vt_no_color)},
```
cz4rs, line 1163: Incorrect negation.

src/vt/configs/arguments/args.cc
```diff
@@ -696,6 +1134,225 @@ void addThreadingArgs(
...
+    // The rest are straightforward
+    else {
+      auto current_node = output_config_yaml[yaml_node];
+      addVariantToNode(current_node, yaml_key, yaml_val);
+    }
+  }
+  std::ostringstream yaml_stream;
+  yaml_stream << output_config_yaml;
+  std::string output_config_string = yaml_stream.str();
+  return output_config_string;
```
cz4rs, line 1353:
```suggestion
  return yaml_stream.str();
```

src/vt/configs/arguments/args.cc
cwschilly: Fixed

src/vt/configs/arguments/args.cc
cwschilly: Fixed

src/vt/configs/arguments/args.cc
cwschilly: Fixed

src/vt/configs/arguments/args.cc
cwschilly: Fixed

src/vt/configs/arguments/args.cc
cwschilly: Uncommented; fixed

src/vt/configs/arguments/args.cc
cwschilly: This parameter isn't present in `appConfig` so I've removed this line

src/vt/configs/arguments/args.cc
cwschilly: Fixed, thanks!

src/vt/configs/arguments/args.cc
cwschilly: Fixed

src/vt/configs/arguments/args.cc
cwschilly: Fixed

src/vt/configs/arguments/args.cc
cwschilly: Fixed

src/vt/configs/arguments/args.cc
```diff
@@ -696,6 +1134,225 @@ void addThreadingArgs(
...
+  // Create an empty node that we will populate
+  YAML::Node output_config_yaml;
+
+  // Then convert to YAML
+  for (const auto& yaml_data : cli_to_yaml_args) {
+
+    // Unpack the yaml data
+    auto yaml_node = std::get<0>(yaml_data);
+    auto yaml_key = std::get<1>(yaml_data);
+    auto yaml_val = std::get<2>(yaml_data);
```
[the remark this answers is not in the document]
cwschilly: Fixed

src/vt/configs/arguments/args.cc
cwschilly: Fixed

src/vt/configs/arguments/args.cc
cwschilly: Fixed

cz4rs (approved):
Looks good to me!

lifflander (approved):
This is ready to merge.

lifflander:
@cwschilly Can you edit the wiki to add yaml-cpp as one of our dependencies?

lifflander:
Can you rebase this?

PR #2242, opened 2024-01-17
2241 update version file and add asm change

### LINKED ISSUE #2241: Update VERSION file and add ASM change
Update from `1.3.0` to `1.4.0` and backport ASM change

stmcgovern:
Fixes #2241.


lifflander (approved):

nlslatt (approved):

nmm0:
Why does the value in ASM need to be increased? We should document it since it's a bit of magic number.

PR #2244, opened 2024-01-25
2243 allow replay to repeat phases using modulus

### LINKED ISSUE #2243: Allow replay to repeat phases using modulus
**What Needs to be Done?**
Allow experimenting with iterative load balancers iteration by iteration by using LB data from json as input for multiple phases. Do this using a phase modulus. For example, using a modulus of 2 would use phase 0 json data as input for phases 0, 2, 4, etc. and phase 1 json data as input for phases 1, 3, 5, etc.

nlslatt:
Closes #2243


lifflander (approved):
This looks good to me!

PR #2252, opened 2024-02-28
#2250: Remove count field from phases metadata

### LINKED ISSUE #2250: Remove `count` field from JSON.
**What Needs to be Done?**

Remove `count` from JSON metadata but leave it in JSON validator.

thearusable:
Closes #2250


scripts/JSON_data_files_validator.py
```diff
@@ -54,7 +54,7 @@ def _get_valid_schema(self) -> Schema:
                         'num_nodes': int,
                     },
                     Optional('phases'): {
-                        'count': int,
+                        Optional('count'): int,
```
JacobDomagala, line 57: We don't want to remove it?

thearusable: Yes, if I recall correctly the discussion from few meettings ago, we need to keep the count field in the validator because there was already some files generated and they need to be still valid.

JacobDomagala: I see, thanks for the clarification!

JacobDomagala (approved):

nlslatt (approved):
Looks good

PR #2253, opened 2024-03-13
#2063: Convert buffer pointers to `std::byte*`

### LINKED ISSUE #2063: Convert buffer pointers to std::byte*
**What Needs to be Done?**

C++17 has a safe-ish vocabulary type `std::byte` for referring to objects/object representations that should not be operated on except as uninterpreted bits/bytes. We ought to adopt it.

**Additional context**

thearusable:
Closes #2063


lifflander (approved):
Looks good to me

PR #2255, opened 2024-03-25
#2254: startup: add missing mpi header include

### LINKED ISSUE #2254: Add missing MPI header file
**Describe the bug**
Need to add `mpi.h` include to `startup.h` for the `MPI_Comm` types.

nlslatt:
Fixes #2254.

### DIFF
--- a/src/vt/collective/startup.h
+++ b/src/vt/collective/startup.h
@@ -47,6 +47,7 @@
 #include "vt/config.h"
 #include "vt/runtime/runtime_headers.h"
 #include "vt/configs/arguments/argv_container.h"
+#include <mpi.h>
 
 namespace vt {
 


JacobDomagala:
Would forward declaring `MPI_Comm` resolve the issue, or we actually need the header there?

nlslatt:
> Would forward declaring `MPI_Comm` resolve the issue, or we actually need the header there?

This was a change I made in my working copy a while back. Now that I think about it, the missing header file may not have been related to my problem at all. I think this was when a problem with modules in my environment was mixing mpich into my OpenMPI and breaking everything. We can probably get away with either your suggestion or doing nothing at all.

lifflander:
> > Would forward declaring `MPI_Comm` resolve the issue, or we actually need the header there?
> 
> This was a change I made in my working copy a while back. Now that I think about it, the missing header file may not have been related to my problem at all. I think this was when a problem with modules in my environment was mixing mpich into my OpenMPI and breaking everything. We can probably get away with either your suggestion or doing nothing at all.

I think it does not hurt to just explicitly include it here. It will get included later on without question, so it shouldn't increase compile times to include it here I believe.

lifflander (approved):

PR #2263, opened 2024-03-27
#2260: Enable one verbose build for each compiler type

### LINKED ISSUE #2260: One CI build should compile with verbose debug prints to check that they compile (at the least)
**What Needs to be Done?**

Add to one of the CI builds the flag that enables verbose debug prints to check that they compile.

thearusable:
Closes #2260


JacobDomagala:
Should we use `PR tests (gcc-12, ubuntu, mpich)` for verbose? It's the PR that runs even in draft mode.

thearusable:
> Should we use `PR tests (gcc-12, ubuntu, mpich)` for verbose? It's the PR that runs even in draft mode.

@JacobDomagala Yes, I think that's a great idea. I'll make this gcc pipeline a verbose one.

thearusable:
Debug verbose logs are enabled for:
- clang-14
- gcc-12
- icpx
- nvcc-12

Verbose debug logs are being compiled for jobs on those compilers but they won't be printed on CI. Enabling verbose logging for all modules with `--vt_debug_all` will result in very verbose output which is hard to analyze.

thearusable:
@lifflander / @nlslatt - I poked a little bit to check why you have the issue with missing verbose logs. 

I only found that not all of the parameters are being passed when building docker images. But that's all I found. All configuration in CMake looks good to me. Every option is properly stored in CMake's cache and it's properly overridden when we pass a different value for a flag.

ci/azure/azure-intel-oneapi-icpx-ubuntu-mpich.yml
```diff
@@ -48,11 +48,12 @@ variables:
   VT_PRODUCTION_BUILD: 0
   VT_FCONTEXT: 0
   VT_ZOLTAN: 0
-  VT_CI_BUILD: 1
+  VT_CI_BUILD: 0
```
lifflander, line 51: I really don't like lying about whether this is a CI build, since that option may be used in the future for other things. Is there a different approach we can take to this?

lifflander (changes requested):

lifflander (approved):

PR #2278, opened 2024-04-30
2201 implement memory aware temperedlb in vt rebased (new version)

### LINKED ISSUE #2201: Implement memory-aware TemperedLB in VT
**What Needs to be Done?**

- [ ] Start with the swapping approach.

lifflander:
Fixes #2201


nlslatt:
We need to make running with `SwapClusters` and rollback enabled an invalid configuration. Either error out or turn off rollback when using `SwapClusters`.

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
```diff
@@ -1295,14 +2150,584 @@ void TemperedLB::decide() {
...
+    auto const src_shared_id = std::get<0>(best_swap);
+    auto const try_shared_id = std::get<1>(best_swap);
+
+    vt_debug_print(
+      normal, temperedlb,
+      "best_c_try={}, swapping {} for {} on rank ={}\n",
+      best_c_try, src_shared_id, try_shared_id, try_rank
+    );
+
+    auto const& give_data = removeClusterToSend(src_shared_id);
```
cz4rs, line 2394:
```suggestion
    // FIXME C++20: use structured binding
    auto const& give_data = removeClusterToSend(src_shared_id);
```

cz4rs:
Note: `TestCheckpoint.test_checkpoint_in_place_2` starts failing with https://github.com/DARMA-tasking/vt/pull/2278/commits/f9ba1180e3a724a650aa45f4b58ec6331415ed93.

cz4rs:
e7581a5 seems to fix the failing tests, although I'm not certain if this is the right solution.
PR also requires a rebase, but lets sync first to make sure no one has anything in progress.

cz4rs:
TODO:
- make sure that objects that are not migratable are also accounted for (load)
- do not migrate objects that are not migratable (see `CollectionManager::migrateOut`)
- do a sanity check and provide a clear error message in `BaseLB::normalizeReassignments`

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
```diff
@@ -1218,28 +2073,28 @@ void TemperedLB::decide() {
...
           verbose, temperedlb,
-          "TemperedLB::decide: selected_node={}, load_info_.size()={}\n",
+          "TemperedLB::originalTransfer: selected_node={}, load_info_.size()={}\n",
           selected_node, load_info_.size()
         );
 
+	// Find load of selected node
         auto load_iter = load_info_.find(selected_node);
         vtAssert(load_iter != load_info_.end(), "Selected node not found");
+        auto& selected_load = load_iter->second.load;
```
lifflander, line 2089: Check to make sure that the object is migratable (should be accessible in `load_iter`) and set `eval` to false if it isn't so it doesn't attempt to transfer it.

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
ppebay, line 1: `452/462: if (avg > 0.0000000001)`
this lower bound should not be hardcoded, but rather use either a machine-dependent absolute quantity (i.e., square root of machine epsilon) or, better yet, a lowed bound that uses first principle: one such example would be to use the smallest object size divided by the number of ranks (or nodes).

ppebay, line 1: `245:    "targetpole"`
I do not think we have a reference to the concept of _long pole_ in any of our publications (conference proceedings & tech reports) on the tempered load-balancer. I remember having discussed it, but I do not recall having seen any real justification written for it, nor even demonstrated evidence that this yielded better LB outcomes in such configurations with large objects ("long poles"). 
If we want to retain this alternate target we should document its relevance; otherwise we should simply drop it.

ppebay, line 1: `897/907: std::vector<double> TemperedLB::createCMF(NodeSetType const& under) {` (and subsequent lines)
The basis for this CMF construction is in Lifflander _et al._, Cluster'21, Algorithm 2:
Screenshot 2024-05-29 at 09 05 24

We further observe that the "original" CMF in the above (ll. 22--23) corresponds to the empirical PDF construction of the seminal Menon _et al._ 2013 paper:
Screenshot 2024-05-29 at 09 37 35

* regarding  `case CMFTypeEnum::Original:`
notwithstanding the caveat in an above comment regarding the "long pole", 
```
913/923: factor = 1.0 / target_max_load_;
``` 
ensure that `factor` is the inverse of the average load in the "non-pole" case.
```
if (target_pole_) {
    ...
    target_max_load_ = (pole > avg ? pole : avg);
  } else {
    target_max_load_ = avg;
  }
```
Thus comparing Algorithm 2 for the "original" case [in black] with the `createCMF()` implementation [in blue]
is made by remarking the following identities in the "non-long-pole" case [in red]: 

Screenshot 2024-05-29 at 10 19 34

This allows for the following substitution in the RHS of the `sum_p` entries:

Screenshot 2024-05-29 at 10 25 33

Thanks to this reformulation of the z normalization factor of Algorithm 2, in terms of entries in the `sum_p` vector, finally yields:

Screenshot 2024-05-29 at 10 35 29

In other words, thanks to the normalization in ll. 946/956--949/959:
```
  // Normalize the CMF
  for (auto& elm : cmf) {
    elm /= sum_p;
  }
```
the computed `elm` vector entries are indeed equal to the $\varphi_j$ in Algorithm 2, for the last entry in the `cmf` vector had indeed and index of $\lvert S^p \rvert$, the size of:
Screenshot 2024-05-29 at 10 42 19
i.e., the size of the `for (auto&& pe : under)` loop that incrementally builds `cmf`.

We can thus conclude that, in the *"non-long-pole" case*, the implementation is correct and corresponds to the "original" CMF construction of Algorithm 2. For the "long-pole" case, we do not have any theoretical case to compare against, as discussed in earlier comment.

nlslatt: Phil's justification for experimenting with it was that, when the long pole load exceeds the rank-average load, there is likely no benefit to reducing any rank's load below the long pole. However, that assumes that there are no synchronization points (and thus idle time) within a phase. That assumption does not hold true with EMPIRE, which is the only app we ever used to experiment with `targetpole`. There might be memory benefits to using `targetpole`...

ppebay: Thanks @nlslatt for the comment regarding the original rationale (and also the lack of actual validation with an application using it). I will defer it to @lifflander and you to decide as to whether this should be removed or not.

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
```diff
@@ -449,10 +537,12 @@ void TemperedLB::runLB(LoadType total_load) {
     target_max_load_ = avg;
   }
 
-  if (avg > 0.0000000001) {
+  // Use an absolute minimal bound on average load to load-balance
+  if (avg > 1e-10) {
```
ppebay, line 541: Cf. above comment regarding this usage of hard-coded lower-bounds.

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
ppebay: Moving this discussion to inside code.

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
```diff
@@ -880,14 +1721,26 @@ void TemperedLB::propagateIncomingSync(LoadMsgSync* msg) {
...
+
+  for (auto&& elm : msg->getNodeInfo()) {
     if (new_load_info_.find(elm.first) == new_load_info_.end()) {
       new_load_info_[elm.first] = elm.second;
 
-      if (isUnderloaded(elm.second)) {
+      if (isUnderloaded(elm.second.load)) {
         new_underloaded_.insert(elm.first);
       }
     }
```
ppebay, line 1746: Concerning `std::vector<double> TemperedLB::createCMF(NodeSetType const& under) {` below, and subsequent lines, the basis for this CMF construction is in Lifflander _et al._, Cluster'21, Algorithm 2:
Screenshot 2024-05-29 at 09 05 24

We further observe that the "original" CMF in the above (ll. 22--23) corresponds to the empirical PDF construction of the seminal Menon _et al._ 2013 paper:
Screenshot 2024-05-29 at 09 37 35

* regarding  `case CMFTypeEnum::Original:`
notwithstanding the caveat in an above comment regarding the "long pole", 
```
factor = 1.0 / target_max_load_;
```
ensure that `factor` is the inverse of the average load in the "non-pole" case.
```
if (target_pole_) {
    ...
    target_max_load_ = (pole > avg ? pole : avg);
  } else {
    target_max_load_ = avg;
  }
```
Thus comparing Algorithm 2 for the "original" case [in black] with the `createCMF()` implementation [in blue]
is made by remarking the following identities in the "non-long-pole" case [in red]: 

Screenshot 2024-05-29 at 10 19 34

This allows for the following substitution in the RHS of the `sum_p` entries:

Screenshot 2024-05-29 at 10 25 33

Thanks to this reformulation of the z normalization factor of Algorithm 2, in terms of entries in the `sum_p` vector, finally yields:

Screenshot 2024-05-29 at 10 35 29

In other words, thanks to the normalization in ll. 946/956--949/959:
```
  // Normalize the CMF
  for (auto& elm : cmf) {
    elm /= sum_p;
  }
```
the computed `elm` vector entries are indeed equal to the $\varphi_j$ in Algorithm 2, for the last entry in the `cmf` vector had indeed and index of $\lvert S^p \rvert$, the size of:
Screenshot 2024-05-29 at 10 42 19
i.e., the size of the `for (auto&& pe : under)` loop that incrementally builds `cmf`.

We can thus conclude that, in the *"non-long-pole" case*, the implementation is correct and corresponds to the "original" CMF construction of Algorithm 2. For the "long-pole" case, we do not have any theoretical case to compare against, as discussed in earlier comment.

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
```diff
@@ -1219,47 +2074,50 @@ void TemperedLB::decide() {
         }
         // Rebuild the CMF with the new loads taken into account
         auto cmf = createCMF(under);
+
```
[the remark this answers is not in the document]
nlslatt: See lines 1768 and 1771.

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
nlslatt: I'm not remembering all the details right now, but I think `NormBySelf` was something I suggested trying when the normalizing factor we had been using was displaying pathological behavior. I don't think `NormBySelf` ended up working well.

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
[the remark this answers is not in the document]
ppebay: Sorry I realize this was unclear. 

These enum values are indeed used to compute the CMF (cf. my comments above regarding how they are computed, and possibly validated -- or not in the case of `NormBySelf`).

What I meant here is "not used in `originalTransfer()`, as a result the `if... else if` dichotomy between ` CMFTypeEnum::Original` and `CMFTypeEnum::NormByMaxExcludeIneligible`: for the 2 other possible values, `under` won't be defined here. Is this normal/expected?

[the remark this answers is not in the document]
nlslatt: `under` is computed for all cases on line 2046. It is *re-computed* between transfers (with some redundancy before the first transfer) only for a subset of the CMF types, in the lines you noticed.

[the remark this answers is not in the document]
ppebay: Yes, but what is the rationale for doing this re-compute between transfers **only** for these two types, and not for the others?

Let me ask why I am thinking this may be incorrect: in the case of `NormBySelf()`, the CMF uses a denominator in one of its factors, that depends on a quantity that can change between transfers:
```
factor = 1.0 / this_new_load_;
```
This new (post-transfer) load becomes outdated as soon as one object is either sent out or received in. Therefore, shouldn't the set of "under" loaded peers be updated in this case as well? If not what is the reason for not doing it?

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
ppebay: Understood, thank you. Shouldn't we thus deprecate it?

nlslatt: Getting rid of `NormBySelf` is fine with me.

ppebay: Making a note at top of issue and resolving this discussion.

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
lifflander: I don't think this should be removed. For the load-only case, the `targetpole` is truly the best you can do. This PR doesn't change this part of the code (beyond mechanical changes) so I think removing this would be outside the scope of this PR anyway.

lifflander: The intention wasn't for you to review the whole TemperedLB, it was to review the new work-base algorithm.

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
```diff
@@ -469,9 +559,512 @@ void TemperedLB::runLB(LoadType total_load) {
...
+    to_add.home_node != uninitialized_destination
+  ) {
+    node_work += delta * to_add.edge_weight;
+  }
+
+  double node_inter_send = info.inter_send_vol;
+  double node_inter_recv = info.inter_recv_vol;
+  node_work -= beta * std::max(node_inter_send, node_inter_recv);
+
+  // All edges outside the to_remove cluster that are also off the node need to
```
ppebay, line 1009: We never wrote the update formulæ anywhere in a reference document, so here is the derivation:

Screenshot 2024-06-04 at 05 37 55

For reference this is how this was done in LBAF for **non-local** (i.e., "inter") communications (prior to the current, full-information method that has been used since in order to deal with memory updates but at the cost of losing the update formulæ for communication volumes):
```
        v_src_to_src = 0.
        v_src_to_dst = 0.
        v_src_to_oth = 0.
        v_src_from_src = 0.
        v_src_from_dst = 0.
        v_src_from_oth = 0.

        # Iterate over objects to be transferred
        for o in objects:
            # Skip objects without a communicator
            comm = o.get_communicator()
            if not isinstance(comm, ObjectCommunicator):
                continue

	    # Tally sent communication volumes by destination
            for k, v in comm.get_sent().items():
                # Skip items sent to object list
                if k in objects:
                    continue

                # Categorize sent volume
                if k.get_rank_id() == src_id:
                    v_src_to_src += v
                elif k.get_rank_id() == dst_id:
                    v_src_to_dst += v
                else:
                    v_src_to_oth += v

            # Tally received communication volumes by source
            for k, v in comm.get_received().items():
                # Skip items received from object list
                if k in objects:
                    continue

                # Categorize received volume
                if k.get_rank_id() == src_id:
                    v_src_from_src += v
                elif k.get_rank_id() == dst_id:
                    v_src_from_dst += v
                else:
                    v_src_from_oth += v

        # Initialize volumes with pre-transfer communications
        values_src["sent volume"] = r_src.get_sent_volume()
        values_dst["sent volume"] = r_dst.get_sent_volume()
        values_src["received volume"] = r_src.get_received_volume()
        values_dst["received volume"] = r_dst.get_received_volume()

        # Update volumes by transferring non-local communications
        values_src["sent volume"] -= v_src_to_dst + v_src_to_oth
        values_dst["sent volume"] += v_src_to_src + v_src_to_oth
        values_src["received volume"] -= v_src_from_dst + v_src_from_oth
        values_dst["received volume"] += v_src_from_src + v_src_from_oth

        # Swap sent/received volumes for local commmunications
        values_src["sent volume"] += v_src_from_src
        values_dst["sent volume"] -= v_src_from_dst 
        values_src["received volume"] += v_src_to_src
        values_dst["received volume"] -= v_src_to_dst
```

@lifflander @nlslatt we should discuss how this maps into the 4 update loops below

ppebay (changes requested):
I believe that at least one error was found; and at least one other comment deserves further discussion.

Other comments are optional.

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
cz4rs: Done, the decision is now made based on `eval` and `is_migratable`.

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
```diff
@@ -1219,47 +2074,50 @@ void TemperedLB::decide() {
...
+          is_migratable,
+          obj_load,
+          target_max_load_,
+          this_new_load_,
           eval
         );
 
-        if (eval) {
+	// Decide about proposed migration based on criterion evaluation
+        if (is_migratable and eval) {
```
cz4rs, line 2118: We could skip the evaluation of criterion if the object is not migratable. I have kept those separate for the sake of the debug print above.

src/vt/vrt/collection/balance/baselb/baselb.cc
```diff
@@ -143,6 +143,12 @@ std::shared_ptr<const balance::Reassignment> BaseLB::normalizeReassignments() {
     auto const new_node = std::get<1>(transfer);
     auto const current_node = obj_id.curr_node;
 
+    if (not obj_id.isMigratable()) {
+      vt_debug_print(
+        verbose, lb, "BaseLB::normalizeReassignments(): obj not migratable\n"
+      );
+      continue;
```
cz4rs, line 150: Should this be changed to an assertion?

cz4rs: TODO: use abort if

lifflander: Let's make this a `vtAbortIf`

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
```diff
@@ -469,9 +559,512 @@ void TemperedLB::runLB(LoadType total_load) {
...
+  double node_work = info.work;
+
+  // Remove/add clusters' load factor from work model
+  node_work -= alpha * to_remove.load;
+  node_work += alpha * to_add.load;
+
+  // Remove/add clusters' intra-comm
+  double const node_intra_send = info.intra_send_vol;
+  double const node_intra_recv = info.intra_recv_vol;
+  node_work -= delta * std::max(node_intra_send, node_intra_recv);
```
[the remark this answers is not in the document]
cz4rs: FIXME: confirmed this should be `gamma`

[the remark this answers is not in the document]
lifflander: @cz4rs Let's make this change

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
```diff
@@ -469,9 +559,512 @@ void TemperedLB::runLB(LoadType total_load) {
...
+    node_intra_recv - to_remove.intra_recv_vol + to_add.intra_recv_vol
+  );
+
+  // Uninitialized destination means that the cluster is empty
+  // If to_remove was remote, remove that component from the work
+  if (
+    to_remove.home_node != node and
+    to_remove.home_node != uninitialized_destination
+  ) {
+    node_work -= delta * to_remove.edge_weight;
```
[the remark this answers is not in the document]
lifflander: I believe this is correct.

src/vt/vrt/collection/balance/baselb/baselb.cc
cz4rs: Done.

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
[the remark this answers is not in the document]
cz4rs: Done.

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
```diff
@@ -469,9 +559,512 @@ void TemperedLB::runLB(LoadType total_load) {
...
+
+  // Remove/add clusters' load factor from work model
+  node_work -= alpha * to_remove.load;
+  node_work += alpha * to_add.load;
+
+  // Remove/add clusters' intra-comm
+  double const node_intra_send = info.intra_send_vol;
+  double const node_intra_recv = info.intra_recv_vol;
+  node_work -= delta * std::max(node_intra_send, node_intra_recv);
+  node_work += delta * std::max(
```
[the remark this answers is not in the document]
cz4rs: Done.

cz4rs:
I have applied all the fixes discussed yesterday:
- removed `CMFTypeEnum::NormBySelf` and related code
- replaced `delta` with `gamma` in `TemperedLB::computeWorkAfterClusterSwap`
- switched to `vtAbortIf` in `BaseLB::normalizeReassignments`

CI seems fine, this still needs a rebase, but I'm holding off until the reviews are finished.

cz4rs:
@lifflander Please have a look at @ppebay comments and resolve them as needed. Maybe we need to create follow up issues based on some of them?

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
```diff
@@ -469,9 +559,512 @@ void TemperedLB::runLB(LoadType total_load) {
...
+              serialized_bytes = *val;
+            } else {
+              vtAbort(
+                "\"task_serialized_bytes\" in variant does not match double"
+              );
+            }
+          }
+          if (key == "rank_working_bytes") {
+            if (BytesType const* val = std::get_if<BytesType>(&variant)) {
+              rank_bytes_ = *val;
```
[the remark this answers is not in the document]
lifflander: Yes, it does.

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
```diff
@@ -469,9 +559,512 @@ void TemperedLB::runLB(LoadType total_load) {
...
+  }
+
+  // If to_add is now remote, add that component to the work
+  if (
+    to_add.home_node != node and
+    to_add.home_node != uninitialized_destination
+  ) {
+    node_work += delta * to_add.edge_weight;
+  }
+
```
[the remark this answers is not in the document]
lifflander: @cz4rs Can you do this?

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
```diff
@@ -536,7 +1253,42 @@ void TemperedLB::doLBStages(LoadType start_imb) {
...
+        }
+      }
+
+      // Execute transfer stage
+      switch (transfer_type_) {
+      case TransferTypeEnum::Original:
+        originalTransfer();
+        break;
+      case TransferTypeEnum::Recursive:
+	vtAbort("TemperedLB:: Unimplemented transfer type: Recursive");
```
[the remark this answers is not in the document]
lifflander: Created issue.

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
lifflander: cf. https://github.com/DARMA-tasking/vt/issues/2299

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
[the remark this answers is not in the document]
cz4rs: Done.

tools/1959-tasks/ccm-lb-delta-1e-11.config
```diff
@@ -0,0 +1 @@
+0 TemperedLB transfer=SwapClusters knowledge=Complete rollback=false iters=12 memory_threshold=66000000000.0 delta=0.00000000001
```
cz4rs, line 1: We have 3 commits marked with `NOT to merge`: 4d4238c0d59bd12d5b00f02982c279599973b0d2 af8b4130d91ba9dfa77c6b3ac76545d9d5b8e2f6 0a967017c07f31110a81be17413ffe4eef2e4ba0. Should these be reverted?

src/vt/configs/arguments/args.cc
```diff
@@ -504,6 +505,7 @@ void addLbArgs(CLI::App& app, AppConfig& appConfig) {
   auto lbasm = app.add_flag("--vt_lb_self_migration",   appConfig.vt_lb_self_migration,      lb_self_migration);
   auto lbspec = app.add_flag("--vt_lb_spec",            appConfig.vt_lb_spec,                lb_spec);
   auto lbspecfile = app.add_option("--vt_lb_spec_file", appConfig.vt_lb_spec_file,           lb_spec_file)->capture_default_str()->check(CLI::ExistingFile);
+  auto lb_first_phase = app.add_flag("--vt_lb_run_lb_first_phase", appConfig.vt_lb_run_lb_first_phase, lb_first_phase_info);
```
cz4rs, line 508: This has no corresponding unit test. It should be possible to use existing tests like `test_load_balancer_other_keep_last_elm` does.

cz4rs: See 72d249783.

src/vt/elm/elm_comm.h
```diff
@@ -107,12 +111,25 @@ struct CommKey {
...
+  ) : nto_(in_home), shared_id_(in_shared_id), cat_(CommCategory::ReadOnlyShared)
+  { }
+
   ElementIDStruct from_ = {};
   ElementIDStruct to_   = {};
 
   ElementIDStruct edge_id_ = {};
   NodeType nfrom_          = uninitialized_destination;
   NodeType nto_            = uninitialized_destination;
+  int shared_id_            = -1;
```
cz4rs, line 132: Magic number. Use a named constant instead.

src/vt/vrt/collection/balance/lb_data_holder.cc
```diff
@@ -255,6 +255,22 @@ std::unique_ptr<nlohmann::json> LBDataHolder::toJson(PhaseType phase) const {
...
+        break;
+      }
+      case elm::CommCategory::WriteShared: {
+        j["communications"][i]["type"] = "WriteShared";
+        j["communications"][i]["to"]["type"] = "node";
+        j["communications"][i]["to"]["id"] = key.toNode();
+        j["communications"][i]["from"]["type"] = "shared_id";
+        j["communications"][i]["from"]["id"] = key.sharedID();
+        break;
+      }
```
cz4rs, line 273: Reduce duplication: use `if` `else` for `type` like above (or a ternary expression).

src/vt/vrt/collection/balance/model/load_model.h
```diff
@@ -215,6 +215,17 @@ struct LoadModel
...
+  /**
+   * \brief Provide all the comm info for a given phase
+   *
+   * \param[in] when the interval in which comm is desired
+   *
+   * \return the comm info
+   */
+  virtual CommMapType getComm([[maybe_unused]] PhaseOffset when) const {
+    return CommMapType{};
+  }
```
cz4rs, line 227: Check if this can be made pure virtual:
```suggestion
  virtual CommMapType getComm([[maybe_unused]] PhaseOffset when) const =0;
```

src/vt/vrt/collection/balance/node_lb_data.cc
```diff
@@ -273,6 +273,9 @@ getRecvSendDirection(elm::CommKeyType const& comm) {
   case elm::CommCategory::CollectiveToCollectionBcast:
   case elm::CommCategory::LocalInvoke:
     return std::make_pair(ElementIDType{}, ElementIDType{});
+  case elm::CommCategory::WriteShared:
+  case elm::CommCategory::ReadOnlyShared:
+    return std::make_pair(ElementIDType{}, ElementIDType{});
```
cz4rs, line 278: This is no different that previous cases:
```suggestion
  case elm::CommCategory::CollectiveToCollectionBcast:
  case elm::CommCategory::LocalInvoke:
  case elm::CommCategory::WriteShared:
  case elm::CommCategory::ReadOnlyShared:
    return std::make_pair(ElementIDType{}, ElementIDType{});
```

src/vt/vrt/collection/balance/model/load_model.h
cz4rs: Let's leave it as is, there are some stub models that don't implement it.

src/vt/vrt/collection/balance/temperedlb/temperedlb.h
```diff
@@ -59,6 +59,13 @@
 
 namespace vt { namespace vrt { namespace collection { namespace lb {
 
+struct WorkBreakdown {
+  double work = 0;
+  double intra_send_vol = 0, intra_recv_vol = 0;
+  double inter_send_vol = 0, inter_recv_vol = 0;
+  double shared_vol = 0;
+};
```
cz4rs, line 67: Should we bother with replacing `double` with `LoadType` and `CommBytesType` here?

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
```diff
@@ -469,11 +554,515 @@ void TemperedLB::runLB(LoadType total_load) {
...
+        BytesType serialized_bytes = 0;
+        for (auto const& [key, variant] : data_map) {
+          if (key == "shared_id") {
+            // Because of how JSON is stored this is always a double, even
+            // though it should be an integer
+            if (double const* val = std::get_if<double>(&variant)) {
+              shared_id = static_cast<int>(*val);
+            } else {
+              vtAbort("\"shared_id\" in variant does not match double");
+            }
```
cz4rs, line 596: Since we always want to extract double (either explicitly or with `BytesType`), we can do a single `std::get_if<double>` and `vtAbort` using `key` value in error message instead of repeating this code.

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
```diff
@@ -1296,14 +2146,586 @@ void TemperedLB::decide() {
...
+  if (theConfig()->vt_debug_temperedlb) {
+    int n_rejected = 0;
+    auto remote_block_count = getRemoteBlockCountHere();
+    runInEpochCollective("TemperedLB::swapClusters -> compute rejection", [=] {
+      proxy_.allreduce<&TemperedLB::rejectionStatsHandler, collective::PlusOp>(
+        n_rejected, n_transfers_swap_, remote_block_count
+      );
+    });
+  }
+} // void TemperedLB::originalTransfer()
```
cz4rs, line 2727: This is the end of `swapClusters`.
```suggestion
} // void TemperedLB::swapClusters()
```

cz4rs (approved):
This looks overall good to me.

Apart from some comments above, `TemperedLB` code feels pretty complex in general. This PR also adds some rather lengthy methods (like `swapClusters` ~120 lines, `considerSwapsAfterLock` almost 200 lines) and there are no unit tests targeting specific parts of the machinery.
If we merge this as is, it'd be good to create a follow-up issue with the goal of refactoring `TemperedLB` internals, making them easier to understand and unit test.

tools/1959-tasks/ccm-lb-delta-1e-11.config
nlslatt: Correct, these file additions were for convenience only and should be reverted before the merge.

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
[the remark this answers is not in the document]
ppebay: I agree this works because we move entire block-clusters at once and therefore these formulae simplify out.

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
ppebay: @cz4rs go ahead please then resolve thanks

src/vt/elm/elm_comm.h
cz4rs: Done. Had to push `SharedIdType` alias to `types_type.h` though :shrug:

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
```diff
@@ -469,9 +559,512 @@ void TemperedLB::runLB(LoadType total_load) {
...
+          ) {
+            max_object_serialized_bytes_outside = std::max(
+              max_object_serialized_bytes_outside, it->second
+            );
+          }
+        }
+      }
+    }
+
+    info.cluster_footprint = cluster_footprint;
```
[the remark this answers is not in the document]
ppebay: sum is only over cluster

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
```diff
@@ -469,9 +559,512 @@ void TemperedLB::runLB(LoadType total_load) {
...
+    max_object_working_bytes_ +
+    object_footprint_bytes +
+    max_object_serialized_bytes_;
+}
+
+std::set<SharedIDType> TemperedLB::getSharedBlocksHere() const {
+  std::set<SharedIDType> blocks_here;
+  for (auto const& [obj, _] : cur_objs_) {
+    if (obj_shared_block_.find(obj) != obj_shared_block_.end()) {
+      blocks_here.insert(obj_shared_block_.find(obj)->second);
```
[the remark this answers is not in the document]
lifflander: `obj_shared_block` gets the shared block for a given obj

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
```diff
@@ -1296,14 +2153,586 @@ void TemperedLB::decide() {
...
+  consider_swaps_counter_--;
+
+  if (pending_actions_.size() > 0) {
+    auto action = pending_actions_.back();
+    pending_actions_.pop_back();
+    action();
+  }
+}
+
+void TemperedLB::giveCluster(
```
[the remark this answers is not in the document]
lifflander: Shared blocks do not communicate their home because it's unnecessary. The shared blocks homed on each rank are known by that rank. If a shared block has another home, it's exact location does not need to be known, just the information that it's not homed at a particular rank is sufficient.

tests/unit/lb/test_temperedlb.cc
```diff
@@ -0,0 +1,87 @@
...
+        }
+        cfg_file_.close();
+    }
+    return config_file;
+}
+
+void runTemperedLBTest(std::string config_file, double expected_imb = 0.0) {
+    // Set configuration
+    theConfig()->vt_lb = true;
+    theConfig()->vt_lb_name = "TemperedLB";
```
nlslatt, line 40: You don't need the above line because your `config_file` will set it on appropriate phases.

tests/unit/lb/test_temperedlb.cc
```diff
@@ -0,0 +1,87 @@
...
+    int phases_to_run = 1;
+    int phase_mod = 0;
+    vt::vrt::collection::balance::replay::replayWorkloads(
+        initial_phase, phases_to_run, phase_mod);
+
+    // Get information for the last phase (this problem only has one)
+    auto phase_info = theLBManager()->getPhaseInfo();
+
+    // Assert that temperedLB found the correct imbalance
+    auto imb = (phase_info->max_load / phase_info->avg_load) - 1;
```
nlslatt, line 57: Instead of computing this yourself, you should be able to use `phase_info->imb_load_post_lb`

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
```diff
@@ -414,6 +496,7 @@ void TemperedLB::inputParams(balance::ConfigEntry* config) {
       knowledge_converter_.getString(knowledge_), f_, k_max_, num_iters_,
       criterion_converter_.getString(criterion_), num_trials_, deterministic_,
       inform_type_converter_.getString(inform_type_),
+      transfer_type_converter_.getString(transfer_type_),
```
nlslatt, line 499: This line was added without adding the necessary `transfer={}, ` to the *beginning* of the `"ordering=` line above

tests/unit/lb/test_temperedlb.cc
```diff
@@ -0,0 +1,87 @@
...
+    int this_rank;
+    MPI_Comm_rank(MPI_COMM_WORLD, &this_rank);
+    auto config_file = getUniqueFilename();
+    if (this_rank == 0) {
+        std::ofstream cfg_file_{config_file.c_str(), std::ofstream::out | std::ofstream::trunc};
+        cfg_file_ << "0 TemperedLB transfer=" << transfer_strategy <<
+                     " alpha=1.0" <<
+                     " beta=" << beta <<
+                     " gamma=" << gamma <<
+                     " delta=" << delta;
```
nlslatt, line 28: You'll need to add `rollback=false` when `transfer=SwapClusters`. Maybe that should be turned off automatically?

src/vt/vrt/collection/balance/temperedlb/temperedlb.h
```diff
@@ -157,30 +385,131 @@ struct TemperedLB : BaseLB {
...
   ObjectOrderEnum obj_ordering_                     = ObjectOrderEnum::FewestMigrations;
   CMFTypeEnum cmf_type_                             = CMFTypeEnum::NormByMax;
   KnowledgeEnum knowledge_                          = KnowledgeEnum::Log;
   bool setup_done_                                  = false;
   bool propagate_next_round_                        = false;
+  double alpha = 1.0;
+  double beta = 0.0;
+  double gamma = 0.0;
+  double delta = 0.0;
+  double epsilon = 0.0;
```
nlslatt, line 420: I don't think this should default to zero. The documentation says it should default to 1.0. @PPP Is 1.0 actually big enough?

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
```diff
@@ -252,6 +270,55 @@ Default: false
...
+Values: <double>
+Defaut: 1.0
+Description: α in the work model (load in work model)
+)"
+    },
+    {
+      "beta",
+      R"(
+Values: <double>
+Defaut: 1.0
```
nlslatt, line 294: We need to make sure this matches the true default

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
```diff
@@ -252,6 +270,55 @@ Default: false
...
+Description: ε in the work model (memory term in work model)
+)"
+    },
+    {
+    "delta",
+      R"(
+Values: <double>
+Defaut: 1.0
```
nlslatt, line 310: We need to make sure this matches the true default

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
```diff
@@ -252,6 +270,55 @@ Default: false
...
+Description: δ in the work model (shared-memory-edges in work model)
+)"
+    },
+    {
+    "gamma",
+      R"(
+Values: <double>
+Defaut: 1.0
```
nlslatt, line 318: We need to make sure this matches the true default

src/vt/vrt/collection/balance/temperedlb/temperedlb.h
nlslatt: In the paper, this is infinity

nlslatt: @ppebay

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
```diff
@@ -469,9 +559,512 @@ void TemperedLB::runLB(LoadType total_load) {
...
+  double load, double inter_comm_bytes, double intra_comm_bytes,
+  double shared_comm_bytes
+) const {
+  // The work model based on input parameters
+  return
+    alpha * load +
+    beta * inter_comm_bytes +
+    gamma * intra_comm_bytes +
+    delta * shared_comm_bytes +
+    epsilon;
```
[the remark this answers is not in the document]
nlslatt: I disagree. Epsilon is added here unconditionally, with its value never changing.

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
```diff
@@ -597,6 +1326,13 @@ void TemperedLB::doLBStages(LoadType start_imb) {
         best_trial, new_imbalance_
       );
     }
+
+    auto remote_block_count = getRemoteBlockCountHere();
+    runInEpochCollective("TemperedLB::doLBStages -> compute unhomed", [=] {
+      proxy_.allreduce<&TemperedLB::remoteBlockCountHandler, collective::PlusOp>(
+        remote_block_count
+      );
+    });
```
nlslatt, line 1335: We should skip this when not using `SwapClusters` unless we change the code so that the number of unhomed blocks is properly tracked when `has_memory_data_=false`

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
```diff
@@ -350,12 +417,19 @@ void TemperedLB::inputParams(balance::ConfigEntry* config) {
...
+  beta = config->getOrDefault<double>("beta", beta);
+  gamma = config->getOrDefault<double>("gamma", gamma);
+  delta = config->getOrDefault<double>("delta", delta);
+  epsilon = config->getOrDefault<double>("epsilon", epsilon);
+
   num_iters_     = config->getOrDefault<int32_t>("iters", num_iters_);
   num_trials_    = config->getOrDefault<int32_t>("trials", num_trials_);
 
   deterministic_ = config->getOrDefault<bool>("deterministic", deterministic_);
   rollback_      = config->getOrDefault<bool>("rollback", rollback_);
```
nlslatt, line 430: I think we should force `rollback_ = false` when using `SwapClusters` because rollback could make locality much worse in favor of a tiny improvement in load balance

nlslatt: Rollback was designed for when work was transferred completely asynchronously and so load balance could degrade if one rank was chosen as the destination for transfers from more than one source rank on the same iteration

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
```diff
@@ -1296,9 +2117,582 @@ void TemperedLB::decide() {
...
+  );
+}
+
+bool TemperedLB::memoryTransferCriterion(double try_total_bytes, double src_bytes) {
+  // FIXME: incomplete implementation that ignores memory regrouping
+  auto const src_after_mem = this->current_memory_usage_;
+  auto const try_after_mem = try_total_bytes + src_bytes;
+
+  return not (src_after_mem > this->mem_thresh_ or try_after_mem > this->mem_thresh_);
+}
```
nlslatt, line 2214: Apparently unused, partially-implemented function

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
```diff
@@ -1296,9 +2117,582 @@ void TemperedLB::decide() {
...
+    try_new_mem += std::max(
+      try_cluster.max_object_serialized_bytes_outside,
+      src_cluster.max_object_serialized_bytes
+    );
+    try_new_mem -= try_cluster.cluster_footprint;
+    try_new_mem += src_cluster.cluster_footprint;
+
+    if (try_new_mem > mem_thresh_) {
+      return - std::numeric_limits<double>::infinity();
+    }
```
nlslatt, line 2274: This is used instead of a variable `epsilon`

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
```diff
@@ -1296,9 +2117,582 @@ void TemperedLB::decide() {
...
+    src_new_mem += std::max(
+      src_cluster.max_object_serialized_bytes_outside,
+      try_cluster.max_object_serialized_bytes
+    );
+    src_new_mem += try_cluster.cluster_footprint;
+    src_new_mem -= src_cluster.cluster_footprint;
+
+    if (src_new_mem > mem_thresh_) {
+      return - std::numeric_limits<double>::infinity();
+    }
```
nlslatt, line 2294: And this instead of a variable `epsilon`

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
```diff
@@ -1296,9 +2117,582 @@ void TemperedLB::decide() {
...
+
+    // Necessary but not sufficient check regarding memory bounds
+    if (try_mem - try_cluster.bytes + src_cluster.bytes > mem_thresh_) {
+      return - std::numeric_limits<double>::infinity();
+    }
+
+    auto const src_mem = current_memory_usage_;
+    if (src_mem + try_cluster.bytes - src_cluster.bytes > mem_thresh_) {
+      return - std::numeric_limits<double>::infinity();
+    }
```
nlslatt, line 2606: And these

tests/unit/lb/test_temperedlb.cc
cwschilly: I think this was actually the source of the problem--tests are passing now (all cases find 0.0 imbalance)

tests/unit/lb/test_temperedlb.cc
cwschilly: Fixed

tests/unit/lb/test_temperedlb.cc
cwschilly: I've added `rollback=false` when `transfer=SwapClusters`, but I did not alter the default settings

tests/unit/lb/test_temperedlb.cc
nlslatt: Oh, right. You were computing the pre-LB imbalance, not the post-LB

nlslatt: The post-LB quantities explicitly say `post_lb` in them

cwschilly:
@nlslatt @lifflander @ppebay The tests fail when run on 2 ranks, which makes sense because the problem is specific to 4 ranks. Should I add the `SET_NUM_NODES_CONSTRAINT(4)` guard?

nlslatt:
> SET_NUM_NODES_CONSTRAINT

Yes, that sounds right to me

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
cwschilly: Fixed

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
```diff
@@ -498,17 +1068,104 @@ void TemperedLB::doLBStages(LoadType start_imb) {
         // Copy this node's object assignments to a local, mutable copy
         cur_objs_.clear();
         for (auto obj : *load_model_) {
-          if (obj.isMigratable()) {
-            cur_objs_[obj] = getModeledValue(obj);
+          cur_objs_[obj] = getModeledValue(obj);
```
nlslatt, line 1071: @cwschilly Please revert this change

cwschilly: Fixed

tests/unit/lb/test_temperedlb.cc
cwschilly: I've added a check in `temperedlb.cc` that sets `rollback=false` when the transfer strategy is `SwapClusters`

src/vt/vrt/collection/balance/temperedlb/temperedlb.h
cwschilly: I've changed this the default value of epsilon to `std::numeric_limits<double>::infinity()`

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
```diff
@@ -252,6 +270,55 @@ Default: false
...
+Values: <double>
+Defaut: 1.0
+Description: β in the work model (inter-node communication in work model)
+)"
+    },
+    {
+    "epsilon",
+      R"(
+Values: <double>
+Defaut: 1.0
```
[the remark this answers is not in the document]
cwschilly: Fixed

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
cwschilly: Fixed

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
cwschilly: Fixed

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
cwschilly: I encased this in an `if` block so it only runs if `SwapClusters` is selected as the transfer strategy

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
cwschilly: Fixed

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
cwschilly: I've replaced this with `epsilon`

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
cwschilly: Replaced with `epsilon`

src/vt/vrt/collection/balance/temperedlb/criterion.h
```diff
@@ -63,7 +63,7 @@ struct GrapevineCriterion {
 
 struct ModifiedGrapevineCriterion  {
   bool operator()(LoadType over, LoadType under, LoadType obj, LoadType) const {
-    return obj < over - under;
+    return obj <= over - under;
```
ppebay, line 66: It seems that the change was needed, at least with small test cases meant to exercise such corner cases (where the optimization must go through zero-sum transfers); it will be interesting to see what that does in terms of efficiency with large, non-academic cases.

ppebay (approved):

tests/unit/lb/test_temperedlb.cc
```diff
@@ -0,0 +1,119 @@
...
+            }
+        }
+        cfg_file_.close();
+    }
+    return config_file;
+}
+
+void runTemperedLBTest(std::string config_file, double expected_imb = 0.0) {
+    // Clear the LB config
+    vrt::collection::balance::ReadLBConfig::clear();
```
lifflander, line 44: Longer term, I'd like to re-engineer this so it doens't have to be written out of file each time and read in. That can cause issues with testing and slow down the tests when dealing with slow file systems. But, I don't think that's necessary now.

lifflander:
I approve of this being merged once testing passes. I have some suggestions, mostly minor to improve it. I can't officially approve it because I'm the author of it. I'm going to force push and fix the code commit signing on this.

lifflander:
I guess I can't fix the signatures by force rebasing. @cz4rs do you know what is wrong with your signature?

tests/unit/lb/test_temperedlb.cc
```diff
@@ -0,0 +1,119 @@
...
+    theConfig()->vt_lb_file_name = config_file;
+    theConfig()->vt_lb_data_file_in="synthetic-dataset-blocks.%p.json";
+    theConfig()->vt_lb_data_dir_in="synthetic-blocks-data";
+
+    // Replay load balancing
+    int initial_phase = 0;
+    int phases_to_run = 1;
+    int phase_mod = 0;
+    vt::vrt::collection::balance::replay::replayWorkloads(
+        initial_phase, phases_to_run, phase_mod);
```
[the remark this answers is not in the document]
lifflander: I went ahead and fixed all the formatting for the whole file.

tests/unit/lb/test_temperedlb.cc
```diff
@@ -0,0 +1,119 @@
...
+
+using TestTemperedLB = TestParallelHarness;
+
+std::string writeTemperedLBConfig(std::string transfer_strategy,
+                                  bool mem_constraints,
+                                  double alpha = 1.0,
+                                  double beta = 0.0,
+                                  double gamma = 0.0,
+                                  double delta = 0.0) {
+    int this_rank;
```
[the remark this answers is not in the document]
lifflander: I fixed this.

cz4rs:
> I guess I can't fix the signatures by force rebasing. @cz4rs do you know what is wrong with your signature?

@lifflander Not sure why they are broken, I suspect that something went wrong during rebasing.
I have fixed the signatures and pushed to a new branch: https://github.com/DARMA-tasking/vt/tree/2201-signatures-fixed. If no one has any work in progress, I can force push the fixed version here too.

---
For the record, Git command that worked well for this:
`git rebase --exec 'git commit --amend --no-edit -n -S' -i a26af8aa5c37a68c73960c6327c55eab746e1559`

cwschilly:
@cz4rs It's ok for me if you force push the fixed commits here

cz4rs:
Signatures are fixed.

`gcc-9` build is failing with an error during JSON validation:
```
INFO - JSON_data_files_validator.py:390 - Validating file: /build/vt/examples/collection/jacobi1d_vt_2_LBDatafile.0.json
ERROR - JSON_data_files_validator.py:406 - Invalid JSON schema in /build/vt/examples/collection/jacobi1d_vt_2_LBDatafile.0.json
[JSON_data_files_validator] SchemaError Key 'phases' error:
Or({'id': <class 'int'>, 'tasks': [{'entity': And({Optional('collection_id'): <class 'int'>, 'home': <class 'int'>, Optional('id'): <class 'int'>, Optional('seq_id'): <class 'int'>, Optional('index'): [<class 'int'>], 'type': <class 'str'>, 'migratable': <class 'bool'>, Optional('objgroup_id'): <class 'int'>}, <function validate_ids at 0x7f0033cd7310>), 'node': <class 'int'>, 'resource': <class 'str'>, Optional('subphases'): [{'id': <class 'int'>, 'time': <class 'float'>}], 'time': <class 'float'>, Optional('user_defined'): <class 'dict'>, Optional('attributes'): <class 'dict'>}], Optional('communications'): [{'type': <class 'str'>, 'to': And({'type': <class 'str'>, Optional('id'): <class 'int'>, Optional('seq_id'): <class 'int'>, Optional('home'): <class 'int'>, Optional('collection_id'): <class 'int'>, Optional('migratable'): <class 'bool'>, Optional('index'): [<class 'int'>], Optional('objgroup_id'): <class 'int'>}, <function validate_ids at 0x7f0033cd7310>), 'messages': <class 'int'>, 'from': And({'type': <class 'str'>, Optional('id'): <class 'int'>, Optional('seq_id'): <class 'int'>, Optional('home'): <class 'int'>, Optional('collection_id'): <class 'int'>, Optional('migratable'): <class 'bool'>, Optional('index'): [<class 'int'>], Optional('objgroup_id'): <class 'int'>}, <function validate_ids at 0x7f0033cd7310>), 'bytes': <class 'float'>}], Optional('user_defined'): <class 'dict'>}) did not validate {'communications': [{'bytes': 160.0, 'from': {'collection_id': 7, 'home': 0, 'id': 524291, 'index': [1], 'migratable': True, 'type': 'object'}, 'messages': 1, 'to': {'collection_id': 7, 'home': 0, 'id': 262147, 'index': [0], 'migratable': True, 'type': 'object'}, 'type': 'SendRecv'}, {'bytes': 160.0, 'from': {'collection_id': 7, 'home': 0, 'id': 262147, 'index': [0], 'migratable': True, 'type': 'object'}, 'messages': 1, 'to': {'collection_id': 7, 'home': 0, 'id': 524291, 'index': [1], 'migratable': True, 'type': 'object'}, 'type': 'SendRecv'}, {'bytes': 160.0, 'from': {'collection_id': 7, 'home': 0, 'id': 1048579, 'index': [3], 'migratable': True, 'type': 'object'}, 'messages': 1, 'to': {'collection_id': 7, 'home': 0, 'id': 786435, 'index': [2], 'migratable': True, 'type': 'object'}, 'type': 'SendRecv'}, {'bytes': 160.0, 'from': {'collection_id': 7, 'home': 0, 'id': 786435, 'index': [2], 'migratable': True, 'type': 'object'}, 'messages': 1, 'to': {'collection_id': 7, 'home': 0, 'id': 1048579, 'index': [3], 'migratable': True, 'type': 'object'}, 'type': 'SendRecv'}, {'bytes': 160.0, 'from': {'collection_id': 7, 'home': 0, 'id': 786435, 'index': [2], 'migratable': True, 'type': 'object'}, 'messages': 1, 'to': {'collection_id': 7, 'home': 0, 'id': 524291, 'index': [1], 'migratable': True, 'type': 'object'}, 'type': 'SendRecv'}, {'bytes': 160.0, 'from': {'collection_id': 7, 'home': 0, 'id': 524291, 'index': [1], 'migratable': True, 'type': 'object'}, 'messages': 1, 'to': {'collection_id': 7, 'home': 0, 'id': 786435, 'index': [2], 'migratable': True, 'type': 'object'}, 'type': 'SendRecv'}, {'bytes': 160.0, 'from': {'home': 1, 'id': 262151, 'migratable': True, 'type': 'object'}, 'messages': 1, 'to': {'collection_id': 7, 'home': 0, 'id': 1048579, 'index': [3], 'migratable': True, 'type': 'object'}, 'type': 'SendRecv'}, {'bytes': 112.0, 'from': {'home': 0, 'id': 0, 'migratable': False, 'type': 'object'}, 'messages': 2, 'to': {'home': 1, 'id': 5, 'migratable': False, 'type': 'object'}, 'type': 'SendRecv'}, {'bytes': 960.0, 'from': {'home': 0, 'id': 1, 'migratable': False, 'type': 'object'}, 'messages': 16, 'to': {'home': 1, 'id': 5, 'migratable': False, 'type': 'object'}, 'type': 'SendRecv'}, {'bytes': 64.0, 'from': {'home': 0, 'id': 1, 'migratable': False, 'type': 'object'}, 'messages': 1, 'to': {'home': 0, 'id': 1, 'migratable': False, 'type': 'object'}, 'type': 'SendRecv'}], 'id': 0, 'tasks': [{'entity': {'collection_id': 7, 'home': 0, 'id': 262147, 'index': [0], 'migratable': True, 'type': 'object'}, 'node': 0, 'resource': 'cpu', 'subphases': [{'id': 0, 'time': 1.4700000065204222e-05}], 'time': 1.4700000065204222e-05}, {'entity': {'collection_id': 7, 'home': 0, 'id': 1048579, 'index': [3], 'migratable': True, 'type': 'object'}, 'node': 0, 'resource': 'cpu', 'subphases': [{'id': 0, 'time': 4.0299999909620965e-05}], 'time': 4.0299999909620965e-05}, {'entity': {'collection_id': 7, 'home': 0, 'id': 786435, 'index': [2], 'migratable': True, 'type': 'object'}, 'node': 0, 'resource': 'cpu', 'subphases': [{'id': 0, 'time': 1.2300999969738768e-05}], 'time': 1.2300999969738768e-05}, {'entity': {'home': 0, 'id': 3145740, 'migratable': False, 'objgroup_id': 786435, 'type': 'object'}, 'node': 0, 'resource': 'cpu', 'time': 0.0}, {'entity': {'home': 0, 'id': 4194316, 'migratable': False, 'objgroup_id': 1048579, 'type': 'object'}, 'node': 0, 'resource': 'cpu', 'time': 0.0}, {'entity': {'home': 0, 'id': 5242892, 'migratable': False, 'objgroup_id': 1310723, 'type': 'object'}, 'node': 0, 'resource': 'cpu', 'time': 0.0}, {'entity': {'collection_id': 7, 'home': 0, 'id': 524291, 'index': [1], 'migratable': True, 'type': 'object'}, 'node': 0, 'resource': 'cpu', 'subphases': [{'id': 0, 'time': 6.9000002440589014e-06}], 'time': 6.9000002440589014e-06}, {'entity': {'home': 0, 'id': 1, 'migratable': False, 'type': 'object'}, 'node': 0, 'resource': 'cpu', 'subphases': [{'id': 0, 'time': 0.00012950200061823125}], 'time': 0.00012950200061823125}, {'entity': {'home': 0, 'id': 0, 'migratable': False, 'type': 'object'}, 'node': 0, 'resource': 'cpu', 'time': 0.0}]}
Key 'communications' error:
Or({'type': <class 'str'>, 'to': And({'type': <class 'str'>, Optional('id'): <class 'int'>, Optional('seq_id'): <class 'int'>, Optional('home'): <class 'int'>, Optional('collection_id'): <class 'int'>, Optional('migratable'): <class 'bool'>, Optional('index'): [<class 'int'>], Optional('objgroup_id'): <class 'int'>}, <function validate_ids at 0x7f0033cd7310>), 'messages': <class 'int'>, 'from': And({'type': <class 'str'>, Optional('id'): <class 'int'>, Optional('seq_id'): <class 'int'>, Optional('home'): <class 'int'>, Optional('collection_id'): <class 'int'>, Optional('migratable'): <class 'bool'>, Optional('index'): [<class 'int'>], Optional('objgroup_id'): <class 'int'>}, <function validate_ids at 0x7f0033cd7310>), 'bytes': <class 'float'>}) did not validate {'bytes': 160.0, 'from': {'home': 1, 'id': 262151, 'migratable': True, 'type': 'object'}, 'messages': 1, 'to': {'collection_id': 7, 'home': 0, 'id': 1048579, 'index': [3], 'migratable': True, 'type': 'object'}, 'type': 'SendRecv'}
Key 'from' error:
validate_ids({'home': 1, 'id': 262151, 'migratable': True, 'type': 'object'}) raised ValueError('If an entity is migratable, it must have a collection_id')
```

ppebay:
Following resolution of [ccm-milp #13](https://github.com/DARMA-tasking/ccm-milp/issues/13), I suggest we add this test case to the CI suite, @cwschilly:
`alpha=1.0`, `beta=gamma=delta=0.0`, no memory constraint.

With the `original` strategy (and thus **no** preservation of clusters), we should find **`W_max=4.0`**

In contrast, using the cluster-swapping strategy (and therefore without subclustering), ), we should find **`W_max=6.0`**

nlslatt:
> Following resolution of [ccm-milp #13](https://github.com/DARMA-tasking/ccm-milp/issues/13), I suggest we add this test case to the CI suite, @cwschilly: `alpha=1.0`, `beta=gamma=delta=0.0`, no memory constraint.
> 
> With the `original` strategy (and thus **no** preservation of clusters), we should find **`W_max=4.0`**
> 
> In contrast, using the cluster-swapping strategy (and therefore without subclustering), ), we should find **`W_max=6.0`**

@ppebay Did you mean beta non-zero?

tests/unit/collection/test_lb.extended.cc
```diff
@@ -150,6 +150,11 @@ TEST_P(TestLoadBalancerOther, test_load_balancer_other_keep_last_elm) {
   runTest(GetParam(), "test_load_balancer_other_keep_last_elm");
 }
 
+TEST_P(TestLoadBalancerOther, test_load_balancer_other_run_lb_first_phase) {
+  vt::theConfig()->vt_lb_run_lb_first_phase = true;
+  runTest(GetParam(), "test_load_balancer_other_keep_last_elm");
```
nlslatt, line 155: The string should match the test name.

tests/unit/collection/test_lb.extended.cc
```diff
@@ -159,6 +164,11 @@ TEST_P(TestLoadBalancerGreedy, test_load_balancer_greedy_keep_last_elm) {
   runTest(GetParam(), "test_load_balancer_greedy_keep_last_elm");
 }
 
+TEST_P(TestLoadBalancerGreedy, test_load_balancer_greedy_run_lb_first_phase) {
+  vt::theConfig()->vt_lb_run_lb_first_phase = true;
+  runTest(GetParam(), "test_load_balancer_greedy_keep_last_elm");
```
nlslatt, line 169: The string should match the test name.

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
```diff
@@ -469,11 +558,485 @@ void TemperedLB::runLB(LoadType total_load) {
     }
   }
 
+  // Perform load rebalancing when deemed necessary
   if (should_lb) {
-    doLBStages(imb);
+#if vt_check_enabled(trace_enabled)
+    theTrace()->disableTracing();
+#endif
```
nlslatt, line 565: We might not need to do this anymore now that #2188 has been merged. In the interests of expediency, let's explore that in a follow-on issue.

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
nlslatt: This still needs to be deleted

nlslatt:
> Following resolution of [ccm-milp #13](https://github.com/DARMA-tasking/ccm-milp/issues/13), I suggest we add this test case to the CI suite, @cwschilly: `alpha=beta=1.0`, `gamma=delta=0.0`, no memory constraint.
> 
> With the `original` strategy (and thus **no** preservation of clusters), we should find **`W_max=4.0`**
> 
> In contrast, using the cluster-swapping strategy (and therefore without subclustering), ), we should find **`W_max=6.0`**

@ppebay Can you express the solution in terms of maximum load or load imbalance? Caleb does not have access to the calculation of W_max in CI.

ppebay:
@nlslatt @cwschilly 

* for the `Original` strategy (i.e. with de facto sub-clustering):
```
# Solution summary:
Rank 0: L = 2.5, W = 4.0, unhomed: 1
Rank 1: L = 1.5, W = 4.0, unhomed: 1
Rank 2: L = 2.0, W = 4.0, unhomed: 1
Rank 3: L = 2.0, W = 4.0, unhomed: 2
W_max = 4.0
```
thus `I_L = L_max / L_ave  - 1.0 = 2.5 / 2.0 - 1.0 =`**`0.25`**

* for the `SwapClusters` strategy (i.e. with which sub-clustering is not possible):
# Solution summary:
```
Rank 0: L = 4.0, W = 6.0, unhomed: 1
Rank 1: L = 1.0, W = 1.5, unhomed: 1
Rank 2: L = 1.5, W = 4.0, unhomed: 0
Rank 3: L = 1.5, W = 3.0, unhomed: 1
W_max = 6.0
```
thus `I_L = L_max / L_ave  - 1.0 = 4.0 / 2.0 - 1.0 =`**`1.0`**

tests/unit/collection/test_lb.extended.cc
cwschilly: Fixed

tests/unit/collection/test_lb.extended.cc
cwschilly: Fixed

src/vt/vrt/collection/balance/temperedlb/temperedlb.cc
cwschilly: Fixed

tests/unit/lb/test_temperedlb.cc
```diff
@@ -0,0 +1,119 @@
...
+#include "test_helpers.h"
+#include "test_parallel_harness.h"
+
+namespace vt { namespace tests { namespace unit { namespace lb {
+
+#if vt_check_enabled(lblite)
+
+using TestTemperedLB = TestParallelHarness;
+
+std::string writeTemperedLBConfig(std::string transfer_strategy,
```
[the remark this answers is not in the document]
cwschilly: Fixed

nlslatt (approved):
If the tests pass this time, I think it's ready

nlslatt:
> @ppebay @cwschilly For the case that is still failing, we should check if there is a second solution with the same Wmax but that has a worse imbalance. Perhaps we should try adding `trials=3` (make 3 attempts and keep the one that yields the best _load imbalance_) to see if it gets the right imbalance with a different random seed. Someday we might want to extend the trials feature to target the best Wmax instead.

@ppebay Not sure if you were notified of this because I mentioned you incorrectly the first time.

ppebay:
@nlslatt based on latest discussions, we are going to drop this test (with non-zero `beta`) as a result of the replay not being able to move communications; is this correct?

PR #2282, opened 2024-05-09
#2264: Add stricter warnings for apps that use them

### LINKED ISSUE #2264: Add stricter warnings for apps that use them
**What Needs to be Done?**

Add stricter warnings to cmake build

cz4rs:
Enable more warnings when building the project.

The most common warning here is unused variable / parameter. Silencing it with `[[maybe_unused]]` attribute is explicit and reasonably concise, so I'm going with that in most cases (b832ffc5532d9d43da9dbd4e9913fa5a8d12481a handles cases unsupported in gcc-8).

fixes #2264


src/vt/rdma/channel/rdma_channel.cc
```diff
@@ -188,6 +188,7 @@ Channel::lockChannelForOp() {
       (not is_target_) ?
       (op_type_ == RDMA_TypeType::Put ? MPI_LOCK_EXCLUSIVE : MPI_LOCK_SHARED) :
       (op_type_ == RDMA_TypeType::Put ? MPI_LOCK_SHARED : MPI_LOCK_SHARED);
+      // FIXME: same expression in both branches of ternary operator
```
cz4rs, line 191: FIXME

cz4rs:
Marked as `Ready for review` to trigger all pipelines, not ready for merging yet.

src/vt/collective/barrier/barrier.cc
```diff
@@ -91,7 +91,9 @@ Barrier::BarrierStateType& Barrier::insertFindBarrier(
 }
 
 void Barrier::removeBarrier(
-  bool const& is_named, bool const& is_wait, BarrierType const& barrier
+  bool const& is_named,
+  [[maybe_unused]] bool const& is_wait,
```
cz4rs, line 95: Maybe this parameter can be simply removed instead (this method is marked as `\internal` in the header).

JacobDomagala (approved):
:rocket:

src/vt/rdma/channel/rdma_channel.cc
thearusable: In the second part we have `MPI_LOCK_SHARED` twice so that could be simplified.

thearusable (approved):

src/vt/rdma/channel/rdma_channel.cc
cz4rs: Yeah, I'm actually wondering if that's a bug :thinking:

lifflander (approved):
Looks great!

lifflander:
> Looks great!

Ready to merge?

PhilMiller:
Looking ahead, you may want to explicitly add warnings about performance-impacting things `-Wredundant-move`, `-Wpessimizing-move`, and `-Wrange-loop-construct`, separate from their current default inclusion in GCC's `-Wall`

PR #2293, opened 2024-05-27
2279: Bug in communication JSON output

### LINKED ISSUE #2279: Bug in communication JSON output
**Describe the bug**
As @pierrepebay and @cwschilly were working on displaying the communication graph in vt-tv, we noticed that there were communication edges in the JSON LB data file output which linked to object IDs that did not exist as tasks. Either the communication link shouldn't exist or more likely a task should be output with zero time if that is what is happening. This is probably due to some bare handler link that is being output.

cz4rs:
fixes #2279


cz4rs:
Fixed version:
[data.0.json](https://github.com/user-attachments/files/15792157/data.0.json)
[data.1.json](https://github.com/user-attachments/files/15792155/data.1.json)

lifflander (approved):

PR #2294, opened 2024-05-28
#2174: Integrate vt-tv api into lb data export

### LINKED ISSUE #2174: Integrate vt-tv API into LB data export
**What Needs to be Done?**

- [x] Move LB data from tvt into the vt-tv interface

pierrepebay:
Fixes #2174


cwschilly:
If `VT_TV_ENABLED` is ON, all tests will be run with `--vt_tv` and the default config file, `test_vttv.yaml`. This will not generate any .pngs or meshes--it only tests that the data is passed from `vt` to `vt-tv` without any errors.

cwschilly:
The `vt-tv` pipeline shows this error when trying to access the cached docker image:

```
#6 [ubuntu-cpp-vtk] importing cache manifest from ***/vt:amd64-ubuntu-22.04-gcc-12-gcc-12-vtk-cpp
#6 ERROR: failed to configure registry cache importer: docker.io/***/vt:amd64-ubuntu-22.04-gcc-12-gcc-12-vtk-cpp: not found
```

I think this will be resolved once we push to develop and the `pushdockerimage.yml` workflow runs. Is that correct?

src/vt/vrt/collection/balance/lb_data_holder.cc
```diff
@@ -275,6 +279,103 @@ std::unique_ptr<nlohmann::json> LBDataHolder::toJson(PhaseType phase) const {
...
+        break;
+      }
+      default:
+        // skip all other communications for now
+        break;
+      }
+    }
+  }
+
+  return std::make_unique<PhaseWork>(phase, objects);;
```
JacobDomagala, line 342: double semicolon at the end : )

src/vt/vrt/collection/balance/lb_data_holder.cc
```diff
@@ -275,6 +279,103 @@ std::unique_ptr<nlohmann::json> LBDataHolder::toJson(PhaseType phase) const {
...
+      ElementIDStruct id = elm.first;
+
+      bool is_collection = false;
+      bool is_objgroup = false;
+
+      std::vector<uint64_t> idx;
+      if (node_idx_.find(id) != node_idx_.end()) {
+        is_collection = true;
+        idx = std::get<1>(node_idx_.find(id)->second);
+      }
```
JacobDomagala, line 360: You should be able to use 'if statements with initializer' like so:
```c++
      if (auto it = node_idx_.find(id); it != node_idx_.end()) {
        is_collection = true;
        idx = std::get<1>(it->second);
      }
```

src/vt/vrt/collection/balance/lb_data_holder.h
```diff
@@ -97,6 +101,26 @@ struct LBDataHolder {
    */
   std::unique_ptr<nlohmann::json> toJson(PhaseType phase) const;
 
+#if vt_check_enabled(tv)
+          /**
```
JacobDomagala, line 105: Extra spaces

src/vt/vrt/collection/balance/lb_invoke/lb_manager.cc
```diff
@@ -474,6 +478,42 @@ void LBManager::fatalError() {
...
+  std::unordered_map<NodeType, tv::Rank> rank_map;
+  std::unordered_map<ElementIDType, tv::ObjectInfo> info;
+  PhaseType phase = 0;
+
+  template <typename SerializerT>
+  void serialize(SerializerT& s) {
+    s | rank_map | info | phase;
+  }
+
+  friend GatherTVInfo operator+(GatherTVInfo a1, GatherTVInfo const& a2) {
```
JacobDomagala, line 494: Shouldn't 'a1' be reference here?

src/vt/vrt/collection/balance/lb_data_holder.cc
cwschilly: Fixed

src/vt/vrt/collection/balance/lb_data_holder.cc
cwschilly: Fixed

src/vt/vrt/collection/balance/lb_data_holder.h
cwschilly: Fixed

src/vt/vrt/collection/balance/lb_invoke/lb_manager.cc
cwschilly: Fixed

src/vt/vrt/collection/balance/lb_invoke/lb_manager.cc
```diff
@@ -491,7 +491,7 @@ struct GatherTVInfo {
     s | rank_map | info | phase;
   }
 
-  friend GatherTVInfo operator+(GatherTVInfo a1, GatherTVInfo const& a2) {
+  friend GatherTVInfo operator+(GatherTVInfo& a1, GatherTVInfo const& a2) {
```
JacobDomagala, line 494: This should probably take both `a1` and `a2` as const& and create new `GatherTVInfo` inside the function:
```c++
friend GatherTVInfo operator+(GatherTVInfo const& a1, GatherTVInfo const& a2) {
    GatherTVInfo result;
    /*
      Do stuff ...
    */
    return result;
    }
```

Also, the current version of this operator looks more like copy than sum, is this intended?

cwschilly: I began implementing this but realized that it was essentially the same as the original function (in which  `a1` was copied instead of passed by reference). I ran into some new bugs as well, so I reverted back to that original signature.

pierrepebay:
@cwschilly we should add `lib/vt-tv` in the `.gitignore` in this PR

src/vt/vrt/collection/balance/lb_invoke/lb_manager.cc
JacobDomagala: I'm fine with current version.

CMakeLists.txt
```diff
@@ -59,16 +59,17 @@ set(PROJECT_TOOLS_DIR   ${CMAKE_CURRENT_SOURCE_DIR}/tools)
...
 
-include(cmake/check_compiler.cmake)
 include(cmake/nvcc_no_deprecated_gpu_targets.cmake)
 include(cmake/load_bundled_libraries.cmake)
 
+# Load packages that are required for core VT build
+# We need to do this after bundled libraries so we avoid library duplication
+include(cmake/load_packages.cmake)
+include(cmake/check_compiler.cmake)
+
```
JacobDomagala, line 72: We have to be careful with that. 
`configure_options.cmake` conditionally enables `vt_mpi_guards` if Perl is found (we call find_package(Perl) inside `load_packages.cmake`)

For CI it doesn't matter that much since we use `vt_trace_only` and that enables`vt_mpi_guard`

cwschilly: Good catch; we've added other guards against library duplication since this change so I went ahead and reverted `CMakeLists.txt` back to its original order.

cwschilly: @JacobDomagala @pierrepebay I've had to update all TPLs to be guarded by `if(NOT TARGET <lib>)`. For example:

```cmake
# EngFormat-Cpp always included in the build
set(ENG_FORMAT_LIBRARY EngFormat-Cpp)
if(NOT TARGET ${ENG_FORMAT_LIBRARY})
  add_subdirectory(${PROJECT_LIB_DIR}/EngFormat-Cpp)
  set_darma_compiler_flags(${ENG_FORMAT_LIBRARY})
endif()

# json library always included in the build
set(JSON_BuildTests OFF)
set(JSON_MultipleHeaders ON)
set(JSON_LIBRARY nlohmann_json)
if(NOT TARGET ${JSON_LIBRARY})
  add_subdirectory(${PROJECT_LIB_DIR}/json)
endif()
```

This seems to work, but I'm not sure how to handle `fmt`, since we also support external fmt libs.

If I do this:

```cmake
# use included fmt or external one
if(${vt_external_fmt})
  # user should provide 'fmt_DIR' or 'fmt_ROOT' to CMake (unless fmt is installed in system libs)
  if(fmt_ROOT)
    message(STATUS "vt_external_fmt = ON. Using fmt located at ${fmt_ROOT}")
  elseif(fmt_DIR)
    message(STATUS "vt_external_fmt = ON. Using fmt located at ${fmt_DIR}")
  else()
    message(STATUS "vt_external_fmt = ON but neither fmt_DIR nor fmt_ROOT is provided!")
  endif()
  find_package(fmt 7.1.0 REQUIRED)

else()
  set(FMT_LIBRARY fmt)
  if(NOT TARGET ${FMT_LIBRARY})
    add_subdirectory(${PROJECT_LIB_DIR}/fmt)
    set_darma_compiler_flags(${FMT_LIBRARY})
  endif()
endif()
```

Then we get the following error if I try to build with an external fmt:

```
CMake Error at /home/calebschilly/Develop/DARMA/fmt_test/install/lib/cmake/fmt/fmt-targets.cmake:42 (message):
  Some (but not all) targets in this export set were already defined.

  Targets Defined: fmt::fmt

  Targets not yet defined: fmt::fmt-header-only

  /home/calebschilly/Develop/DARMA/fmt_test/install/lib/cmake/fmt/fmt-config.cmake:27 (include)
  cmake/load_bundled_libraries.cmake:33 (find_package)
  CMakeLists.txt:70 (include)
```

Alternatively, if I enclose the whole block in the guard:

```cmake
# use included fmt or external one
if(NOT TARGET fmt)
  if(${vt_external_fmt})
    # user should provide 'fmt_DIR' or 'fmt_ROOT' to CMake (unless fmt is installed in system libs)
    if(fmt_ROOT)
      message(STATUS "vt_external_fmt = ON. Using fmt located at ${fmt_ROOT}")
    elseif(fmt_DIR)
      message(STATUS "vt_external_fmt = ON. Using fmt located at ${fmt_DIR}")
    else()
      message(STATUS "vt_external_fmt = ON but neither fmt_DIR nor fmt_ROOT is provided!")
    endif()
    find_package(fmt 7.1.0 REQUIRED)

  else()
    set(FMT_LIBRARY fmt)
    add_subdirectory(${PROJECT_LIB_DIR}/fmt)
    set_darma_compiler_flags(${FMT_LIBRARY})
  endif()
endif()
```

Then we would end up ignoring the user-specified fmt and use the one installed with vt-tv instead.

Is there a better way to fix this problem?

src/vt/vrt/collection/balance/lb_invoke/lb_manager.cc
```diff
@@ -484,6 +524,23 @@ void LBManager::finishedLB(PhaseType phase) {
...
+      map[phase] = std::move(*phase_work);
+      auto this_node = theContext()->getNode();
+      vt::tv::Rank r{this_node, std::move(map)};
+      std::unordered_map<NodeType, tv::Rank> rank_map;
+      rank_map[this_node] = std::move(r);
+      GatherTVInfo gather{rank_map, object_info, phase};
+      proxy_.reduce<collectTVData, vt::collective::PlusOp>(0, std::move(gather));
+    });
+  }
+#endif
```
nlslatt, line 543: Can you make this into a separate function called from here? That might allow us to incorporate vt-tv into the task replay feature someday.

tests/config/test_vttv.yaml
```diff
@@ -0,0 +1,15 @@
...
+  z_ranks: 1
+  object_jitter: 0.5
+  rank_qoi: load
+  object_qoi: load
+  save_meshes: false
+  save_pngs: false
+  force_continuous_object_qoi: true
+
+output:
+  directory: ../../output
```
nlslatt, line 13: We need to be really careful writing to the disk from vt CI. Some tests run with multiple rank counts or are otherwise parameterized, so we need to construct filenames/paths that are guaranteed not to collide. `getUniqueFilenameWithRanks()` and `getUniqueFilename()` exist for this purpose but I'm not sure the best way to make use of those for viz testing.

lib/fmt/CMakeLists.txt
```diff
@@ -12,6 +12,8 @@ endfunction()
 add_headers(FMT_HEADERS core.h format.h format-inl.h ostream.h)
 set(FMT_SOURCES src/format.cc)
 
+set(CMAKE_POSITION_INDEPENDENT_CODE ON)
```
nlslatt, line 15: Why is this now necessary?

tests/config/test_vttv.yaml
cwschilly: In this case, we don't actually write anything out (`save_pngs` and `save_meshes` are False)--the output field is only provided to avoid an error from vt-tv.

@pierrepebay Should we add a something in vt-tv to allow omitting the `output` field as long as nothing is being saved?

nlslatt: Is the test primarily of the interface with vt-tv then, without verifying that any resulting viz is reasonable?

pierrepebay: > @pierrepebay Should we add a something in vt-tv to allow omitting the `output` field as long as nothing is being saved?
@cwschilly yes ideally in vt-tv we should try to access these sorts of fields only if the "directing" one is on. Also, for non critical fields, if not in the yaml we should just use a default value. We should create an issue in vt-tv for this.

pierrepebay: > Is the test primarily of the interface with vt-tv then, without verifying that any resulting viz is reasonable?

Yes that's the idea. Ideally we would test generated images as well, but I don't think this will be part of this PR.

cwschilly: @nlslatt Yes, exactly, for now we're only checking that the data is passed from vt to vt-tv without error.

CMakeLists.txt
nlslatt: Can you make the conditional depend on whether vt_tv is turned on? If so, let vt-tv handle it either internally or externally, otherwise let vt itself handle it either internally or externally?

JacobDomagala: One way to do this is probably to modify CMake files inside bundled libs and rename the targets.

Example for `fmt`: 
- `vt`: **fmt** -> **fmt-vt** 
- `vt-tv`: **fmt** -> **fmt-vt-tv**

This way we can use isolated version for both vt and its libraries. 

We should probably discuss this on tomorrow's meeting.

lib/fmt/CMakeLists.txt
cwschilly: It is only required by `vt-tv` when we build the bindings. I've opened a [PR in vt-tv](https://github.com/DARMA-tasking/vt-tv/pull/100) to enable it conditionally based on this, so that vt doesn't have to worry about it. Once that's merged I can remove this line.

nlslatt: That PR has now been merged.

src/vt/vrt/collection/balance/lb_invoke/lb_manager.cc
cwschilly: Fixed

lib/fmt/CMakeLists.txt
cwschilly: This line has been deleted :+1:

CMakeLists.txt
cwschilly: @nlslatt @lifflander  I ended up following an approach recommended by @JacobDomagala to move `load_local_packages.cmake` (which loads vttv) after `load_bundled_libraries.cmake`. That way vt-tv and checkpoint use any libararies provided by vt, instead of the other way around.

tests/config/test_vttv.yaml
nlslatt: Should we remove the output directory now that it's no longer required?

cwschilly: Fixed :+1:

cwschilly:
@nlslatt @lifflander @JacobDomagala I want to point out [this comment](https://github.com/DARMA-tasking/vt/pull/2294#issuecomment-2236535967) before we merge--will this problem be resolved once we push to develop, or is there a fix that I need to do?

cwschilly:
Rerunning CI after merging the [vttv#85](https://github.com/DARMA-tasking/vt-tv/pull/85)

cwschilly:
CI fails on the `gcc-9, ubuntu, mpich, zoltan` pipeline (due to `VT_CI_TEST_LB_SCHEMA=1` flag):

```
##[error]Traceback (most recent call last):
##[error]  File "/vt/scripts/compare_lb_data_file.py", line 4, in <module>
##[error]    from deepdiff import DeepDiff
##[error]  File "/usr/local/lib/python3.8/dist-packages/deepdiff/__init__.py", line 10, in <module>
##[error]    from .diff import DeepDiff
##[error]  File "/usr/local/lib/python3.8/dist-packages/deepdiff/diff.py", line 30, in <module>
##[error]    from deepdiff.distance import DistanceMixin, logarithmic_similarity
##[error]  File "/usr/local/lib/python3.8/dist-packages/deepdiff/distance.py", line 1, in <module>
##[error]    import numpy as np
##[error]ModuleNotFoundError: No module named 'numpy'
##[error]The process '/usr/bin/docker' failed with exit code 2
```

This error exists for me on `develop` too. To recreate, run from `vt` dir:

```sh
docker-compose run -e BUILD_TYPE=debug -e VT_CI_TEST_LB_SCHEMA=1 ubuntu-cpp
```

thearusable:
@cwschilly I reran the failing gcc-9 job after pushing a new Docker image to Docker Hub. It seems to work now.

cwschilly:
@thearusable Awesome, thanks! I'll rebase and then this should be ready for SNL reviews

ppebay (approved):
Also submitting for final SNL review

ci/deps/vtk.sh
```diff
@@ -0,0 +1,35 @@
...
+rm ${vtk_tar_name}
+
+mkdir -p ${build_root}
+cd ${build_root}
+
+mkdir -p build
+mkdir -p install
+
+cd build
+rm -rf ./*
```
nlslatt, line 31: I think the other deps scripts remove the source and build directories after install. If you do that, you shouldn't need to clear the build directory out here.

nlslatt (commented):
Almost there

ci/deps/vtk.sh
cwschilly: Since vt-tv uses the vtk build directory (and not the install), I've kept this `rm -rf ./*` line. But I did make the following changes:
- Remove the source directory after we've built
- No longer install vtk (since only the build directory is needed)

@pierrepebay @nlslatt Let me know what you think of these changes.

cwschilly: Nevermind, I've changed it so that we use vtk's install dir instead of the build directory. Now we properly remove the source and build directories after installation.

src/vt/vrt/collection/balance/lb_invoke/lb_manager.cc
```diff
@@ -474,6 +478,57 @@ void LBManager::fatalError() {
...
+  std::unordered_map<NodeType, tv::Rank> rank_map;
+  std::unordered_map<ElementIDType, tv::ObjectInfo> info;
+  PhaseType phase = 0;
+
+  template <typename SerializerT>
+  void serialize(SerializerT& s) {
+    s | rank_map | info | phase;
+  }
+
+  friend GatherTVInfo operator+(GatherTVInfo a1, GatherTVInfo const& a2) {
```
lifflander, line 494: I don't see the reference added

lifflander (changes requested):

src/vt/vrt/collection/balance/lb_invoke/lb_manager.cc
cwschilly: I looked into this with @pierrepebay and we thought that the current implementation was best. We discussed with @JacobDomagala here: https://github.com/DARMA-tasking/vt/pull/2294#discussion_r1701833168

cwschilly:
@lifflander @nlslatt

I think we should wait to merge this PR until there is a stable vt-tv release so that we can target that release when we build vt-tv within vt (we currently just build `master`).

This would protect against future pushes to vt-tv's `master` branch that may break the pipeline. Then, before every subsequent release of both vt and vt-tv, we would just need to reassess that vt-tv is still compatible with vt.

lifflander (approved):

PR #2304, opened 2024-06-12
#2302: Add papi as dependency

### LINKED ISSUE #2302: Integrate PAPI in LB data export
This would require using [PAPI's "Low-Level" API](https://github.com/icl-utk-edu/papi/wiki/PAPI-LL).

pierrepebay:
Fixes #2302


nlslatt:
This branch needs a lot more testing before it is merged. When I put more than 4 items in `VT_EVENTS`, I'm seeing memory corruption after the first task collects PAPI events. I'm not sure where in the software stack the problem lies.

nlslatt:
> This branch needs a lot more testing before it is merged. When I put more than 4 items in `VT_EVENTS`, I'm seeing memory corruption after the first task collects PAPI events. I'm not sure where in the software stack the problem lies.

I also got memory corruption after many tasks when using only 3 items (`cache-misses,page-faults,fp_assist.any`). I think this PR still needs work.

nlslatt:
Also, this PR unconditionally requires PAPI. We need an option that turns on PAPI support instead of failing configure in all cases where PAPI is not found.

src/vt/context/runnable_context/perf_data.h
```diff
@@ -0,0 +1,198 @@
...
+            vtAbort("The PerfEventMap event map is empty; initialize vt::thePerfEventMap->getEventMap(...)");
+        }
+        else
+        {
+            const char* env_p = getenv("VT_EVENTS");
+
+            // Check if the environment variable is set
+            if (env_p == nullptr) {
+                vtWarn("Warning: Environment variable VT_EVENTS not set, defaulting to 'instructions' for the PAPI event set.\n");
+                event_names_.push_back("instructions");
```
nlslatt, line 79: This warning is printed for *every single task*, which is unscalable. That makes me wonder if the all of the other code in `PerfData` needs to be run separately for every single task or if some of it can be optimized down to once per rank.

nlslatt:
Before this can be merged, it needs to be refactored so that the user interface for perf is very similar to the existing PAPI interface within `RunnableNew`. In fact, perf and PAPI should share a single interface in `RunnableNew` (so please rename those member functions to not include `PAPI` in the name). When perf is chosen at configure time instead of PAPI, the `RunnableNew` functions will forward the requests to `vt::thePerfData()->startTaskMeasurement()` and the like. Please let me know if this is not clear.

src/vt/runnable/runnable.cc
```diff
@@ -199,6 +199,39 @@ void RunnableNew::run() {
 #endif
 }
 
+void RunnableNew::startMetrics() {
+#if vt_check_enabled(papi)
+  contexts_.lb.startPAPIMetrics();
+#elif vt_check_enabled(perf)
+  vt::thePerfData()->startTaskMeasurement();
+#else
+  #error "vt wasn't configured with measurement options"
```
cz4rs, line 208: This breaks all builds without papi or perf enabled.

src/vt/metrics/perf_data.cc
```diff
@@ -0,0 +1,202 @@
...
+  for (size_t i = 0; i < event_fds_.size(); ++i) {
+    uint64_t count = 0;
+
+    if (event_fds_[i] != -1) {
+      if (sizeof(count) != sizeof(uint64_t)) {
+          vtAbort("Buffer size mismatch: expected " + std::to_string(sizeof(uint64_t)) +
+                  " bytes, but got " + std::to_string(sizeof(count)));
+      }
+
+      ssize_t bytesRead = read(event_fds_[i], &count, sizeof(count));
```
cz4rs, line 154: FIXME: Codacy warning.

src/vt/metrics/perf_data.cc
```diff
@@ -0,0 +1,202 @@
...
+namespace vt { namespace metrics {
+
+PerfData::PerfData()
+  : event_map_(example_event_map)
+{
+  const char* env_p = getenv("VT_EVENTS");
+
+  // Check if the environment variable is set
+  if (env_p == nullptr) {
+    // vtWarn("Warning: Environment variable VT_EVENTS not set, defaulting to 'instructions' for the perf event set.\n");
```
cz4rs, line 55: Remove or uncomment.

ci/build_cpp.sh
```diff
@@ -114,6 +114,28 @@ then
     fi
 fi
 
+if test "$VT_PAPI_ENABLED" -eq 1 && "$VT_PERF_ENABLED" -eq 1
+then
+    echo "Both PAPI and perf measurements are enabled; this isn't possible, please turn off one of these options. Exiting."
```
cz4rs, line 119:
```suggestion
    echo "Both PAPI and perf measurements are enabled; this is not supported. Please turn off one of these options. Exiting."
```

ci/build_cpp.sh
```diff
@@ -114,6 +114,28 @@ then
...
+fi
+
+if test "${VT_PAPI_ENABLED:-0}" -eq 1
+then
+    if test -d "${source_dir}/lib/papi"
+    then
+        { echo "papi already in lib... not downloading, building, and installing"; } 2>/dev/null
+    else
+        cd "${source_dir}/lib"
+        git clone --depth 1 https://github.com/icl-utk-edu/papi.git
```
cz4rs, line 130: Consider pinning PAPI to a specific release or commit.

ci/azure/azure-gcc-12-ubuntu-perf.yml
```diff
@@ -0,0 +1,214 @@
...
+            val=$(./scripts/check_containers.sh)
+            echo "##vso[task.setvariable variable=compose_command]$val"
+      - task: DockerCompose@1
+        displayName: Pull container
+        inputs:
+          projectName: darma-tasking-vt
+          containerregistrytype: 'Container Registry'
+          dockerComposeFile: '**/docker-compose.yml'
+          action: 'Run a Docker Compose command'
+          dockerComposeCommand: '$(compose_command) ubuntu-cpp-clean --privileged'
```
cz4rs, line 142: Is `--privileged` the only difference vs typical Azure workflow?

ci/azure/azure-gcc-12-ubuntu-perf.yml
```diff
@@ -0,0 +1,214 @@
+###############################################################################
+############## Warning this is a generated file---do not modify ###############
+###############################################################################
```
cz4rs, line 3: This looks like it was actually handcrafted. How much effort would it take to generate it with the workflow scripts?

cmake/load_perf.cmake
```diff
@@ -0,0 +1,48 @@
...
+                # )
+                # if (PERF_STAT_RESULT EQUAL 0)
+                    message(STATUS "Perf measurements enabled.")
+                    set(vt_perf_enabled "1")
+                    set(vt_perf_found "1")
+                # else ()
+                #     message(WARNING "Perf measurements enabled but couldn't run perf stat successfully. Disabling perf measurements.")
+                #     set(vt_perf_enabled "0")
+                #     set(vt_perf_found "0")
+                # endif ()
```
cz4rs, line 32: Remove or uncomment dead code.

src/vt/context/runnable_context/papi_data.h
```diff
@@ -0,0 +1,154 @@
...
+
+    start_real_cycles = PAPI_get_real_cyc();
+    start_real_usec = PAPI_get_real_usec();
+    start_virt_cycles = PAPI_get_virt_cyc();
+    start_virt_usec = PAPI_get_virt_usec();
+  }
+
+  void stop()
+  {
+    retval = PAPI_stop(EventSet, reinterpret_cast<long long*>(values.data()));
```
cz4rs, line 139: Any issues expected with values being stored as `uint64_t`? :thinking:

src/vt/context/runnable_context/lb_data.cc
```diff
@@ -78,4 +78,18 @@ typename LBData::ElementIDStruct const& LBData::getCurrentElementID() const {
   return cur_elm_id_;
 }
 
+#if vt_check_enabled(papi)
+std::unordered_map<std::string, uint64_t> LBData::getPAPIMetrics() {
+  std::unordered_map<std::string, uint64_t> papi_metrics = {};
+  for (size_t i = 0; i < papiData_->native_events.size(); i++) {
+    papi_metrics[papiData_->native_events[i]] = papiData_->values[i];
+  }
+  papi_metrics[std::string("real_time")] = papiData_->end_real_usec - papiData_->start_real_usec;
```
cz4rs, line 87: I don't think we need to be explicit with `std::string` here:
```suggestion
  papi_metrics["real_time"] = papiData_->end_real_usec - papiData_->start_real_usec;
```

src/vt/context/runnable_context/lb_data.cc
```diff
@@ -78,4 +78,18 @@ typename LBData::ElementIDStruct const& LBData::getCurrentElementID() const {
   return cur_elm_id_;
 }
 
+#if vt_check_enabled(papi)
+std::unordered_map<std::string, uint64_t> LBData::getPAPIMetrics() {
+  std::unordered_map<std::string, uint64_t> papi_metrics = {};
+  for (size_t i = 0; i < papiData_->native_events.size(); i++) {
+    papi_metrics[papiData_->native_events[i]] = papiData_->values[i];
+  }
```
cz4rs, line 86: This seems potentially problematic (if the indexes ever get out of sync) and potentially a data duplication. Can we just an unordered_map instead of two vectors in `PAPIData`?

src/vt/context/runnable_context/papi_data.h
```diff
@@ -0,0 +1,154 @@
...
+        vtAbort(fmt::format("Couldn't event_name_to_code for {}: PAPI error {}: {}\n", native_events[i].c_str(), retval, PAPI_strerror(retval)));
+      }
+      retval = PAPI_add_event(EventSet, native);
+      if (retval != PAPI_OK) {
+        vtAbort(fmt::format("Couldn't add {} to the PAPI Event Set: PAPI error {}: {}\n", native_events[i].c_str(), retval, PAPI_strerror(retval)));
+      }
+    }
+  }
+
+  void handle_error (std::string info) const
```
cz4rs, line 120: Pass the string as a const ref.

src/vt/context/runnable_context/papi_data.h
```diff
@@ -0,0 +1,154 @@
...
+      retval = PAPI_event_name_to_code(native_events[i].c_str(), &native);
+      if (retval != PAPI_OK) {
+        vtAbort(fmt::format("Couldn't event_name_to_code for {}: PAPI error {}: {}\n", native_events[i].c_str(), retval, PAPI_strerror(retval)));
+      }
+      retval = PAPI_add_event(EventSet, native);
+      if (retval != PAPI_OK) {
+        vtAbort(fmt::format("Couldn't add {} to the PAPI Event Set: PAPI error {}: {}\n", native_events[i].c_str(), retval, PAPI_strerror(retval)));
+      }
+    }
+  }
```
cz4rs, line 118: This is long enough to be moved into .cc file.

src/vt/context/runnable_context/papi_data.h
```diff
@@ -0,0 +1,154 @@
...
+
+  void handle_error (std::string info) const
+  {
+    vtAbort(fmt::format("{}: PAPI error {}: {}\n", info.c_str(), retval, PAPI_strerror(retval)));
+  }
+
+  void start()
+  {
```
cz4rs, line 126:
```suggestion
  void start() {
```
Function's opening bracket [should be placed in the same line](https://github.com/DARMA-tasking/vt/wiki/Code-Style-Guidelines). Update other functions as well.

cz4rs: BTW there's a `.clang-format` file available in the repository. It doesn't reflect vt's preferred style ideally, but for most cases it should be close enough.

src/vt/metrics/perf_data.h
```diff
@@ -0,0 +1,106 @@
...
+  template <typename SerializerT>
+  void serialize(SerializerT& s);
+
+private:
+  std::unordered_map<std::string, std::pair<uint64_t,uint64_t>> event_map_;
+  std::vector<std::string> event_names_;
+  std::vector<int> event_fds_;
+
+  void cleanupBeforeAbort();
+  static long perf_event_open(struct perf_event_attr *hw_event, pid_t pid, int cpu, int group_fd, unsigned long flags);
```
cz4rs, line 95:
```suggestion
  static long perfEventOpen(struct perf_event_attr *hw_event, pid_t pid, int cpu, int group_fd, unsigned long flags);
```
It's not pretty, but let's do it for the sake of consistency.

src/vt/context/context.cc
```diff
@@ -90,6 +90,20 @@ Context::Context([[maybe_unused]] bool const is_interop, MPI_Comm comm) {
...
   numNodes_ = static_cast<NodeType>(numNodesLocal);
   thisNode_ = static_cast<NodeType>(thisNodeLocal);
+
+#if vt_check_enabled(papi)
+  int retval;
+
+  /* Initialize the PAPI library */
+  retval = PAPI_library_init(PAPI_VER_CURRENT);
+  if (retval != PAPI_VER_CURRENT)
+      handle_papi_error(retval);
```
cz4rs, line 100: You can make `handle_papi_error` a free function inside `context.cc` if this is the only place where it's being used. Less noise and no need to include `<papi.h>` in the header.

ci/azure/azure-gcc-12-ubuntu-perf.yml
pierrepebay: Yes here and line 150 as well. To your question about using the workflow generation script, should I modify the generation script to allow to turn on/off priviliged modes?

cz4rs: Yes, this should help with keeping the workflows in sync :+1:

cz4rs:
@pierrepebay It seems like you need a rebase here, there are still conflicts showing up.

scripts/workflows-azure.ini
```diff
@@ -241,3 +244,33 @@ distro = 22.04
...
+distro = 22.04
+output_name = ci/azure/azure-gcc-12-ubuntu-papi.yml
+vt_papi_enabled = 1
+
+[PR-tests-clang-14]
+test_configuration = "clang-14, ubuntu, mpich, papi"
+compiler_type = clang
+compiler = clang-14
+distro = 22.04
+output_name = ci/azure/azure-clang-14-ubuntu-mpich-papi.yml
```
cz4rs, line 276: Let's try with `gcc-14` ([available in Ubuntu 24.04](https://packages.ubuntu.com/noble/gcc-14)), `clang-17` and `clang-18` ([again Ubuntu 24.04](https://packages.ubuntu.com/noble/clang)). This way we can improve compiler coverage too.

cz4rs: @lifflander Are we worried about hitting any limit in Azure because of increasing number of builds?

src/vt/context/runnable_context/lb_data.h
```diff
@@ -115,10 +123,35 @@ struct LBData {
...
+   */
+  std::unordered_map<std::string, uint64_t> getPAPIMetrics();
+#endif
+
 private:
   ElementLBData* lb_data_ = nullptr;     /**< Element LB data */
   ElementIDStruct cur_elm_id_ = {};   /**< Current element ID  */
   bool should_instrument_ = false;    /**< Whether we are instrumenting */
+#if vt_check_enabled(papi)
+  std::unique_ptr<PAPIData> papiData_;
```
cz4rs, line 153: Consistency:
```suggestion
  std::unique_ptr<PAPIData> papi_data_;
```

scripts/workflows-azure.ini
lifflander: I think the number of builds is not limited. The limit is on how long then run.

src/vt/context/runnable_context/papi_data.h
cz4rs: Yep, I think this should be enough :+1:

cmake/load_perf.cmake
```diff
@@ -0,0 +1,48 @@
...
+                    message(STATUS "Perf measurements enabled.")
+                    set(vt_perf_enabled "1")
+                    set(vt_perf_found "1")
+                else ()
+                    message(WARNING "Perf measurements enabled but couldn't run perf stat successfully. Disabling perf measurements.")
+                    set(vt_perf_enabled "0")
+                    set(vt_perf_found "0")
+                endif ()
+            else ()
+                message(WARNING "Perf measurements enabled but Kernel major version is less than 4. Disabling perf measurements.")
```
lifflander, line 34:
```suggestion
                message(WARNING "Perf disabled: kernel major version is less than 4.")
```

cmake/load_perf.cmake
```diff
@@ -0,0 +1,48 @@
...
+                set(vt_perf_enabled "0")
+                set(vt_perf_found "0")
+            endif ()
+        else ()
+            message(WARNING "Perf measurements enabled but couldn't find perf_event.h. Disabling perf measurements.")
```
lifflander, line 39:
```suggestion
            message(WARNING "Perf disabled: could not find \"perf_event.h\".")
```

cmake/load_perf.cmake
```diff
@@ -0,0 +1,48 @@
...
+            set(vt_perf_enabled "0")
+            set(vt_perf_found "0")
+        endif ()
+    else ()
+        message(WARNING "Perf measurements enabled but system is not Linux. Disabling perf measurements.")
```
lifflander, line 44:
```suggestion
        message(WARNING "Perf disabled: system name is not Linux.")
```

docker-compose.yml
```diff
@@ -103,6 +103,7 @@ x-vtopts: &vtopts
   VT_EXTENDED_TESTS_ENABLED: ${VT_EXTENDED_TESTS:-1}
   VT_CODE_COVERAGE: ${VT_CODE_COVERAGE:-0}
   VT_KOKKOS_ENABLED: ${VT_KOKKOS_ENABLED:-0}
+  VT_PERF_ENABLED: ${VT_PERF_ENABLED:-0}
```
lifflander, line 106: Do we need `VT_PAPI_ENABLED` here also?

examples/collection/CMakeLists.txt
```diff
@@ -12,6 +12,12 @@ set(
   transpose
   4d_collection
 )
+if ((vt_papi_enabled AND vt_papi_found))
```
lifflander, line 15:
```suggestion
if (vt_papi_enabled AND vt_papi_found)
```

examples/collection/CMakeLists.txt
```diff
@@ -12,6 +12,12 @@ set(
...
+  list(APPEND COLLECTION_EXAMPLES do_flops_papi)
+endif()
+if ((vt_perf_enabled AND vt_perf_found))
```
lifflander, line 18:
```suggestion
if (vt_perf_enabled AND vt_perf_found)
```

examples/collection/do_flops_papi.cc
```diff
@@ -0,0 +1,291 @@
...
+//
+// Questions? Contact darma@sandia.gov
+//
+// *****************************************************************************
+//@HEADER
+*/
+
+#include <vt/transport.h>
+
+#include <cstdlib>
```
lifflander, line 46: These headers are repeated down below.

examples/collection/do_flops_papi.cc
```diff
@@ -0,0 +1,291 @@
...
+  size_t maxIter_ = 8;
+  NodeObjProxy objProxy_;
+
+public:
+  explicit GenericWork() :
+    iter_(0), msgReceived_(0), totalReceive_(0),
+    numObjs_(1), flopsPerIter_(default_flops_per_iter), maxIter_(8)
+  { }
+
+  using BlankMsg = vt::CollectionMessage<GenericWork>;
```
lifflander, line 98: Blank messages are never needed, just have an empty parameter list.

examples/collection/do_flops_papi.cc
```diff
@@ -0,0 +1,291 @@
...
+  double sum = 0.0;
+  int sign = 1;
+  for (uint64_t i = 0; i < n; ++i) {
+    sum += sign/(2.0*i+1.0);
+    sign *= -1;
+  }
+  return 4.0*sum;
+}
+
+struct NodeObj {
```
lifflander, line 75: This code seems to have been copied from an old Jacobi example and uses a lot of anti-patterns in VT. None of the messages in this example are needed, parameters should be used instead. The `NodeObj` pattern is also unnecessary along with spinning on a variable to check completion.

examples/collection/do_flops_perf.cc
```diff
@@ -0,0 +1,291 @@
...
+// Questions? Contact darma@sandia.gov
+//
+// *****************************************************************************
+//@HEADER
+*/
+
+#include <vt/transport.h>
+#include <vt/metrics/perf_data.h>
+
+#include <cstdlib>
```
lifflander, line 47: A lot of the same comments here... I think this needs to be stripped down and simplified.

src/vt/context/context.cc
```diff
@@ -90,6 +90,20 @@ Context::Context([[maybe_unused]] bool const is_interop, MPI_Comm comm) {
   communicator_ = comm;
   numNodes_ = static_cast<NodeType>(numNodesLocal);
   thisNode_ = static_cast<NodeType>(thisNodeLocal);
+
+#if vt_check_enabled(papi)
```
lifflander, line 94: I am thinking this might belong elsewhere. Does PAPI really have to do with the context? I think we should probably initialize PAPI in a new method in runtime.cc instead of the context component. There are a couple complex reasons regarding initialization that might make this location problematic, but in general the context component holds contextual state on each node and probably shouldn't be used for side effects like this in the constructor.

src/vt/context/runnable_context/papi_data.cc
```diff
@@ -0,0 +1,136 @@
...
+#include "vt/context/runnable_context/papi_data.h"
+
+namespace vt { namespace ctx {
+
+PAPIData::PAPIData()
+  : EventSet(PAPI_NULL), retval(0), 
+    start_real_cycles(0), end_real_cycles(0), start_real_usec(0), end_real_usec(0),
+    start_virt_cycles(0), end_virt_cycles(0), start_virt_usec(0), end_virt_usec(0) 
+{
+  const char* env_p = getenv("VT_EVENTS");
```
lifflander, line 52:
```suggestion
 char const* env_p = getenv("VT_EVENTS");
```

Following style guidelines, types are read from right-to-left so we put the `const` on the RHS. A pointer to a const char.

docker-compose.yml
pierrepebay: We do. Fixing.

lifflander:
When I try to build the container manually for gcc-14 with ubuntu 24.04, I get this error:
```
19.95 E: Couldn't find any package by glob 'linux-tools-6.4.16-linuxkit'
19.95 E: Couldn't find any package by regex 'linux-tools-6.4.16-linuxkit'
```
Any ideas?

pierrepebay:
Putting this into draft as the work containing perf measurements has been added with #2381. We can pick up the remaining work needed to integrate PAPI later whenever this is a priority.

PR #2305, opened 2024-06-20
#2291: Make LB data examples always up to date

### LINKED ISSUE #2291: make sure that the LB Data JSON examples are always up to date in documentation
A follow up to #2124 to avoid docs getting out of sync again. This should also help catching any accidental changes in the generated files.

**What Needs to be Done?**
- add a reference LB Data JSON file
- add a unit test / example that will generate LB Data JSON file and compare it with the reference
- make sure that the example is checked with schema validator
- embed the reference JSON file in `vt/docs/node_lb_data.md`
- make sure the JSON Python validator stays up to date also by including the segment of code

thearusable:
Closes #2291 
~~Depends on #2307~~

This PR adds:
- `lb_data_file_generator` example to create an example LBDatafile,
- an already generated file for later comparing,
- python script which will compare those two LBDatafiles,
- updates in CMake files to enable generation of the LBDatafiles by all vt examples,
- updates to documentation to insert example LBDatafile and schema which is used for validation,


scripts/LBDatafile_schema.py
```diff
@@ -0,0 +1,84 @@
...
+                            Optional('objgroup_id'): int,
+                        },
+                        'bytes': float
+                    }
+                ],
+                Optional('user_defined'): dict
+            },
+        ]
+    }
+)
```
cz4rs, line 84: missing newline

examples/lb_data/lb_data_file_generator.cc
```diff
@@ -0,0 +1,98 @@
...
+      vt::runInEpochCollective(
+        [=] { proxy.broadcastCollective<&WorkCol::someWork>(sub); });
+    }
+
+    vt::thePhase()->nextPhaseCollective();
+  }
+
+  vt::finalize();
+  return 0;
+}
```
cz4rs, line 98: missing newline

scripts/generate_and_validate_lb_data_file.py
```diff
@@ -0,0 +1,37 @@
...
+            sys.stderr.write("\n")
+            sys.stderr.flush()
+            sys.exit(1)
+        else:
+            print("Comparison OK.")
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--file-to-check", "-f", dest='file', required=True)
+    parser.add_argument("--reference-file", "-r", dest='reference_file', required=False)
```
cz4rs, line 30: This looks like a required argument.
```suggestion
    parser.add_argument("--reference-file", "-r", dest='reference_file', required=True)
```

scripts/generate_and_validate_lb_data_file.py
```diff
@@ -0,0 +1,37 @@
...
+        if not is_valid:
+            sys.stderr.write("Detected differences:\n")
+            json.dump(str(diff), sys.stderr, indent=4)
+            sys.stderr.write("\n")
+            sys.stderr.flush()
+            sys.exit(1)
+        else:
+            print("Comparison OK.")
+
+def main():
```
cz4rs, line 27: The name of the file should not contain `generate`.

scripts/generate_and_validate_lb_data_file.py
```diff
@@ -0,0 +1,37 @@
...
+
+    with open(file_to_validate) as val_file, open(reference_file) as ref_file:
+        to_validate = json.load(val_file)
+        reference = json.load(ref_file)
+        diff = DeepDiff(to_validate, reference, report_repetition=True, math_epsilon=0.1)
+        is_valid = not len(diff.affected_paths)
+
+        if not is_valid:
+            sys.stderr.write("Detected differences:\n")
+            json.dump(str(diff), sys.stderr, indent=4)
```
cz4rs, line 20: Can you post an example of error output for a small difference?

scripts/check_lb_data_files.sh
```diff
@@ -1,13 +1,14 @@
...
-set -xo pipefail
+set -exo pipefail
 
 path_to_vt_build_dir=${1}
 path_to_vt_src_dir=${2}
 cd "$path_to_vt_build_dir" || exit 1
 
 function run_schema_validator() {
     file=$1
+    echo ""
```
cz4rs, line 11: The output gets rather noisy with multiple files. I am actually trying to make it more concise in [PR #3215](https://github.com/DARMA-tasking/vt/pull/2315/files#diff-349a86e2d39b7498f5fb4f00a743737304fd1161fe56908dd81de180df840841).

scripts/check_lb_data_files.sh
```diff
@@ -1,13 +1,14 @@
 #!/usr/bin/env bash
 
-set -xo pipefail
+set -exo pipefail
```
cz4rs, line 3: Consider output for a single file. First there's an intro from the testing script:
```
+ echo 'Validating schema of json files...'
+ /vt/scripts/check_lb_data_files.sh /build/vt /vt
Validating schema of json files...
```

And then for each file:
```
+ path_to_vt_build_dir=/build/vt
+ path_to_vt_src_dir=/vt
+ cd /build/vt
+ find . -iname '*.json'
+ read f
+ grep -v compile_commands
+ run_schema_validator ./tests/TestInitialization_test_initialize_with_lb_data_and_offline_lb_1.0.json
+ file=./tests/TestInitialization_test_initialize_with_lb_data_and_offline_lb_1.0.json
+ echo ''
+ echo 'Running schema validator on: ./tests/TestInitialization_test_initialize_with_lb_data_and_offline_lb_1.0.json'
+ python3 /vt/scripts/JSON_data_files_validator.py --file_path=./tests/TestInitialization_test_initialize_with_lb_data_and_offline_lb_1.0.json

Running schema validator on: ./tests/TestInitialization_test_initialize_with_lb_data_and_offline_lb_1.0.json
Validating file: /build/vt/tests/TestInitialization_test_initialize_with_lb_data_and_offline_lb_1.0.json
Valid JSON schema in /build/vt/tests/TestInitialization_test_initialize_with_lb_data_and_offline_lb_1.0.json
Valid file
```

We can probably disable printing commands to terminal:
```suggestion
set -eo pipefail
```
And then there's still some duplication between `echo`s here and `print`s from Python code.

scripts/check_lb_data_files.sh
```diff
@@ -23,7 +24,20 @@ do
     run_schema_validator "$f"
 done
 
+find "${path_to_vt_src_dir}/examples" -iname "*.json" | while read f
```
cz4rs, line 27: `find` can take multiple directories as arguments as in `find /usr/bin/ /bin/ /usr/local/bin/ -name "*php*"`.

cz4rs: Multiple search strings too. Something like this could (maybe?) work:
```suggestion
find . "${path_to_vt_src_dir}/examples" -iname "*.json" -o -iname "*.json.br" | grep -v "compile_commands" | while read f
```

scripts/generate_and_validate_lb_data_file.py
thearusable: This example if for different values in the json:
```
Detected differences:
"{'values_changed': {\"root['phases'][0]['communications'][0]['bytes']\": {'new_value': 152.0, 'old_value': 160.0}}}"
```
There will be similar output when field is removed/added

scripts/generate_and_validate_lb_data_file.py
thearusable: Fixed.

scripts/generate_and_validate_lb_data_file.py
thearusable: Fixed.

scripts/check_lb_data_files.sh
thearusable: Ok, I removed the `-x` option and reduced the logs from comparing script

ci/build_cpp.sh
```diff
@@ -151,6 +151,7 @@ cmake -G "${CMAKE_GENERATOR:-Ninja}" \
       -Dvt_tests_num_nodes="${VT_TESTS_NUM_NODES:-}" \
       -Dvt_external_fmt="${VT_EXTERNAL_FMT:-0}" \
       -Dvt_no_color_enabled="${VT_NO_COLOR_ENABLED:-0}" \
+      -Dvt_test_lb_schema="${TEST_LB_SCHEMA:-0}" \
```
cz4rs, line 154: We should probably rename `TEST_LB_SCHEMA` to `VT_TEST_LB_SCHEMA` for consistency.

scripts/compare_lb_data_file.py
```diff
@@ -0,0 +1,36 @@
...
+
+        if not is_valid:
+            sys.stderr.write("Comparing '" + file_to_validate + "' with reference file '" + reference_file + "'... Failed!\n")
+            sys.stderr.write("Detected differences:\n")
+            json.dump(str(diff), sys.stderr, indent=4)
+            sys.stderr.write("\n")
+            sys.stderr.flush()
+            sys.exit(1)
+        else:
+            print("Comparing '" + file_to_validate + "' with reference file '" + reference_file + "'... Status OK.")
```
cz4rs, line 24: Use an f-string here:
```suggestion
            print(f"Comparing '{file_to_validate}' with reference file '}'... Status OK.")
```

cz4rs: Also maybe move the common `print("Comparing...")` part before the `if` statement.

scripts/compare_lb_data_file.py
```diff
@@ -0,0 +1,36 @@
...
+    """
+    Compares file to validate wih reference
+    """
+    with open(file_to_validate) as val_file, open(reference_file) as ref_file:
+        to_validate = json.load(val_file)
+        reference = json.load(ref_file)
+        diff = DeepDiff(to_validate, reference, report_repetition=True, math_epsilon=0.1)
+        is_valid = not len(diff.affected_paths)
+
+        if not is_valid:
```
cz4rs, line 16: I think checking if `diff` is not empty should be enough:
```suggestion
        if diff:
```

cz4rs (commented):
Looks pretty good, minor comments posted.

`nvcc` builds have failed with `Error: retrieving gpg key timed out.`. I have restarted both of them, let's see if this is enough.

thearusable:
> Looks pretty good, minor comments posted.
> 
> `nvcc` builds have failed with `Error: retrieving gpg key timed out.`. I have restarted both of them, let's see if this is enough.

One of them passed so it looks like a temporary connectivity issue.

scripts/compare_lb_data_file.py
thearusable: Fixed.

scripts/compare_lb_data_file.py
thearusable: Fixed.

ci/build_cpp.sh
thearusable: I changed it to `VT_CI_TEST_LB_SCHEMA` and the cmake parameter is now `vt_ci_generate_lb_files`

cz4rs (approved):
Looks good to me!

lifflander (approved):
Looks good to me!

PR #2315, opened 2024-07-04
#2300: Create a JSON validator that runs on all JSON produced by CI

### LINKED ISSUE #2300: Create a JSON validator that runs on all JSON produced by CI
**What Needs to be Done?**

- [ ] The check we want will verify that comm links reference tasks.

cz4rs:
Validate that comm links correctly reference tasks.

Fixes #2300 
Depends on #2305

### DIFF
--- a/scripts/JSON_data_files_validator.py
+++ b/scripts/JSON_data_files_validator.py
@@ -386,7 +386,7 @@ class JSONDataFilesValidator:
             logging.warning(f"Schema type not found in file: {file_path}. \nPassing by default when schema type not found.")
 
         if validate_comm_links:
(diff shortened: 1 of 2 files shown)


cwschilly (approved):
lgtm

cz4rs (commented):
Looks good to me.
Note that validator script is [used in LBAF](https://github.com/DARMA-tasking/LB-analysis-framework?tab=readme-ov-file#json-data-files-validator), so we need to keep an eye on LBAF CI after merging.

nlslatt (approved):
I haven't tested this, but it looks very reasonable.

PR #2329, opened 2024-07-17
#2321: Build system is mixing the system googletest headers with our internal googletest

### LINKED ISSUE #2321: Build system is mixing the system googletest headers with our internal googletest
**Describe the bug**
On the Gitlab CI, our build system finds the googletest installation that is part of the anaconda module, which causes our internal googletest to fail to compile.

cwschilly:
Fixes #2321 

---

### Explanation

The problem was in `load_zlib_package.cmake`:

```cmake
# Original load_zlib_package.cmake
find_package(ZLIB REQUIRED)
if(ZLIB FOUND)
    include_directories(${ZLIB_INCLUDE_DIRS})
...
```

The `zlib` includes found by CMake were in a directory that also contained `gtest` includes.

- By setting `include_directories`, this directory was added as an include dir for every target that vt builds. This is what caused the conflict when we tried to build the in-house gtest.

To fix this, I removed the call to `include_directories`; the correct includes will be propagated correctly, as @cz4rs [pointed out](https://github.com/DARMA-tasking/vt/pull/2329#discussion_r1681420310)

### DIFF
--- a/cmake/link_vt.cmake
+++ b/cmake/link_vt.cmake
@@ -165,6 +165,9 @@ function(link_target_with_vt)
     target_link_libraries(
       ${ARG_TARGET} PUBLIC ${ARG_BUILD_TYPE} ${ZLIB_LIBRARIES}
     )
+    target_include_directories(
+      ${ARG_TARGET} PUBLIC $<BUILD_INTERFACE:${ZLIB_INCLUDE_DIRS}>
+    )
   endif()
 
   if (NOT DEFINED ARG_LINK_FMT AND ${ARG_DEFAULT_LINK_SET} OR ARG_LINK_FMT)
--- a/cmake/link_vt.cmake
+++ b/cmake/link_vt.cmake
@@ -165,9 +165,6 @@ function(link_target_with_vt)
     target_link_libraries(
       ${ARG_TARGET} PUBLIC ${ARG_BUILD_TYPE} ${ZLIB_LIBRARIES}
     )
-    target_include_directories(
-      ${ARG_TARGET} PUBLIC $<BUILD_INTERFACE:${ZLIB_INCLUDE_DIRS}>
-    )
   endif()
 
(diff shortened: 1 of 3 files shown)


cmake/load_zlib_package.cmake
```diff
@@ -4,8 +4,6 @@
 #
 
 find_package(ZLIB REQUIRED)
```
cz4rs, line 6: That's the only relevant line here. The check below is a dead code, CMake will error out when Zlib is not found (because of `REQUIRED`). You can move this directly into `load_packages.cmake` and get rid of `load_zlib_package.cmake` completely.

cwschilly: Fixed

cmake/link_vt.cmake
```diff
@@ -164,6 +164,9 @@ function(link_target_with_vt)
     target_link_libraries(
       ${ARG_TARGET} PUBLIC ${ARG_BUILD_TYPE} ${ZLIB_LIBRARIES}
     )
+    target_include_directories(
+      ${ARG_TARGET} PUBLIC $<BUILD_INTERFACE:${ZLIB_INCLUDE_DIRS}>
+    )
```
cz4rs, line 169: Actually it might be enough to use `target_link_libraries` - if Zlib defines it's library correctly, the include paths will be propagated. Can you try removing `target_include_directories` completely?

cwschilly: You're right, include paths are propagated correctly without specifying `target_include_directories` :+1:

nlslatt: @cwschilly Was this tested outside of CI and found to work?

cwschilly: Yes I tested locally and on SRN

nlslatt (approved):
Looks good

cz4rs (approved):

nlslatt:
Please rebase this.

lifflander (approved):
Looks good

PR #2333, opened 2024-08-06
#2299: Do not deploy LB when average load is smaller than estimated load balancing cost

### LINKED ISSUE #2299: Do not deploy LB when average load is too small
This lower bound should not be hardcoded, but rather use either a machine-dependent absolute quantity (i.e., square root of machine epsilon) or, better yet, a lowed bound that uses first principle: one such example would be to use the smallest object size divided by the number of ranks (or nodes).

_Originally posted by @ppebay in https://github.com/DARMA-tasking/vt/pull/2278#discussion_r1619468807_

thearusable:
Closes #2299

### DIFF
--- a/src/vt/vrt/collection/balance/baselb/baselb.h
+++ b/src/vt/vrt/collection/balance/baselb/baselb.h
@@ -190,12 +190,29 @@ protected:
   );
 
   /**
-   * \brief Get the estimated time needed for load balancing
+   * \brief Get the estimated time needed for load-balancing
    *
    * \return the estimated time
    */
   auto getCollectiveEpochCost() const {
     return std::chrono::nanoseconds(100);
+  }
+
+  /**
+   * \brief Check if load-balancing should be done
+   *
+   * \return true when the maximum load exceeds the cost of load balancing; false otherwise
+   */
+  bool maxLoadExceedsLBCost() const {
+    auto const max = base_stats_->at(lb::Statistic::Rank_load_modeled).at(
+      lb::StatisticQuantity::max
+    );
+    auto max_in_ns = std::chrono::duration_cast<std::chrono::nanoseconds>(
+      std::chrono::duration<double>(max)
+    );
+
+    // Compare the maximum rank load to the estimated load-balancing cost
+    return max_in_ns > getCollectiveEpochCost();
   }
 
 private:
--- a/src/vt/vrt/collection/balance/baselb/baselb.h
+++ b/src/vt/vrt/collection/balance/baselb/baselb.h
@@ -164,16 +164,6 @@ struct BaseLB {
   bool isCommAware() const { return comm_aware_; }
   void recvSharedEdges(CommMsg* msg);
 
-  /**
-   * \brief Get the estimated time needed for load balancing
-   *
-   * \return the estimated time
-   */
-  double getCollectiveEpochCost() const {
-    // 100 ns
-    return 0.0000001;
-  }
-
 protected:
   void getArgs(PhaseType phase);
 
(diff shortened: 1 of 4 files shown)


cz4rs:
@thearusable `invalid project name "DARMA-tasking/vt": must consist only of lowercase alphanumeric characters, hyphens, and underscores as well as start with a letter or number` strikes again :shrug:

nlslatt (changes requested):
PR needs to target `develop`

JacobDomagala:
> > @thearusable `invalid project name "DARMA-tasking/vt": must consist only of lowercase alphanumeric characters, hyphens, and underscores as well as start with a letter or number` strikes again 🤷
> 
> @cz4rs The target branch of this PR is `2201-implement-memory-aware-temperedlb-in-vt-rebased`, which is over two months old. This likely explains the error.

@cz4rs @thearusable Not related to this PR but most recent (on develop) spack workflow failed with that message (https://dev.azure.com/DARMA-tasking/DARMA/_build/results?buildId=60702&view=logs&j=3dc8fd7e-4368-5a92-293e-d53cefc8c4b3&t=1364e02f-4e1e-51cf-2821-fd252a3c0e34&l=10)

thearusable:
> > > @thearusable `invalid project name "DARMA-tasking/vt": must consist only of lowercase alphanumeric characters, hyphens, and underscores as well as start with a letter or number` strikes again 🤷
> > 
> > 
> > @cz4rs The target branch of this PR is `2201-implement-memory-aware-temperedlb-in-vt-rebased`, which is over two months old. This likely explains the error.
> 
> @cz4rs @thearusable Not related to this PR but most recent (on develop) spack workflow failed with that message (https://dev.azure.com/DARMA-tasking/DARMA/_build/results?buildId=60702&view=logs&j=3dc8fd7e-4368-5a92-293e-d53cefc8c4b3&t=1364e02f-4e1e-51cf-2821-fd252a3c0e34&l=10)

@JacobDomagala I will take a look on that

cz4rs (approved):
Looks good to me.

src/vt/vrt/collection/balance/baselb/baselb.h
```diff
@@ -188,6 +188,16 @@ struct BaseLB {
...
   );
 
+  /**
+   * \brief Get the estimated time needed for load balancing
+   *
+   * \return the estimated time
+   */
+  double getCollectiveEpochCost() const {
+    // 100 ns
+    return 0.0000001;
```
PhilMiller, line 198: This ideally would use std chrono types and literals. Barring that, at least use scientific notation and specify the units.

Also, I think 100ns is on the low side, but it really doesn't matter.

src/vt/vrt/collection/balance/greedylb/greedylb.cc
```diff
@@ -160,7 +160,8 @@ void GreedyLB::loadStats() {
   bool should_lb = false;
   this_load_begin = this_load;
 
-  if (avg_load > 0.0000000001) {
+  // Use an estimated load-balancing cost on average rank load to load-balance
+  if (avg_load > getCollectiveEpochCost()) {
```
PhilMiller, line 164: I came here to note that I think calculation like this should use max load, rather than average. Otherwise, if a large-scale job has maximal imbalance of a very fine-grain problem, averaging might mask it.

src/vt/vrt/collection/balance/baselb/baselb.h
thearusable: Ok, I changed it to `std::chrono::nanoseconds`.

src/vt/vrt/collection/balance/greedylb/greedylb.cc
thearusable: @lifflander What should we use in this case?

thearusable: After the discussion on the meeting, we should use the `max` load instead of `avg` load and move the logic for checking that to `BaseLB` to be used only by the real load balancers.

lifflander (approved):
Looks good to me!

thearusable:
@lifflander PR rebased. Ready to merge.

PR #2339, opened 2024-08-26
#2338: Fix the object ID to have to the correct features in the JSON reader

### LINKED ISSUE #2338: Fix the object ID to have to the correct features in the JSON reader
**What Needs to be Done?**
Instead of directly using the object ID from the json file, the json reader should generate a vt-compatible object ID based on what appears in the json file. This would allow it to encode object type, home rank, if the object is migratable or not, etc.

**Additional context**
Related to testing in #2278

cwschilly:
Fixes #2338

### DIFF
--- a/src/vt/vrt/collection/balance/lb_data_holder.cc
+++ b/src/vt/vrt/collection/balance/lb_data_holder.cc
@@ -319,12 +319,12 @@ LBDataHolder::LBDataHolder(nlohmann::json const& j)
                                       vt::elm::elm_id_num_bits,
                                       Field
                                     >(static_cast<Field>(object));
-              auto cid = task["entity"]["collection_id"];
-              auto idx = task["entity"]["index"];
               elm = elm::ElmIDBits::createCollectionImpl(migratable,
                                                          strippedObject,
                                                          home,
                                                          node);
+              auto cid = task["entity"]["collection_id"];
+              auto idx = task["entity"]["index"];
               if (cid.is_number() && idx.is_array()) {
                 std::vector<uint64_t> arr = idx;
                 auto proxy = static_cast<VirtualProxyType>(cid);
--- a/src/vt/vrt/collection/balance/lb_data_holder.cc
+++ b/src/vt/vrt/collection/balance/lb_data_holder.cc
@@ -42,10 +42,8 @@
 */
 
 #include "vt/vrt/collection/balance/lb_data_holder.h"
-#include "vt/vrt/collection/balance/node_lb_data.h"
 #include "vt/context/context.h"
 #include "vt/elm/elm_id_bits.h"
-#include "vt/elm/elm_comm.h"
 
 #include <nlohmann/json.hpp>
 
--- a/src/vt/vrt/collection/balance/lb_data_holder.cc
+++ b/src/vt/vrt/collection/balance/lb_data_holder.cc
@@ -42,7 +42,10 @@
 */
 
 #include "vt/vrt/collection/balance/lb_data_holder.h"
+#include "vt/vrt/collection/balance/node_lb_data.h"
 #include "vt/context/context.h"
+#include "vt/elm/elm_id_bits.h"
+#include "vt/elm/elm_comm.h"
 
 #include <nlohmann/json.hpp>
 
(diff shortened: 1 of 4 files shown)


src/vt/vrt/collection/balance/lb_data_holder.cc
```diff
@@ -296,14 +299,18 @@ LBDataHolder::LBDataHolder(nlohmann::json const& j)
           auto node = task["node"];
           auto time = task["time"];
           auto etype = task["entity"]["type"];
+          auto home = task["entity"]["home"];
+          bool migratable = task["entity"]["migratable"];
+          auto seq_id = theNodeLBData()->getNextElm();
```
nlslatt, line 304: If you use `getNextElm()` here, you've essentially thrown out the ID as given in the json file. That's a problem because different ranks will then refer to the same object by different IDs. For json files where the IDs are just 0, 1, 2, etc., you could use the ID from the json file (from `object`) in place of `seq_id`. However, I'm not sure if this will cause problems with vt-generated json files, where `object` could be very must larger because the bits for migratable, home, and type are already set. You might need to strip off any high bits before using `object`.

Have you checked here that the task belongs to a collection and is not an objgroup or bare handler? For those types of tasks, it isn't right to use `createCollectionImpl()`, but I don't know if it will actually cause a problem.

@lifflander??

lifflander: Agreed, we need to get the ID from the file, and call the appropriate method of get the bare ID without the bits set. There are different bit patterns used in collections, objgroups, etc. For now, we could just do this for collections, perhaps? 

You should call the `get` version of this call (which is not implemented anywhere right now):

https://github.com/DARMA-tasking/vt/blob/96689a3ea6eb1d6e984f698aefb2ec868852e264/src/vt/elm/elm_id_bits.cc#L112-L114

cwschilly: @lifflander Here's a test for `getField`. How does this translate to what I'm trying to do?

```cpp
TEST_F(TestBitPacking, test_bit_packing_1) {
  using vt::utils::BitPacker;
  using Field = uint64_t;
  Field x = 0xFAADFEEDFBBBFCCDull;
  auto x0 = BitPacker::getField<0, 16, Field>(x);
  auto x1 = BitPacker::getField<16, 16, Field>(x);
  auto x2 = BitPacker::getField<32, 16, Field>(x);
  auto x3 = BitPacker::getField<48, 16, Field>(x);

  EXPECT_EQ(x0, 0xFCCDull);
  EXPECT_EQ(x1, 0xFBBBull);
  EXPECT_EQ(x2, 0xFEEDull);
  EXPECT_EQ(x3, 0xFAADull);
}
```

For example, given an `elm` that already contains the encoded info for migratability, home, and node, would the first 16 bits (e.g. `getField<0,16,Field>(elm)`) correspond to the original ID?

cwschilly: Here is how I've implemented the stripping:

```cpp
using Field = uint64_t;
auto strippedObject = BitPackerType::getField<
                        vt::elm::eElmIDProxyBitsNonObjGroup::ID,
                        vt::elm::elm_id_num_bits,
                        Field
                      >(static_cast<Field>(object));
auto elm = elm::ElmIDBits::createCollectionImpl(migratable,
                                                strippedObject,
                                                home,
                                                node);
```

tests/unit/utils/test_bit_packing.nompi.cc
```diff
@@ -125,4 +126,20 @@ TEST_F(TestBitPacking, test_bit_packing_dynamic_check_masking_5) {
   EXPECT_EQ(x, 0x0000FCCD0000FEEDull);
 }
 
+TEST_F(TestBitPacking, test_bit_packing_and_create_collection) {
+  using Field = uint64_t;
+  bool migratable = true;
+  int home = 0;
+  int node = 0;
```
nlslatt, line 133: This check would be a little more robust if the `node` was non-zero

nlslatt: Sorry, I meant `home`

cwschilly: Fixed

nlslatt (approved):
Looks good

cz4rs (approved):

lifflander (approved):

PR #2343, opened 2024-09-05
#2342: Change schema to allow for `id` or `seq_id`

### LINKED ISSUE #2342: Change schema to allow for `id` or `seq_id`
**What Needs to be Done?**

Motivated by #2201 testing

Change the schema to allow either `id` or `seq_id` in the JSON. `id` should be assumed bit-packed, while `seq_id` is assumed not bit-packed.

cwschilly:
Fixes #2342


src/vt/vrt/collection/balance/lb_data_holder.cc
```diff
@@ -41,14 +41,72 @@
...
+    is_collection = true;
+  } else {
+    is_collection = false;
+  }
+}
+
+ElementIDStruct get_elm_from_object_info_(
+  const nlohmann::json& object, bool is_bitpacked, bool is_migratable,
+  const nlohmann::json& home) {
+  using Field = uint64_t;
```
nlslatt, line 72: This should really be `ElementIDType` rather than the current definition of `ElementIDType`, I think

src/vt/vrt/collection/balance/lb_data_holder.cc
```diff
@@ -41,14 +41,72 @@
...
+  if (is_bitpacked) {
+    object_id = BitPackerType::getField<
+      vt::elm::eElmIDProxyBitsNonObjGroup::ID, vt::elm::elm_id_num_bits, Field>(
+      static_cast<Field>(object));
+  } else {
+    object_id = static_cast<Field>(object);
+  }
+
+  return elm::ElmIDBits::createCollectionImpl(
+    is_migratable, object_id, home, theContext()->getNode());
```
nlslatt, line 84: This is more complicated than it needs to be. It should be simple enough to leave inline in the constructor. When bit-packed, just do `elm = ElementIDStruct{object, node};` like you're doing when it's not a collection element. When it's not bit-packed, do `elm::ElmIDBits::createCollectionImpl(
    is_migratable, static_cast<Field>(object), home, node);` (use the `node` variable from the json file, not the current rank, even though they should be the same).

src/vt/vrt/collection/balance/lb_data_holder.cc
```diff
@@ -41,14 +41,72 @@
...
+  } else {
+    object = field["seq_id"];
+    is_bitpacked = false;
+  }
+  if (field.find("collection_id") != field.end()) {
+    is_collection = true;
+  } else {
+    is_collection = false;
+  }
+}
```
nlslatt, line 67: You should probably check if the `object` is a number here for consistency and completeness. Please follow the vt conventions for naming functions, and probably make it a static member function of the `LBDataHolder` since it's only used from there.

nlslatt (commented):
Some comments below. Please familiarize yourself with the vt style guidelines (function naming, etc.). I'm still looking at the tests but wanted to give you feedback on the main implementation right away.

src/vt/vrt/collection/balance/lb_data_holder.cc
cwschilly: Almost fixed.

> When bit-packed, just do `elm = ElementIDStruct{object, node};` like you're doing when it's not a collection element. When it's not bit-packed, do `elm::ElmIDBits::createCollectionImpl( is_migratable, static_cast<Field>(object), home, node);`

This is done.

> (use the `node` variable from the json file, not the current rank, even though they should be the same).

This isn't.

I've added a `this_node_` member variable to the LBDataHolder class. On instantiation, it is populated with `theContext()->getNode()`. This is what is used to created every `elm` in the file.

If we wanted to use the `node` from the JSON, we'd need to store which node corresponds to each task, since `node` is not included in the `entity` block (and therefore is not accessible from the communications field).

The best approach I could think of would be an unordered map of object IDs to their node. @nlslatt Should I implement this?

nlslatt: What you've done makes sense given the comm issue.

src/vt/vrt/collection/balance/lb_data_holder.h
```diff
@@ -135,6 +158,8 @@ struct LBDataHolder {
   void readMetadata(nlohmann::json const& j);
 
 public:
+  /// The current node
+  NodeType this_node_;
```
nlslatt, line 162: This should be initialized to something here because of the default constructor. `vt::uninitialized_destination` is a good option.

nlslatt (commented):
Almost there

src/vt/vrt/collection/balance/lb_data_holder.h
cwschilly: Fixed, thanks

PR #2393, opened 2025-01-10
#2388: LB stat files should not be written out every phase and should be marked with a trace event

### LINKED ISSUE #2388: LB stat files should not be written out every phase and should be marked with a trace event
**Describe the bug**
I think we should output them every 100 or so phases, maybe with a user-settable flag? We should mark these with a bracketed trace user event so we know how much it costs.

cwschilly:
Fixes #2388

### DIFF
--- a/src/vt/trace/trace.cc
+++ b/src/vt/trace/trace.cc
@@ -84,7 +84,6 @@ void Trace::initialize() /*override*/ {
 
   // Register a trace user event to demarcate flushes that occur
   flush_event_ = trace::registerEventCollective("trace_flush");
-  write_stats_event_ = trace::registerEventCollective("write_lb_stats");
 #endif
 }
 
--- a/src/vt/trace/trace.cc
+++ b/src/vt/trace/trace.cc
@@ -84,6 +84,7 @@ void Trace::initialize() /*override*/ {
 
   // Register a trace user event to demarcate flushes that occur
   flush_event_ = trace::registerEventCollective("trace_flush");
+  write_stats_event_ = trace::registerEventCollective("write_lb_stats");
 #endif
 }
 
--- a/src/vt/vrt/collection/balance/lb_invoke/lb_manager.cc
+++ b/src/vt/vrt/collection/balance/lb_invoke/lb_manager.cc
@@ -667,7 +667,7 @@ void LBManager::commitPhaseStatistics(PhaseType phase) {
   }
 
   #if vt_check_enabled(trace_enabled)
-  auto write_lb_stats_event = theTrace()->registerUserEventColl("write_lb_stats");
+  auto write_lb_stats_event = theTrace()->getWriteStatsEvent();
   theTrace()->addUserEventBracketedBegin(write_lb_stats_event);
   #endif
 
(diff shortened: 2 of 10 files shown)


src/vt/vrt/collection/balance/lb_invoke/lb_manager.cc
```diff
@@ -658,11 +658,18 @@ void LBManager::stagePostLBStatistics(
...
+  // Statistics output when LB is enabled, appropriate flag is enabled,
+  // and the user-defined frequency is respected
+  if (theContext()->getNode() != 0 \
+      or !theConfig()->vt_lb_statistics \
+      or phase % theConfig()->vt_lb_statistics_freq != 0) {
     return;
   }
 
+  #if vt_check_enabled(trace_enabled)
+  theTrace()->addUserEventBracketedBegin(1);
```
lifflander, line 670: I think we need to actually register a user event with an associated name and pass it here instead of just passing `1`.

lifflander (changes requested):

src/vt/vrt/collection/balance/lb_invoke/lb_manager.cc
cwschilly: Fixed

src/vt/vrt/collection/balance/lb_invoke/lb_manager.cc
```diff
@@ -658,11 +658,19 @@ void LBManager::stagePostLBStatistics(
...
+  // Statistics output when LB is enabled, appropriate flag is enabled,
+  // and the user-defined frequency is respected
+  if (theContext()->getNode() != 0 \
+      or !theConfig()->vt_lb_statistics \
+      or phase % theConfig()->vt_lb_statistics_freq != 0) {
     return;
   }
 
+  #if vt_check_enabled(trace_enabled)
+  auto write_lb_stats_event = theTrace()->registerUserEventColl("write_lb_stats");
```
lifflander, line 670: We need to do this once the first time it's used (or on startup) and then reuse the event.

lifflander (changes requested):

src/vt/vrt/collection/balance/lb_invoke/lb_manager.cc
cwschilly: I added a check to the registerEvents function so that it will first see if the event already exists in `user_event_` before creating a new event hash.

lifflander: There is still an issue here that this is a collective call that all nodes must make. It seems this is conditional on the node, which will cause a problem I think.

lifflander (changes requested):

src/vt/trace/trace.cc
```diff
@@ -84,6 +84,7 @@ void Trace::initialize() /*override*/ {
 
   // Register a trace user event to demarcate flushes that occur
   flush_event_ = trace::registerEventCollective("trace_flush");
+  write_stats_event_ = trace::registerEventCollective("write_lb_stats");
```
lifflander, line 87: Let's move this to LBManager

lifflander (changes requested):

src/vt/trace/trace.cc
cwschilly: Fixed

lifflander (approved):

PR #2394, opened 2025-01-10
#2390: getRecentTime() should not be used when tracing is enabled

### LINKED ISSUE #2390: getRecentTime() should not be used when tracing is enabled
**Describe the bug**
In runnable.cc, using the "recent time" to avoid calling the timer which is expensive leads to highly inaccurate traces. We should get the current time when tracing is compile-time enabled and being used at runtime (i.e., the trace context is enabled for a runnable).

cwschilly:
Fixes #2390

### DIFF
--- a/src/vt/runnable/runnable.cc
+++ b/src/vt/runnable/runnable.cc
@@ -124,7 +124,16 @@ void RunnableNew::run() {
   {
     needs_time = contexts_.lb.needsTime();
   }
-  TimeType start_time = needs_time ? theSched()->getRecentTime() : TimeType{NAN};
+
+  auto getTime = [&]() {
+#if vt_check_enabled(trace_enabled)
+  return theTrace() ? timing::getCurrentTime() : theSched()->getRecentTime();
+#else
+  return theSched()->getRecentTime();
+#endif
+};
+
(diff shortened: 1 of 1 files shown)


src/vt/runnable/runnable.cc
```diff
@@ -124,7 +124,16 @@ void RunnableNew::run() {
...
     needs_time = contexts_.lb.needsTime();
   }
-  TimeType start_time = needs_time ? theSched()->getRecentTime() : TimeType{NAN};
+
+  auto getTime = [&]() {
+#if vt_check_enabled(trace_enabled)
+    if (theTrace()) return timing::getCurrentTime();
+    else
+#endif
+    return theSched()->getRecentTime();
```
PhilMiller, line 133:
```suggestion
    {
      return theSched()->getRecentTime();
    }
```
Makes it more clear that there's a control flow statement preceding it.

This also better matches the style adopted for conditionally-compiled conditions.

cwschilly: Fixed, thanks

cz4rs (approved):

lifflander (approved):

PR #2396, opened 2025-01-13
#2389: trace: start tracing invoke calls 

### LINKED ISSUE #2389: Make invoke without a message work with tracing
**What Needs to be Done?**

Right now, if the user wants to do `proxy[x].invoke<&Test::xyz>(some args)` (not the message variant) we don't get tracing on these. We want to put these in the registry when tracing is enabled and add the trace context to them in that case.

lifflander:
Fixes #2389


cwschilly (commented):
LGTM

Issue #2387, opened 2025-01-08
Hashed trace user events should do an all-reduce at the end instead of sending pt2pts to rank 0 during the run


lifflander:
**Describe the bug**
When using `trace::TraceScopedEventHash scope("XXX");`, the node that calls this immediately sends a message to rank 0 with the hashed event, causing tons of messages that interfere with the application run. We should all-reduce these at the end of the run when we need to write out the STS file.

PR #2404, opened 2025-02-26
#2403: runtime: seperate out runtime dependencies from startup dependencies

### LINKED ISSUE #2403: Distinguish startup dependencies from use dependencies in VT components
**What Needs to be Done?**

Right now, `ComponentPack` has a struct `Deps` which holds a set of dependencies which determines startup order and which components have to be constructed when a component is added to the runtime. We need two different sets of dependencies: ones that matter for startup, i.e., the component can't be initialized correctly until after other components are initialization; and, ones that are use dependencies which don't matter for startup ordering but should be added to the runtime component set if a component is added since they might use them in the component code.

lifflander:
Fixes #2403

### DIFF
--- a/src/vt/runtime/component/component.h
+++ b/src/vt/runtime/component/component.h
@@ -100,14 +100,14 @@ struct Component : BaseComponent {
    * \brief Constructor for a new component with the appropriate typed
    * dependencies
    *
-   * \param[in] DepsPack a pack of typed dependencies for the given component
-   * that it depends on
+   * \param[in] StartupDepsPack a pack of typed dependencies for the given
+   * component that it depends on
    *
    * \return the new component
    */
-  template <typename... Deps>
-  Component(DepsPack<Deps...>) {
-    ComponentRegistry::dependsOn<T, Deps...>();
+  template <typename... StartupDeps, typename... RuntimeDeps>
+  Component(StartupDepsPack<StartupDeps...>, RuntimeDepsPack<RuntimeDeps...>) {
+    ComponentRegistry::dependsOn<T, StartupDeps...>();
   }
 
   /**
--- a/src/vt/runtime/component/component.h
+++ b/src/vt/runtime/component/component.h
@@ -97,20 +97,6 @@ struct Component : BaseComponent {
   Component() = default;
 
   /**
-   * \brief Constructor for a new component with the appropriate typed
-   * dependencies
-   *
-   * \param[in] StartupDepsPack a pack of typed dependencies for the given
-   * component that it depends on
-   *
-   * \return the new component
-   */
-  template <typename... StartupDeps, typename... RuntimeDeps>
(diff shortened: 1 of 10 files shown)


src/vt/runtime/runtime.cc
```diff
@@ -835,91 +885,127 @@ void Runtime::initializeComponents() {
...
+      ctx::Context,                        // Everything depends on theContext
+      collective::CollectiveAlg,           // Depends on collective scope
+      objgroup::ObjGroupManager            // Since it's an objgroup
+    >{},
+    RuntimeDeps<
       messaging::ActiveMessenger,         // Depends on active messenger for messaging
       vrt::collection::CollectionManager, // For RDMA on collection elements
-      objgroup::ObjGroupManager,          // For RDMA on objgroups
-      collective::CollectiveAlg           // Depends on collective scope
+      objgroup::ObjGroupManager           // For RDMA on objgroups
```
nlslatt, line 946: Does `objgroup::ObjGroupManager` need to be in both sections?

lifflander: No that was a mistake.

tests/unit/runtime/test_component_construction.cc
```diff
@@ -154,4 +160,86 @@ TEST_F(TestComponentConstruction, test_component_construct_dispatch_move_3) {
 
 ////////////////////////////////////////////////////////////////////////////////
 
+////////////////////////////////////////////////////////////////////////////////
+// Test dependencies -- runtime and startup to ensure they are being followed
```
cwschilly, line 164: Since we ran into this on #2387, it might be good to test that cyclical dependencies are handled correctly as well

lifflander: I was a little nervous about an expect fail test, but I'll give it a try

src/vt/runtime/component/component.h
```diff
@@ -100,14 +100,14 @@ struct Component : BaseComponent {
...
+   * component that it depends on
    *
    * \return the new component
    */
-  template <typename... Deps>
-  Component(DepsPack<Deps...>) {
-    ComponentRegistry::dependsOn<T, Deps...>();
+  template <typename... StartupDeps, typename... RuntimeDeps>
+  Component(StartupDepsPack<StartupDeps...>, RuntimeDepsPack<RuntimeDeps...>) {
+    ComponentRegistry::dependsOn<T, StartupDeps...>();
```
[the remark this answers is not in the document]
lifflander: Removed this constructor all together as it's not necessary. I meant to do that before actually but forgot.

tests/unit/runtime/test_component_construction.cc
lifflander: Done

cwschilly (approved):

PR #2412, opened 2025-03-17
#2411: Unregister locations of destroyed collection elements

### LINKED ISSUE #2411: Unregister entity locations when destroyed
**What Needs to be Done?**
The EntityLocationCoord objects have a function for unregistering entities that is supposed to be called whenever an entity is destroyed (not migrated), but the function is never used in the codebase. See this paragraph: https://github.com/DARMA-tasking/vt/blob/eae6fa90c7b446dad7561a99def90f06c1c09b79/src/vt/topos/location/location.h#L86

Not unregistering can cause an assertion failure when restoring a destroyed collection: https://github.com/DARMA-tasking/vt/blob/eae6fa90c7b446dad7561a99def90f06c1c09b79/src/vt/topos/location/location.impl.h#L103

The current checkpoint restore function does not trigger this issue, because it allows restored collections to be given a new proxy ID. However, the new checkpoint/restore functions using the latest Magistrate features will restore to the same proxy ID.

Call the unregister function whenever we are destroying a collection element.

We could also accept that duplicate registrations are acceptable and remove the assertion. No idea if this has any other consequences though.

Matthew-Whitlock:
Closes #2411

### DIFF
--- a/src/vt/runtime/runtime.cc
+++ b/src/vt/runtime/runtime.cc
@@ -920,17 +920,17 @@ void Runtime::initializeComponents() {
   p_->registerComponent<vrt::collection::CollectionManager>(
     &theCollection,
     StartupDeps<
-      ctx::Context                         // Everything depends on theContext
+      ctx::Context,             // Everything depends on theContext
+      location::LocationManager // For element location
     >{},
     RuntimeDeps<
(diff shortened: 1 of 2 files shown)


lifflander (approved):
Looks good to me

PR #2425, opened 2025-04-24
#2424 Update fmt version

### LINKED ISSUE #2424: Update CMake for external fmt 11.1.3
**Describe the bug**
VT is now designed for fmt 11.1.3, but requests version 7.1.0, so building fails when using the spack package.

**To Reproduce**

`spack install darma-vt@develop`

Matthew-Whitlock:
Closes #2424

### DIFF
--- a/cmake/load_bundled_libraries.cmake
+++ b/cmake/load_bundled_libraries.cmake
@@ -30,7 +30,7 @@ if(${vt_external_fmt})
   else()
     message(STATUS "vt_external_fmt = ON but neither fmt_DIR nor fmt_ROOT is provided!")
   endif()
-  find_package(fmt 11.1.3 REQUIRED)
+  find_package(fmt 10.2.1 REQUIRED)
 
 else()
   set(FMT_LIBRARY fmt)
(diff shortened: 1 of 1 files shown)


cmake/load_bundled_libraries.cmake
```diff
@@ -30,7 +30,7 @@ if(${vt_external_fmt})
   else()
     message(STATUS "vt_external_fmt = ON but neither fmt_DIR nor fmt_ROOT is provided!")
   endif()
-  find_package(fmt 7.1.0 REQUIRED)
+  find_package(fmt 11.1.3 REQUIRED)
```
cz4rs, line 33:
```suggestion
  find_package(fmt 10.2.1 REQUIRED)
```

We should consider using a lower requirement. For example `distBVH` uses `10.2.1` ([see CI setup](https://github.com/sandialabs/distBVH/blob/9a296b5cf7886d75da051b39c5705bb6c0e6c1cc/spack.yaml#L8)), which builds and runs fine with the latest `develop`.

cz4rs (approved):
Looks good to me, you just need to slap `#2424` at the beginning of the commit message.

nmm0 (approved):

lifflander (approved):

PR #2429, opened 2025-04-24
#2428 Checkpoint using latest Magistrate trait-based API

### LINKED ISSUE #2428: Checkpoint using new Magistrate trait-based API
**What Needs to be Done?**
Use the new trait-based API to allow more flexible checkpoint/recovery patterns

KokkosResilience checkpoint/restart needs more control over checkpointing objgroups, individual elements of collections, and writing to different locations (e.g. streams vs VT's file directory approach).

Matthew-Whitlock:
Closes #2428


src/vt/vrt/collection/types/storage/storable.impl.h
```diff
@@ -50,7 +50,11 @@ namespace vt { namespace vrt { namespace collection { namespace storage {
 
 template <typename SerializerT>
 void Storable::serialize(SerializerT& s) {
-  s | map_;
+  //We need a traitless serializer, since StoreElmBase uses virtual
+  //serialization. It doesn't use any traits to specialize serializing
+  //though, so this is fine.
+  auto clean_s = s.setTraits();
```
PhilMiller, line 56: Is `setTraits()` actually just getting `s` with an empty set of traits, or is it mutating `s` itself? This looks like either mis-use, or mis-naming.

Matthew-Whitlock: This is getting `s` with an empty set of traits. The traits are immutable, since they're types. I can see how the naming may be misleading, but `setTraits()` (which gets a serializer with only the specified traits) needs to be distinct from `withTraits()` and `withoutTraits()` (which get serializers with the specified traits added-to/removed-from the existing traits). Happy to rename if you have any suggestions, but that'd be a PR in Magistrate

Matthew-Whitlock: `asTraits()` seems a little clearer

PR #2440, opened 2025-06-06
#2438:  [CI] Use docker bake to define the matrix of vt builds

### LINKED ISSUE #2438: [CI] Use docker bake to define the matrix of `vt` builds
Use bake to define the matrix of VT builds that target the WF base containers.

Follow-up to #2400.
See #2407 for context.

cz4rs:
fixes #2438


cz4rs:
List available workflow images:
```
workflows  docker buildx bake --print build-all | grep "lifflander1/vt:"
        "lifflander1/vt:wf-amd64-alpine-3.16-clang-cpp"
        "lifflander1/vt:wf-amd64-ubuntu-20.04-clang-10-cpp"
        "lifflander1/vt:wf-amd64-ubuntu-20.04-clang-9-cpp"
        "lifflander1/vt:wf-amd64-ubuntu-20.04-gcc-10-cpp"
        "lifflander1/vt:wf-amd64-ubuntu-20.04-gcc-10-openmpi-cpp"
        "lifflander1/vt:wf-amd64-ubuntu-20.04-gcc-9-cpp"
        "lifflander1/vt:wf-amd64-ubuntu-20.04-gcc-9-cuda-11.4.3-cpp"
        "lifflander1/vt:wf-amd64-ubuntu-20.04-gcc-9-cuda-12.2.0-cpp"
        "lifflander1/vt:wf-amd64-ubuntu-20.04-icpx-cpp"
        "lifflander1/vt:wf-amd64-ubuntu-22.04-clang-11-cpp"
        "lifflander1/vt:wf-amd64-ubuntu-22.04-clang-12-cpp"
        "lifflander1/vt:wf-amd64-ubuntu-22.04-clang-13-cpp"
        "lifflander1/vt:wf-amd64-ubuntu-22.04-clang-14-cpp"
        "lifflander1/vt:wf-amd64-ubuntu-22.04-clang-15-cpp"
        "lifflander1/vt:wf-amd64-ubuntu-22.04-gcc-11-cpp"
        "lifflander1/vt:wf-amd64-ubuntu-22.04-gcc-12-cpp"
        "lifflander1/vt:wf-amd64-ubuntu-22.04-gcc-12-vtk-cpp"
        "lifflander1/vt:wf-amd64-ubuntu-22.04-gcc-12-zoltan-cpp"
        "lifflander1/vt:wf-amd64-ubuntu-24.04-clang-16-cpp"
        "lifflander1/vt:wf-amd64-ubuntu-24.04-clang-16-vtk-cpp"
        "lifflander1/vt:wf-amd64-ubuntu-24.04-clang-16-zoltan-cpp"
        "lifflander1/vt:wf-amd64-ubuntu-24.04-clang-17-cpp"
        "lifflander1/vt:wf-amd64-ubuntu-24.04-clang-18-cpp"
        "lifflander1/vt:wf-amd64-ubuntu-24.04-gcc-13-cpp"
        "lifflander1/vt:wf-amd64-ubuntu-24.04-gcc-14-cpp"
```

ci/docker/vt.dockerfile
```diff
@@ -0,0 +1,44 @@
...
+ARG VT_TESTS_NUM_NODES
+ARG VT_TRACE_ENABLED
+ARG VT_TRACE_RUNTIME_ENABLED
+ARG VT_TV_ENABLED
+ARG VT_UBSAN_ENABLED
+ARG VT_UNITY_BUILD_ENABLED
+ARG VT_WERROR_ENABLED
+ARG VT_ZOLTAN_ENABLED
+
+COPY . /vt
```
cz4rs: TODO:
We use `RUN --mount=type=bind,source=ci,target=${WF_TMP_DIR}` instead of `COPY` in workflows. Double check which is better here.

docker-bake.hcl
```diff
@@ -0,0 +1,353 @@
...
+    VT_WERROR_ENABLED              = vt_werror(item)
+    VT_ZOLTAN_ENABLED              = vt_zoltan(item)
+  }
+
+  # to get the list of available images from DARMA-tasking/workflows:
+  # workflows > docker buildx bake --print build-all | grep "lifflander1/vt:"
+  matrix = {
+    item = [
+      {
+        image = "amd64-alpine-3.16-clang-cpp"
```
cz4rs, line 239: Using something close to actual docker tag here instead of constructing it from compiler/distro/distro-version etc. This avoids code duplication (from `workflows/docker-bake.hcl`) and lets us simply rely on produced images.

ci/docker/vt.dockerfile
```diff
@@ -0,0 +1,44 @@
...
+RUN /vt/ci/build_cpp.sh /vt /build
+RUN /vt/ci/test_cpp.sh /vt /build
+RUN /vt/ci/build_vt_sample.sh /vt /build
```
cz4rs: FIXME
`vt_sample` script runs, but does not actually build the sample:
```

#8 [5/5] RUN /vt/ci/build_vt_sample.sh /vt /build
#8 0.053 + source_dir=/vt
#8 0.053 + build_dir=/build
#8 0.053 + inclusion_type=
#8 0.053 + test '' = TPL
#8 0.053 + test '' = EXT_LIB
#8 0.054 ++ grep ID
#8 0.054 ++ grep -c alpine
#8 0.055 ++ true
#8 0.056 + is_alpine=0
#8 0.056 + test 0 -eq 0
#8 0.056 + test 0 -eq 1
#8 DONE 0.1s
```

docker-bake.hcl
```diff
@@ -0,0 +1,353 @@
...
+        image = "amd64-ubuntu-24.04-clang-18-cpp"
+      },
+      {
+        image = "amd64-ubuntu-24.04-gcc-13-cpp"
+        vt_trace = 1
+        vt_pool = 0
+        vt_asan = 1
+        vt_unity_build = 0
+        #FIXME
+        lsan_options = "suppressions=/vt/tests/lsan.supp"
```
cz4rs: This should be passed as an environment variable. There was no need for additional code in Dockerfile when we were using `docker run` to execute the tests. Now it seems like we'll have to add `ARG` + `ENV` combo.

ci/docker/vt.dockerfile
cz4rs: This is less elegant than multiple stages as originally intended:
```Dockerfile
COPY . /vt
RUN /vt/ci/build_cpp.sh /vt /build

FROM build AS test
RUN /vt/ci/test_cpp.sh /vt /build

FROM test AS test_sample
RUN /vt/ci/build_vt_sample.sh /vt /build
```

The issue is that AFAIK there's no way to inherit from matrix in docker-bake.hcl, so the following does not work:
```
target "vt-test" {
  inherits = ["vt-build-all"]
  target = "test"
}
```
And if we use `inherits = ["vt-build"]`, we'd have to define the matrix again.

cz4rs:
`Build spack-package` build was not migrated at this stage (it's heavily customized compared to others). I suggest handling the transition in a separate issue. Another thing is that it's currently failing on `develop` anyways :shrug:

.github/workflows/build-docker-images.yml
```diff
@@ -0,0 +1,53 @@
...
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Build
+        uses: docker/bake-action@v6
+        with:
+          targets: ${{ matrix.target }}
+          files: docker-bake.hcl
+          # push: true
```
cz4rs: This should be something along the lines of:
```suggestion
          push: ${{ github.ref == 'refs/heads/develop' }}
```
We need to login into DockerHub first though.

.github/workflows/build-docker-images.yml
```diff
@@ -0,0 +1,53 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        target: ${{ fromJson(needs.prepare.outputs.targets) }}
+
+    steps:
+      # - name: Login to DockerHub
+      #   uses: docker/login-action@v3
+      #   with:
+      #     username: ${{ secrets.DOCKER_USERNAME }}
```
cz4rs: @lifflander This needs correct secret names, can you please check?

.github/workflows/build-docker-images.yml
```diff
@@ -0,0 +1,53 @@
+name: Build & Push Docker images
```
cz4rs: We need a shorter name, tags are pretty informative so the PR summary is fine I think:
image

docker-bake.hcl
cz4rs: Seems that ASAN build is running fine without this line.

.dockerignore
```diff
@@ -1,2 +1 @@
-.git
```
cz4rs, line 1: Seems necessary to get correct git revision during build. The repository status is reported as `DIRTY` for some reason :thinking: 
```
 #9 30.16 -- Reading head file /vt/.git/HEAD
#9 30.16 -- Using git executable at "/usr/bin/git"
#9 30.16 -- GIT_REFSPEC: ""
#9 30.16 -- GIT_SHA1: "ed76fa3e89c2531e2bf1ccbbe04cb81301874647"
#9 30.16 -- GIT_EXACT_TAG: "remotes/pull/2440/merge"
#9 30.16 -- GIT_DESCRIPTION: "remotes/pull/2440/merge-0-ged76fa3e89"
#9 30.16 -- Git Clean Status: "DIRTY"
#9 30.16 -- Configuring /vt/vt_git_revision.cc.in to generate /build/vt/src/vt/configs/generated/vt_git_revision.cc.
```

cz4rs: Another option is to copy `.git` explicitly in the Dockerfile.

ci/docker/vt.dockerfile
cz4rs: This is now building correctly except for the `EXT_LIB` inclusion case.

.github/workflows/build-docker-images.yml
lifflander: Yes, I think they should be `DOCKER_USERNAME` and `DOCKER_PASSWORD`

ci/docker/vt.dockerfile
lifflander: I think mounting is better since the copy operation isn't actually necessary. Also makes interactive jobs easier since we can modify the code in the container. That's essentially what docker compose is doing.

lifflander (approved):

ci/docker/vt.dockerfile
lifflander: Yes, I noticed that limitation also.

lifflander (approved):
Overall, looks great

ci/docker/vt.dockerfile
cz4rs: Done.

ci/docker/vt.dockerfile
```diff
@@ -0,0 +1,46 @@
...
+ARG VT_TESTS_NUM_NODES
+ARG VT_TRACE_ENABLED
+ARG VT_TRACE_RUNTIME_ENABLED
+ARG VT_TV_ENABLED
+ARG VT_UBSAN_ENABLED
+ARG VT_UNITY_BUILD_ENABLED
+ARG VT_WERROR_ENABLED
+ARG VT_ZOLTAN_ENABLED
+
+RUN --mount=target=/vt \
```
cz4rs: My thinking is that we should use something like this for caching:
```suggestion
RUN --mount=target=/vt --mount=type=cache,source=/ccache,target=/build/ccache \
```
(in combination with the usual `actions/cache@v4`).
Unfortunately I haven't been able to get it to work yet.

docker-bake.hcl
```diff
@@ -0,0 +1,365 @@
...
+}
+
+function "vt_build_shared_libs" {
+  params = [item]
+  result = lookup(item, "vt_build_shared_libs", "")
+}
+
+function "vt_build_trace_only" {
+  params = [item]
+  result = lookup(item, "vt_build_trace_only", "1")
```
cz4rs, line 22: I tried to use empty string by default for all those helpers, which allows the default values from `build_cpp` script to take precedence. For `VT_TRACE_ONLY` the default value is `0` in the script and `1` in `workflows-azure.ini`.

cz4rs:
@lifflander Everything seems to work except for `ccache`. I went through the builds and as far as I could see the configurations are matching Azure.
I will give caching another look today in the evening.

TODO:
- add separate issue for caching if needed
- create a follow-up issue to remove Azure builds once we decide that this is stable
- add separate issue / update existing one to fix and migrate spack build

ci/docker/vt.dockerfile
```diff
@@ -0,0 +1,46 @@
...
+ARG VT_TESTS_NUM_NODES
+ARG VT_TRACE_ENABLED
+ARG VT_TRACE_RUNTIME_ENABLED
+ARG VT_TV_ENABLED
+ARG VT_UBSAN_ENABLED
+ARG VT_UNITY_BUILD_ENABLED
+ARG VT_WERROR_ENABLED
+ARG VT_ZOLTAN_ENABLED
+
+RUN --mount=target=/vt --mount=type=cache,target=/build/ccache \
```
cz4rs, line 43: `--mount=type=cache,target=/build/ccache` helps when I rerun builds locally, but I'm not sure how to connect it with `actions/cache@v4`.

docker-bake.hcl
```diff
@@ -0,0 +1,362 @@
...
+  dockerfile = "ci/docker/vt.dockerfile"
+  platforms = [
+    "linux/amd64",
+    # "linux/arm64"
+  ]
+  ulimits = [
+    "core=0"
+  ]
+  # FIXME: verify that caching works as intended
+  # cache-from = [
```
cz4rs, line 178: Fixed the typo (`cache_from` -> `cache-from`) and the directory. This doesn't work for me locally when uncommented:
```
ERROR: Cache export is not supported for the docker driver.
Switch to a different driver, or turn on the containerd image store, and try again.
Learn more at https://docs.docker.com/go/build-cache-backends/
```

ci/docker/vt.dockerfile
lifflander: I think this is the right direction with `mount-type=cache`. You need to tell ccache that the cache is located there. Then do something like this:

```
- uses: actions/cache@v4
    id: cache
    with:
      path: ?? how to specify where the host puts the cache ??
      key: /* generate a key based on build/date/time */ like we do with Azure
```

lifflander: From what I'm reading, mount-type=cache won't work with github because it's not intended for this use case.

docker-bake.hcl
lifflander: I'm not sure what the problem is here.. maybe the `~/ccache`? Probably should be a full path.

lifflander: Strange.. I am getting the same error locally with different paths.

lifflander: I changed my local settings to enable containerd image store and now it might be working.. not sure. But it's running.

ci/docker/vt.dockerfile
cz4rs: > From what I'm reading, mount-type=cache won't work with github because it's not intended for this use case.

Yeah, I've seen it too. There are some hacks discussed in the internet, but nothing standard it seems.

PR #2444, opened 2025-06-17
#2443: location: add options to keep cache valid and create independent instances

### LINKED ISSUE #2443: Add location manager mode to guarantee cache is valid
**What Needs to be Done?**

The location in the cache may be invalided during migrations. If migrations are only allowed between phases, we can create a mode where the cache is always valid or invalided if we don't want to update it.

There are several design considerations here. With the current default cache size of 2k records, updating the whole cache everywhere might be expensive. Namely, after a migration, no rank knows all the other ranks that have it in cache. Perhaps we want to limit the size by default in this mode or fully invalidate the cache after migrations?

lifflander:
Fixes #2443

### DIFF
--- a/src/vt/collective/reduce/operators/functors/plus_op.h
+++ b/src/vt/collective/reduce/operators/functors/plus_op.h
@@ -91,6 +91,15 @@ struct PlusOp< std::array<T,N> > {
   }
 };
 
+template <typename T, typename U>
+struct PlusOp<std::unordered_map<T, U>> {
+  void operator()(std::unordered_map<T, U>& v1, std::unordered_map<T, U> const& v2) {
+    for (auto const& [key, value] : v2) {
+      vtAssert(v1.find(key) == v1.end(), "Key must not exist in other map");
+      v1[key] = value;
+    }
+  }
+};
 
 }}}} /* end namespace vt::collective::reduce::operators */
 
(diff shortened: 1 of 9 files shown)


cz4rs:
@lifflander `vt_werror` needs to be set to 1 by default in `docker-bake.hcl`:
```
function "vt_werror" {
  params = [item]
  result = lookup(item, "vt_werror", "1")
}
```

cz4rs:
Here's the list of differences in configuration found for `gcc-12, vt-tv` Azure and `gcc-12-vtk` Bake build:

```
Azure:
-Dvt_trace_only=0 
-Dvt_werror_enabled=1
-Dvt_tv_enabled=1 
-Dvt_unity_build_enabled=1 
-Dvt_mpi_guards=1
-Dvt_tests_num_nodes=2
-Dvt_no_color_enabled=1

Bake:
-Dvt_trace_only=1 
-Dvt_werror_enabled=0
-Dvt_tv_enabled=0 
-Dvt_unity_build_enabled=0 
-Dvt_mpi_guards=0
-Dvt_tests_num_nodes=
-Dvt_no_color_enabled=0
```

This needs to be fixed by setting correct defaults in `docker-bake.hcl`, possibly in a separate PR.

cz4rs:
@lifflander This should work fine after rebasing (#2450 is done).

cz4rs:
@lifflander Bake builds are passing fine now. Signature check is failing after my rebase, so you should probably do a rebase of you own and maybe squash my last commit as well.

cwschilly (approved):
LGTM

PR #2445, opened 2025-06-20
#2448: use docker bake - caching experiments

### LINKED ISSUE #2448: [CI] enable caching for docker bake builds
Follow-up to #2438.

Enable caching with `ccache` in a similar fashion as is used in Azure builds.

cz4rs:
fixes #2448


cz4rs:
Testing the changes for cache here to avoid polluting original PR. Builds need to be re-run after finishing to see if the execution time drops.

Switching to containerd image store for local development is a nuisance, but I think we can live with it if CI works well.

cz4rs:
`Warning: Path Validation Error: Path(s) specified in the action for caching do(es) not exist, hence no cache is being saved.`
Duh :/ Maybe an absolute path is the way to go.

PR #2449, opened 2025-06-24
#2448: ci: bake: enable local output and use it for caching

### LINKED ISSUE #2448: [CI] enable caching for docker bake builds

cz4rs:
fixes #2448

This is pretty simple and straightforward approach, with the main downside being that local output means exporting entire container (while we are only interested in `/build/ccache`). Using `docker cp` would probably achieve the same result with an additional step in `build-docker-images.yml`.

Other possibilities:
- look for the cache location for type=cache mounts: https://stackoverflow.com/a/79489859
- copy files from container to host: https://stackoverflow.com/a/51186557

### DIFF
--- a/.github/workflows/build-docker-images.yml
+++ b/.github/workflows/build-docker-images.yml
@@ -41,7 +41,7 @@ jobs:
 
       - uses: actions/cache@v4
         with:
-          path: ~/ccache
+          path: docker-output/build/ccache
           key: ${{ matrix.target }}-${{ github.ref }}-${{ github.sha }}
           restore-keys: ${{ matrix.target }}-${{ github.ref }}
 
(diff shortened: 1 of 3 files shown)


cz4rs:
`ccache` does not work correctly in Alpine build (even though the output from previous build is correctly stored / restored with `actions/cache@v4`):
```
#8 [build 2/3] RUN --mount=target=/vt <<EOF (if [ -d vt/docker-output/build/ccache ]; then...)
#8 0.159 Summary:
#8 0.159   Hits:               0 /    0
#8 0.159     Direct:           0 /    0
#8 0.159     Preprocessed:     0 /    0
#8 0.159   Misses:             0
#8 0.159     Direct:           0
#8 0.159     Preprocessed:     0
#8 0.159 Primary storage:
#8 0.159   Hits:               0 /    0
#8 0.159   Misses:             0
#8 0.159   Cache size (GB): 0.00 / 0.70 (0.00 %)
#8 0.159 
#8 0.159 Use the -v/--verbose option for more details.
#8 DONE 2.0s
```

Can be fixed in a separate issue.

lifflander (approved):
This looks great for CI. Good job finding a solution for that!

PR #2458, opened 2025-07-18
#2457: Don't install fmt files when we're using external fmt for `vt-trace`

### LINKED ISSUE #2457: `vt-trace-only` target only installs `fmt/core.h`
This is causing compile issues as core.h file tries to include format.h.

JacobDomagala:
Fixes #2457 

We don't want to copy bundled `fmt` files if we're using external `fmt`. 

Also in case of bundled fmt, we want to also copy the rest of the headers but do we actually need to copy the files at all? We link against `fmt` target that is configured to install fmt files (to a parent directory, for vt-runtime export set), so it makes sense that a single location for multiple targets should be fine :thinking:

### DIFF
--- a/cmake/trace_only_functions.cmake
+++ b/cmake/trace_only_functions.cmake
@@ -92,10 +92,16 @@ function(create_trace_only_target)
   # we don't use INSTALL_DIR/include/ as include directory
   # we use INSTALL_DIR/include/vt-trace instead
   # so we have to install FMT lib aswell
-  install(
-    FILES "${CMAKE_CURRENT_SOURCE_DIR}/../lib/fmt/include/fmt-vt/core.h"
-    DESTINATION "include/vt-trace/fmt"
-  )
+  if (NOT vt_external_fmt)
+    install(
+      FILES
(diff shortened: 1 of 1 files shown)


cz4rs (approved):
Looks good to me.

Using
> single location for multiple targets

sounds even better, although the comment in the code suggests that going this way was a conscious choice. I'd be happy with merging this as is and posting another PR to explore single location.

cz4rs:
@JacobDomagala If you posted this as a draft and only later changed to `Ready for review`, most of the Azure jobs won't start (they don't start for drafts and you need to push again from `Ready to review` state).

Also I've had to restart `Unsigned commits` check again, it's been failing recently because of some API limits. If it keeps occuring we'll need to investigate.

JacobDomagala:
> Also I've had to restart `Unsigned commits` check again, it's been failing recently because of some API limits. If it keeps occuring we'll need to investigate.

We could enforce signed commits via github's settings (https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches#require-signed-commits)

lifflander (approved):

PR #2481, opened 2025-09-24
#2479: CI: add ASAN options to always detect leaks and abort on error (non-zero code)

### LINKED ISSUE #2479: CI does not fail when ASAN fails
**What Needs to be Done?**

- [ ] First do this
- [ ] Then do that.
- [ ] Also this other thing.

**Additional context**

lifflander:
Fixes #2479


docker-bake.hcl
```diff
@@ -184,6 +184,11 @@ function "vt_ldms" {
   result = lookup(item, "vt_ldms", "")
 }
 
+function "vt_asan_options" {
+  params = [item]
+  result = lookup(item, "vt_asan_options", "detect_leaks=1 abort_on_error=1")
```
nlslatt: I get more informative leak stack traces when I add the following arguments to this list: `fast_unwind_on_malloc=0 malloc_context_size=150 max_stack_depth=0`. I'm not sure if all three of these are strictly necessary.

cwschilly: I've added these options. They seem to pick up on the error that was not triggered [in this PR](https://github.com/DARMA-tasking/vt/pull/2480)

antoinemeyer5 (approved):
LGTM

PR #2483, opened 2025-09-30
#2482: lib: update bundled libs cmake version requirements to avoid failure with CMake 4.0

### LINKED ISSUE #2482: Fix mac os x build on github
**Describe the bug**
The build is failing due to a incompatibility with libfort's specified cmake version.

lifflander:
Fixes #2482

This fixes the Mac build which has been consistently failing due to a much later cmake version:
```
  Compatibility with CMake < 3.5 has been removed from CMake.
```

### DIFF
--- a/lib/EngFormat-Cpp/CMakeLists.txt
+++ b/lib/EngFormat-Cpp/CMakeLists.txt
@@ -1,5 +1,5 @@
 project(EngFormat-Cpp CXX)
-cmake_minimum_required(VERSION 3.0)
+cmake_minimum_required(VERSION 3.0...4.0)
 
 add_library(${PROJECT_NAME} include/EngFormat-Cpp/eng_format.hpp src/eng_format.cpp)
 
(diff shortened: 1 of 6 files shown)


cz4rs (approved):
This seems alright as a temporary measure, but might cause issues for users with CMake > 4.0. As long as the CI passes, we can have reasonable confidence that no bugs were introduced by touching `cmake_minimum_required`.

I think we should upgrade the libs in follow-up issues, this should help with everything except for libfort:
- mimalloc [3.18](https://github.com/microsoft/mimalloc/blob/main/CMakeLists.txt#L1)
- EngFormat-Cpp - to activity upstream, but the lib is so small that we should be able to just raise min required version
- brotli [3.15](https://github.com/google/brotli/blob/master/CMakeLists.txt)
- json is the one that is using [3.5...4.0](https://github.com/nlohmann/json/blob/develop/CMakeLists.txt) actually, not sure about this one
- libfort - no activity upstream, stuck with [3.0](https://github.com/seleznevae/libfort/blob/develop/CMakeLists.txt)
- yaml-cpp - [3.5...3.30](https://github.com/jbeder/yaml-cpp/blob/master/CMakeLists.txt) here with the explicit comment about avoid CMake warnings :shrug:

nlslatt:
@lifflander This is now critical: `Compatibility with CMake < 3.5 has been removed from CMake.`

nlslatt (approved):

PR #2488, opened 2025-10-06
#2487: CI: switch one build to use the C++20 standard

### LINKED ISSUE #2487: Support building under the C++20 standard
When dropped into an application build system that has `set(CMAKE_CXX_STANDARD 20)`, the vt build produces many `is not usable in a constant expression` (fmt-related) and `is not a constant expression` (at least partly fmt-related) errors.

lifflander:
Fixes #2487


nlslatt (approved):
Looks good

PR #2498, opened 2026-01-07
#2497: cmake: add option to disable find_package(MPI)

### LINKED ISSUE #2497: Add option to disable finding MPI with cmake
**What Needs to be Done?**

- [ ] This is causing issues when VT is used as a TPL and it's unnecessary.

lifflander:
Fixes #2497


nlslatt (changes requested):
Changes also need to be made in link_vt.cmake so that vt doesn't try to explicitly link against MPI. Is there a way to do CI testing with the MPI compiler wrappers?

nlslatt (approved):
Looks good but I have not tested it yet

nlslatt (approved):

PR #2501, opened 2026-01-08
#2500: Implement MPI_Improbe/MPI_Imrecv for active and data messages

### LINKED ISSUE #2500: Reduce network pressure with unexpected send/recvs
We are getting non-deterministic failures with a large MPI application running on a supercomputer with exit code 134. This is using MPICH. Sometimes the application will complete, but most of the times it fails. The failure location seems to vary also in the code.

libfabric:34975:1767894133::cxi:core:cxip_ux_onload_cb():2659<warn> nid001043: RXC (0x9f1:0) PtlTE 20:[Fatal] LE resources not recovered during flow control. FI_CXI_RX_MATCH_MODE=[hybrid|software] is required

lifflander:
Fixes #2500

Large-scale runs on HPE Cray Slingshot (libfabric CXI provider) hit non-deterministic aborts under heavy unexpected-message pressure due to hardware match-table exhaustion and Iprobe→Irecv race conditions. Switch to MPI_Improbe/MPI_Imrecv (MPI 3+) to claim messages from unexpected queue immediately.

- Updated tryProcessIncomingActiveMsg to use MPI_Improbe+MPI_Imrecv on MPI 3+
- Updated recvDataMsgBuffer to use MPI_Improbe for data messages (enqueue=false)
- Extended recvDataDirect signature with optional first_msg and first_chunk_bytes parameters
- Implemented matched-first-chunk path in recvDataDirect using MPI_Imrecv
- Added MPI_VERSION >= 3 guards for all new code paths
- Maintained backward compatibility with MPI < 3 using existing Iprobe/Irecv
- Add new broker to post Irecv so messages are not unexpected partitioned by size using tags

### DIFF
--- a/src/vt/messaging/active.cc
+++ b/src/vt/messaging/active.cc
@@ -210,9 +210,17 @@ MsgSizeType ActiveMessenger::packMsg(
   return size + ptr_bytes;
 }
 
+// Choose active message tag based on final size
+/*static*/ MPI_TagType ActiveMessenger::selectActiveTag(MsgSizeType size) {
+  if (size <= ActiveRecvBroker::caps_[0]) return static_cast<MPI_TagType>(MPITag::ActiveMsgS);
+  if (size <= ActiveRecvBroker::caps_[1]) return static_cast<MPI_TagType>(MPITag::ActiveMsgM);
+  if (size <= ActiveRecvBroker::caps_[2]) return static_cast<MPI_TagType>(MPITag::ActiveMsgL);
+  if (size <= ActiveRecvBroker::caps_[3]) return static_cast<MPI_TagType>(MPITag::ActiveMsgXL);
+  return static_cast<MPI_TagType>(MPITag::ActiveMsgTag); // fallback
+}
+
 EventType ActiveMessenger::sendMsgBytesWithPut(
-  NodeType const& dest, MsgSharedPtr<BaseMsgType> const& base,
-  TagType const& send_tag
+  NodeType const& dest, MsgSharedPtr<BaseMsgType> const& base
 ) {
   auto msg = base.get();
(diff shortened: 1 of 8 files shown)


PhilMiller:
The len -> sublen change in the diagnostic call is a distinct bug fix of the existing code, right? If so, I think it ought to be in its own commit, rather than mixed into the broader feature addition of the first commit in this PR.

lifflander:
> The len -> sublen change in the diagnostic call is a distinct bug fix of the existing code, right? If so, I think it ought to be in its own commit, rather than mixed into the broader feature addition of the first commit in this PR.

Done

PhilMiller:
In recvDataDirect, if you declare start_chunk unconditionally, then the for loop doesn't need to be part of the conditional compilation.

src/vt/messaging/active.cc
```diff
@@ -190,8 +196,8 @@ MsgSizeType ActiveMessenger::packMsg(
 ) {
   vt_debug_print(
     verbose, active,
-    "packMsg: msg_size={}, put_size={}, ptr={}\n",
-    size, ptr_bytes, print_ptr(ptr)
+    "packMsg: msg_size={}, put_size={}, ptr={}, ptr_bytes={}\n",
+    size, ptr_bytes, print_ptr(ptr), ptr_bytes
```
nlslatt: What's the reason for the redundant print of `ptr_bytes`?

nlslatt (approved):
Looks good and seems to work for my use case.

src/vt/messaging/active.cc
lifflander: This was a mistake

lifflander: Fixed

PR #2503, opened 2026-01-29
2502 fix cmake ordering and quiet mpi output when not testing

### LINKED ISSUE #2502: Clean up MPI-related warnings and configuration messages when testing is not enabled
**What Needs to be Done?**

- [ ] First do this
- [ ] Then do that.
- [ ] Also this other thing.

**Additional context**

nlslatt:
Closes #2502


nlslatt:
@lifflander Although this PR allows printing of the MPI config when examples are enabled but the build of tests is not, I see that we don't set up ctest when only examples are being built. Is that the way we want it, or should it be changed? In particular, should we add an `OR vt_build_examples` to this conditional?
```
if (vt_build_tests
   AND "${CMAKE_PROJECT_NAME}" STREQUAL "${PROJECT_NAME}")
  # CTest implies enable_testing() and defines the BUILD_TESTING option.
  # The default of BUILD_TESTING is ON.
  # Testing is only enabled if the actual project being built is VT.
  include(CTest)
endif()
```

nlslatt:
The pipelines are failing because `cmake/load_mpi_package.cmake` knows about `vt_build_tests` and `vt_build_examples` when they are explicitly defined on the cmake command-line but but not when they are defaulted.

nlslatt:
> The pipelines are failing because `cmake/load_mpi_package.cmake` knows about `vt_build_tests` and `vt_build_examples` when they are explicitly defined on the cmake command-line but but not when they are defaulted.

I believe this is because we load packages before defining configuration options:
```
# Load packages that are required for core VT build
include(cmake/load_packages.cmake)

# Some options use TPL information (whether Perl was found etc.)
include(cmake/configure_options.cmake)
```

@lifflander Should this order be reversed?

lifflander:
> @lifflander Although this PR allows printing of the MPI config when examples are enabled but the build of tests is not, I see that we don't set up ctest when only examples are being built. Is that the way we want it, or should it be changed? In particular, should we add an `OR vt_build_examples` to this conditional?
> 
> ```
> if (vt_build_tests
>    AND "${CMAKE_PROJECT_NAME}" STREQUAL "${PROJECT_NAME}")
>   # CTest implies enable_testing() and defines the BUILD_TESTING option.
>   # The default of BUILD_TESTING is ON.
>   # Testing is only enabled if the actual project being built is VT.
>   include(CTest)
> endif()
> ```

I think we should probably enable CTest in that case so we can run the examples as pseudo-tests. It's sort of a edge case, because I don't think it will impact many pipelines (except maybe sample project where I think we might disable tests but keep examples building).

lifflander:
> > The pipelines are failing because `cmake/load_mpi_package.cmake` knows about `vt_build_tests` and `vt_build_examples` when they are explicitly defined on the cmake command-line but but not when they are defaulted.
> 
> I believe this is because we load packages before defining configuration options:
> 
> ```
> # Load packages that are required for core VT build
> include(cmake/load_packages.cmake)
> 
> # Some options use TPL information (whether Perl was found etc.)
> include(cmake/configure_options.cmake)
> ```
> 
> @lifflander Should this order be reversed?

Yes, that is a bug. We should always get our options before we do anything. Maybe check to see if it should be moved even earlier?

lifflander (approved):
Looks good!

PR #2505, opened 2026-02-16
2504 fix CI failures

### LINKED ISSUE #2504: Fix CI failures
When upgrading the CI pipeline there seem to be more failures now.

lifflander:
Fixes #2504

This fixes two issues;
1. the format specifiers for the ZoltanLB (conditionally compiled) do not conform to the compile-time specification with the new FMT that VT is now using for all vt_debug_print
2. Location manager: Fixes the test failure and potential performance issue with cache updates. With this change, we now connect eager location updates to the message’s epoch and send the first update directly back to the original sender, rather than cascading via all the intermediate hops. This fixes the first iteration cache miss (in the test) by ensuring the origin sees the cache update before epoch termination.


cwschilly (approved):
LGTM

PR #2510, opened 2026-06-11
#2508: CI: replace copy of entire folder with copies of specified files

### LINKED ISSUE #2508: Docs CI pipeline failing
I've been noticing that the docs CI pipeline is failing: `CI / bake (vt-build-amd64-ubuntu-20-04-gcc-9-cpp-docs)`. Can you investigate why and if we can upgrade m.css to make it more stable?

antoinemeyer5:
Fixes #2508


lifflander (approved):
LGTM

PR #2523, opened 2026-08-06
#2514: Start using LB repository instead of the custom TemperedLB here

### LINKED ISSUE #2514: Start using LB repository instead of the custom TemperedLB here

antoinemeyer5:
Fixes #2514 

Notes:
- clang 9/10 cannot build LB, so `vt_lb = 0`
- gcc 9 CUDA job raise "standards after C++17 are not supported", so `vt_lb = 0`
- "amd64-alpine-3.16-clang-cpp" fail with an upstream DARMA-tasking/LB issue: CMF selection returned an invalid rank and triggered "Selected node not found" so put `vt_extended_tests = 0`


docker-bake.hcl
```diff
@@ -310,16 +312,20 @@ target "vt-build-all" {
       },
       {
         image = "amd64-ubuntu-20.04-gcc-10-openmpi-cpp"
-        vt_lb = 1
+        # The Spack package does not yet stage the external LB and comm sources.
+        vt_lb = 0
```
antoinemeyer5: Changes to make in Spack:
- Use Spack `resource()` directives
- They fetch extra source repositories
- And place them inside VT’s staged source tree before CMake runs

antoinemeyer5:
## Description 

The remaining failure is a mismatch between VT’s Spack CI script and the variants declared by `DARMA-tasking/spack-package`.
It is unrelated to the external LB implementation.

VT’s CI requests six variants that the recipe does not define:
- `doxygen_enabled`
- `fcontext_enabled`
- `diagnostics_enabled`
- `diagnostics_runtime_enabled`
- `unity_build_enabled`
- `zoltan_enabled`

## Short-term fix

Update VT’s `ci/test_spack_package.sh` and remove the six unsupported entries.

## Long-term fix

If those six options are intended to be supported through Spack, declare them in `DARMA-tasking/spack-package/packages/darma-vt/package.py`.
So we should add the corresponding `variant()` declarations and CMake mappings to `package.py`.

cwschilly:
> ## Short-term fix
> 
> Update VT’s `ci/test_spack_package.sh` and remove the six unsupported entries.

I think this is the correct, long-term fix. Those variants are no longer defined in the Spack package

lifflander:
> Fixes #2514
> 
> Notes:
> 
>     * clang 9/10 cannot build LB, so `vt_lb = 0`
> 
>     * gcc 9 CUDA job raise "standards after C++17 are not supported", so `vt_lb = 0`

Do we need to add more pipelines to replace these or do you think we have enough coverage. Can you look into that?

antoinemeyer5:
> > Fixes #2514
> > Notes:
> > ```
> > * clang 9/10 cannot build LB, so `vt_lb = 0`
> > 
> > * gcc 9 CUDA job raise "standards after C++17 are not supported", so `vt_lb = 0`
> > ```
> 
> Do we need to add more pipelines to replace these or do you think we have enough coverage. Can you look into that?

## Current
- General `LB` coverage is already strong (26 targets total / 17 building with `LB` enabled).
- Disabled legacy jobs still verify that core `VT` remains buildable with older C++17-only toolchains when `LB` is disabled (so we can keep those).

## Ideas
- `LB` + shared libraries:
   - Add => "vt_build_shared_libs = 1" to one modern LB-enabled Clang target
   - Note: currently shared-library coverage is attached only to Clang 9, where LB is disabled.

antoinemeyer5:
Reviewer: @lifflander 

## Observation

Check `CI / bake (vt-build-amd64-ubuntu-20-04-gcc-10-openmpi-cpp-spack` still fail.

VT CI fails with:
```
KeyError: No spec with name fmt
```

## Explanation

`spack-package/ci/add_vt_branch.py` creates a Spack version for the VT branch, but that version does not inherit dependencies declared for `@develop`, notably:
```
depends_on("darma-magistrate@develop", when="@develop")
depends_on("fmt@12", when="@develop")
```

Later, `package.py` accesses `self.spec["fmt"]`, causing the failure.

## Proposition

Update `add_vt_branch.py` so every generated branch version receives all dependencies declared with `when="@develop"`.
This belongs in `DARMA-tasking/spack-package`; no VT CI workaround should be needed.

antoinemeyer5:
Waiting for https://github.com/DARMA-tasking/spack-package/pull/36 to be merged.

ci/test_spack_package.sh
```diff
@@ -5,27 +5,23 @@ set -exo pipefail
...
 vt_spack_package="$cur_path/spack-package"
 
-git clone --branch v0.23.1 --depth=2 https://github.com/spack/spack.git
+# DARMA-tasking/spack-package uses Spack's v2 repository API, including the
+# spack_repo.builtin build-system imports introduced in Spack 1.x.
+git clone --branch v1.0.4 --depth=2 https://github.com/spack/spack.git
 . spack/share/spack/setup-env.sh
 
-git clone -b master https://github.com/DARMA-tasking/spack-package.git
+git clone -b 35-spack-package-fix https://github.com/DARMA-tasking/spack-package.git
```
antoinemeyer5: redo `master`, just for testing purpose

antoinemeyer5:
Fix is working:
image

I redo the `master` branch for `spack-package`.
It will work when https://github.com/DARMA-tasking/spack-package/pull/36 is merge.

lifflander:
@lifflander Need to do deeper integration testing for this.

PR #2525, opened 2026-08-25
#2524: Start integrating the new loc repo with VT replacing the location coordinator

antoinemeyer5:
Fixes #2524

PR #2527, opened 2026-09-16
#2526: constexpr breaks CUDA relaxed-constexpr builds

### LINKED ISSUE #2526: constexpr getFunction() breaks CUDA relaxed-constexpr builds
Building DistBVH with `Kokkos_ENABLE_CUDA_CONSTEXPR=ON`  (which adds `-expt-relaxed-constexpr`) fails on files that register a VT handler. Downstream libraries set this flag, so BVH needs to support it.

The error is:

    error: identifier "vt::location::EntityLocationCoord<...>::updateLocation"
           is undefined in device code
    error: identifier "MyObjGroup::my_handler" is undefined in device code

The problem is that the adapters in `src/vt/registry/auto/auto_registry_general.h` (lines 90, 102, 114, 152, 184, 216) declare `static constexpr FunctionPtrType getFunction() { return f; }`. Under `-expt-relaxed-constexpr` nvcc treats every `constexpr` function as implicitly `__host__ __device__` and compiles the device body where the address of a host-only handler doesn't exist.

The `constexpr` is unused: all six call sites assign to a runtime local (one via `reinterpret_cast`, which can't be constant-evaluated anyway), and there are no uses in `tests/`, `examples/`, or `tutorial/`. Dropping it fixes the build with no linkage change (in-class static member functions are already implicitly inline).

cwschilly:
Fixes #2526


cwschilly:
CI will be fixed by #2525

lifflander (approved):
LGTM

